PR #112, opened 2015-01-28
Some additional readme info

ceiroa:
How does this look, @jlantz? Should we also talk about the other env variables, like OAUTH_CLIENT_ID, etc?


jlantz:
Hmm... so those variables are very different in Jenkins and how they're configured.  I think we need to have separate documents for the different CI systems we've done integrations with.

ceiroa:
Oh, ok. Never mind then. I mainly wanted to mention the ant dependencies and the deployCI target (as in https://github.com/SalesforceFoundation/NPSP-API-Examples), then I got caught into trying to give additional info...

PR #118, opened 2015-05-08
Add PersonAccount to whitelist

powlo:
This fixes the issue where calls to the ant target "uninstallUnpackagedPost" will fail if unpackaged/post contains PersonAccount.object (error: "The CustomObject called 'PersonAccount' is standard and cannot be deleted")


jlantz:
@powlo Thanks for the contribution :)  This makes sense to me so merging it.

PR #121, opened 2015-07-24
Add AuraBundleDefinition types to buildPackageXmlFromDir

dextermilo:
This simply allows CumulusCI to traverse the aura folder in src and add the bundles to package.xml.


dextermilo:
@jlantz would you take a look at this?  I am just want to allow the aura bundles to get added to package.xml

PR #122, opened 2015-07-29
Added CustomMetadata for DestructiveChanges.xml Creation

triestelaporte:
Working a lot with CMDTs and had some stragglers that were not deleting.  Looks like it was never set up to recognize them.  Adding the included change made it all work without further changes, for me.


triestelaporte:
Something weird going on with this version.  Further testing needed.

PR #149, opened 2016-08-25
Support namespace injection in filenames.

paulmalenke:
Support the injection of the namespace in the filename during pre and post deploy. I have tested this in an unpackaged org. I am not sure how to test this in a packaged org. @jlantz, I will chat with you about it tomorrow.

### DIFF
--- a/build/cumulusci.xml
+++ b/build/cumulusci.xml
@@ -1051,6 +1051,9 @@ project.setProperty("downgradePackage", result);
               <delete dir="@{stagedir}" />
               <copy todir="@{stagedir}">
                 <fileset dir="@{subdir}"/>
+                <filtermapper>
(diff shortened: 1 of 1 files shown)


build/cumulusci.xml
```diff
@@ -1051,6 +1051,9 @@ project.setProperty("downgradePackage", result);
               <delete dir="@{stagedir}" />
               <copy todir="@{stagedir}">
                 <fileset dir="@{subdir}"/>
+                <filtermapper>
+                  <replacestring from="___NAMESPACE___" to="@{nsprefix}"/>
```
paulmalenke: I chose to use underscores instead of percents because a quick search labelled it as a symbol to not use in filenames.

PR #166, opened 2016-10-19
Add apiVersion to sf:retrieve calls

cdcarter:
Pass the cumulusci configured api version to all sf:retrieve calls, instead of using the default from the force.com migration tool jar.


jlantz:
This is awesome!  I had this on my backlog as something to do but you beat me to it :)

PR #168, opened 2016-10-19
pass through sessionId, serverURL, and apiVersion in all sf: ant calls

cdcarter:
review with w=1 ;)

### DIFF
--- a/build/build.xml
+++ b/build/build.xml
@@ -42,6 +42,21 @@
         <var name="sf.serverurl" value="${env.SF_SERVERURL}" />
       </then>
     </if>
+    
+    <if>
+      <isset property="env.SF_SESSIONID" />
+      <then>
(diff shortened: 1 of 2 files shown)


cdcarter:
note: in order to use a sf.sessionId attribute, you do need to make sure sf.username and sf.password are blank (NOT unassigned)

jlantz:
Would it make sense to add an environment variable mapping like SF_USERNAME and SF_PASSWORD and potentially override SF_USERNAME and SF_PASSWORD to force them to be a blank string if the sessionId is available around here:

https://github.com/SalesforceFoundation/CumulusCI/blob/master/build/build.xml#L45

cdcarter:
Like that? :)

jlantz:
Perfect :)

PR #176, opened 2016-10-28
add the userinfo endpoint to the orgconfig

cdcarter:
adds the userInfo endpoint to OrgConfig() as described here: https://developer.salesforce.com/page/Inside_OpenID_Connect_on_Force.com


cdcarter:
merp, totally breaks the tests. do not merge, will be back with a mock of the userinfo

jlantz:
Cool.  Here's the mocking library we're using:
https://github.com/getsentry/responses

PR #230, opened 2017-02-10
Feature/username

### LINKED ISSUE #204: C2 - Print org name/username at start of task/flow
Would be helpful to see the org name/username at the beginning of the flow. This is known when user provides --org option but could be useful when using the default org.

cdcarter:
# Critical Changes

# Changes
Include `userinfo` in the OrgConfig, and print username and org id at the beginning of every task run.

# Issues Closed
#204 Print org name/username at start of task/flow


cdcarter:
this maybe is more efficiently resolved with the conversation we had about logging, but this may still be useful in the interim.

PR #238, opened 2017-02-24
Feature/task context in flow

cdcarter:
# Critical Changes

# Changes
- Tasks that are not salesforce tasks will no longer log the org_config details even if an org_config was passed.
- The flow engine will print the org_config details at the top of a flow.
- When being run by the flow engine, tasks will not individually print their org_config details. This means, the org id and username are only printed once in the flow log. Yay!
- BaseTask has a new optional "flow" attribute which will be set to the BaseFlow instance when the task is executed from the flow runner. 

# Issues Closed

### DIFF
--- a/cumulusci/core/tests/test_flows.py
+++ b/cumulusci/core/tests/test_flows.py
@@ -1,6 +1,7 @@
 """ Tests for the Flow engine """
 
 import unittest
+import logging
 
 from collections import Callable
 
@@ -10,10 +11,11 @@ from cumulusci.core.config import BaseGlobalConfig
 from cumulusci.core.config import BaseProjectConfig
 from cumulusci.core.config import FlowConfig
 from cumulusci.core.config import OrgConfig
+from cumulusci.core.utils import MockLoggingHandler
+import cumulusci.core
 
 
 class _TaskReturnsStuff(BaseTask):
-
     def _run_task(self):
         self.return_values = {'name': 'supername'}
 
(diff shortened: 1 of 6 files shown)


cumulusci/core/utils.py
```diff
@@ -1,6 +1,45 @@
...
 def import_class(path):
+    """ Import a class from a string module class path """
     components = path.split('.')
     module = components[:-1]
     module = '.'.join(module)
     mod = __import__(module, fromlist=[components[-1]])
     return getattr(mod, components[-1])
+
+
+class MockLoggingHandler(logging.Handler):
```
cdcarter: moved it into cumulusci.core.tests.utils, since it doesn't actually contain tests to run.

cumulusci/core/tests/test_tasks.py
```diff
@@ -68,3 +93,45 @@ def _run_task(self):
...
+        """ A salesforce_task will not log the org id if run from a flow """
+
+        task = _SfdcTask(
+            self.project_config,
+            self.task_config,
+            self.org_config,
+            flow=BaseFlow(self.project_config, FlowConfig(), self.org_config)
+        )
+        task()
+        self.assertFalse(any(
```
cdcarter: now 

`    def test_task_logs_name_not_org(self):
        """ A task logs the task class name to info (and not creds) """
`

cumulusci/core/tests/test_flows.py
```diff
@@ -133,3 +172,37 @@ def test_call_task_not_found(self):
         flow = BaseFlow(self.project_config, flow_config, self.org_config)
 
         self.assertRaises(AttributeError, flow)
+
+    def test_flow_prints_org_id(self):
```
cdcarter:
```
    def test_flow_no_org_no_org_id(self):
        """ A flow without an org does not print the org ID """
```
added

cdcarter:
updated, there's probably more consistency work to do around test strings / test flow configs ...

cpolcyn:
Thanks for the contribution!

PR #246, opened 2017-03-01
Escape special XML characters in package name

### LINKED ISSUE #240: Zip file extraction fails due to package name mismatch
When extracting a zip file from Contacts & Organizations package using our utility function zip_subfolder() there is a name mismatch because in that project's cumulusci.yml the project name is "Contacts &amp;amp; Organizations". The function checks each file in the zip to see if the filename starts with this path, and in the zip the filenames are under path "Contacts & Organizations". This results in no files extracted, and exception is thrown when the expected package.xml is not found in the extracted files directory.

reproduce:
From inside Contacts & Organizations repo with packaging org already set up: `cci flow run ci_master --org <packaging org>`

cpolcyn:
# Critical Changes

# Changes

# Issues Closed
Fixes #240

### DIFF
--- a/cumulusci/salesforce_api/metadata.py
+++ b/cumulusci/salesforce_api/metadata.py
@@ -337,7 +337,7 @@ class ApiRetrievePackaged(BaseMetadataApiCall):
     def _build_envelope_start(self):
         return self.soap_envelope_start.format(
             self.api_version,
-            self.package_name,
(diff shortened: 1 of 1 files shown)


cpolcyn:
Can't use urlencode because the spaces are fine. This may need work in the future for other special XML characters.

PR #277, opened 2017-03-28
Push API enhancements

### LINKED ISSUE #254: Retry push without failing orgs
If we try to write PackagePushJob objects (link a request to an individual org) in batches of 200, sometimes one or more records in a batch will fail for some reason or another.  We've determined that the whole batch gets failed.  I think the solution is to get the list of failing record ids from the batch, remove them, then resubmit the batch.  If we can fix that bug, we can use batches of 200 and avoid blowing API limits in the org.

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/responses_composite_sobject_tree.htm#topic-title

cpolcyn:
# Critical Changes

# Changes

# Issues Closed
Fixes #254

### DIFF
--- a/cumulusci/tasks/push/push_api.py
+++ b/cumulusci/tasks/push/push_api.py
@@ -515,14 +515,12 @@ class SalesforcePushApi(object):
         for batch in batch_list(orgs, self.batch_size):
 
             batch_data = {'records': []}
-            i = 0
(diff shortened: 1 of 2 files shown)


Joble:
Thank you for taking care of this. It's much appreciated.

PR #286, opened 2017-04-03
report push status if start time is less than 1 minute in the future

### LINKED ISSUE #207: C2 - Exit when push is scheduled in the future
We use the start_time arg to schedule a push upgrade for a future time, but the CCI session enters a loop polling for status which would continue until the push job completes. Currently I address the issue by ctrl-c killing the process once I'm in the polling loop.

We should address this by detecting that the time is in the future and exit cleanly.

cpolcyn:
# Critical Changes

# Changes

# Issues Closed
Fixes #207

### DIFF
--- a/cumulusci/tasks/push/tasks.py
+++ b/cumulusci/tasks/push/tasks.py
@@ -298,10 +298,11 @@ class SchedulePushOrgList(BaseSalesforcePushTask):
         )
 
         # Report the status if start time is less than 1 minute from now
-        if start_time - datetime.datetime.utcnow() > datetime.timedelta(
-                    minutes=1
-                ):
+        if (start_time - datetime.datetime.utcnow() <
+                datetime.timedelta(minutes=1)):
             self._report_push_status(self.request_id)
+        else:
+            self.logger.info('Exiting early since request is in the future')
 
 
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/push/tasks.py
```diff
@@ -1,12 +1,13 @@
...
 
 from cumulusci.core.tasks import BaseTask
-from cumulusci.tasks.salesforce import BaseSalesforceApiTask
 from cumulusci.tasks.push.push_api import SalesforcePushApi
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
 
 class BaseSalesforcePushTask(BaseSalesforceApiTask):
-    completed_statuses = ['Succeeded','Failed','Cancelled']
+    completed_statuses = ['Succeeded', 'Failed', 'Cancelled']
```
Joble: It has been my experience that the last status is saved as **Canceled** in the database.

cpolcyn: Good catch, but that is outside the scope of this PR (this was just a formatting change). I will file a new issue for this.

cpolcyn: See #287

PR #289, opened 2017-04-03
pyinstaller support

cpolcyn:
# Critical Changes

# Changes
- Support for pyinstaller. Run `pyinstaller cci.spec` to generate an executable file version of CumulusCI.

# Issues Closed


cpolcyn:
We may need to add more hidden imports as they are discovered.

cpolcyn:
I've added all the data files and hidden imports I could think of, and did lots of testing to make sure those paths are found.

cdcarter:
HOLY. COW.

PR #307, opened 2017-04-12
Push enhancements

### LINKED ISSUE #294: Retry a batch when "something bad has happened."
Occasionally when scheduling an upgrade, every org in a batch throws the _"Something bad has happened. Please verify your input and try again."_ exception. The entire batch should be retried as the something bad exception seems to be intermittent. Retrying may prevent orgs from being excluded due to a random problem.

### LINKED ISSUE #306: Push script crashes when 0 orgs scheduled
If orgs are removed from the push request and we end up with a request with 0 orgs, we should cancel it and exit cleanly.

### LINKED ISSUE #308: The schedule command should fail if start time in the past is supplied.
When scheduling an upgrade with a specified start time, the command should fail if the time provided is in the past.

Why? Because a time in the past will immediately start the upgrade, which wasn't the intent because a time was given. A single misplaced number can lead to a premature upgrade.

cpolcyn:
# Critical Changes

# Changes
* Push org lists (text files with one org ID per line) can now have comments and blank lines. The first word on the line is assumed to be the org ID and anything after that is ignored.

# Issues Closed
Fixes #294 
Fixes #306 
Fixes #308

### DIFF
--- a/cumulusci/tasks/push/push_api.py
+++ b/cumulusci/tasks/push/push_api.py
@@ -545,7 +545,7 @@ class SalesforcePushApi(object):
                 scheduled_orgs,
             )
         )
-        return request_id
+        return request_id, scheduled_orgs
 
     def _add_batch(self, batch, request_id):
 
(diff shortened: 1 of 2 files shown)


Joble:
Hooray for the batch retry functionality! Thanks for this.

PR #316, opened 2017-05-12
New exception for browser tests

cpolcyn:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/command.py
+++ b/cumulusci/tasks/command.py
@@ -58,11 +58,11 @@ class Command(BaseTask):
     def _process_output(self, line):
         self.logger.info(line.rstrip())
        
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/command.py
```diff
@@ -127,3 +122,11 @@ def _get_env(self):
         else:
             env['RUN_LOCAL'] = 'True'
         return env
+
+    def _handle_returncode(self, returncode, stderr):
+        if returncode:
```
jlantz: We should probably only throw the browser test error with an exit status of 1.  Any non-zero exit status is considered an error but it's probably safe to assume most browser test runner commands will return an exit status of 1 for test failures and might return a different exit code for other types of errors.

PR #326, opened 2017-05-26
add task before/after debug

cdcarter:
# Critical Changes

# Changes
add `--debug-before` and `--debug-after` options to task execution via the cli, which drops the user into pdb, the Python debugger, to debug your task.

# Issues Closed


cumulusci/cli/cli.py
```diff
@@ -647,6 +655,10 @@ def task_run(config, task_name, org, o, debug):
     if exception:
         raise exception
 
+    if debug_after:
```
cpolcyn: Seems this should be before raising exception so user can inspect. Not sure if you had a reason in mind though.

cdcarter: debug_after probably should imply debug

cpolcyn: I like them as independent from each other. Setting debug this way I think would be confusing to the user. Seems reasonable to keep them separate so that you can `debug` at end without worrying about the exceptions. But in that case you'd need to move it before the `raise exception' above.

PR #359, opened 2017-06-27
New SFDX Command Wrapper

jlantz:
# Critical Changes

* Removes `cumulusci.tasks.salesforcedx.*` which is now replaced by `cumulusci.tasks.sfdx.*`

# Changes

* New test classes `cumulusci.tasks.sfdx.SFDXBaseTask` and `cumulusci.tasks.sfdx.SFDXOrgTask` wrap the sfdx cli.  `SFDXOrgTask` passes the current cci keychain org's access token and instance url via environment variable so sfdx runs its commands against the cci org.

# Issues Closed


cumulusci/cumulusci.yml
```diff
@@ -51,18 +51,6 @@ tasks:
             managed: True
             namespace_token: '%%%NAMESPACE%%%'
             filename_token: '___NAMESPACE___'
-    dx_push:
-        description: Uses Salesforce DX to push code to a scratch workspace org
-        class_path: cumulusci.tasks.salesforcedx.BaseSalesforceDXTask
-        options:
-            command: force:src:push
-            options: '-a'
-    filter_subscribers:
```
cpolcyn: Why are filter_subscribers and generate_apex_docs deleted? Seems they should be left in.

cumulusci/tasks/sfdx.py
```diff
@@ -0,0 +1,184 @@
...
+Actually do this in Command. have it expose command segments.
+
+Then here in SFDX we will add an additional metalayer for
+how the CLI formats args opts commands etc.
+"""
+import json
+
+from cumulusci.core.config import ScratchOrgConfig
+from cumulusci.tasks.command import Command
+from cumulusci.tasks.command import SalesforceCommand
```
cpolcyn: Unnecessary import

cumulusci/tasks/sfdx.py
```diff
@@ -0,0 +1,184 @@
...
+
+class SFDXBaseTask(Command):
+    """ Call the sfdx cli with params and no org """
+
+    task_options = {
+        'command': {
+            'description': 'The full command to run with the sfdx cli.',
+            'required': True,
+        },
+        'extra': {
```
cpolcyn: Why is `extra` separate from `command`? Seems it could all just be `command`.

cumulusci/tasks/sfdx.py
```diff
@@ -0,0 +1,184 @@
...
+    def _get_env(self):
+        env = super(SFDXOrgTask, self)._get_env()
+        if not isinstance(self.org_config, ScratchOrgConfig):
+            # For non-scratch keychain orgs, pass the access token via env var
+            env['SFDX_INSTANCE_URL'] = self.org_config.instance_url
+            env['SFDX_USERNAME'] = self.org_config.access_token
+        return env
+
+class SFDXJsonTask(SFDXOrgTask):
+    command = 'force:mdapi:deploy --json'
```
cpolcyn: Here you include `--json` which seems it would go in the `extra` arg. See comment above about consolidating these args.

requirements_dev.txt
```diff
@@ -10,4 +10,5 @@ flake8==2.6.0
 tox==2.3.1
 coverage==4.1
 Sphinx==1.4.8
+testfixtures==4.13.4
```
cpolcyn: Why different version from setup.py?

cumulusci/tasks/command.py
```diff
@@ -68,28 +81,46 @@ def _handle_returncode(self, returncode, stderr):
             self.logger.error(message)
             raise CommandException(message)
 
-    def _run_command(self, env):
+    def _run_command(self, env, command=None, output_handler=None, return_code_handler=None):
```
cpolcyn: Why the optional args? The only one I see used to override is `command` and in that case it's confusing because you also have a `command` option available to the user, but it is hard-coded to override.

cumulusci/cumulusci.yml
jlantz: filter_subscribers and get_subscribers both raise NotImplemented errors from the task class. I'm not sure why we mapped those two through in the cumulusci.yml.  I'm going to remove both from this branch.  I'll put the generate_apex_docs task back though.  This was caused by a merge conflict.

cumulusci/tasks/sfdx.py
jlantz: I'll remove

cumulusci/tasks/sfdx.py
jlantz: The purpose of extra is to allow for this:
cci task run TASK_NAME -o extra "-a some_value -b some_other_value"

Basically, command defines the core command you have to run and that allows you to pass extra args via the CLI

requirements_dev.txt
jlantz: I'll fix that.

cumulusci/tasks/command.py
jlantz: Check out `_poll_action` on SFDXJsonPollingTask.  It injects a different command into _run_command to do a status check on the deployment.  In this case you don't want to change the command in self.options b/c that's the main command you're running.

PR #381, opened 2017-07-28
New task: commit Apex docs to GitHub

cpolcyn:
Given a local dir containing the output of ApexDoc tool, upload to a target repo+branch+subdir.

Execution process:
1. HEAD tree of branch is used as basis for comparison to determine update/add/delete
2. New blobs are created for updated and added files
3. New tree is constructed, leaving out items to delete. Also leaves out subtree items as they are implied by blob paths (API will reject call if subtrees are present).
4. New commit to point at new tree
5. Update HEAD to point at new commit

Verified using gh-pages branch of CumulusCI-Test repo.

Known issues:
* Author info should not need to be given as it should be filled in by GitHub after the API call. We supply our own author info from cci github config as a workaround.
* When using GitHub user "Mr Belvedere" GitHub is not recognizing the user. Need someone with 2FA account access for that user to confirm email is set correctly in account settings.

Future enhancements:
* Use regex to specify local files to skip. Better yet, read from gitignore. Currently skips all files starting with `.`.


cumulusci/tasks/github.py
```diff
@@ -207,3 +210,51 @@ def _run_task(self):
                 )
 
                 self.logger.info('Merge conflict on branch {}: created pull request #{}'.format(branch.name, pull.number))
+
+
+class CommitApexDocs(BaseGithubTask):
+
+    task_options = {
```
cdcarter: Can any of these task options receive default values from the ProjectConfig in cumulusci.yml? 

We have a config section for ApexDocs as well as one where we list git branches. It'd be nice if this task worked "automatically" if you have the github service configured and your apexdocs setup. 

I bet we could make very reasonable defaults for `branch` and `dir_local` that inherit from the global yaml

cumulusci/tasks/github.py
```diff
@@ -207,3 +210,51 @@ def _run_task(self):
...
+        'branch': {
+            'description': 'branch name',
+            'required': True,
+        },
+        'dir_local': {
+            'description': 'local dir of ApexDocs (contains index.html)',
+            'required': True,
+        },
+        'dir_repo': {
```
cdcarter: This option name is confusing. not sure what's better tho. `commit_path`?

cumulusci/tasks/github.py
```diff
@@ -207,3 +210,51 @@ def _run_task(self):
...
+        # get API
+        repo = self.get_repo()
+
+        # commit
+        author = {
+            'name': self.github.user().name,
+            'email': self.github_config.email,
+        }
+        commit_dir = CommitDir(repo, self.logger, author)
+        commit_dir(local_dir, branch, repo_dir, dry_run)
```
cdcarter: can the commit sha be a task return value? just add it to self.return_values (https://github.com/SalesforceFoundation/CumulusCI/blob/master/cumulusci/tasks/github.py#L143)

cumulusci/tasks/github_commit.py
```diff
@@ -0,0 +1,188 @@
...
+        commit = self.repo.git_commit(head.object.sha)
+
+        # get tree of commit
+        tree = self.repo.tree(commit.tree.sha) # shallow tree
+        tree = tree.recurse() # recurse/flatten
+        tree = tree.to_json() # convert to native types
+
+        # create new tree (delete, update)
+        new_tree_list = []
+        changed = False
```
cdcarter: can this be something like `changes_to_commit`? when reading the loop, it was hard to tell what level this flag was for

cumulusci/tasks/github_commit.py
```diff
@@ -0,0 +1,188 @@
...
+import logging
+import os
+
+from cumulusci.core.exceptions import GithubException
+
+
+class CommitDir(object):
+    """Commit all changes in local_dir to branch/repo_dir"""
+
+    def __init__(self, repo, logger=None, author={}):
```
cdcarter: for author, I think you should use a default arg val of None and then check for that in the initializer, since {} is mutable.

cumulusci/tasks/github_commit.py
```diff
@@ -0,0 +1,188 @@
...
+            if not new_tree:
+                raise GithubException('Failed to create tree')
+
+        # create new commit
+        if dry_run:
+            self.logger.info('[dry_run] Skipping creation of new commit')
+        else:
+            self.logger.info('Creating new commit')
+            new_commit = self.repo.create_commit(
+                message='Update Apex docs',
```
cdcarter: can we pass this commit message from the calling task, in case we end up wanting to use this utility to commit other things?

cumulusci/tasks/github.py
cpolcyn: I named it this way because I wanted to make it clear that the dir on local machine would be copied exactly into the dir at the remote repo. How about `dir_remote`?

cumulusci/tasks/github.py
cpolcyn: Agreed for `branch` value but for `dir_local` this is the output of the apex docs command, not sure we'd want that set by default. If so, the value should also be used in generation of ApexDocs.

cdcarter: I'm not super familiar with how you run the apexdocs task. It looks like it doesn't have a default either.

I think it'd be great if the apexdocs task and this task shared a value there. This could be with a project config value that both tasks use as a default, or have the apexdocs task add the dir to return_values, and use that as the task option in a flow that generates *and* uploads.

cumulusci/tasks/github_commit.py
cpolcyn: Yes, good idea.

cumulusci/tasks/github.py
cdcarter: either that or `dir_branch` maybe? since this is mostly used for branches that don't have `master` as a parent.

cumulusci/tasks/github_commit.py
cpolcyn: Sure. I hesitated to use flag at all for this, would be nice to just compare before-and-after but they will always be different due to remove of subtree items. But even so, I'm now thinking it is better to refactor and do a comparison at the end, it will be cleaner and hopefully clearer to read.

cumulusci/tasks/github.py
cpolcyn: Correct, no default there either. Will set up a shared default value in yml.

cumulusci/tasks/github.py
cpolcyn: Could certainly be master branch if a user keeps their docs in master, or could otherwise share history with master. And remote doesn't really imply any branch since it is independent of branches in a way. Let's go with `dir_remote` if that's okay with you.

cumulusci/tasks/github_commit.py
cpolcyn: Not sure what the concern is with mutable default values, but happy to change to None at least for consistency/readability.

cpolcyn:
@cdcarter 
> interested to hear about why you chose to make CommitDir a Callable

As a class it can hold persistent info like repo API and other init vars, then running the command takes fewer args. Also creates space for adding complexity later.

cumulusci/tasks/github.py
cdcarter: Sounds good.

cumulusci/tasks/github_commit.py
cpolcyn: Found some info on the concern here: http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments

cumulusci/tasks/github.py
cpolcyn: Thinking about this more, it really has nothing to do with remote vs local either, it is all about where in the repo you want it to be (repo is same whether local or remote). I'd like to leave it as `dir_repo` for clarity.

cumulusci/tasks/github.py
cpolcyn: This would mean piping it out of the CommitDir class as well. What is the need for the commit sha?

cdcarter: I don't know if there's a compelling usecase other than logging it for debugging / a quick link to the commit in github. so, i'm not committed to it.

cdcarter: get it? committed? 😹

cumulusci/tasks/github.py
cdcarter: yea, the more i think about it, it works.

cpolcyn:
@cdcarter I just pushed a bunch of commits to address your review comments

cumulusci/tasks/github_commit.py
```diff
@@ -0,0 +1,217 @@
...
+        if not os.path.isdir(local_dir):
+            raise GithubException('Not a dir: {}'.format(local_dir))
+        if repo_dir is None:
+            repo_dir = ''
+        if repo_dir.startswith('.'):
+            repo_dir = repo_dir[1:]
+        if repo_dir.startswith('/'):
+            repo_dir = repo_dir[1:]
+        if repo_dir.endswith('/'):
+            repo_dir = repo_dir[:-1]
```
cdcarter: Is there something in the stdlib that can do this?

cumulusci/tasks/github_commit.py
```diff
@@ -0,0 +1,217 @@
...
+            self.logger.info('[dry_run] Skipping creation of new tree')
+        else:
+            self.logger.info('Creating new tree')
+            new_tree = self.repo.create_tree(new_tree_list, None)
+            if not new_tree:
+                raise GithubException('Failed to create tree')
+
+        # create new commit
+        if commit_message is None:
+            commit_message = 'Commit dir {} to {}/{}/{} via CumulusCI'.format(
```
cdcarter: Nice touch!

cumulusci/tasks/github_commit.py
cpolcyn: We don't want to use anything from os.path because this is not dealing with local paths. Also we want to preserve any "inner" dir seps e.g. `./foo/bar/` should be converted to `foo/bar`.

cdcarter: makes a lot of sense. a comment reminding the reader that its not a local path and don't use os.path might be helpful to prevent a mistake in the future.

PR #385, opened 2017-08-02
Specify release notes parser config in yml

cpolcyn:
Need to support per-project parser customizations for release notes. YAML to the rescue!

Once this is merged and deployed to mrbelvedereci, SAL has an associated change to take advantage (https://github.com/SalesforceFoundation/sal/commit/a61684fc871954dab029debe1907c2903a303293). Otherwise all other repos should work without any modifications.

Tested in multiple repos with lots of option toggling.

### DIFF
--- a/cumulusci/tasks/release_notes/parser.py
+++ b/cumulusci/tasks/release_notes/parser.py
@@ -130,6 +130,7 @@ class GithubLinesParser(ChangeNotesLinesParser):
 
     def __init__(self, release_notes_generator, title):
         super(GithubLinesParser, self).__init__(release_notes_generator, title)
+        self.link_pr = release_notes_generator.link_pr
         self.pr_number = None
         self.pr_url = None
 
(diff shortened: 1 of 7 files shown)


cumulusci/tasks/release_notes/parser.py
```diff
@@ -188,16 +187,24 @@ def github_info(self):
 
 
 class GithubIssuesParser(IssuesParser, ParserGithubApiMixin):
+    message_prod = 'Included in production release'
```
cdcarter: format these as a `CONSTANT`?

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,8 +12,8 @@ class GithubReleaseNotes(BaseGithubTask):
             'required': True,
         },
         'publish': {
```
cdcarter: I wouldn't change it here, but we have a task with `dry run` (default True) and a task with `publish` (default False) and i wonder if we should try to be consistent if thats a positive or a negative.

i could see that becoming confusing if we wanted to be able to run a release flow that does everything dry first.

PR #388, opened 2017-08-04
don't try to escape None strings

### LINKED ISSUE #376: AttributeError: 'NoneType' object has no attribute 'replace'
I've seen this happen a couple times on builds with test failures where something needs to be recompiled. Exception comes from task run_tests_debug. For example see the original build here:
https://mrbelvedereci.herokuapp.com/builds/5237

cdcarter:
@cpolcyn - easy review.

closes #376

### DIFF
--- a/cumulusci/tasks/apex/testrunner.py
+++ b/cumulusci/tasks/apex/testrunner.py
@@ -319,10 +319,13 @@ class RunApexTests(BaseSalesforceApiTask):
                     s += ' time="{}"'.format(result['Stats']['duration'])
                 if result['Outcome'] in ['Fail', 'CompileFail']:
                     s += '>\n'
-                    s += ('    <failure type="failed" ' +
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/apex/testrunner.py
```diff
@@ -319,10 +319,13 @@ def _write_output(self, test_results):
                     s += ' time="{}"'.format(result['Stats']['duration'])
                 if result['Outcome'] in ['Fail', 'CompileFail']:
                     s += '>\n'
-                    s += ('    <failure type="failed" ' +
-                          'message="{}"><![CDATA[{}]]></failure>\n'.format(
-                              cgi.escape(result['Message']),
-                              cgi.escape(result['StackTrace'])))
+                    s += '    <failure type="failed" ' 
+                    if result['Message']:
```
cpolcyn: If this is False, "<failure " tag is not closed.

PR #399, opened 2017-08-17
Restructure github tasks and provide tests for MergeBranch

jlantz:
This branch current contains a restructuring of tasks/github.py into tasks/github/*.py so task files are more manageable. The branch also includes 88% test coverage for the MergeBranch github task class to enable refactoring.

Other changes are related to building a better infrastructure for testing task classes. The new `cumulusci.test.util.create_project_config` method can quickly create a dummy project config to use in tests.


cumulusci/tasks/github/base.py
```diff
@@ -0,0 +1,20 @@
+from github3 import login
+
+from cumulusci.core.exceptions import GithubException
```
cpolcyn: unused import

cumulusci/tasks/github/base.py
```diff
@@ -0,0 +1,20 @@
...
+from cumulusci.core.tasks import BaseTask
+from cumulusci.core.utils import process_bool_arg
```
cpolcyn: unused import

cumulusci/tests/util.py
```diff
@@ -0,0 +1,61 @@
...
+
+import yaml
+
+from cumulusci.core.config import BaseGlobalConfig
+from cumulusci.core.config import BaseProjectConfig
+from cumulusci.core.keychain import BaseProjectKeychain
+from cumulusci.core.config import OrgConfig
+from cumulusci import __location__
+
+def random_sha():
```
cpolcyn: This should probably be seeded with a constant, so that the random sequence is the same each test run.

cumulusci/tests/util.py
```diff
@@ -0,0 +1,61 @@
...
+    hash = random.getrandbits(128)
+    return "%032x" % hash
+
+def get_base_config():
+    path = os.path.abspath(os.path.join(
+            __location__,
+            'cumulusci.yml'
+        ))
+    f = open(path, 'r')
```
cpolcyn: use context manager `with open() as f:`

cumulusci/tests/util.py
jlantz: I'm not quite sure the implications of doing that but I'm open to discussing it.  The functionality works fine without a consistent random sequence each test run.

cpolcyn: I'm sure it does work fine but this is more about being able to reproduce any failures. You can imagine with a randomly seeded random generator, a test failure could happen in one test run and not another even though the code is the same. I think it's a good policy in general to seed random number generators for testing with a constant. Not a blocker though.

PR #402, opened 2017-08-28
Support feature/parent__child branch naming and automate merges

jlantz:
This branch implements the logic to handle the new parent/child branch naming format for `feature/parent__child` and includes the following functional changes to CumulusCI's automated branch merging logic:

- The `github_master_to_feature` task will now skip any child branches that have a parent branch.  Given parent branch `feature/parent` exists and `feature/not-a` does not exist:
  - `feature/parent__child` is skipped
  - `feature/not-a__child` uses the child syntax but the parent branch does not exist.  The merge is run against this branch
- The new `github_parent_to_children` task merges from the current branch to any child branches.  If the current branch is `feature/parent` this would merge to `feature/parent__child`
- Task option values can now dynamically lookup attributes on project_config with the syntax `$project_config.attr_name`.  For example, you could get the current branch via `$project_config.repo_branch`
- `ci_feature` flow now runs `github_parent_to_children` at the end of the flow

This branch also includes 100% test coverage for the changes :)

### DIFF
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -67,6 +67,16 @@ class BaseTask(object):
         if kwargs:
             self.options.update(kwargs)
 
+        # Handle dynamic lookup of project_config values via $project_config.attr
+        for option, value in self.options.items():
+            try:
+                if value.startswith('$project_config.'):
+                    attr = value.replace('$project_config.', '', 1)
+                    self.options[option] = getattr(self.project_config, attr, None)
+            except AttributeError:
+                pass
+                
+
     def _validate_options(self):
         missing_required = []
         for name, config in self.task_options.items():
--- a/cumulusci/tasks/github/merge.py
+++ b/cumulusci/tasks/github/merge.py
@@ -15,57 +15,166 @@ class MergeBranch(BaseGithubTask):
         'branch_prefix': {
             'description': "The prefix of branches that should receive the merge.  Defaults to project__git__prefix_feature",
         },
+        'parent_child_only': {
+            'description': "If True, only the parent to child branch merge is run.  Defaults to False",
+        },
     }
 
+    def _init_options(self, kwargs):
+        super(MergeBranch, self)._init_options(kwargs)
+
+        if 'commit' not in self.options:
+            self.options['commit'] = self.project_config.repo_commit
+        if 'branch_prefix' not in self.options:
+            self.options['branch_prefix'] = self.project_config.project__git__prefix_feature
+        if 'source_branch' not in self.options:
+            self.options['source_branch'] = self.project_config.project__git__default_branch
+        if 'parent_child_only' not in self.options:
+            self.options['parent_child_only'] = False
+        elif self.options['parent_child_only'] in [True, 'True']:
+            self.options['parent_child_only'] = True
+
     def _run_task(self):
-        repo = self.get_repo()
+        self.repo = self.get_repo()
 
-        commit = self.options.get('commit', self.project_config.repo_commit)
-        branch_prefix = self.options.get('branch_prefix', self.project_config.project__git__prefix_feature)
-        source_branch = self.options.get('source_branch', self.project_config.project__git__default_branch)
+        commit = self.options['commit']
+        branch_prefix = self.options['branch_prefix']
+        source_branch = self.options['source_branch']
+        parent_child_only = self.options['parent_child_only']
 
-        head_branch = repo.branch(source_branch)
+        head_branch = self.repo.branch(source_branch)
         if not head_branch:
             message = 'Branch {} not found'.format(source_branch)
             self.logger.error(message)
             raise GithubApiNotFoundError(message)
 
         # Get existing pull requests targeting a target branch
-        existing_prs = []
-        for pr in repo.iter_pulls(state='open'):
+        self.existing_prs = []
+        for pr in self.repo.iter_pulls(state='open'):
             if pr.base.ref.startswith(branch_prefix):
-                existing_prs.append(pr.base.ref)
-        
-        targets = []
-        for branch in repo.iter_branches():
+                self.existing_prs.append(pr.base.ref)
+       
+        # Create list and dict of all target branches 
+        branches = []
+        branches_dict = {}
+        for branch in self.repo.iter_branches():
+            if branch.name == source_branch:
+                self.logger.debug('Skipping branch {}: is source branch'.format(branch.name))
+                branches_dict[branch.name] = branch
+                continue
             if not branch.name.startswith(branch_prefix):
                 self.logger.info('Skipping branch {}: does not match prefix {}'.format(branch.name, branch_prefix))
                 continue
-            if branch.name == source_branch:
-                self.logger.info('Skipping branch {}: is source branch'.format(branch.name))
-                continue
-               
-            compare = repo.compare_commits(branch.commit.sha, commit)
-            if not compare.files:
-                self.logger.info('Skipping branch {}: no file diffs found'.format(branch.name))
-                continue
-   
-            try: 
-                result = repo.merge(branch.name, source_branch)
-                self.logger.info('Merged {} commits into {}'.format(compare.behind_by, branch.name))
-            except GitHubError as e:
-                if e.code != 409:
-                    raise
+            branches.append(branch)
+            branches_dict[branch.name] = branch
 
-                if branch.name in existing_prs:
-                    self.logger.info('Merge conflict on branch {}: merge PR already exists'.format(branch.name))
-                    continue 
+        # Identify parent/child branches
+        possible_children = []
+        possible_parents = []
+        parents = {}
+        children = []
+        for branch in branches:
+            parts = branch.name.replace(branch_prefix, '', 1).split('__', 1)
+            if len(parts) == 2:
+                possible_children.append(parts)
+            else:
+                possible_parents.append(branch.name)
 
-                pull = repo.create_pull(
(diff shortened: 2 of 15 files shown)


cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
-        source_branch = self.options.get('source_branch', self.project_config.project__git__default_branch)
+        if 'commit' not in self.options:
+            self.options['commit'] = self.project_config.repo_commit
+        if 'branch_prefix' not in self.options:
+            self.options['branch_prefix'] = self.project_config.project__git__prefix_feature
+        if 'source_branch' not in self.options:
+            self.options['source_branch'] = self.project_config.project__git__default_branch
+        if 'children_only' not in self.options:
+            self.options['children_only'] = False
+        elif self.options['children_only'] in [True, 'True']:
```
cdcarter: use `cumulusci.core.utils.process_bool_arg()` ?

cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
+                branch_type,
+                branch,
+            ))
+            if children and not self.options['children_only']:
+                self.logger.info('  Skipping merge into the following child branches:')
+                for child in children:
+                    self.logger.info('    {}'.format(child.name))
+
+        except GitHubError as e:
+            if e.code != 409:
```
cdcarter: use httplib.CONFLICT

cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
+                raise
```
cdcarter: is there a CCI wrapped exception type we should use?

cumulusci/tasks/github/tests/test_merge.py
```diff
@@ -192,9 +202,11 @@ def test_no_feature_branch(self):
             expected = [
                 ('INFO', 'Beginning task: MergeBranch'),
                 ('INFO', ''),
+                ('DEBUG', 'Skipping branch master: is source branch'),
+                ('DEBUG', 'Skipping branch not-a-feature-branch: does not match prefix feature/'),
             ]
-            self.assertEquals(log_lines, expected)
-        self.assertEquals(len(responses.calls), 4)
+            self.assertEquals(expected, log_lines)
```
cdcarter: is there an easy way to just assert that each of the expected lines is IN the log? i don't like that adding additional logging could break the test, but I don't feel that strongly abt this.

cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
+                if not self.options['children_only']:
+                    self.logger.debug('Skipping branch {}: does not match prefix {}'.format(branch.name, self.options['branch_prefix']))
                 continue
-            if branch.name == source_branch:
-                self.logger.info('Skipping branch {}: is source branch'.format(branch.name))
+            branches.append(branch)
+            branches_dict[branch.name] = branch
+
+        # Identify parent/child branches
+        possible_children = []
```
cpolcyn: Might be nice to encapsulate the building of possible parents/children into another function. Or go big and encapsulate all of building the `branch_tree`.

cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
-        targets = []
-        for branch in repo.iter_branches():
-            if not branch.name.startswith(branch_prefix):
-                self.logger.info('Skipping branch {}: does not match prefix {}'.format(branch.name, branch_prefix))
+        self.existing_prs = []
+        for pr in self.repo.iter_pulls(state='open'):
+            if pr.base.ref.startswith(self.options['branch_prefix']):
+                self.existing_prs.append(pr.base.ref)
+       
+        # Create list and dict of all target branches
```
cpolcyn: We're looping over a list of branches twice in a row (L55, L73). Could we instead do it all in one loop, adding parent/child info along the way?

cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
+        branches = []
+        branches_dict = {}
+        for branch in self.repo.iter_branches():
+            if branch.name == self.options['source_branch']:
+                if not self.options['children_only']:
+                    self.logger.debug('Skipping branch {}: is source branch'.format(branch.name))
+                    branches_dict[branch.name] = branch
+                    continue
+            if not branch.name.startswith(self.options['branch_prefix']):
+                if not self.options['children_only']:
```
cpolcyn: Is this `if` needed? Seems that children_only or not, we should be printing a message about skipping a branch due to not matching prefix.

cumulusci/tasks/github/merge.py
cpolcyn: Use httplib constants

cumulusci/tasks/github/tests/test_merge.py
```diff
@@ -176,14 +184,16 @@ def test_branch_does_not_exist(self):
         task = self._create_task()
         with self.assertRaises(GithubApiNotFoundError):
             task()
-        self.assertEquals(len(responses.calls), 2)
+        self.assertEquals(2, len(responses.calls))
```
cpolcyn: Just curious… Why switch these around?

cumulusci/tasks/github/merge.py
```diff
@@ -15,57 +15,187 @@ class MergeBranch(BaseGithubTask):
...
+            self.logger.info('{}Merged {} commits into {} {}'.format(
+                indent,
+                compare.behind_by,
+                branch_type,
+                branch,
+            ))
+            if children and not self.options['children_only']:
+                self.logger.info('  Skipping merge into the following child branches:')
+                for child in children:
+                    self.logger.info('    {}'.format(child.name))
```
cpolcyn: Should be using `indent` value here?

cumulusci/tasks/github/merge.py
jlantz: Probably not.  github3.py is pretty good about throwing specific exceptions so the exception should provide enough info to figure out what's going on.

cumulusci/tasks/github/tests/test_merge.py
jlantz: I kind of like that it's an all or nothing thing.  That allows us to test that content is not shown in the log as well as that it is.

cumulusci/core/tasks.py
```diff
@@ -67,6 +67,16 @@ def _init_options(self, kwargs):
...
             self.options.update(kwargs)
 
+        # Handle dynamic lookup of project_config values via $project_config.attr
+        for option, value in self.options.items():
+            try:
+                if value.startswith('$project_config.'):
+                    attr = value.replace('$project_config.', '', 1)
+                    self.options[option] = getattr(self.project_config, attr, None)
+            except AttributeError:
+                pass
```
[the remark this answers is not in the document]
jlantz: The value might not be a string in which case startswith() will throw an AttributeError.  Basically, it's a conservative call that if you can't even call startswith you should probably just leave the value alone.

cumulusci/tasks/github/tests/test_merge.py
```diff
@@ -117,7 +124,7 @@ def _mock_branch_does_not_exist(self, branch):
     def _mock_merge(self, conflict=None):
         api_url = '{}/merges'.format(self.repo_api_url)
         expected_response = self._get_expected_merge(conflict)
-        status = 200
+        status = 201
```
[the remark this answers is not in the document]
jlantz: Done

cumulusci/tasks/github/tests/test_merge.py
jlantz: The output seems to be nicer if you put expected first in the assertEquals call.  Definitely the diff is nicer that way.

cumulusci/tasks/github/merge.py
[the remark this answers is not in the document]
jlantz: Done

cumulusci/tasks/github/merge.py
jlantz: You have to identify all children first and then go searching for their parents.  You might have a child early in the list whose parent you haven't get iterated to in the loop.  So you need a loop to build a list of possible children and possible parents and then a list to loop through the child to try to find a parent.

cumulusci/tasks/github/merge.py
jlantz: I skipped output here because in the case of a children_only merge you're merging from a parent feature branch to child feature branches.  It's impossible for a child branch to not have the feature prefix.  In the case of children_only merges, I wanted to just have it show the info on child branches it found rather than all the other branches.

cumulusci/tasks/github/merge.py
jlantz: Done

cumulusci/tasks/github/merge.py
jlantz: Actually, indent was from the previous incarnation where _merge was _merge_recursive and did the recursive merge from master -> parent -> child all in one run.  I just removed the indent altogether.

cumulusci/tasks/github/merge.py
jlantz: Went even bigger and encapsulated almost all of _run_task :)

PR #418, opened 2017-09-05
100% test coverage for core/flow.py

jlantz:
One thing I'm not happy about is that `test_task_raises_exception_fail` for some reason prints the traceback from the raised exception it's testing.  I'm not sure if that's worth trying to figure out now or if we just chalk it up to a cost of higher coverage for now.

### DIFF
--- a/cumulusci/core/tests/test_flows.py
+++ b/cumulusci/core/tests/test_flows.py
@@ -29,6 +29,15 @@ class _TaskResponseName(BaseTask):
     def _run_task(self):
         return self.options['response']
 
+class _TaskRaisesException(BaseTask):
+    task_options = {
+        'exception': {'description': 'The exception to raise'},
+        'message': {'description': 'The exception message'},
+    }
+
+    def _run_task(self):
+        raise self.options['exception'](self.options['message'])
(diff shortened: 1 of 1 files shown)


cumulusci/core/tests/test_flows.py
```diff
@@ -111,6 +128,181 @@ def test_pass_around_values(self):
...
+        """ A flow aborts when a task raises an exception """
+
+        flow_config = FlowConfig({
+            'description': 'Run a task',
+            'tasks': {
+                1: {'task': 'raise_exception'},
+            }
+        })
+        flow = BaseFlow(self.project_config, flow_config, self.org_config)
+        self.assertRaises(Exception, flow)
```
cpolcyn: Shouldn't assertRaises be used in context manager? E.g. `with self.assertRaises(Exception):`

jlantz: The context manager is only needed if you want to check attributes on the exception that's raised.  self.assertRaises will call the second arg and assert that it raises an instance of the first arg.  I tried both ways and the traceback still gets printed out.

PR #443, opened 2017-09-19
Allow for namespaced and non-namespaced scratch orgs in the same project

### LINKED ISSUE #441: Support namespaced and non-namespaced scratch org definitions in cumulusci.yml
sfdx force:org:create supports the -n option which will create a scratch org without a namespace applied.  If a project's cumulusci.yml has a namespace value, all scratch orgs will be created with that namespace applied by default unless -n is passed in the cli.

For cci, we want projects to be able to configure the default org shapes to include the namespace or not include the namespace.  For example, Project A might want the `dev` scratch config to not use the namespace but would want to have `dev_managed` defined to apply the project's namespace.  Project B might want their `dev` scratch config to create namespaced orgs but would want `beta` orgs to not have the namespace to avoid naming conflicts.

This could be done via a `managed: True` option on the org in the cumulusci.yml which should be stored in the ScratchOrgConfig's dictionary.

jlantz:
Closes #441 by allowing projects to specify that certain scratch org configurations defined in cumulusci.yml should have the namespace applied.  The default still remains that all orgs are created without a namespace even if the sfdx-project.json has a namespace configured.  This is because in a full feature + beta test environment, you need orgs without the namespace applied to install a managed version.

Since most projects using CumulusCI were previously set up to not hard code namespaces in order to enable unmanaged feature builds, we keep this as the default behavior.  Individual projects could easily flip the default `dev` and `feature` orgs to use the namespace while other orgs remain non-namespaced with the following in their cumulusci.yml file:

```
orgs:
  scratch:
    dev:
      namespaced: True
    feature:
      namespaced: True
```

### DIFF
--- a/cumulusci/core/config.py
+++ b/cumulusci/core/config.py
@@ -802,11 +802,15 @@ class ScratchOrgConfig(OrgConfig):
         if self.devhub:
             devhub = ' --targetdevhubusername {}'.format(self.devhub)
 
+        namespaced = ''
+        if not self.namespaced:
+            namespaced = ' -n'
+
         # This feels a little dirty, but the use cases for extra args would mostly
         # work best with env vars
         extraargs = os.environ.get('SFDX_ORG_CREATE_ARGS', '')
-        command = 'sfdx force:org:create -f {}{} {}'.format(
-            self.config_file, devhub, extraargs)
+        command = 'sfdx force:org:create -f {}{}{} {}'.format(
+            self.config_file, devhub, namespaced, extraargs)
         self.logger.info(
             'Creating scratch org with command {}'.format(command))
         p = sarge.Command(command, stdout=sarge.Capture(buffer_size=-1))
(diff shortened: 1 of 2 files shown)


cumulusci/core/config.py
```diff
@@ -802,11 +802,15 @@ def create_org(self):
         if self.devhub:
             devhub = ' --targetdevhubusername {}'.format(self.devhub)
 
+        namespaced = ''
```
cpolcyn: Would be nicer as 1-liner:
`namespaced = '' if self.namespaced else ' -n'`

jlantz: I find that format hard to read.  It's similar to [x.y for x in foo] vs a normal for loop.  Perhaps that's just that I commonly forget the syntax for it.  I'd personally prefer to leave it this way as we use this pattern all over the place in cumulusci already but I'm open to discussion on that.

cpolcyn: I like doing it this way but certainly won't block the PR for it.

PR #444, opened 2017-09-19
cci org scratch should use config_name for org name by default

### LINKED ISSUE #436: cci org scratch should default to using the scratch config name
The most common use of `cci org scratch <config> <name>` is `cci org scratch dev dev` to create a scratch org using the `dev` config and name that org `dev`.  It makes sense that you'd also want to have similarly simple org names for the different scratch definitions, so `cci org scratch` should really default to using the config's name but allow the user to specify a different name if they want to.

Options:

1. Remove the <org_name> arg and replace with `--name <org_name>`.  This risks breaking anyone who uses the old format of the command.
2. Figure out how to accept org_name as an optional arg.  This is tricky with Click but should be possible.

jlantz:
Closes #436 

This branch changes `cci org scratch` from `cci org scratch <config_name> <org_name>` to `cci org scratch <config_name>` or optionally `cci org scratch <config_name> --org <org_name>`.  The default org name for scratch orgs is the config name unless you optionally override the name.

One concern is that the previous command `cci org scratch dev <org_name>` will now throw a usage error.  I think that's acceptable as any CI builds should be using the EnvironmentProjectKeychain instead of calling `cci org scratch` themselves.  The usage will show how to use the command and we can announce the breaking change in release notes and other channels.

The other concern is if existing users already have an org name `dev` and call the `cci org scratch dev` command without the `--org <org_name>` option, their existing `dev` org would be overwritten in the keychain possibly without their knowing.  To address this, `cci org scratch` and `cci org connect` now prompt for overwrite confirmation before overwriting an org and support the `--overwrite` option flag to bypass the prompt in headless scripts.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,12 @@
 History
 =======
 
+2.0.0-beta62 (2017-09-19)
+-------------------------
+
+* BREAKING CHANGE: The `cci org scratch` command has changed.  It used to be `cci org scratch <config_name> <org_name>` which often led most users to run the command `cci org scratch dev dev` which is redundant.  The new command format is `cci org scratch <config_name>` which will set the org name to the config name.  If you want to set a different name than the config name, you can do that with `cci org scratch <config_name> --org <org_name>`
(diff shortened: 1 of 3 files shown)


jlantz:
Closing this as we will likely implement differently via auto-creating scratch org configs in the project keychain

PR #447, opened 2017-09-21
Add cryptography to requirements

cpolcyn:


cpolcyn:
Already installed as dependency of requests[security]

PR #452, opened 2017-09-26
Auto Detection of CI environments

### LINKED ISSUE #451: Support environment variable injection of repo info in auto-detected known CI environments
Currently we rely on parsing the .git folder to get repo info. However, in a CI environment, there are environment variables set to provide much of the same information. CumulusCI should be able to detect specific environment variables from known CI environments and use those instead of parsing the .git files.

Heroku CI doesn't clone the repository and as a result there's no .git folder in /app where the repo commit being tested lives. Hence, it is the first proposed implementation for this CI environment auto-detection framework. Additional CI environments should be pretty simple to wire in as it's mostly setting values in a dictionary from environment variables in an if statement.

Heroku CI does set environment variables HEROKU_TEST_RUN_BRANCH and HEROKU_TEST_RUN_COMMIT but doesn't provide a way to get the repo url. However, we should also implement the following CumulusCI specific environment variables to allow manual override:

CUMULUSCI_REPO_BRANCH
CUMULUSCI_REPO_COMMIT
CUMULUSCI_REPO_URL
Then, in Heroku CI you would just need to define the CUMULUSCI_REPO_URL in the test environment which can even be done via the app.json file.

We should also provide warning output in the log file to show that we're using environment variables to override the repository information.

jlantz:
Fixes #451 by adding support for auto detecting CI system build environments via environment variables.  If a known CI environment is detected, its repository information will be injected into the project config instead of relying on parsing the files in the .git directory.

An initial implementation is included for Heroku CI which should allow Heroku CI to run cci builds easily.

### DIFF
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -19,6 +19,7 @@ from cumulusci.core.config import TaskConfig
 from cumulusci.core.config import YamlGlobalConfig
 from cumulusci.core.config import YamlProjectConfig
 from cumulusci.core.exceptions import ApexTestException
+from cumulusci.core.exceptions import ConfigError
 from cumulusci.core.exceptions import FlowNotFoundError
 from cumulusci.core.exceptions import KeychainConnectedAppNotFound
 from cumulusci.core.exceptions import KeychainKeyNotFound
@@ -73,6 +74,8 @@ class CliConfig(object):
             pass
         except NotInProject as e:
             raise click.UsageError(e.message)
+        except ConfigError as e:
+            raise click.UsageError('Config Error: {}'.format(e.message))
 
     def _load_keychain(self):
         self.keychain_key = os.environ.get('CUMULUSCI_KEY')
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cli.py
```diff
@@ -86,16 +89,35 @@ def _load_keychain(self):
                 self.project_config, self.keychain_key)
             self.project_config.set_keychain(self.keychain)
 
+def make_pass_instance_decorator(obj, ensure=False):
```
cpolcyn: `ensure` unused?

cumulusci/core/config.py
```diff
@@ -149,7 +149,119 @@ def config_global(self):
...
+                'ci': 'heroku',
+                'root': '/app',
+            }
+
+        # Other CI environment implementations can be implemented here...
+
+        # Apply CUMULUSCI_REPO_* environment variables last so they can
+        # override and fill in missing values from the CI environment
+        repo_branch = os.environ.get('CUMULUSCI_REPO_BRANCH')
+        if repo_branch:
```
cpolcyn: Could probably encapsulate this repeated code somehow.

cumulusci/core/config.py
```diff
@@ -149,7 +149,119 @@ def config_global(self):
...
+                # <key>, <env var to manually override>
+                ('branch', 'CUMULUSCI_REPO_BRANCH'),
+                ('commit', 'CUMULUSCI_REPO_COMMIT'),
+                ('name', 'CUMULUSCI_REPO_URL'),
+                ('owner', 'CUMULUSCI_REPO_URL'),
+                ('root', 'CUMULUSCI_REPO_ROOT'),
+                ('url', 'CUMULUSCI_REPO_URL'),
+            ))
+            for key, env_var in validate.items():
+                if key not in info or not info[key]:
```
cpolcyn: could be `if not info.get(key):`

PR #455, opened 2017-09-27
Check PR head branch before skipping

### LINKED ISSUE #454: github_master_to_feature skips parent branch when child branch has open PR
When checking if a branch has any open PRs against it in the master-to-feature merge task, only the base branch of the PR is checked. If there is a child branch with an open PR to its parent branch, that will prevent master being merged into the parent branch.

We need to also check that the source branch is master before skipping.

cpolcyn:
Fixes #454


force2b:
@cpolcyn Was this PR related to the issue I had with feature/rollups not being sync'd with Master?

cpolcyn:
@force2b yes your issue should be fixed now.

PR #462, opened 2017-09-29
Revert version check

cpolcyn:
The first time I hit the check it took a long time to finish:

```bash
$ time cci task list
…
real	1m21.941s
user	0m1.826s
sys	0m0.347s
```

Called again right afterward and got better results due to the caching:
```bash
real	0m2.496s
user	0m1.516s
sys	0m0.197s
```

I'd like to revert this until we can reduce the time it takes to check the version.

### DIFF
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -4,18 +4,8 @@ import sys
 import webbrowser
 import code
 import yaml
-import time
-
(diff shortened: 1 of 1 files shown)


cpolcyn:
Reverts https://github.com/SalesforceFoundation/CumulusCI/pull/450

PR #476, opened 2017-10-17
cci org * enhancements

### LINKED ISSUE #313: Use CCI org name as scratch org alias when creating scratch orgs
Sometimes I use sfdx directly to work with my scratch orgs, but that requires looking up the scratch org username using `cci org info` (or `sfdx force:org:list`) and copy/pasting the username as the `--targetusername` argument every time.

It would be awesome if CumulusCI passed along the org name to sfdx when creating the scratch org to use as the org alias (`sfdx force:org:create --setalias foo`).

Then I could use the same logical org name for both tools.

Example:

```
$ cci org scratch dev my-dev-org
$ cci org info my-dev-org # creates scratch org
$ sfdx force:data:soql:query --targetusername my-dev-org -q "SELECT Id FROM Contact"
```

### LINKED ISSUE #474: Auto-create scratch org configs in keychain
After the keychain orgs are loaded, we should auto-create a scratch org config in the keychain for each scratch config defined in the cumulusci.yml if an org by the config's name doesn't already exist in the keychain.  The main goal is to eliminate the need for calling `cci org scratch` for simple use cases and improve the onboarding experience.

CumulusCI comes preconfigured with 4 scratch org configs useful in the full project workflow:
* **dev**: A developer org
* **feature**: A feature branch test org used by CI to test feature commits
* **beta**: A beta release test org used by CI to test beta managed package versions
* **release**: A release test org used by CI to test production managed package versions

Projects can also define their own scratch orgs under `orgs -> scratch` in the project's cumulusci.yml file.

This change would automatically create all these scratch org configs for each project anytime the keychain is loaded (on every command).  Once created, the configs will persist in the keychain.

The auto-created scratch orgs are simply a lazy config object for creating a scratch org.  No actual scratch org gets created until one of the orgs is used in commands like `cci org info`, `cci task run`, etc.

Example use cases:
* In a new CumulusCI project, `cci org list` would show the 4 default scratch org configs.
* In a project with 3 additional scratch org configs defined in the project's cumulusci.yml file, the first time you run `cci org list` you would see 7 scratch orgs available for use: 4 from CumulusCI's default and 3 specific to the project
* For existing CumulusCI users, if you already have an org named `dev` and do not have a orgs named `beta`, `feature`, and `release`, this change will not touch your `dev` org but will add `beta`, `feature`, and `release` to your org list.

### LINKED ISSUE #475: cci org remove
Add command for completely removing an org from the keychain.  Currently the only way to do this is to delete the file from ~/.cumulusci/PROJECT_NAME/org_name.org

jlantz:
Fixes #474 
Fixes #475 
Fixes #313 

This branch implements a few enhancements to the workflow of using the `cci org` grouping of commands.  More detail is available in the linked issues above but a quick summary:

* `cci org remove` now exists to remove an org from the keychain
* CumulusCI now sets an alias in sfdx for all created scratch orgs using the naming format `ProjectName__orgname`
* Named scratch org configs are now automatically created in the keychain.  If an org with the same name already exists, the existing org is not changed.  CumulusCI includes 4 by default and projects can define their own.

### DIFF
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -56,11 +56,14 @@ def dbm_cache():
     context manager for accessing simple dbm cache
     located at ~/.cumlusci/cache.dbm
     """
-    db = dbm.open(os.path.join(
+    config_dir = os.path.join(
         os.path.expanduser('~'),
         YamlGlobalConfig.config_local_dir,
-        'cache.dbm'
-    ), 'c',)
+        
+    if not os.path.exists(config_dir):
+        os.mkdir(config_dir)
+
+    db = dbm.open(os.path.join(config_dir, 'cache.dbm'), 'c',)
     yield db
     db.close()
 
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -604,8 +604,6 @@ def org_list(config):
         row.append('*' if org_config.scratch else '')
         row.append(org_config.config_name if org_config.config_name else '')
         username = org_config.config.get('username', org_config.userinfo__preferred_username)
-        if not username:
-            username = org_config.userinfo__preferred_username
         row.append(username if username else '')
         data.append(row)
     table = Table(data, headers)
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -59,6 +59,7 @@ def dbm_cache():
     config_dir = os.path.join(
         os.path.expanduser('~'),
         YamlGlobalConfig.config_local_dir,
+    )
         
     if not os.path.exists(config_dir):
         os.mkdir(config_dir)
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -34,6 +34,7 @@ from cumulusci.core.exceptions import ConfigError
 from cumulusci.core.exceptions import FlowNotFoundError
 from cumulusci.core.exceptions import KeychainConnectedAppNotFound
 from cumulusci.core.exceptions import KeychainKeyNotFound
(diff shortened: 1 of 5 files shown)


cumulusci/cli/cli.py
```diff
@@ -591,16 +596,32 @@ def org_info(config, org_name):
...
-            data.append((org, '*'))
-        else:
-            data.append((org, ''))
+        row = [org]
+        row.append('*' if org_config.default else '')
+        row.append('*' if org_config.scratch else '')
+        row.append(org_config.config_name if org_config.config_name else '')
+        username = org_config.config.get('username')
+        if not username:
+            username = org_config.userinfo__preferred_username
```
cpolcyn: Could be 2nd arg to get() above

cumulusci/cli/cli.py
```diff
@@ -591,16 +596,32 @@ def org_info(config, org_name):
...
     click.echo(table)
 
+@click.command(name='remove', help="Removes an org from the keychain")
+@click.argument('org_name')
+@click.option('--global-org', is_flag=True, help="Set this option to force remove a global org.  Default behavior is to error if you attempt to delete a global org.")
+@pass_config
+def org_remove(config, org_name, global_org):
+    check_connected_app(config)
+    try:
+        org_config = config.keychain.get_org(org_name)
```
cpolcyn: Unused output var?

jlantz: I'm confused on this one... do you mean global_org isn't used?

cdcarter: org_config isn't used :)

PR #477, opened 2017-10-17
Feature/py3 core

cdcarter:
Initial pass of futurize on cumulusci/core, cumulusci/cli, and cumulusci/oauth.

NOTE: if you use this branch and python3 executable against an existing keychain, that keychain will be updated to a py3 pickle format and python2 executables won't be able to read it anymore.

### DIFF
--- a/cumulusci/core/config.py
+++ b/cumulusci/core/config.py
@@ -1,3 +1,5 @@
+from __future__ import unicode_literals
+from builtins import object
 import base64
 import datetime
 import json
@@ -102,7 +104,7 @@ class BaseTaskFlowConfig(BaseConfig):
     def list_tasks(self):
         """ Returns a list of task info dictionaries with keys 'name' and 'description' """
         tasks = []
-        for task in self.tasks.keys():
+        for task in list(self.tasks.keys()):
             task_info = self.tasks[task]
             if not task_info:
                 task_info = {}
@@ -221,7 +223,7 @@ class BaseProjectConfig(BaseTaskFlowConfig):
                 ('root', 'CUMULUSCI_REPO_ROOT'),
                 ('url', 'CUMULUSCI_REPO_URL'),
             ))
-            for key, env_var in validate.items():
+            for key, env_var in list(validate.items()):
                 if key not in info or not info[key]:
                     message = 'Detected CI on {} but could not determine the repo {}'.format(
                         info['ci'],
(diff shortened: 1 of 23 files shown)


cdcarter:
@cpolcyn i'm gonna go over the build failures that I didn't get locally, and then I'd love to grab time tomorrow or this week to go through this together a little carefully and make sure nothing seems wrong.

cdcarter:
Okay, so this branch contains a pile of changes to allow the core test suite to execute under both python2 and python3.6. There's a lot of work to be done to get the metadata client and all our tasks python3 friendly, but now you can manage your keychain and run *some* tasks in python3 as well as import cumulusci.core in python3 apps.

To get here, I relied pretty heavy on [future](http://python-future.org/) and friends.

First, I ran each stage of the `futurize` script.

[Stage 1 Fixes](http://python-future.org/futurize.html#stage-1-safe-fixes) were totally safe, and mostly fixed up some import syntax, some exception syntax, print function, etc...

[Stage 2 Fixes](http://python-future.org/futurize.html#stage-2-py3-style-code-with-wrappers-for-py2) added a dependency on the future package, and primarily fixed up our dictionary access. Dictionaries [work a bit different](http://python-future.org/compatible_idioms.html#dictionaries) in python3, but for the most part it boiled down to wrapping any calls to `dict.items()`, `dict.keys()`, and `dict.values()` in a `list()` to convert the View to a real iterable list.

Then, the next step was to fix our uses of strings and bytes to be python3 friendly. I decided [to import unicode_literals](http://python-future.org/imports.html#should-i-import-unicode-literals) in core, which took care of most of the changes needed.

Some manual handling of decode/encode was required in `EnvironmentProjectKeychain` to handle raw strings coming from the default Linux locale.

Reading/writing of encrypted keychain files needed to be changed to binary mode.

StringIO needed to be replaced with BytesIO when handling ZipFiles.

Import paths changed to be more python3 friendly. These imports are autofixed up by futurize wherever possible, and in try/except blocks where not.

cdcarter:
@cpolcyn - this is ready for review! a tiny bit early :)

cumulusci/core/config.py
```diff
@@ -102,7 +104,7 @@ class BaseTaskFlowConfig(BaseConfig):
     def list_tasks(self):
         """ Returns a list of task info dictionaries with keys 'name' and 'description' """
         tasks = []
-        for task in self.tasks.keys():
+        for task in list(self.tasks.keys()):
```
cpolcyn: Might be nicer to use `iteritems()` instead of `list()`?

cdcarter: iteritems is dead in py3

cdcarter:
it's merged. wow. one step closer! @jlantz if you run into any immediate issues as you're releasing for metaci we can rollback easily..

PR #479, opened 2017-10-18
Feature/namespaced org token

### LINKED ISSUE #478: Add support for new tokens in namespace tokenization logic to handle use cases of namespaced scratch orgs
We've always operated in a mostly binary world.  Your org is either unmanaged with no namespace or it has a managed version of the package installed (namespaced) and the rest of the metadata is not namespaced.  There's actually a third type of situation which applies to both packaging orgs and namespaced scratch orgs.  In this third situation, all unmanaged metadata in the org has the namespace prefix applied.

This gets challenging if, for example, you use an unpackaged/post folder to tell CumulusCI to deploy some record types and then deploy Profiles which reference those record types.  The problem is that in this third type of org, this "unmanaged" metadata has the namespace prefix applied to it.

The proposed solution is:
* Add a new option, `namespaced_org`, to all tasks with the `namespace_inject` option.  The default would be False which would be backwards compatible with previous functionality.  
* If `namespaced_org` is set to True, all instances of the token `%%%NAMESPACED_ORG%%%` in file contents and `___NAMESPACED_ORG___` in file names will get replaced with the namespace prefix (i.e. yournamespace__).  
* If `namespaced_org` is False, these new tokens are simply stripped and replaced with an empty string.

This allows tokenized metadata to handle the third org situation where everything is automatically namespace prefixed.

jlantz:
Fixes #478 

Details on the change are in the issue.  This should be backwards compatible with previous functionality and thus a purely additive, opt in feature.

### DIFF
--- a/cumulusci/utils.py
+++ b/cumulusci/utils.py
@@ -121,7 +121,7 @@ def zip_subfolder(zip_src, path):
     return zip_dest
 
 
-def zip_inject_namespace(zip_src, namespace=None, managed=None, filename_token=None, namespace_token=None):
+def zip_inject_namespace(zip_src, namespace=None, managed=None, filename_token=None, namespace_token=None, namespaced_org=None):
     """ Replaces %%%NAMESPACE%%% for all files and ___NAMESPACE___ in all 
         filenames in the zip with the either '' if no namespace is provided
         or 'namespace__' if provided.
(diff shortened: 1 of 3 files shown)


cumulusci/utils.py
```diff
@@ -135,10 +135,17 @@ def zip_inject_namespace(zip_src, namespace=None, managed=None, filename_token=N
     else:
         namespace = ''
 
+    # Handle tokens %%%NAMESPACED_ORG%%% and ___NAMESPACED_ORG___
+    namespaced_org_token = '%%%NAMESPACED_ORG%%%'
+    namespaced_org_file_token = '___NAMESPACED_ORG___'
```
cdcarter: make overrideable in cumulusci.yml?

jlantz: I thought about that... for NAMESPACE I could see a use case where you want to inject multiple namespaces via different tasks.  However, for NAMESPACED_ORG, you'd only ever want one since an org can only ever have one namespace.

PR #487, opened 2017-10-27
Enhancements to cci project init

jlantz:
I had previously left `cci project init` alone in favor of using a cookiecutter template to launch the project config.  The problem is that we don't have as much control over the process with cookiecutter, cookiecutter is another dependency, and cookiecutter can't run against an existing directory.

The two most common use cases we really want to handle are:
1. You're starting a new project
2. You have a Github repo with metadata under `src/` and you want to enable that repo for CumulusCI

Cookiecutter won't help in use case #2.  This new task will run against any local git repository that doesn't already have a `cumulusci.yml` file and configures the following:

* `cumulusci.yml`
* `sfdx-project.json` if none exists already, otherwise leaves existing file alone
* `orgs/*.json` Starter files for defining org shapes

### DIFF
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -317,8 +317,9 @@ cli.add_command(service)
 @click.command(name='init',
                help="Initialize a new project for use with the cumulusci toolbelt",
                )
+@click.option('--extend', help="If set to the url of another Github repository configured for CumulusCI, creates this project as an extension which depends on the other Github project and all its dependencies")
 @pass_config
-def project_init(config):
+def project_init(config, extend):
     if not os.path.isdir('.git'):
         raise click.ClickException("You are not in the root of a Git repository")
 
@@ -363,6 +364,10 @@ def project_init(config):
     if package_config:
         yml_config.append('    package:')
         yml_config.extend(package_config)
+
+    if extend:
+        yml_config.append('    dependencies:')
+        yml_config.append('        github: {}'.format(extend))
 
     #     git:
     git_config = []
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -101,11 +101,6 @@ def check_latest_version():
             click.echo(
                 "An update to CumulusCI is available. Use pip install --upgrade cumulusci to update.")
 
-def pretty_dict(data):
-    if not data:
-        return ''
-    return json.dumps(data, sort_keys=True, indent=4, separators=(',', ': '))
-
 def handle_exception_debug(config, debug, e, throw_exception=None, no_prompt=None):
     if debug:
         import pdb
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cli.py
```diff
@@ -305,115 +317,153 @@ def service(config):
...
-        package_config.append(
-            '        namespace: {}'.format(package_namespace))
+
+    click.echo()
+    package_namespace = None
+    if click.confirm(click.style("Is this a managed package project?", bold=True), default=False):
+        click.echo('Enter the namespace assigned to the managed package for this project')
+        package_name = click.prompt(click.style('Package Namespace', bold=True), default=name)
+        if package_namespace and package_namespace != config.global_config.project__package__namespace:
+            package_config.append(
```
cdcarter: I have to ask if theres a way to build this config in memory as a dict and just dump to yaml?

cdcarter: i know the yaml strings are't new but...

cumulusci/cli/cli.py
```diff
@@ -305,115 +317,153 @@ def service(config):
...
+        "namespace": package_namespace,
+        "sourceApiVersion": package_api_version,
+}
+    with open('sfdx-project.json','w') as f:
+        f.write(json.dumps(sfdx_project))
+
+    if not os.path.isdir('orgs'):
+        os.mkdir('orgs')
+      
+    org_content_url = 'https://raw.githubusercontent.com/SalesforceFoundation/sfdo-package-cookiecutter/master/%7B%7Bcookiecutter.project_name%7D%7D/orgs/{}.json'
```
cdcarter: this...worries me...

I know it was simple but maybe lets just include these files in cci distribution?

cumulusci/cumulusci.yml
```diff
@@ -462,8 +462,20 @@ services:
...
+        required: False
+      url:
+        description: The main url to the MetaCI site
+        required: True
+      token:
+        description: Your API token to the MetaCI site (get from SITE_URL/api/token)
+        required: True
   mrbelvedere:
-    description: Configure connection for mrbelvedere tasks, e.g. MrbelvederePublish
+    descript: Configure connection for mrbelvedere tasks, e.g. MrbelvederePublish
```
cdcarter: description? descript?

jlantz: Oops, fixed

cumulusci/cli/cli.py
jlantz: The yaml dump doesn't respect ordereddict and even the ordereddict dumper in HiYaPyCo which we have available doesn't handle recursive dictionaries very well.  I do want to revisit this at some point, probably with building a model of some sort for the cci config and then being able to render that as json, yaml, etc.   We could use that model and its validation in a bunch of places.

cumulusci/cli/cli.py
jlantz: Probably something we should do eventually but can be done in a future commit.  This definitely works for now at least.

cumulusci/cli/cli.py
cdcarter: marshmallow!

PR #493, opened 2017-11-01
Add Windows sections to tutorial

cpolcyn:
Continuation of https://github.com/SalesforceFoundation/CumulusCI/pull/492

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -31,7 +31,7 @@ Windows
 
 #. Install Python 2: https://www.python.org/downloads/release/python-2714/
 
-#. Update Environment Path to include Python 2.7 install folders (C:\Python27 and C:\Python27\Scripts)
+#. Update Environment Path to include Python 2.7 install folders (C:\\Python27 and C:\\Python27\Scripts)
 
 Create Virtual Environment
 --------------------------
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -2,14 +2,19 @@
 Tutorial
 ========
 
-This tutorial is for macOS. Linux and Windows are not yet officially supported but should work for the most part.
+This tutorial is for macOS and Windows. Linux is not yet officially supported but should work for the most part.
```
cdcarter: i don't think we should say Windows is "officially supported" since we don't actually support anybody on the team on win and we don't have a test VM yet.

cpolcyn:
Thanks for the suggestion @cdcarter – I clarified that Windows is still not officially supported. Please have another look.

docs/tutorial.rst
```diff
@@ -21,23 +26,40 @@ Due to an issue regarding TLS support in the Python included in macOS it is nece
 
     $ brew install python
 
+Windows
+^^^^^^^
+
+#. Install Python 2: https://www.python.org/downloads/release/python-2714/
+
+#. Update Environment Path to include Python root and Scripts folder.  
+  **Update system path to include Python 2.7 install folders (C:\Python27 and C:\Python27\Scripts)
```
daveb-501commons: @cdcarter 
When viewing the tutorial.rst it views as 
**Update system path to include Python 2.7 install folders (C:Python27 and C:Python27\Scripts)

The backslash after C: is missing \ when viewing the rst in a browser.

cpolcyn:
Good catch @daveb-501commons – just fixed it

PR #503, opened 2017-11-06
pass on unicode decode failures during tokenization

### LINKED ISSUE #489: namespace injection fails if a binary file is in the dir
See https://gist.github.com/jcschultz/87e7d5189ef9f31ae8dcbb1a3bda4092 for a traceback.

zip_inject_namespace tries to read each file in the given tree, and replace the namespace token with a namespace. 

this fails if the file cannot be safely decoded to ascii.

some options:

- catch `UnicodeDecodeError` and pass
- check the file type from the zip listing and skip binary
- whitelist the types of files that SHOULD receive namespace injection that we can guarantee we can decode.

cdcarter:
closes #489

### DIFF
--- a/cumulusci/utils.py
+++ b/cumulusci/utils.py
@@ -95,7 +95,7 @@ def removeXmlElement(name, directory, file_pattern, logger=None):
 
 def download_extract_zip(url, target=None, subfolder=None):
     resp = requests.get(url)
-    zip_content = io.StringIO(resp.content)
+    zip_content = io.BytesIO(resp.content)
     zip_file = zipfile.ZipFile(zip_content)
     if subfolder:
         zip_file = zip_subfolder(zip_file, subfolder)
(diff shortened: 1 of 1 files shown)


cumulusci/utils.py
```diff
@@ -146,11 +146,15 @@ def zip_inject_namespace(zip_src, namespace=None, managed=None, filename_token=N
     zip_dest = zipfile.ZipFile(io.BytesIO(), 'w', zipfile.ZIP_DEFLATED)
 
     for name in zip_src.namelist():
-        content = zip_src.read(name).replace(namespace_token, namespace)
-        content = content.replace(namespaced_org_token, namespaced_org)
-        name = name.replace(filename_token, namespace)
-        name = name.replace(namespaced_org_file_token, namespaced_org)
-        zip_dest.writestr(name, content)
+        try:
+            content = zip_src.read(name).replace(namespace_token, namespace)
```
cpolcyn: Could we limit the number of statements in the try block to just what is needed? In other words, which statement(s) do we actually expect to raise this? Seems it would only be the `zip_src.read(name)`.

Szandor72:
would love to see this merged - currently a show stopper for us :(

cpolcyn:
@Szandor72 this has been released in v2.0.0b76

PR #510, opened 2017-11-16
Get QA org ids from MetaCI

cdcarter:
requires the MetaCI OrgMart PR to merge.

usage (for cumulus) is
```
cci service connect metaci
cci task run get_org_list -o repo_id 4 -o output_file qa_push.txt
cat qa_push.txt
```

Very barebones because I don't actually know what you need or want in your qa push flow, chris. but. here's a start!


cpolcyn:
Shouldn't this be in MetaCI CLI instead?

cdcarter:
closing this, will re-open in metaci-cli

PR #511, opened 2017-11-22
 Add support for setting durationdays on scratch orgs and tracking org expiration

jlantz:
- Scratch org configs in cumulusci.yml can now specify a days attribute that sets the number of days the scratch org should persist (1-30)
- `cci org scratch` now supports the `--days N` option to override the days value on the scratch config
- Scratch orgs now have `date_created` stored in their config
- Scratch orgs now have new calculated properties: `expired`, `expires`, `days_alive`
- `cci org list` now shows days and if orgs are expired
- All cci commands that interact with an org from the keychain now check if the org is expired.  If the org is expired, the user is prompted to recreate the org config to continue

### DIFF
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -113,7 +113,6 @@ def check_org_expired(config, org_name, org_config):
             config.keychain.create_scratch_org(
                 org_name,
                 org_config.config_name,
-                {'namespaced': org_config.namespaced},
                 org_config.days,
             )
             click.echo('Org config was refreshed, attempting to recreate scratch org')
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -116,8 +116,8 @@ def check_org_expired(config, org_name, org_config):
                 org_config.days,
             )
             click.echo('Org config was refreshed, attempting to recreate scratch org')
-            org = config.keychain.get_org(org_name)
-            org.create_org()
+            org_config = config.keychain.get_org(org_name)
+            org_config.create_org()
         else:
             raise click.ClickException('The target scratch org is expired.  You can use cci org remove {} to remove the org and then recreate the config manually'.format(org_name))
 
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -744,7 +744,7 @@ def org_scratch(config, config_name, org_name, default, delete, devhub, days):
     if devhub:
         scratch_config['devhub'] = devhub
 
-    config.keychain.create_scratch_org(org_name, config_name, scratch_config, days)
+    config.keychain.create_scratch_org(org_name, config_name, days)
 
     if default:
         org = config.keychain.set_default_org(org_name)
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -107,22 +107,20 @@ def check_latest_version():
                 "An update to CumulusCI is available. Use pip install --upgrade cumulusci to update.")
 
 def check_org_expired(config, org_name, org_config):
-    if org_config.scratch and org_config.expires and org_config.expired:
+    if org_config.scratch and org_config.date_created and org_config.expired:
         click.echo(click.style('The scratch org is expired', fg='yellow'))
         if click.confirm('Attempt to recreate the scratch org?', default=True):
-            config.keychain.set_org(
+            config.keychain.create_scratch_org(
                 org_name,
-                config.keychain.create_scratch_org(
-                    org_name,
-                    org_config.config_name,
-                    {'namespaced': org_config.namespaced},
-                    org_config.days,
-                )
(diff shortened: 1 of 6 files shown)


cumulusci/cli/cli.py
```diff
@@ -106,6 +106,23 @@ def check_latest_version():
...
+                org_config.config_name,
+                org_config.days,
+            )
+            click.echo('Org config was refreshed, attempting to recreate scratch org')
+            org = config.keychain.get_org(org_name)
+            org.create_org()
+        else:
+            raise click.ClickException('The target scratch org is expired.  You can use cci org remove {} to remove the org and then recreate the config manually'.format(org_name))
+
+    return org_config
```
cdcarter: is there a reason this is called in an `org_config = check_org_expired(config, org_name, org_config)` kinda way?

if it's not intended to mutate the state of the org_config var, lets not give it that power. can we just not return a value?

```
org_config = config.project_config.get_org(org_name)
check_org_expired(config, org_name, org_config)
org_config.refresh_oauth_token(config.keychain.get_connected_app())
```
instead of 
```
org_config = config.project_config.get_org(org_name)
org_config = check_org_expired(config, org_name, org_config)
org_config.refresh_oauth_token(config.keychain.get_connected_app())
```

cumulusci/cli/cli.py
```diff
@@ -659,6 +678,9 @@ def org_info(config, org_name, print_json):
     else:
         click.echo(render_recursive(org_config.config))
 
+    if org_config.scratch and org_config.expires:
+        click.echo('Org expires on {}'.format(org_config.expires.strftime('%c')))
```
cdcarter: click.echo('Org expires on {:%c}'.format(org_config.expires))

cumulusci/core/config.py
```diff
@@ -930,6 +930,25 @@ def password(self):
             password = self.scratch_info['password']
         return password
 
+    @property
+    def days(self):
+        return self.config.get('days', 7)
```
cdcarter: i'd do `self.config.setdefault('days',7)` just to be extra safe, in case anybody tries to do `if days in self.config`

cumulusci/core/config.py
```diff
@@ -946,15 +965,19 @@ def create_org(self):
...
         if self.sfdx_alias:
-            alias = ' -a "{}"'.format(self.sfdx_alias)
+            alias = ' -a {}'.format(self.sfdx_alias)
 
         # This feels a little dirty, but the use cases for extra args would mostly
         # work best with env vars
         extraargs = os.environ.get('SFDX_ORG_CREATE_ARGS', '')
-        command = 'sfdx force:org:create -f {}{}{}{} {}'.format(
-            self.config_file, devhub, namespaced, alias, extraargs)
+        command = 'sfdx force:org:create -f {}{}{}{}{} {}'.format(
```
cdcarter: starting to wonder if we should name the fields in the format string...

cumulusci/cumulusci.yml
```diff
@@ -57,6 +57,21 @@ tasks:
...
+    dx_convert_from:
+        description: Converts force-app directory in sfdx format into metadata format under src
+        class_path: cumulusci.tasks.sfdx.SFDXBaseTask
+        options:
+            command: 'force:mdapi:convert -r force-app -d src'
+    dx_push:
+        description: Uses sfdx to push the force-app directory metadata into a scratch org
+        class_path: cumulusci.tasks.sfdx.SFDXOrgTask
+        options:
+            command: 'force:source:push'
```
cdcarter: remove these tasks from PR

cumulusci/cumulusci.yml
```diff
@@ -564,9 +579,13 @@ orgs:
     scratch:
         dev:
             config_file: orgs/dev.json
+            days: 14
```
cdcarter: I think 7 is still an appropriate default, unless we've heard from devs they want longer lasting ones.

cumulusci/cli/cli.py
jlantz: There was a bug in that check_org_expired wasn't returning the newly built org_config.  We want it return the new org_config so the click command function doesn't need to query the keychain again for the org.  I'll push a fix shortly.

cumulusci/cli/cli.py
jlantz: Nice :)

cumulusci/core/config.py
jlantz: I've never used setdefault but it makes sense to me.

cumulusci/cumulusci.yml
jlantz: I changed the default for dev to 7 and the global default from 7 to 1 for all orgs.

cumulusci/core/config.py
jlantz: Yep, changed in a new commit

PR #512, opened 2017-11-22
Avoid orphan scratch orgs from flow and task run

jlantz:
This branch changes flows and tasks so they will always write a modified org config back to the keychain.  This helps avoid the scenario where `cci flow run` or `cci task run` creates a scratch org, fails, then fails to write the scratch org info into cci keychain which effectively creates an orphaned scratch org.

### DIFF
--- a/cumulusci/cli/cli.py
+++ b/cumulusci/cli/cli.py
@@ -929,10 +929,6 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
         except Exception as e:
             handle_exception_debug(config, debug, e, no_prompt=no_prompt)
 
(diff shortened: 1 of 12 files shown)


cumulusci/cli/cli.py
```diff
@@ -137,6 +137,16 @@ def render_recursive(data, indent=None):
             else:
                 click.echo('{}{} {}'.format(indent_str, key_str, value))
 
+def check_org_overwrite(config, org_name):
+    try:
+        org = config.keychain.get_org(org_name)
+        raise click.ClickException(
```
cdcarter: can we prompt the user if they want to delete and move forward. i usually do want to...

PR #531, opened 2018-01-09
filter sobjects by record type in bulkdataretrieve

### LINKED ISSUE #525: Bulk Data QueryData task doesn't filter on Record Type
Using SAL as my base repo and adding a block to the mapping file like:
```
Insert Business Organization Accounts:
    sf_object: Account
    table: business_organizations
    fields:
        Id: sf_id
        Name: name
    record_type: Business_Organization
```
The query still looks like ` Submitting query: SELECT Id, Name FROM Account`

cdcarter:
closes #525 - if a mapping defines a record_type, when retrieving that sobject, filter by that record type.

additionally fixes a bug where a second mapping on the same sqlite table would not properly add newly mapped fields to the table in retrieve mode.

adds an `api` flag on the mapping, currently doesn't do anything, i can roll that back if you want a clean merge, but eventually it will support loading objects that bulk doesn't.

### DIFF
--- a/docs/bulk_data.rst
+++ b/docs/bulk_data.rst
@@ -40,6 +40,7 @@ Mapping File
             FieldAPIName: True
             AnotherFieldAPIName: MonthlyLiteral
 
+
 Example
 ^^^^^^^
 
@@ -77,3 +78,4 @@ Example
                 table: organizations
                 join_field: organization_id
                 value_field: sf_id
+
--- a/docs/bulk_data.rst
+++ b/docs/bulk_data.rst
@@ -11,6 +11,9 @@ Runs the mapping YAML in order, from top to bottom, selecting data from the spec
 sf_object and inserting it into the local table.
 
 LoadData
+^^^^^^^^
+Runs the mapping YAML in order, selecting data from the local table and inserting into the
+specified sf_object 
 
 
 Mapping File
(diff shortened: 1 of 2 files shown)


docs/bulk_data.rst
```diff
@@ -0,0 +1,62 @@
+=========
+Bulk Data
+=========
+
+Mapping File
+===========
+
+Step Name: (must be unique)
```
jlantz: This should be wrapped in `.. code-block:: yaml` with all sub lines indented.  See the features.rst file for examples.

docs/bulk_data.rst
```diff
@@ -0,0 +1,62 @@
...
+    api: [(bulk)|sobject]
```
jlantz: Add a note here that sobject isn't yet implemented but you can use this to start stubbing out your yaml files for when it is?

docs/bulk_data.rst
```diff
@@ -0,0 +1,83 @@
...
+
+.. code-block:: yaml
+
+    Step Name: (must be unique)
+        api: [(bulk)|sobject] (not yet implemented, specify which API to use to load data)
+        sf_object: API Name for sfdc object
+        table: full table name in sqlite3
+        filters: used to filter the sqlite3 table when loading data
+            - 'sql string' 
+        record_type: API Name (for insert/query)
```
davidjray: @cdcarter isn't this actually the developer name?

PR #546, opened 2018-02-01
Feature/salesforce bulk 2 0 0

### LINKED ISSUE #545: update salesforce-bulk dependency
update salesforce-bulk package to 2.0.0 (from 1.1.0). breaking changes to the python api but it is py3 compatible and is the official version now.

cdcarter:
closes #545 by updating to salesforce-bulk 2.0.0 -- tested with NPSP and SAL dev_data set.

### DIFF
--- a/cumulusci/tasks/bulkdata.py
+++ b/cumulusci/tasks/bulkdata.py
@@ -139,6 +139,7 @@ class LoadData(BaseSalesforceApiTask):
     def _run_task(self):
         self._init_mapping()
         self._init_db()
+        self.logger.debug('here')
 
         for name, mapping in self.mapping.items():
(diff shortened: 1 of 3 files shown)


cumulusci/tasks/bulkdata.py
```diff
@@ -139,6 +141,7 @@ class LoadData(BaseSalesforceApiTask):
     def _run_task(self):
         self._init_mapping()
         self._init_db()
+        self.logger.debug('here')
```
jlantz: Should this be here?

cdcarter: nope. and was gone at one point but merged its way back in...

PR #558, opened 2018-02-21
Only run meta.xml file cleaning on classes/* and triggers/* directory

jlantz:
meta.xml files


e02d96ec16:
Wouldn't `pages/*` also qualify for this cleansing?

PR #569, opened 2018-03-12
Initial Integration with Robot Framework

jlantz:
This branch represents the initial framework and proof of concept for integrating Robot Framework with CumulusCI.  See docs/robotframework.rst for more information about the integration.


cumulusci/robotframework/CumulusCI.py
```diff
@@ -0,0 +1,160 @@
...
+    ROBOT_LIBRARY_SCOPE = "GLOBAL"
+
+    def __init__(self, org_name=None):
+        if not org_name:
+            org_name = 'dev'
+        self.org_name = org_name
+        self.sf = self._init_api()
+        self.tooling = self._init_api('tooling/')
+        # Turn off info logging of all http requests 
+        logging.getLogger('requests.packages.urllib3.connectionpool').setLevel(logging.WARN)
```
cdcarter: i don't love this global side effect in this class, and I really really don't love it happening in the initializer. can this be handled in the logging setup for cci?

cumulusci/robotframework/CumulusCI.py
```diff
@@ -0,0 +1,160 @@
...
+        task_config = self._parse_task_options(options, task_class, task_config)
+        return task_class, task_config
+
+    def _parse_task_options(self, options, task_class, task_config):
+        # Parse options and add to task config
+        if options:
+            if 'options' not in task_config:
+                task_config['options'] = {}
+            for name, value in options.items():
+                # Validate the option
```
cdcarter: little surprised to see that BaseTask doesn't do this validation already. consider moving it there?

cumulusci/robotframework/Salesforce.py
```diff
@@ -0,0 +1,477 @@
...
+            except SalesforceMalformedRequest as e:
+                self.builtin.log('    {type} {id} could not be deleted:'.format(**record), level='WARN')
+                self.builtin.log('      {}'.format(e), level='WARN')
+        
+    def get_current_record_id(self):
+        """ Parses the current url to get the object id of the current record.
+            Expects url format like: [a-zA-Z0-9]{15,18}
+        """
+        url = self.selenium.get_location()
+        for part in url.split('/'):
```
cdcarter: for vf pages it could also be in a query string parameter

cumulusci/robotframework/Salesforce.py
```diff
@@ -0,0 +1,477 @@
...
+        if filter_name:
+            url += '?filterName={}'.format(filter_name)
+        self.selenium.go_to(url)
+        time.sleep(5)
+        self._wait_until_loading_is_complete()
+
+    def go_to_record_home(self, obj_id, filter_name=None):
+        """ Navigates to the Home view of a Salesforce Object """
+        url = self.cumulusci.org.lightning_base_url
+        url = '{}/one/one.app#/sObject/{}/view'.format(url, obj_id)
```
cdcarter: can we pull all the different URL strings we're using into module constants? url v2 compatibilit is also something we need to think about.

cumulusci/robotframework/Salesforce.py
```diff
@@ -0,0 +1,477 @@
...
+
+    def populate_lookup_field(self, name, value):
+        self.populate_field(name, value)
+        locator = locators['object']['field_lookup_value'].format(value)
+        self._call_selenium('_populate_lookup_field', True, locator)
+    
+    def _populate_lookup_field(self, locator):
+        self.selenium.set_focus_to_element(locator)
+        self.selenium.get_webelement(locator).click()
+        time.sleep(.5)
```
cdcarter: can we pull out `SLEEP_SHORT=0.5` and `SLEEP_LONG=5` as module constants and use in all method?

cumulusci/robotframework/Salesforce.py
```diff
@@ -0,0 +1,477 @@
...
+            timeout=15,
+        )
+        time.sleep(3)
+
+    def wait_until_loading_is_complete(self):
+        """ EXPERIMENTAL!!! """
+        self._call_selenium('_wait_until_loading_is_complete', True)
+
+    def _wait_until_loading_is_complete(self):
+        self.selenium.wait_until_element_is_not_visible(
```
cdcarter: are these locators we can put in the locators.py file? it seems weird that they're here. idk much about robot.

cumulusci/robotframework/CumulusCI.py
jlantz: It's only needed in the context of robot.  Without this line, robot outputs every HTTPS call made which pollutes the logs.  cci never seemed to have this issue, perhaps because its logger doesn't log anything from other packages.

cumulusci/robotframework/Salesforce.py
jlantz: Don't have to know much about robot to notice that this doesn't make sense... I was just lazy :)

cumulusci/robotframework/Salesforce.py
jlantz: I've been thinking about having a library init option for LEX url format version and handling it that way.  We'd need to build up a test suite that could validate that.

cumulusci/robotframework/Salesforce.py
jlantz: Looking through this, we're trying to keep these adjusted based on the performance of the individual method.  There are too many possible sleep values already to justify storing them all as separate variables.  Since the test is tied to the keyword's method, an unstable test tells us we might want to adjust that individual sleep.

Also note, sleeping is an interim hack until we can figure out how to better tap into component states via $A.

PR #598, opened 2018-04-06
Switch interactive task feature name from use_std to interactive.

topherlandry:
Sorry, I don't have an active way to update the PR in question. But this incorporates the change requested in the PR to master.


cpolcyn:
No problem, this is fine. We merge to a feature branch so that it can be tested by our CI system before merging to master.

PR #602, opened 2018-04-13
Make Command tasks work on Windows

### LINKED ISSUE #379: Command class does not work on Windows
In command.py/Command we specify to Popen the arg `executable='/bin/bash'`. This makes Command unusable for systems that do not have bash at that location. Can we rewrite this to take advantage of Popen's cross-platform nature?

cpolcyn:
Fixes #379 

As far as I know there's no reason we need to specify /bin/bash in the Popen command. Removing this allows Windows to use the command task.


cdcarter:
i'm assuming this runs locally?

PR #607, opened 2018-04-19
support nesting flows

cpolcyn:
Allow flows to call other flows. Example usage in `cumulusci.yml`:

```
flows:
    example_flow:
        tasks:
            1:
                task: example_task
            2:
                flow: another_flow
```

- A given flow may only occur once in the tree, to prevent infinite recursion
- Nested flows inherit the project config, org config, flow options, and skipped tasks from parent flows
- Return values from nested flows can be referenced just like tasks: flow1.flow2.task.key

### DIFF
--- a/cumulusci/core/flows.py
+++ b/cumulusci/core/flows.py
@@ -175,7 +175,12 @@ class BaseFlow(object):
             self._run_task(flow_task_config)
 
     def _run_flow(self, flow_task_config):
-        flow = BaseFlow(
+        class_path = flow_task_config['task_config'].config.get(
+            'class_path',
+            'cumulusci.core.flows.BaseFlow',
+        )
+        flow_class = import_class(class_path)
+        flow = flow_class(
             self.project_config,
             flow_task_config['task_config'],
             self.org_config,
--- a/cumulusci/core/flows.py
+++ b/cumulusci/core/flows.py
@@ -25,13 +25,10 @@ class BaseFlow(object):
         self.skip = skip
         self.task_options = {}
         self.skip_tasks = []
-        self.task_return_values = []
-        """ A collection of return_values dicts in task execution order """
-        self.task_results = []
-        """ A collection of result objects in task execution order """
-        self.tasks = []
-        """ A collection of configured task objects, either run or failed """
-        self.nested = nested
+        self.task_return_values = []  # A collection of return_values dicts in task execution order
+        self.task_results = []  # A collection of result objects in task execution order
+        self.tasks = []  # A collection of configured task objects, either run or failed
+        self.nested = nested  # indicates if flow is called from another flow
         self._init_options()
         self._init_skip(skip)
         self._init_logger()
--- a/cumulusci/core/flows.py
+++ b/cumulusci/core/flows.py
@@ -100,6 +100,7 @@ class BaseFlow(object):
                 raise ConfigError('"flow" and "task" in same config item')
             if (('flow' in config and config['flow'] == 'None') or
                 ('task' in config and config['task'] == 'None')):
+                # allows skipping flows/tasks using YAML overrides
                 continue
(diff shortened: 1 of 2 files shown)


cpolcyn:
Thinking about it some more, I think that task options should not be passed through to nested flows. They are particular to a task within a given flow and not meant to be applied to the flow as a whole.

cumulusci/core/flows.py
```diff
@@ -27,15 +31,16 @@ def __init__(self, project_config, flow_config, org_config, options=None, skip=N
         """ A collection of result objects in task execution order """
         self.tasks = []
         """ A collection of configured task objects, either run or failed """
-        self._init_options(options)
+        self.nested = nested
```
cdcarter: can you add a docstring? i'm  pretty sure this indicates we are on an inner flow?

cumulusci/core/flows.py
```diff
@@ -72,26 +77,46 @@ def _init_flow(self):
...
+        if not self.nested:
+            self._check_infinite_flows(self.flow_config.tasks)
         tasks = []
         for step_num, config in list(self.flow_config.tasks.items()):
-            if config['task'] == 'None':
+            if 'flow' in config and 'task' in config:
+                raise ConfigError('"flow" and "task" in same config item')
+            if (('flow' in config and config['flow'] == 'None') or
+                ('task' in config and config['task'] == 'None')):
                 continue
```
cdcarter: we support no-op tasks?

cumulusci/core/flows.py
```diff
@@ -72,26 +77,46 @@ def _init_flow(self):
...
-                    'task_config': self.project_config.get_task(
-                        config['task']
-                    ),
+                    'task_config': task_config,
                 }
             ))
-        tasks.sort()
-        tasks = [task_info[1] for task_info in tasks]
+        tasks.sort()  # sort by step number
+        tasks = [task[1] for task in tasks]  # drop the sort keys
```
cdcarter: thank you for these clarifying comments, this was magic wasn't it...

cumulusci/core/flows.py
```diff
@@ -167,8 +224,15 @@ def _run_task(self, flow_task_config):
             if str(value).startswith('^^'):
                 value_parts = value[2:].split('.')
                 parent = self._find_task_by_name(value_parts[0])
-                for attr in value_parts[1:]:
-                    parent = parent.return_values.get(attr)
+                n = 0
+                while isinstance(parent, BaseFlow):
+                    n += 1
+                    parent = parent._find_task_by_name(value_parts[n])
+                for attr in value_parts[(n + 1):]:
```
cdcarter: this took me a second to figure out but is really smart. nice.

cumulusci/core/flows.py
cpolcyn: Sure, you mean a docstring for `__init__`?

cumulusci/core/flows.py
cpolcyn: Hmm good catch, I suppose this should raise a config exception?

cpolcyn: On second thought this keeps the functionality we had before (see deleted line 82) so I'll leave it like this. I think the idea is to no-op when overriding a task in another flow, like using a deploy+test flow but skipping tests.

cdcarter: that makes a ton of sense. can you add a comment that this is to allow overrides in yaml?

cumulusci/core/flows.py
cdcarter: just like above tasks or options

see http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html and cmd-f for "self.attr5 = None" to see how it looks i guess

cpolcyn:
@cdcarter I addressed your comments with some new commits. Please re-review.

PR #612, opened 2018-04-26
add task to report info from GitHub release notes

cpolcyn:
`cci task info github_release_report`

### DIFF
--- a/cumulusci/tasks/github/release_report.py
+++ b/cumulusci/tasks/github/release_report.py
@@ -1,4 +1,3 @@
-from datetime import datetime
 import json
 import re
 
@@ -43,11 +42,12 @@ class ReleaseReport(BaseGithubTask):
         releases = []
         last_time = None
         repo = self.get_repo()
-        regex_compiled = re.compile(
-            r'^((?P<sandbox>{})|(?P<production>{}))\s*(?P<date>\d\d\d\d-\d\d-\d\d)'.format(
+        regex_compiled_prefix = re.compile(
+            r'^((?P<sandbox>{})|(?P<production>{}))(?P<remaining>.*)$'.format(
                 self.project_config.project__git__push_prefix_sandbox,
                 self.project_config.project__git__push_prefix_production,
             ))
+        regex_compiled_date = re.compile(r'\s*(?P<date>\d\d\d\d-\d\d-\d\d)')
         for release in repo.iter_releases():
             if release.prerelease and not self.options['include_beta']:
                 continue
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/github/release_report.py
```diff
@@ -0,0 +1,83 @@
...
+    @staticmethod
+    def _parse_datetime(dt_str):
+        t = time.strptime(dt_str, '%Y-%m-%d')
+        return datetime(t[0], t[1], t[2], t[3], t[4], t[5], t[6], pytz.UTC)
+
+    def _run_task(self):
+        releases = []
+        last_time = None
+        repo = self.get_repo()
+        prog = re.compile(
```
cdcarter: can you help me out with this regex's name? whats prog? it's already a bit of a wild RE to parse...

cumulusci/tasks/github/release_report.py
```diff
@@ -0,0 +1,83 @@
...
+                self.project_config.project__git__push_prefix_production,
+            ))
+        for release in repo.iter_releases():
+            if release.prerelease and not self.options['include_beta']:
+                continue
+            if self.options['date_start'] and release.created_at < self.options[
+                    'date_start']:
+                continue
+            if self.options[
+                    'date_end'] and release.created_at > self.options['date_end']:
```
cdcarter: should these ranges be inclusive? (i.e. >= ) I think so...? at least one side should be right?

cumulusci/tasks/github/release_report.py
```diff
@@ -0,0 +1,83 @@
...
+        self.options['date_end'] = self._parse_datetime(
+            self.options['date_end'],
+        ) if 'date_end' in self.options else None
+        self.options['include_beta'] = process_bool_arg(
+            self.options.get('include_beta', False))
+        self.options['print'] = process_bool_arg(
+            self.options.get('print', False))
+
+    @staticmethod
+    def _parse_datetime(dt_str):
```
cdcarter: please throw a `TaskOptionsError` if invalid format (or something else under the cci base exception class)

cdcarter: also consider putting this in utils.py with `process_bool_arg`

cumulusci/tasks/github/release_report.py
cpolcyn: `prog` is the compiled expression (var name came from help docs). It is compiled here and used repeatedly in the releases loop.

cumulusci/tasks/github/release_report.py
cpolcyn: This is basically "use release if start_date <= release_date <= end_date". It's just flipped around to the negative form because the `continue` statements skip the releases.

cumulusci/tasks/github/release_report.py
cdcarter: the var name is weird to me, can we consider...compiled_regex? branch_regex? what why regex

cumulusci/tasks/github/release_report.py
cdcarter: got it

cpolcyn:
@cdcarter I addressed all your requests, please re-review. The only thing I did a bit different was warning instead of raising an exception when date format isn't found. This way the task can continue and find info on other releases. Here's an example of the warning:

`2018-05-02 16:58:41: [https://github.com/SalesforceFoundation/Cumulus/releases/tag/rel/3.112] Could not find date in line: Production orgs: ~~2017-11-14~~ canceled
Use the format %Y-%m-%d`

PR #622, opened 2018-05-08
validate CUMULUSCI_KEY

cpolcyn:
Currently just checks that bool(key) is True and that len(key) is 16. Please let me know if any other checks would be helpful.


cpolcyn:
Not quite ready, this fails for EnvironmentProjectKeychain use cases where the CCI key is not used.

cdcarter:
will you assign me when this is ready, chris?

PR #635, opened 2018-05-21
Connected App as a CCI service

### LINKED ISSUE #244: connected_app as a service?
Jason and I were discussing offline that, for consistencies sake, the connected_app should probably be a special type of service instead of it's own ... thing. this would affect both the cli usage (register with service add, not config_connected_app) and the internal API (the current connected_app accessors would be re-written to access the service API) but not the task api (i.e. projectconfig would still have normal ol connected_app methods).

cpolcyn:
Fixes #244

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -508,6 +508,18 @@ flows:
                 task: update_admin_profile
 
 services:
+  connected_app:
+    description: Configure connected app for connecting to Salesforce orgs
+    attributes:
+      callback_url:
+        description: Callback URL configured on the Connected App
+        required: True
+      client_id:
+        description: Client ID from the Connected App
+        required: True
+      client_secret:
+        description: Client Secret from the Connected App
+        required: True
   github:
     description: Configure connection for github tasks, e.g. Create Release
     attributes:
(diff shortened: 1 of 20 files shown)


cdcarter:
a quick grep of the metaci codebase for `get_connected_app` reveals a few usages, we'll need to plan those updates accordingly

cumulusci/cumulusci.yml
```diff
@@ -508,6 +508,18 @@ flows:
                 task: update_admin_profile
 
 services:
+  connected_app:
+    description: Configure connected app for connecting to Salesforce orgs
```
cdcarter: Consider:
"A connected app is required to connect to and run commands against persistent orgs. See http://cumulusci.readthedocs.io/en/latest/tutorial.html#creating-a-connected-app for more."

cumulusci/cumulusci.yml
```diff
@@ -508,6 +508,18 @@ flows:
...
+    attributes:
+      callback_url:
+        description: Callback URL configured on the Connected App
+        required: True
+      client_id:
+        description: Client ID from the Connected App
```
cdcarter: consider "Client ID/Consumer Key from the Connected App"

cumulusci/cumulusci.yml
```diff
@@ -508,6 +508,18 @@ flows:
...
+        required: True
+      client_secret:
+        description: Client Secret from the Connected App
```
cdcarter: Client Secret/Consumer Secret

PR #640, opened 2018-05-28
Allowed default org to be used in cci org info and cci org browser.

topherlandry:
Implements https://github.com/SalesforceFoundation/CumulusCI/issues/435

Added in error messaging similar to running a flow without a specified or default org.
Brought the oauth refresh error handling that existed in `org_info` into `org_browser` where it didn't exist.


topherlandry:
Also implements https://github.com/SalesforceFoundation/CumulusCI/issues/378

cpolcyn:
Will continue in new internal PR

Issue #641, opened 2018-05-28
Scratch org password not being updated where created in a flow


topherlandry:
Since this change (https://github.com/SalesforceFoundation/CumulusCI/commit/fde28ca7345e9ef052dffdbebc20d5247910a261) I've not been able to get a new scratch org password generated when creating an org with the `dev_org` flow. When the new scratch org is created, it is created with a new password, but org info does not display the new password.

`cci org info dev` will give me the password of a previous org instead of the new one and I find that I'm not actually able to log in with the new credentials.

Based on what I'm looking at, it looks like when `BaseFlow` goes to refresh the oauth token and finds no scratch org, it runs `create_org` in `ScratchOrgConfig`. `cci org scratch dev org_name` uses `create_scratch_org` in the keychain, so it makes sense that the password generation is being handled differently.

I've reverted this change locally and everything has gone back to working as expected.
This is still true after the changes in #639.

cpolcyn:
I'm unable to reproduce the issue as of 4370d869b07877d22d2e7840db5fd85d4b376155. Here are my steps:

```bash
$ cci org scratch dev org_name
$ cci org info org_name
$ cci org scratch_delete org_name
$ cci flow run dev_org --org org_name
$ cci org info org_name
```

When I do this I can see the username and password have been updated, and logging in with the new credentials works.

I'm closing this one but please feel free to continue conversation here if needed.

PR #642, opened 2018-05-28
Fixing issue where new orgs made with flow have stale passwords in org info.

topherlandry:
Fixes https://github.com/SalesforceFoundation/CumulusCI/issues/641

Stale passwords were being shown in `cci org info` after a flow created a new org when it found none. Fixed this by specifying creation options on the org when a flow is executed and finds an org hasn't been created.

Unlike `cci org scratch dev org_name` I am not providing an option to not set a password, although this is an option that could be added in the future.


cpolcyn:
Closing, see comments in #641

PR #645, opened 2018-05-29
Py3 Compats & Some Miscellaney

cdcarter:
Like I mentioned, over the weekend I worked on this megabranch. I can split it up into smaller PRs but it's just a few small things. Make sure to review with ?w=1.
- Replace all print statements w/ print() calls and change one encoding method.
- Various fixes that Py3 work revealed: fix some imports for util tasks, weirdness w/ setup script.
- Adds three new utility task classes: PassOptionAsResult, PassOptionAsReturnValue, LogLine
- Flow executions (inner or outer) now know their flow name, and inner flows now know their step number.
- Adds pre-/post- hooks for flow & subflow executions.
- Add a `prep` argument to skip _init_flow when instantiating a flow class. The flow cannot execute til it is `prepped`.
- Sub-flow executions will use the same flow runner class as their parent flow if a flow class is not specified.

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -269,7 +269,7 @@ tasks:
         options:
             seconds: 5
 
-    info:
+    log:
         description: Log a line at the info level.
         class_path: cumulusci.tasks.util.LogLine
         options:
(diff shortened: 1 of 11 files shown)


cumulusci/cumulusci.yml
```diff
@@ -269,6 +269,12 @@ tasks:
         options:
             seconds: 5
 
+    info:
```
jlantz: Can we change this task name to log instead?  That way you could call log and override the level via options

PR #646, opened 2018-05-29
Inject cci project config into CumulusCI robot library from the robot task

jlantz:
- Use a listener in the `robot` task to inject the already initialized project_config from the task and the correct org name
- Fix task_options `test`, `include`, and `exclude` in the `robot` task

### DIFF
--- a/cumulusci/core/tests/test_keychain.py
+++ b/cumulusci/core/tests/test_keychain.py
@@ -9,6 +9,8 @@ import mock
 import nose
 import yaml
 
+from cumulusci.core.tests.utils import EnvironmentVarGuard
+
 from cumulusci.core.config import BaseGlobalConfig
 from cumulusci.core.config import BaseProjectConfig
 from cumulusci.core.config import OrgConfig
@@ -258,12 +260,24 @@ class TestEnvironmentProjectKeychain(TestBaseProjectKeychain):
 
     def setUp(self):
         super(TestEnvironmentProjectKeychain, self).setUp()
-        self.env = {}
+        self.env = EnvironmentVarGuard()
         self._clean_env(self.env)
-        self.env['{}test'.format(self.keychain_class.org_var_prefix)] = json.dumps(self.org_config.config)
-        self.env['{}connected_app'.format(self.keychain_class.service_var_prefix)] = json.dumps(self.services['connected_app'].config)
-        self.env['{}github'.format(self.keychain_class.service_var_prefix)] = json.dumps(self.services['github'].config)
-        self.env['{}mrbelvedere'.format(self.keychain_class.service_var_prefix)] = json.dumps(self.services['mrbelvedere'].config)
+        self.env.set(
+            '{}test'.format(self.keychain_class.org_var_prefix),
+            json.dumps(self.org_config.config)
+        )
+        self.env.set(
+            '{}connected_app'.format(self.keychain_class.service_var_prefix),
+            json.dumps(self.services['connected_app'].config)
+        )
+        self.env.set(
+            '{}github'.format(self.keychain_class.service_var_prefix),
+            json.dumps(self.services['github'].config)
+        )
+        self.env.set(
+            '{}mrbelvedere'.format(self.keychain_class.service_var_prefix),
+            json.dumps(self.services['mrbelvedere'].config)
+        )
 
     def _clean_env(self, env):
         for key, value in list(env.items()):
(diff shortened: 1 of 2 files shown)


cumulusci/robotframework/CumulusCI.py
```diff
@@ -32,23 +32,38 @@ def __init__(self, org_name=None):
...
     @property
     def org(self):
         if not hasattr(self, '_org'):
-            self._org = self.config.keychain.get_org(self.org_name)
+            self._org = self.keychain.get_org(self.org_name)
         return self._org
 
+    @property
+    def sf(self):
+        if not hasattr(self, '_sf'):
```
cdcarter: this is maybe a matter of style, but I'd prefer to see all these instance attrs set to None in `__init__` and these checks be e.g. `if self._sf is None`. if the instance attr isn't declared in __init__, ide complete, pylint, etc... struggle with it.

cumulusci/robotframework/CumulusCI.py
```diff
@@ -32,23 +32,38 @@ def __init__(self, org_name=None):
...
+            self._sf = self._init_api()
+        return self._sf
+
+    @property
+    def tooling(self):
+        if not hasattr(self, '_tooling'):
```
cdcarter: move to initializer.

cumulusci/robotframework/CumulusCI.py
```diff
@@ -32,23 +32,38 @@ def __init__(self, org_name=None):
...
         # Turn off info logging of all http requests 
         logging.getLogger('requests.packages.urllib3.connectionpool').setLevel(logging.WARN)
 
     @property
-    def config(self):
-        if not hasattr(self, '_config'):
-            self._config = self._init_config()
-        return self._config
+    def project_config(self):
+        if not hasattr(self, '_project_config'):
```
cdcarter: move to initializer

cumulusci/robotframework/CumulusCI.py
```diff
@@ -32,23 +32,38 @@ def __init__(self, org_name=None):
...
+            self._project_config = CliConfig().project_config
```
cdcarter: can you confirm that `CliConfig` will instantiate from ENV vars if present?

cumulusci/robotframework/CumulusCI.py
```diff
@@ -84,7 +99,7 @@ def run_task(self, task_name, **options):
             | Run Task  | deploy      |                            | Run deploy with standard options |
             | Run Task  | deploy      | path=path/to/some/metadata | Run deploy with custom path      |
         """
-        task_config = getattr(self.config.project_config, 'tasks__{}'.format(task_name))
+        task_config = getattr(self.project_config, 'tasks__{}'.format(task_name))
```
cdcarter: `self.project_config.get_task(task_name)` instead

cumulusci/robotframework/CumulusCI.py
```diff
@@ -152,7 +163,7 @@ def _run_task(self, task_class, task_config):
         task_config = TaskConfig(task_config)
         exception = None
 
-        task = task_class(self.config.project_config,
+        task = task_class(self.project_config,
```
cdcarter: task also takes a `name`

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -34,6 +38,22 @@ def _init_options(self, kwargs):
         if 'suites' not in self.options:
             self.options['suites'] = 'tests'
 
+        # Initialize the include list and add the org name to it
```
cdcarter: "add the org name to it" ??

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -34,6 +38,22 @@ def _init_options(self, kwargs):
...
+        if 'include' in self.options:
+            if not isinstance(self.options['include'], list):
+                new_includes = []
+                for include in self.options['include'].split(','):
+                    new_includes.append(include.strip())
```
cdcarter: you could (should?) use `shlex` instead of split and strip

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -34,6 +38,22 @@ def _init_options(self, kwargs):
...
+                self.options['include'] = new_includes
+
+        # Initialize the exclude list and add the org name to it
```
cdcarter: "and add the org name to it" ??

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -51,13 +71,33 @@ def _init_options(self, kwargs):
...
+        # from self.project_config instead of reinitializing CumulusCI's config
+        listener = CumulusCIRobotListener(self.project_config, self.org_config.name)
+        settings['Listeners'] = [listener]
 
+        # Build the top level test suite
+        suite = TestSuiteBuilder().build(options['SuiteNames'])
+        suite.configure(**settings.suite_config)
+
+        # Run the test suite
+        with pyloggingconf.robot_handler_enabled(settings.log_level):
```
cdcarter: can you add a comment of why we're creating the suites ourself? and link back to robot.run? 

is it because of the include/exlude? robot run should be able to take a listener.

cumulusci/robotframework/CumulusCI.py
jlantz: I think the only ENV var that's relevant there is CUMULUSCI_KEYCHAIN_CLASS which would allow you to swap out the keychain implementation to use EnvironmentProjectKeychain.  The EnvironmentProjectKeychain is then responsible for handling the rest of the env vars.

Issue #664, opened 2018-06-08
CCI creates scratch org with SFDX alias in quotes


dospromptman:
It used to be that the scratch org name given to CCI would get passed along to sfdx to use as the alias when creating the scratch org.  It was possible to use the alias with the project name prefixed to conveniently invoke sfdx against a scratch org.

However, the aliases are now created with quotes, such that the alias name is surrounded in literal quote characters.  This requires escaping the quotes when trying to use sfdx with that alias.

This seems like a bug to me.  Why is it wrapping it in quotes now?  Why aren't the quotes getting removed from the argument when the sfdx command is invoked from cci?

```
cci org scratch dev itchy
cci org info itchy
```

This should work, but it does not:

```
sfdx force:org:display -u SAL__itchy
```

Instead, only this works:

```
sfdx force:org:display -u \"SAL__itchy\"
```

See attached screenshot.

<img width="974" alt="cci_alias_quotes_bug" src="https://user-images.githubusercontent.com/1058323/41166864-ee114744-6b06-11e8-88ae-2ed976f2b1c0.png">

force2b:
I noticed this yesterday, but wasn't sure if it was me and haven't had a chance to get back to it. Thanks for documenting and posting!

cdcarter:
Oh well thats annoying!

jlantz:
Fixed in 2.0.4

PR #666, opened 2018-06-13
fix: replace sarge with subprocess to fix quote escaping

jstvz:
Some additional work to replace `sarge` throughout `ScratchOrgConfig` is required, so only a fix for issue #664 is included in this commit.

### DIFF
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -3,6 +3,8 @@ import datetime
 import json
 import os
 import re
+import shlex
+import subprocess
 
 import sarge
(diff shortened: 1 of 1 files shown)


davisagli:
@jstvz I tried this out and it still creates a scratch org with quotes in the alias (presumably because you aren't passing shell=True to subprocess.check_output, so the quotes around the alias make it through to sfdx instead of being interpreted by the shell). I think we want to use the shell=True flag with sarge but be sure to escape user input like this: http://sarge.readthedocs.io/en/latest/tutorial.html#formatting-commands-with-placeholders-for-safe-usage (or do something similar with subprocess).

jstvz:
Closing this PR for now, will re-submit.

PR #669, opened 2018-07-09
Update PyPI URLs

davisagli:
The canonical domain for PyPI changed to pypi.org a few months ago. This should avoid getting redirected.


davisagli:
Will re-open from a local branch

PR #673, opened 2018-07-17
Expand test coverage for mdapi client

jlantz:
This branch adds a test suite with 77% coverage on `cumulusci.salesforce_api.metadata` and removes some lines of code which were unused and thus untestable.


cumulusci/salesforce_api/metadata.py
```diff
@@ -509,69 +483,6 @@ def _process_response(self, response):
...
-    def __init__(self, task, version, purge_on_delete=True):
-        self.version = version
-        if not version.number:
-            self.package_zip = None
-        else:
-            self.package_zip = PackageZipBuilder(
-                self.version.package.namespace).uninstall_package()
-        super(ApiUninstallVersion, self).__init__(
-            task, self.package_zip, purge_on_delete)
-
```
davisagli: These are going away because they weren't used?

cumulusci/salesforce_api/soap_envelopes.py
```diff
@@ -123,7 +123,7 @@
   </soap:Header>
   <soap:Body>
     <checkStatus xmlns="http://soap.sforce.com/2006/04/metadata">
-      <asyncProcessId>%(process_id)s</asyncProcessId>
+      <asyncProcessId>{process_id}</asyncProcessId>
     </checkStatus>
```
davisagli: Not strictly related to this PR, but while reviewing it I noticed that `checkStatus` is documented as having been removed in MD API version 31 and higher. So now I'm wondering why continuing to use it hasn't been a problem.

cumulusci/salesforce_api/metadata.py
jlantz: Yes, I was trying to figure out how to get test coverage and realized there was no where that called this.  An earlier check, if I recall, already ensured this condition so it was impossible to have a test ever hit these lines.

cumulusci/salesforce_api/soap_envelopes.py
jlantz: We do override the status check envelope on the ApiDeploy subclass: https://github.com/SalesforceFoundation/CumulusCI/blob/b42bd95fd9fbf1fab0a8e7e6af7832af6f0513ce/cumulusci/salesforce_api/metadata.py#L341

My hunch is that we might be best to change them all to the new status check envelopes for each type.  Generally, all we do is deploy, retrieve, and list metadata.  Looks like retrieves are using the generic checkStatus still

PR #681, opened 2018-07-23
Configure github client to retry requests

### LINKED ISSUE #676: Retry when we get a 502 response from github

davisagli:
Fixes #676


davisagli:
Should I do more to test this? I'm pretty much relying on the retry feature from urllib3 to do what it should.

davisagli:
Added a test of the actual retry functionality -- had to go lower level than where `responses` patches.

PR #682, opened 2018-07-24
Windows 10 fix - Add shell=true to command

daveb-501commons:
cci Connected App Error - no such command
Getting an error in cci when I enter cci org config_connected_app

Error: No such command "config_connected_app"

Debugged the issue and indeed is related cci interop with sfdx. I'm on Windows 10 which doesn't work unless you explicitly specify access to the shell (i.e., shell=True on the Sarge.Command call). This is contrary to the python docs which says you shouldn't need to include shell=True for bat files like sfdx.

To get cci org info dev to work I had to make the following changes to ScratchOrgConfig.py by adding shell=True

line 32: (sfdx display call)
p = sarge.Command(
command,
stderr=sarge.Capture(buffer_size=-1),
stdout=sarge.Capture(buffer_size=-1),
shell=True,
)

line: 175 (sfdx create call)
p = sarge.Command(command, stdout=sarge.Capture(buffer_size=-1), shell=True)


davisagli:
Hi @daveb-501commons, thanks for taking the time to submit a pull request. But I'm afraid I don't understand how the proposed change relates to the `No such command "config_connected_app"` error.

The changelog for 2.0.0-beta97 says: Salesforce Connected App is now a CCI Service! Instead of using `cci org config_connected_app` you can use the familiar `cci service` commands. See http://cumulusci.readthedocs.io/en/latest/tutorial.html#creating-a-connected-app for current instructions.

As for adding `shell=True`, it seems to me that that would only help if the usernames are quoted, which sounds like the bug discussed in #664. I think we should aim to fix things so that we are creating scratch org aliases that don't include quotes rather than adding `shell=True` when we call commands.

davisagli:
@daveb-501commons fyi we merged a modified version of your fix (#693) which is included in cumulusci 2.0.4

PR #683, opened 2018-07-25
Make sure ParseErrors from metadata xml include the filename

davisagli:


cumulusci/tasks/metadata/package.py
```diff
@@ -242,6 +242,16 @@ class MissingNameElementError(Exception):
 class ParserConfigurationError(Exception):
     pass
 
+
+def elementtree_parse_file(path):
+    try:
+        root = ET.parse(path)
+    except ET.ParseError as err:
+        err.filename = path
```
cdcarter: neat usage! 

can you put this (elementtree_parse_file) helper in `cumulusci.utils`? https://github.com/SalesforceFoundation/CumulusCI/blob/master/cumulusci/utils.py

davisagli: ParseError is just an empty subclass of Python's builtin SyntaxError!

I moved it to cumulusci.utils

PR #702, opened 2018-08-03
Be consistent about passing a keychain to refresh_oauth_token

davisagli:
Previously, the refresh_oauth_token method did not require a keychain param for DummyOrgConfig and ScratchOrgConfig, but it did for OrgConfig. As a result, I think this token refresh in metadata.py would fail with a persistent org. This standardizes the interface to always pass a keychain.

### DIFF
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -284,7 +284,7 @@ class ScratchOrgConfig(OrgConfig):
             message = 'Message: {}'.format('\n'.join(stdout_list))
             raise ScratchOrgException(message)
 
-    def refresh_oauth_token(self, keychain=None):
+    def refresh_oauth_token(self, keychain):
         """ Use sfdx force:org:describe to refresh token instead of built in OAuth handling """
         if hasattr(self, '_scratch_info'):
             # Cache the scratch_info for 1 hour to avoid unnecessary calls out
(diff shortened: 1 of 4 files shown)


cdcarter:
I'm gonna thumbs up this, but also... 

the invalid session logic in the MDAPI client is weird.... It's from before the CCI2 keychain existed and was handling refresh/token logic. 

I'm not fully convinced we should be trying a refresh if the request fails, **but** I am fully convinced that even if we do, it should be handled by throwing an exception in the metadata api client and handling it in BaseTask in such a way that other tasks could throw InvalidSessionPleaseRetry or whatever....  thoughts @davisagli ?

davisagli:
Hmm yeah, let me take a look at the bigger picture here.

jlantz:
So, the legacy backstory for refreshing oauth token in the MDAPI client relates back to mrbelvedere where the MDAPI client originally came from.

We had some install failures on long running deployments where the oauth token had expired in the middle of the install, usually against a production org where unmanaged deploys must run all apex tests for each deploy.  In mrbelvedere, we don't split things up into tasks like we do in CumulusCI.  In CumulusCI, each task refreshes the token at start which makes  this perhaps a bit less necessary.

The challenge with having the task handle it is that the current logic will pick up in the middle of an MDAPI request (usually during polling for an operation to complete) and continue on seamlessly after refreshing the token.  If we raised an exception to the task level, the task would likely need to restart the whole operation rather than just pick up from polling.

Another consideration: even though tasks refresh the token, if the token was created an hour ago I don't think it would get refreshed until it expires.  Thus, we can't really know when a token will expire on us.  I could be wrong on this though, but I seem to remember always getting back the same token on subsequent refresh calls.

davisagli:
@jlantz's explanation makes sense for why we should handle this here. Maybe we can revisit if we find a way to separate out polling tasks, but for now, can we merge this?

PR #705, opened 2018-08-07
Add test coverage for apex tasks

jstvz:
- test anonymous apex task
- add tests for apex.anon and apex.batch tasks
- Update test_salesforce name to reflect purpose


cumulusci/tasks/apex/tests/test_apex_tasks.py
```diff
@@ -0,0 +1,348 @@
...
+
+    def _get_url_and_task(self):
+        task = AnonymousApexTask(self.project_config, self.task_config, self.org_config)
+        url = task.tooling.base_url + "executeAnonymous"
+        return task, url
+
+    @responses.activate
+    def test_run_anonymous_apex_status_fail(self):
+        task, url = self._get_url_and_task()
+        responses.add(responses.GET, url, status=418, body="I'm a teapot")
```
davisagli: 🍵😂

cumulusci/tasks/apex/tests/test_apex_tasks.py
```diff
@@ -0,0 +1,348 @@
...
+        with assert_raises(SalesforceGeneralError) as cm:
```
davisagli: Is there a reason to use `nose.assert_raises` instead of `with self.assertRaises` from unittest? If no difference, I'd prefer the latter since it's in the stdlib.

PR #708, opened 2018-08-09
Coverage for cumulus.cli.cci

davisagli:
This one's a bit of a doozy. Some changes that warrant special attention:
- Moved global functions taking an initial 'config' param from cci.py to become methods of the CliConfig class
- Removed the unused 'e' arg for handle_exception_debug
- Made render_recursive handle lists better.
- Stop passing the return from render_recursive to click.echo, because it takes care of echoing itself.
- Make check_latest_version handle its own exceptions
- Removed unused --extend option for `project init`
- Removed unimplemented project_list and project_cd commands
- Don't echo scratch org expiration when using `org info` with the --json flag
- Delegate check for whether an org can be deleted to a `can_delete` method on the org config
- Remove unused --delete option for `org scratch`
- Fix exception handling for `org scratch_delete
- Simplify exception handling for `task run` and `flow run`

### DIFF
--- a/cumulusci/cli/tests/test_cci.py
+++ b/cumulusci/cli/tests/test_cci.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from datetime import date
 import json
 import os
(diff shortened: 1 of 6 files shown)


davisagli:
Looks like I need to work on a few test failures; I'll let you know when this is ready for review.

davisagli:
Okay I think I've resolved those failures.

cumulusci/cli/tests/test_cci.py
```diff
@@ -224,6 +225,7 @@ def test_project_init(self, click):
             run_click_command(cci.project_init, obj=config)
 
             # Make sure expected files/dirs were created
+            print recursive_list_files()
```
cdcarter: use python3 friendly print fn plz

davisagli: Heh, this was just meant for temporary debugging anyway. Will remove.

PR #714, opened 2018-08-15
Feature/taskflowconfig

cdcarter:
Implements the missing BaseTaskFlowConfig.list_flows() method, adds unit tests for that class, and splits test_config.py into two files, one that does YAML parsing and one thats fast.


davisagli:
Is splitting the tests out just intended to make it easier to run the fast tests without the slow ones? It would also be possible to use two test classes in the same module and tell the test runner to run one of them. I kind of like having the organization of tests mirror the organization of code (i.e. one test module per module of actual code).

cumulusci/core/tests/test_config.py
```diff
@@ -310,519 +252,46 @@ def test_process_github_dependency(self):
...
+        self.assertEqual(len(tasks), 3)
+        deploy = [task for task in tasks if task['name'] == 'deploy'][0]
+        self.assertEqual(deploy['description'], 'Deploy Task')
+
+    def test_get_task(self):
+        task = self.task_flow_config.get_task('deploy')
+        self.assertIsInstance(task, BaseConfig)
+        self.assertDictContainsSubset({'description': 'Deploy Task'}, task.config)
+
+    @nose.tools.raises(TaskNotFoundError)
```
davisagli: Let's avoid using utils imported from nose since we have some interest in switching to pytest. For this I'd do:

    with self.assertRaises(TaskNotFoundError):
        self.task_flow_config.get_task('robotic_superstar')

cumulusci/core/tests/test_config.py
```diff
@@ -310,519 +252,46 @@ def test_process_github_dependency(self):
...
+    def test_no_task(self):
+        _ = self.task_flow_config.get_task('robotic_superstar')
+
+    def test_get_flow(self):
+        flow = self.task_flow_config.get_flow('coffee')
+        self.assertIsInstance(flow, BaseConfig)
+        self.assertDictContainsSubset({'description': 'Coffee Flow'}, flow.config)
+
+    @nose.tools.raises(FlowNotFoundError)
```
davisagli: Ditto

cumulusci/core/config/BaseTaskFlowConfig.py
```diff
@@ -32,9 +32,19 @@ def get_task(self, name):
     def list_flows(self):
         """ Returns a list of flow info dictionaries with keys 'name' and 'description' """
         flows = []
+        for flow in list(self.flows.keys()):
```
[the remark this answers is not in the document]
cdcarter: roger, will change upclass also

cdcarter:
Yes, that was the purpose for splitting out the tests. There are multiple TestCase classes in each file. I don't know a ton about setting up testrunners... I do think that we want the granularity to run subsets of cumulusci.core.config, since its really quite big. Thoughts on how you'd split up the existing TestCases? Part of me wants to recombine some of the smaller classes in cumulusci.core.config back into a larger module... oy...

Yes, I do want to refactor cci.py to use some of these now... I can rescind this PR if you want to wait for that, or if you are ok with multiple parts, your comments have been addressed.

davisagli:
You can run e.g `nosetests cumulusci/core/tests/test_config.py:NameOfTestCase` to run just the tests in one TestCase class.

I'm fine with cci adjustments happening later.

PR #716, opened 2018-08-16
PyYAML 3.12 -> 3.13

jlantz:
Installing metaci_cli is currently failing due to a mismatch between CumulusCI's dependency on PyYAML being set to an older version.

Increase to 3.13 to resolve.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ plaintable==0.1.1
 pycrypto==2.6.1
 PyGithub==1.39
 pytz==2018.4
-PyYAML==3.12
+PyYAML==3.13
 raven==6.7.0
 requests[security]==2.18.4
 responses==0.9.0
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,7 @@ requirements = [
     'pycrypto==2.6.1',
     'PyGithub==1.39',
     'pytz==2018.4',
-    'PyYAML==3.12',
+    'PyYAML==3.13',
     'raven==6.7.0',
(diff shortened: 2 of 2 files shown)


davisagli:
Why are we pinning specific versions rather than minimum versions in setup.py?

jlantz:
I'm open to pinning minimum versions but wonder if that should be done as a more holistic change to ensure we understand the implications and set minimum versions in the right way to avoid major releases breaking things unexpectedly.

davisagli:
Hmm, we want specific version pins when installing cumulusci as a CLI tool, but when installing as a dependency of something like MetaCI we'd like the downstream package to be able to override the pins when necessary. This is tricky...

jlantz:
Yeah, I'm not sure that's the ideal but it's kinda the contract we've created with anything built on top of cumulusci for now.  It's probably something we should change but we need to ensure that doesn't break other tooling using cumulusci like MetaCI and the metaci_cli.

davisagli:
The thing I've seen that we could consider would be to use minimum pins in setup.py, but avoid accidental version bumps for direct consumers by recommending that people install using a requirements.txt file that we publish, instead of just doing pip install cumulusci

cdcarter:
I'm thinking we're getting close to having reason to split `cci` and `CumulusCI` into separate eggs...

davisagli:
Hmm, maaayybe -- but if we do that I'd strongly advocate for keeping them in the same repository since changes in one often require changes in the other.

PR #718, opened 2018-08-20
Handle RobotTestFailure in cci task run and cci flow run

jlantz:
Outputs a single line rather than a python traceback


cumulusci/cli/cci.py
```diff
@@ -780,8 +781,6 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
             import pdb
             pdb.set_trace()
 
-        task()
-
```
davisagli: I guess you removed this for debugging and forgot to put it back?

cumulusci/cli/cci.py
```diff
@@ -872,13 +871,12 @@ def flow_run(config, flow_name, org, delete_org, debug, o, skip, no_prompt):
...
         flow = flow_class(config.project_config, flow_config,
                           org_config, options, skip, name=flow_name)
-
         flow()
     except (TaskRequiresSalesforceOrg, TaskOptionsError) as e:
         exception = click.UsageError(e.message)
         handle_exception_debug(config, debug, throw_exception=exception)
     except (ApexTestException, BrowserTestFailure, MetadataComponentFailure,
-            MetadataApiError, ScratchOrgException) as e:
+            MetadataApiError, RobotTestFailure, ScratchOrgException) as e:
```
davisagli: Rather than hardcoding a big list of exceptions here, how about if we define CumulusCIExpectedFailure and make the exceptions listed here subclass it? Then we can just catch CumulusCIExpectedFailure here, which makes it simpler for CumulusCI extensions to define their own exceptions (like if we want to refactor the robot support into a separate package)

For the usage errors, maybe TaskRequiresSalesforceOrg and TaskOptionsError should actually be subclasses of click.UsageError

cdcarter: chris always had dreams of making our exceptions more of a hierarchy, this is probably a good idea.

davisagli: I'm not sure what we'd do if we want to handle an exception that we don't define ourselves as an expected failure, though.

cumulusci/cli/cci.py
jlantz: Oops, fixed in latest push

cumulusci/cli/cci.py
jlantz: I changed this to use the new exception classes CumulusCIFailure and CumulusCIUsageError.  I also added a new section to exception handling in task_run and flow_run which currently only has ScratchOrgException but could be used to handle other exceptions from outside of CumulusCI.  ScratchOrgException is a special case where it seems to make sense to allow the implementation to decide how they want to handle the exception.

I didn't have CumulusCIUsageError subclass click.UsageError since click is only used in the CLI and not in other implementations (such as MetaCI).  In MetaCI, we'll likely want to have CumulusCIUsageError create an Error status on a build while all exceptions in CumulusCIFailure will create a Fail status.

cumulusci/core/exceptions.py
```diff
@@ -2,13 +2,18 @@
 class CumulusCIException(Exception):
     pass
 
+class CumulusCIUsageError(CumulusCIException):
+    pass
+
+class CumulusCIFailure(CumulusCIException):
+    pass
```
davisagli: Can we give these docstrings explaining that they're meant to be abstract exceptions and when to use them?

cumulusci/cli/cci.py
```diff
@@ -786,15 +782,17 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
...
         handle_exception_debug(config, debug, throw_exception=exception)
-    except (ApexTestException, BrowserTestFailure, MetadataComponentFailure,
-            MetadataApiError, ScratchOrgException) as e:
+    except CumulusCIFailure as e:
         # Expected failure; report without traceback
         exception = click.ClickException('Failed: {}'.format(e.__class__.__name__))
         handle_exception_debug(config, debug, throw_exception=exception)
+    except ScratchOrgException as e:
+        exception = click.ClickException('Failed: {}'.format(e.__class__.__name__))
+        handle_exception_debug(config, debug, throw_exception=exception)
```
davisagli: I don't understand your point about why ScratchOrgException should be handled separately. Currently you're handling it exactly the same as CumulusCIFailure and it's a subclass of CumulusCIFailure, so I don't think adding this block does anything.

davisagli: In fact, since the `except CumulusCIFailure` will match `ScratchOrgException` before getting to this block, I think this is unreachable code.

davisagli: Gah, what am I talking about -- ScratchOrgException extends CumulusCIException, not CumulusCIFailure. Sorry, need more ☕️...

davisagli: But I'm still not clear on why you don't want ScratchOrgException to be a CumulusCIFailure. And if it does need to be separate, you can still match both CumulusCIFailure and ScratchOrgException in the same `except` clause given that you're handling them in the same way.

davisagli: Okay, I get it.

cumulusci/cumulusci.yml
```diff
@@ -502,6 +502,18 @@ flows:
...
+        description: Retrieves declarative changes made in a scratch org and converts to src directory
+        steps:
+            1:
+                task: dx_convert_to
+            2:
+                task: dx_pull
+            3:
+                task: dx_convert_from
+            4:
+                task: update_package_xml
```
davisagli: This was meant for a different branch, right?

jlantz: yes, it's in a separate PR which should be merged before this PR and would cause this one to match master.  I accidentally pushed to this branch :)

jlantz: After merging the other PR and merging master into this branch, the diff is now gone

PR #719, opened 2018-08-20
Add Flows to mapping for package.xml generation

jlantz:


jlantz:
Looks like this was already done via a recent merge.

PR #720, opened 2018-08-20
set client id in call options

cdcarter:
For all SimpleSalesforce instances we summon up, set the Sforce-Call-Options header to include a client id when registered or the CumulusCI clientName.

### DIFF
--- a/cumulusci/tasks/salesforce/BaseSalesforceApiTask.py
+++ b/cumulusci/tasks/salesforce/BaseSalesforceApiTask.py
@@ -3,6 +3,7 @@ from simple_salesforce import Salesforce
 
 from cumulusci.tasks.salesforce import BaseSalesforceTask
 
+CALL_OPTS_HEADER_KEY = 'Sforce-Call-Options'
 
 class BaseSalesforceApiTask(BaseSalesforceTask):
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/salesforce/BaseSalesforceApiTask.py
```diff
@@ -28,6 +28,9 @@ def _init_api(self, base_url=None):
         )
         if base_url is not None:
             rv.base_url += base_url
+
+        rv.headers.setdefault(CALL_OPTS_HEADER_KEY, self._get_client_name())
```
davisagli: Should be in the form `client=CLIENT_NAME`, not simply the CLIENT_NAME

cdcarter: I woke up, hopped out of bed, and realized that this wasn't correct :)

cdcarter: I don't know what that says about me.

cdcarter:
API Calls will be grouped properly when reporting.

cdcarter:
And yea, it is a little weird that its a header and its mostly undocumented behavior...

PR #722, opened 2018-08-21
rely on the project_config for more

cdcarter:
the CLI was doing stuff that the project_config can already do: get_task, list_flows, get_flow.

i thought this branch had something wrong with it, but then when I went to check it out today all my tests were passing...soo.... I'm gonna say ready for review?


cumulusci/cli/tests/test_cci.py
```diff
@@ -773,7 +759,7 @@ def test_flow_run(self):
         DummyTask._run_task.assert_called_once()
         org_config.delete_org.assert_called_once()
 
-    def test_flow_run_delete_non_scratch(self):
+    def test_flow_run_delete_non_scratch(self, ):
```
davisagli: 😮

PR #726, opened 2018-09-04
Complete coverage for cli

davisagli:


cci_rth.py
```diff
@@ -2,7 +2,10 @@
```
cdcarter: I'm not certain that this module, or the pyinstaller dep in general are still in use...

cumulusci/cli/tests/test_cci.py
```diff
@@ -159,19 +159,22 @@ def test_handle_sentry_event_no_event(self):
...
     @mock.patch('cumulusci.cli.cci.CliConfig')
     @mock.patch('cumulusci.cli.cci.check_latest_version')
-    def test_main(self, check_latest_version, CliConfig):
+    def test_main(self, check_latest_version, CliConfig, init_logger):
         CliConfig.return_value = _marker = object()
 
         ctx, result = run_click_command(cci.main)
 
         check_latest_version.assert_called_once()
+        init_logger.assert_called_once()
```
cdcarter: this comment probably would have been better suited during the initial refactor but...

i'd appreciate a change in test name, or perhaps a comment (here or in cli.py), to the effect of "we're testing the stuff that should happen before every command execution" -- theres some magic to understanding how pass_context works and the CliConfig/logger are initialized

cumulusci/cli/tests/test_config.py
```diff
@@ -0,0 +1,156 @@
...
+    @classmethod
+    def setUpClass(cls):
+        os.chdir(os.path.dirname(cumulusci.__file__))
+        os.environ['CUMULUSCI_KEY'] = '1234567890abcdef'
+
+    def test_init(self):
+        config = CliConfig()
+
+        self.assertSetEqual(
+            {'cumulusci', 'tasks', 'flows', 'services', 'orgs', 'project'},
```
cdcarter: i think we should probably assert that this is a subset of the keys, the `config` object is sorta like `__dict__`, i don't like asserting about its exact structure

cumulusci/cli/tests/test_config.py
```diff
@@ -0,0 +1,156 @@
...
+            set(config.global_config.config.keys()),
+        )
+        self.assertEqual('CumulusCI', config.project_config.project__name)
+        self.assertSetEqual(
+            {'services', 'orgs', 'app'},
```
cdcarter: see above

cci_rth.py
davisagli: I guess not, given that the import here was out of date! I'll remove the files related to pyinstaller.

cdcarter: you rock

PR #738, opened 2018-09-07
New Flow: retrieve_scratch

jlantz:
- Add new flow `retrieve_scratch` to retrieve declarative changes from  scratch org via sfdx force:source:pull and convert to src directory via sfdx force:source:convert
- Decode command output as utf-8 in Command task

One drawback is that it reorders some of the elements in .object files.  It moves the `<fields>` elements to the end rather than keeping in alphabetical order.

Another drawback is that it outputs a lot of log lines but that's minor.


davisagli:
@jlantz shoot, I think there's a problem with our current appveyor workaround -- I don't think it's getting triggered when we merge master to a branch via the github UI instead of doing it locally and pushing. I guess we need auto-merging to feature branches for CumulusCI...

PR #741, opened 2018-09-10
Complete coverage for cumulusci.core.config

davisagli:


cumulusci/core/config/BaseProjectConfig.py
```diff
@@ -194,7 +194,7 @@ def repo_url(self):
                     in_remote_origin = True
                     continue
                 if in_remote_origin and "url = " in line:
-                    return line[7:]
+                    return line[6:]
```
cdcarter: `len("url = ")` ? since you're already making the fix

cumulusci/core/tests/test_config.py
```diff
@@ -210,20 +248,328 @@ def get_service(self, name):
...
+            'HEROKU_TEST_RUN_COMMIT_VERSION': 'HEAD',
+            'CUMULUSCI_REPO_BRANCH': 'feature/test',
+            'CUMULUSCI_REPO_COMMIT': 'HEAD~1',
+            'CUMULUSCI_REPO_ROOT': '.',
+            'CUMULUSCI_REPO_URL': 'https://github.com/SalesforceFoundation/CumulusCI-Test.git',
+        }
+        config = BaseProjectConfig(BaseGlobalConfig())
+        with mock.patch.dict(os.environ, env):
+            result = config.repo_info
+            self.assertEqual({
```
cdcarter: does this need to be inside the patch?

cumulusci/core/tests/test_config.py
```diff
@@ -210,20 +248,328 @@ def get_service(self, name):
...
+
+    def test_repo_owner_no_repo_root(self):
+        config = BaseProjectConfig(BaseGlobalConfig())
+        with temporary_dir() as d:
+            cwd = os.getcwd()
+            os.chdir(d)
+            self.assertIsNone(config.repo_owner)
+            os.chdir(cwd)
+
+    def test_repo_branch_from_repo_info(self):
```
cdcarter: this is repetitive that it feels like maybe we should do `test_from_repo_info` `test_from_git` `test_no_repo_root` that has an assert for each property instead of doing the setup code over and over again.

cumulusci/core/tests/test_config.py
```diff
@@ -210,20 +248,328 @@ def get_service(self, name):
...
+                f.write('# pack-refs with: peeled fully-peeled sorted\n')
+                f.write('#\n')
+                f.write('8ce67f4519190cd1ec9785105168e21b9599bc27 refs/remotes/origin/master\n')
+
+            self.assertIsNotNone(config.repo_commit)
+
+            os.chdir(cwd)
+
+    def test_use_sentry(self):
+        config = BaseProjectConfig(BaseGlobalConfig())
```
cdcarter: i don't completely understand why this test passes? uses the cumlusci root cumulusci.yml?

cumulusci/core/tests/test_config.py
```diff
@@ -210,20 +248,328 @@ def get_service(self, name):
...
+                f.write('abcdef')
+
+            self.assertIsNone(config.repo_commit)
+
+            os.chdir(cwd)
+
+    def test_repo_commit_packed_refs(self):
+        config = BaseProjectConfig(BaseGlobalConfig())
+        with temporary_dir() as d:
+            cwd = os.getcwd()
```
cdcarter: didn't we just add a cwd context manager? oh wait was that metadeploy?

cumulusci/core/tests/test_config.py
```diff
@@ -210,20 +248,328 @@ def get_service(self, name):
...
+
+    def test_check_keychain(self):
+        config = BaseProjectConfig(BaseGlobalConfig())
+        with self.assertRaises(KeychainNotFound):
+            config._check_keychain()
+
+    def test_list_orgs(self):
+        config = BaseProjectConfig(BaseGlobalConfig())
+        config.keychain = mock.Mock()
+        config.keychain.list_orgs.return_value = _marker = object()
```
cdcarter: should use `unittest.mock.sentinel` for clarity on what we're testing

cumulusci/core/tests/test_config.py
davisagli: That would have been a good idea, but given that I've already written it this way, I'm not sure it's worth the effort to refactor? This way has the small benefit that a failure on one of the properties doesn't keep it from testing the others.

cumulusci/core/tests/test_config.py
davisagli: It passes because the mock keychain's `get_service` method returns another mock object without raising an exception.

cumulusci/core/tests/test_config.py
davisagli: It was metadeploy. But it looks like most places that we're using `temporary_dir` in CumulusCI we do want to chdir to it, so I'll adjust it to handle that...

cumulusci/core/tests/test_config.py
davisagli: Nice, I didn't know that one.

cumulusci/core/tests/test_config.py
```diff
@@ -504,14 +478,14 @@ def test_check_keychain(self):
...
-        self.assertIs(_marker, config.list_orgs())
+        config.keychain.list_orgs.return_value = mock.sentinel.orgs
+        self.assertIs(mock.sentinel.orgs, config.list_orgs())
 
     def test_get_org(self):
         config = BaseProjectConfig(BaseGlobalConfig())
         config.keychain = mock.Mock()
-        config.keychain.get_org.return_value = _marker = object()
-        self.assertIs(_marker, config.get_org('test'))
+        config.keychain.get_org.return_value = mock.sentinel.org = object()
```
cdcarter: i don't think the `= object()` is necessary

cumulusci/core/tests/test_config.py
cdcarter: i know you're just fixing up code in place but this indicates something fishy smelling in the code, I think... it's very hard to correlate the setup of the mock to the rest of the code and the observed behavior. I think a comment is probably useful but its also probably a sign we should come back here and do some cleanup.

cumulusci/core/tests/test_config.py
cdcarter: no prob.

PR #742, opened 2018-09-11
 Add file support to AnonymousApexTask

davisagli:
We should be able to replace the similar tasks in SAL and Cumulus with this.

**lurch: attach W-018728

### DIFF
--- a/cumulusci/tasks/apex/anon.py
+++ b/cumulusci/tasks/apex/anon.py
@@ -1,35 +1,112 @@
 from simple_salesforce.util import exception_handler
 
-from cumulusci.tasks.salesforce import BaseSalesforceApiTask
 from cumulusci.core.exceptions import ApexCompilationException
 from cumulusci.core.exceptions import ApexException
+from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.utils import in_directory
+
+EMAIL_TOKEN = 'mrbelvedere@salesforce.org'
 
 
 class AnonymousApexTask(BaseSalesforceApiTask):
-    """ Executes a string of anonymous apex. """
+    """ Executes anonymous apex from a file or string."""
 
-    task_options = {"apex": {"description": "The apex to run.", "required": True}}
+    task_options = {
+        "path": {
+            "description": "The path to an Apex file to run.",
+            "required": False,
+        },
+        "apex": {
+            "description": "A string of Apex to run (after the file, if specified).",
+            "required": False,
+        },
+        "managed": {
+            "description": (
+                "If True, will insert the project's namespace prefix.  "
+                "Defaults to False or no namespace."
+            ),
+            "required": False,
+        },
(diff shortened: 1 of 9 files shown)


LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000006POTmAAO" target="_blank">W-018728</a>

cumulusci/tasks/apex/anon.py
```diff
@@ -1,35 +1,110 @@
...
         )
-        if result.status_code != 200:
-            raise SalesforceGeneralError(url, path, result.status_code, result.content)
+        self._check_result(result)
+
+        self.logger.info("Anonymous Apex Success")
+
+    def _prepare_apex(self, apex):
+        # Replace the email string only if it exists
+        admin_email = self.project_config.dev_config__admin_email
```
cdcarter: can we add this convention to the docs somewhere?

cumulusci/tasks/apex/anon.py
```diff
@@ -1,35 +1,110 @@
...
+        managed = self.options.get('managed') or False
+        namespaced = self.options.get('namespaced') or False
+        namespace = self.project_config.project__package__namespace
+        namespace_prefix = ''
+        record_type_prefix = ''
+        if managed or namespaced:
+            namespace_prefix = namespace + '__'
+        if namespaced:
+            record_type_prefix = namespace + '.'
+        apex = apex.replace('%%%NAMESPACE%%%', namespace_prefix)
```
cdcarter: run `black`

cumulusci/tasks/apex/anon.py
```diff
@@ -1,35 +1,110 @@
...
+        apex = apex.replace('%%%NAMESPACED_ORG%%%', namespace_prefix)
+        apex = apex.replace('%%%NAMESPACED_RT%%%', record_type_prefix)
```
cdcarter: other tasks take these tokens as options. we should probably do the same here, or consider adding these to the project_config. very magic stringy rn. but, a backlog item and not required for merge unless you want to do it now.

cumulusci/tasks/apex/anon.py
davisagli: Okay, actually got it set up to format on save now.

cumulusci/tasks/apex/anon.py
davisagli: I think I'll leave it until there's an explicit need. This is just copied over from what's in SAL and Cumulus.

cumulusci/tasks/apex/anon.py
cdcarter: a new block in `features.rst` called Anonymous Apex that explains this email behavior, as well as the concat behavior and how it is used to call functions of a shared file?

davisagli: I made the task doc generator include extra docs from a `task_docs` attribute on the class, and added a few pointers there. Sorry for putting this in the same commit that includes running black...

PR #743, opened 2018-09-11
Update github urls

davisagli:


cdcarter:
@davisagli we may need to do some work to get readthedocs updated again?

davisagli:
@cdcarter the readthedocs link seems to be still working fine.

cdcarter:
nice!

PR #744, opened 2018-09-13
Add CircleCI config

davisagli:
This sets up a Circle CI workflow that is similar to the existing Heroku CI script, but with several jobs running in parallel: unit tests, robot tests, and CumulusCI-Test release tests.

It also lays the groundwork for running separate python2 and python3 builds -- that just needs to get uncommented in the feature/python3 branch once this is merged.

Once this is merged to master we can follow up with decommissioning the heroku CI setup and removing heroku_ci.sh

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,7 +20,15 @@ jobs:
           - v1-dependencies-
 
       - run:
-          name: install dependencies
+          name: install sfdx
+          command: |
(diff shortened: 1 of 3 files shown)


davisagli:
Oh good catch @jlantz. I was thinking about cci org info entirely with regard to its side effect of creating the org and forgot it prints info.

PR #746, opened 2018-09-13
Set app visibility in update_admin_profile

jlantz:
and set tab visibility to DefaultOn


davisagli:
Looks like a good improvement, but I'm disappointed that there are no tests.

PR #748, opened 2018-09-14
Capture test results in CircleCI

jlantz:


.circleci/config.yml
```diff
@@ -45,8 +47,8 @@ jobs:
   #     - run:
   #         name: Run Python tests
   #         command: |
-  #           mkdir testresults
-  #           venv/bin/tox -e py36 -- --with-coverage --cover-package=cumulusci --with-xunit --xunit-file=testresults/junit.xml
+  #           mkdir -p testresults/nosetests
+  #           venv/bin/tox -e py36 -- --with-coverage --cover-package=cumulusci --with-xunit --xunit-file=testresults/nosetests/junit.xml
```
davisagli: The py3 result file should be named differently than the py2 result file

PR #753, opened 2018-09-17
Python 3 support

davisagli:
Not ready yet -- I'm opening the PR so that the CumulusCI-Test release will run in CI.


cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -224,10 +225,10 @@ def generate_password(self):
         p.run()
 
         stdout = []
-        for line in p.stdout:
+        for line in io.TextIOWrapper(p.stdout):
             stdout.append(line)
         stderr = []
-        for line in p.stderr:
+        for line in io.TextIOWrapper(p.stderr):
```
cdcarter: can we shrink these to the comprehension form used in scratch_info?

cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -282,7 +283,7 @@ def force_refresh_oauth_token(self):
         p.run()
 
         stdout_list = []
-        for line in p.stdout:
+        for line in io.TextIOWrapper(p.stdout):
```
cdcarter: comprehension

cumulusci/core/config/ScratchOrgConfig.py
cdcarter: actually, is there a way we can generalize this collecting of sarge stdout&stderr into arrays of lines into a function or somethin?

cumulusci/salesforce_api/tests/test_metadata.py
```diff
@@ -1,4 +1,12 @@
-import httplib
+from future import standard_library
+
+standard_library.install_aliases()
+from builtins import str
+import ast
+import http.client
+import os
+import shutil
+import tempfile
```
cdcarter: unused imports?

cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
```diff
@@ -49,11 +49,16 @@ def test_run_task(self, ApiRetrieveInstalledPackages):
...
-        with mock.patch.dict(
-            "cumulusci.tasks.salesforce.UpdateDependencies._install_dependency.__func__.__globals__",
-            download_extract_zip=mock.Mock(return_value=zf),
-        ):
+        try:
+            UD_globals = UpdateDependencies._install_dependency.__globals__
+        except AttributeError:  # Python 2
+            UD_globals = UpdateDependencies._install_dependency.__func__.__globals__
+        download_extract_zip = UD_globals["download_extract_zip"]
+        UD_globals["download_extract_zip"] = mock.Mock(return_value=zf)
```
cdcarter: we should probably restructure the task so that we don't need to do this much state-maintenance in a test

cumulusci/core/config/ScratchOrgConfig.py
davisagli: We can probably just do `stdout = io.TextIOWrapper(p.stdout).readlines()`

And sure, refactoring sounds good but let's do it separately

cumulusci/core/keychain/BaseEncryptedProjectKeychain.py
```diff
@@ -71,7 +71,19 @@ def _decrypt_config(self, config_class, encrypted_config, extra=None):
...
+            config_dict = pickle.loads(pickled)
+        else:  # Python 3
+            # Convert pickles created in Python 2
+            config_dict = {}
+            for k, v in unpickled.items():
+                if isinstance(k, bytes):
+                    k = k.decode("utf-8")
+                if isinstance(v, bytes):
+                    v = v.decode("utf-8")
+                config_dict[k] = v
```
davisagli: Note to self: there might be problems going the other direction too if you're trying to switch back and forth between Python 2 and Python 3 on the same machine.

cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
davisagli: The real problem is that the salesforce package imports the task class from a module of the same name so there isn't a good way for us to get the module to patch its download_extract_zip global.

requirements.txt
```diff
@@ -1,5 +1,4 @@
-arrow==0.12.1
-click==6.7
+click==7.0
 coloredlogs==10.0
 docutils==0.14
 future==0.16.0
```
cdcarter: do we need to add unicodecsv

davisagli: oh huh, it was already imported by the bulkdata module so I assumed it was here, but it must be getting pulled in transitively. yeah we should pin it

cumulusci/core/keychain/BaseEncryptedProjectKeychain.py
davisagli: @cdcarter I changed `_encrypt_config` to use pickle protocol 2 which Python 3 can read, and added similar code to translate unicode back to bytes when going in that direction. We may uncover more issues here but it worked going back and forth with a scratch org for me.

cdcarter: @davisagli - I'll give it a test on my very weird-organically-grown-many-years keychain on my main machine on Monday, see what I can and cannot roundtrip...

I'm struggling to find muuuch documentation on pickle protocol 3 as it wasn't introduced with a PEP but just as a part of py3k... I just want to sanity check that we aren't creating a footgun for ourselves in the future by continuing to serialize in protocol 2. Based on my understanding, we should be able to switch to protocol 3 without issue once py3 is a minimum requirement. Is there any reason to put the pickle protocol version into [env var|module global export|global yaml] for configuration or runtime inspection?

davisagli: My understanding is that the main change in protocol 3 was explicit support for serializing bytes as a distinct type from strings. I'm not too worried about using protocol 2; I wouldn't expect Python to remove support for it any time soon -- and yeah, I see this as a temporary measure until we update CumulusCI to only support Python 3.

As written this code is imperfect; it only translates top-level keys/values. Perhaps it should recurse.

jgerigmeyer:
@davisagli (cc @wlonk) MetaDeploy is currently setup to install this now-deleted branch, so the builds are failing. Should we update to install the `v2.1.0b1` tag?

davisagli:
@jgerigmeyer oops, sorry -- you can use the 2.1.0b1 release now instead of a git checkout

PR #754, opened 2018-09-17
Coverage for the rest of cumulusci.core

davisagli:
(except for the retry and polling parts of tasks.py)

Changes made in passing:
- Removed some unused code in BaseFlow
- Fixed EncryptedFileProjectKeychain._load_files when the key doesn't exist yet
- Fixed EncryptedFileProjectKeychain.remove_org to remove the org in the instance's config, not just on disk.

### DIFF
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -23,7 +23,7 @@ class BaseTask(object):
     code.
     """
 
-    task_docs = ''
+    task_docs = ""
     task_options = {}
     salesforce_task = False  # Does this task require a salesforce org?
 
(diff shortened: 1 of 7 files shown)


cumulusci/core/tasks.py
```diff
@@ -151,7 +151,7 @@ def _process_exception(self, e):
 
     def _run_task(self):
         """ Subclasses should override to provide their implementation """
-        pass
+        raise NotImplementedError("Subclasses should provide their own implementation")
```
cdcarter: I'm surprised this didn't cause any test failures, I sorta assumed we were using that implicit pass *somewhere* in the tests! awesome that we weren't.

anyway, this exception message doesn't provide anymore information that raising NotImplementedError bare does...

davisagli: I just copied it from what we're using elsewhere in this file.

PR #758, opened 2018-09-18
Improve coverage for misc tasks

davisagli:


cumulusci/tasks/apex/testrunner.py
```diff
@@ -265,19 +273,18 @@ def _get_test_results(self):
             self.logger.error("-" * 80)
             counter = 0
             for result in test_results:
-                if result["Outcome"] not in ["Fail", "CompileFail"]:
-                    continue
```
cdcarter: ... oh wow.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -308,11 +315,6 @@ def _run_task(self):
             url=self.tooling.base_url + "runTestsAsynchronous",
             json={"classids": ",".join(str(id) for id in ids)},
         )
-        if result.status_code != httplib.OK:
```
cdcarter: por que?

cumulusci/tasks/apex/testrunner.py
davisagli: The logic is the same as before (I inverted the condition and moved the code that was outside it to be inside) but this got me out of adding a second test case to the fixture just to get line coverage here.

cumulusci/tasks/apex/testrunner.py
davisagli: `_call_salesforce` already takes care of checking the status and raising SalesforceGeneralError

cumulusci/tasks/apex/testrunner.py
cdcarter: its smart!

cumulusci/tasks/apex/testrunner.py
```diff
@@ -374,9 +376,8 @@ def _write_output(self, test_results):
                     s += ">\n"
                     s += '    <failure type="failed" '
                     if result["Message"]:
-                        s += 'message="{}">'.format(cgi.escape(result["Message"]))
-                    else:
-                        s += ">"
```
davisagli: @cdcarter this was my other lazy coverage hack.

Once we get line coverage to 100% we can turn on branch coverage and it'll keep us honest.

cdcarter: i mean, this is also more aesthetically pleasing

davisagli: First time I've heard that phrase used to describe code that generates xml :)

PR #764, opened 2018-09-19
Coverage for cumulusci.tasks.{metadata,metaxml,mrbelvedere}

davisagli:


cumulusci/tasks/metadata/package.py
```diff
@@ -199,26 +190,23 @@ def parse_item(self, item):
 
     def _parse_item(self, item):
         "Receives a file or directory name and returns a list of members"
-        raise NotImplemented("Subclasses should implement their parser here")
+        raise NotImplementedError("Subclasses should implement their parser here")
 
     def strip_extension(self, filename):
-        return ".".join(filename.split(".")[:-1])
+        return filename.rsplit(".", 1)[0]
```
jstvz: Another possible, and slightly more verbose approach:
```python
base, ext = os.path.splitext(filename)
return base
```
Admittedly a quibble, but I have a personal preference towards avoiding numeric indexes.

cumulusci/tasks/metadata/package.py
```diff
@@ -396,17 +384,12 @@ class UpdatePackageXml(BaseTask):
 
     def _init_options(self, kwargs):
         super(UpdatePackageXml, self)._init_options(kwargs)
-        if "managed" not in self.options:
-            self.options["managed"] = False
-        if self.options["managed"] in [True, "True"]:
-            self.options["managed"] = True
+        self.options["managed"] = self.options.get("managed") in (True, "True", "true")
```
jstvz: :+1:

cumulusci/tasks/metadata/package.py
davisagli: I share that preference generally, but I'm less strict about it when it comes to 0 and -1 if the semantics are about getting the first/last element.

PR #770, opened 2018-09-20
Feature/push fails

cdcarter:
I didn't wire up the task in cumulusci.yml because its pretty rarely used and you can just do it in your local if you need to. Happy to add it to the global if we want...or strategize about how to register the mostly-releng-tasks....

Also ran black on the stuff that hasn't been formatted yet.

# Changes

- Adds a PushFailTask class that generates a report of push failures from a given request id.

### DIFF
--- a/cumulusci/tasks/push/pushfails.py
+++ b/cumulusci/tasks/push/pushfails.py
@@ -66,4 +66,4 @@ class PushFailTask(BaseSalesforceApiTask):
                 )
 
         self.logger.debug("Written out to {file_name}.".format(file_name=file_name))
-        return records
+        return file_name
--- a/cumulusci/tasks/push/pushfails.py
+++ b/cumulusci/tasks/push/pushfails.py
@@ -9,9 +9,10 @@ import csv
 from cumulusci.tasks.salesforce import BaseSalesforceApiTask
 
 
-class PushFailTask(BaseSalesforceApiTask):
-    """ Produce a report of the failed and otherwise anomalous push jobs from
-    a specified push request id, and write it out to a results CSV.  """
+class ReportPushFailures(BaseSalesforceApiTask):
+    """ Produce a report of the failed and otherwise anomalous push jobs.
+    
+    Takes a push request id and writes results to a CSV file.  """
 
     task_doc = __doc__
     task_options = {
@@ -23,7 +24,9 @@ class PushFailTask(BaseSalesforceApiTask):
     }
     api_version = "43.0"
     query = "SELECT ID, SubscriberOrganizationKey, (SELECT ErrorDetails, ErrorMessage, ErrorSeverity, ErrorTitle, ErrorType FROM PackagePushErrors) FROM PackagePushJob WHERE PackagePushRequestId = '{request_id}' AND Status !='Succeeded'"
-    gack = re.compile(r"error number: ([\d-]+) \(([\d-]+)\)")
+    gack = re.compile(
+        r"error number: (?P<gack_id>[\d-]+) \((?P<stacktrace_id>[\d-]+)\)"
+    )
     headers = [
         "OrganizationId",
         "ErrorSeverity",
@@ -51,6 +54,18 @@ class PushFailTask(BaseSalesforceApiTask):
             w = csv.writer(f)
             w.writerow(self.headers)
             for result in records:
+                if not result.get("PackagePushErrors", {}).get("records", None):
+                    w.writerow(
+                        [
+                            result["SubscriberOrganizationKey"],
+                            None,
+                            None,
+                            None,
+                            None,
+                            None,
+                        ]
+                    )
+                    continue
                 error = result["PackagePushErrors"]["records"][0]
                 m = self.gack.search(error["ErrorMessage"])
                 w.writerow(
@@ -60,8 +75,8 @@ class PushFailTask(BaseSalesforceApiTask):
                         error["ErrorTitle"],
                         error["ErrorType"],
                         error["ErrorMessage"],
-                        m.groups()[0] if m else "",
-                        m.groups()[1] if m else "",
+                        m.group("gack_id") if m else "",
+                        m.group("stacktrace_id") if m else "",
                     ]
                 )
 
--- a/cumulusci/tasks/push/pushfails.py
+++ b/cumulusci/tasks/push/pushfails.py
@@ -1,0 +1,69 @@
+""" simple task(s) for reporting on push upgrade jobs.
+
+this doesn't use the nearby push_api module, and was just a quick ccistyle
+get the job done kinda moment.
+"""
+
+import re
+import csv
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+
+class PushFailTask(BaseSalesforceApiTask):
+    """ Produce a report of the failed and otherwise anomalous push jobs from
+    a specified push request id, and write it out to a results CSV.  """
+
+    task_doc = __doc__
+    task_options = {
+        "request_id": {
+            "description": "PackagePushRequest ID for the request you need to report on.",
+            "required": True,
+        },
+        "result_file": {"description": "Where to write results"},
+    }
+    api_version = "43.0"
+    query = "SELECT ID, SubscriberOrganizationKey, (SELECT ErrorDetails, ErrorMessage, ErrorSeverity, ErrorTitle, ErrorType FROM PackagePushErrors) FROM PackagePushJob WHERE PackagePushRequestId = '{request_id}' AND Status !='Succeeded'"
+    gack = re.compile(r"error number: ([\d-]+) \(([\d-]+)\)")
+    headers = [
+        "OrganizationId",
+        "ErrorSeverity",
+        "ErrorTitle",
+        "ErrorType",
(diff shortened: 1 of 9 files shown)


cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+this doesn't use the nearby push_api module, and was just a quick ccistyle
+get the job done kinda moment.
+"""
+
+import re
+import csv
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+
+class PushFailTask(BaseSalesforceApiTask):
```
davisagli: How about `ReportPushFailures`. Elsewhere we haven't used the Task suffix (tasks are usually addressed via the `tasks` package so it seems redundant) but I don't have an actual problem with doing so.

cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+    """ Produce a report of the failed and otherwise anomalous push jobs from
+    a specified push request id, and write it out to a results CSV.  """
```
davisagli: PEP 257 (how do I know that number without looking it up? scary) recommends starting multi-line docstrings with a one-line summary followed by a blank line and then any details. So here that would look like:

```
"""Produce a report of unsuccessful push jobs.

Takes a push request id and writes results to a CSV file.
"""
```

I don't think we need to be real strict about this, but wanted to make sure you knew the Pythonic convention.

cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+
+        result = self.sf.query(formatted_query)
+        records = result["records"]
+        self.logger.debug(
+            "Query is complete: {done}. Found {n} results.".format(
+                done=result["done"], n=result["totalSize"]
+            )
+        )
+
+        file_name = self.options.get("result_file", "push_fails.csv")
```
davisagli: Maybe writing to stdout would be a better default?

cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+        "ErrorTitle",
+        "ErrorType",
+        "ErrorMessage",
+        "Gack Id",
+        "Stacktrace Id",
+    ]
+
+    def _run_task(self):
+        formatted_query = self.query.format(**self.options)
+        self.logger.debug("Running query: " + formatted_query)
```
davisagli: Nothing to worry about now, but we should probably build this into simple-salesforce or a wrapper thereof.

davisagli: Actually I have some MIT-licensed code I wrote for a former client that does this, as well as automatically retrying calls if the SF session expired.

cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+                done=result["done"], n=result["totalSize"]
+            )
+        )
+
+        file_name = self.options.get("result_file", "push_fails.csv")
+        with open(file_name, "wb") as f:
+            w = csv.writer(f)
+            w.writerow(self.headers)
+            for result in records:
+                error = result["PackagePushErrors"]["records"][0]
```
davisagli: Do you think there's any chance there would be an unsuccessful push but with no error records returned? I feel like I've seen one like that in the UI.

cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+                m = self.gack.search(error["ErrorMessage"])
+                w.writerow(
+                    [
+                        result["SubscriberOrganizationKey"],
+                        error["ErrorSeverity"],
+                        error["ErrorTitle"],
+                        error["ErrorType"],
+                        error["ErrorMessage"],
+                        m.groups()[0] if m else "",
+                        m.groups()[1] if m else "",
```
davisagli: This would be more readable if you use named groups in the regexp e.g. `(?P<gack_id>[\d-]+)` and then use m.group('gack_id'). Readability of the regexp itself is a lost cause, I guess...

cumulusci/tasks/push/pushfails.py
```diff
@@ -0,0 +1,69 @@
...
+                    ]
+                )
+
+        self.logger.debug("Written out to {file_name}.".format(file_name=file_name))
```
davisagli: I have a slight preference for `self.logger.debug("Written out to {}.".format(file_name))` because it feels less redundant, but I guess with your way it's easier to make changes to the string that change the order of params. Let's get to Python 3 so we can use f-strings :)

cumulusci/tasks/tests/test_pushfails.py
```diff
@@ -0,0 +1,66 @@
...
+    TaskConfig,
+    OrgConfig,
+)
+from cumulusci.utils import temporary_dir
+from cumulusci.core.keychain import BaseProjectKeychain
+from cumulusci.tests.util import get_base_config
+from cumulusci.tasks.push.pushfails import PushFailTask
+
+
+def ErrorRecord(gack=False):
```
davisagli: Can we use camel case so it doesn't look like a class?

cumulusci/tasks/tests/test_pushfails.py
```diff
@@ -0,0 +1,66 @@
...
+        task.sf = mock.Mock()
+        task.sf.query.return_value = {
+            "done": True,
+            "totalSize": 2,
+            "records": [ErrorRecord(), ErrorRecord(True)],
+        }
+        with temporary_dir() as d:
+            task()
+            task.sf.query.assert_called_once()
+            self.assert_(os.path.isfile(task.result), "the result file does not exist")
```
davisagli: `assert_` is documented as a deprecated alias for `assertTrue` so I'd probably use the latter

davisagli:
Re whether to register the task...it'd be nice if we had a way to group tasks in the list, and then maybe a way to hide a group unless you have a flag in your personal settings. In other words I want a way to say "show me all the releng tasks" rather than needing to configure each one. This is obviously a separate feature request though.

cumulusci/tasks/push/pushfails.py
cdcarter: I was thinking about a generic BaseQueryTask that gets subclassed like:
```
class PushReportFail(BaseQueryTask):
    task_doc = "<blah>"
    query = "<blah> {using_options}"
    def _query_results(self, result):
        pass
```
and maybe even prints to stdout by default? later tho...

davisagli: But then we're assuming that we only do queries from tasks, that the task does a single query and that it doesn't do other things apart from querying. I think it's better decoupling to build the logging into a query function that can be used from wherever.

davisagli: I guess we only really have an API to use within the context of a task though, so the first of those assumptions is reasonable.

cumulusci/tasks/push/pushfails.py
cdcarter: I tend to use the more verbose form, since it becomes easier to "lift" out either the string literal or the kwargs dict. more greppable. Etc... Python3 yes pls.

cumulusci/tasks/push/pushfails.py
cdcarter: some wrapper thereof is definitely an ineresting idea...

cumulusci/tasks/tests/test_pushfails.py
```diff
@@ -50,7 +50,7 @@ def setUp(self):
         self.task_config = TaskConfig({"options": {"request_id": "123"}})
 
     @mock.patch("cumulusci.tasks.push.pushfails.ReportPushFailures._update_credentials")
-    def test_run_task(self, *mocks):
+    def test_run_task(self, *_):
```
davisagli: pass `mock.Mock()` as a 2nd argument to `mock.patch` and then it won't try to pass it to the function.

cumulusci/tasks/tests/test_pushfails.py
```diff
@@ -67,7 +67,7 @@ def test_run_task(self, *mocks):
                 },
             ],
         }
-        with temporary_dir() as d:
+        with temporary_dir() as temp_dir:
```
davisagli: You don't have to give it a name at all since you're just using it for the side effect and not using it within the with block

cdcarter: right, originally i thought I might need it to read the file, but i'm letting the task give me the name anyway.

jlantz:
I do think we should register this task since we already have 4 push_* tasks and this use seems at least as useful to managed package projects as the other 4.

I agree we should have some support for task grouping at some point but am not too worried about including this one task in the interim.

cumulusci/tasks/push/pushfails.py
cdcarter: leaving this for now, because I want to keep having conversations about return value and result, and how they're treated in CLI. plus, in any use of thsi I can imagine right now, it involves uploading the file to a chatter group if theres an anonamly.

cdcarter:
@davisagli ready for rereview.

PR #771, opened 2018-09-20
Coverage for some salesforce tasks

davisagli:
I started factoring some of the repeated test setup into SalesforceTaskTestCase so I'm curious how you think that looks.

# Critical Changes

# Changes

# Issues Closed


cdcarter:
But approved if you disagree, cause I'd like to use even this in my current PR :)

davisagli:
I'm okay with factoring it out if it's useful elsewhere too, but then I think I'd rather make it a function rather than a mixin. We should also probably reconcile it with the existing `cumulusci.tests.util.create_project_config`

davisagli:
I'll merge this as is and then revisit the base test case class as I work on coverage for the rest of the salesforce tasks.

PR #775, opened 2018-09-21
Feature/retrypolltesting

cdcarter:
Literally just some tests for BaseTask._retry

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/tests/test_tasks.py
```diff
@@ -59,6 +59,35 @@ def setUp(self):
         self._task_log_handler.reset()
         self.task_log = self._task_log_handler.messages
 
+    @mock.patch("cumulusci.core.tasks.time.sleep", mock.Mock())
+    def test_retry_on_exception(self):
+        """ calling _retry() should call try until the task succeeds.  """
+        task = BaseTask(self.project_config, self.task_config, self.org_config)
+        task.options = {"retries": 5, "retry_interval": 1, "retry_interval_add": 1}
```
davisagli: It'd be safer to set the task_config's `config` attr before initializing the task...doesn't matter in this case but in general validating options is a thing that happens during task `__init__`

cumulusci/core/tests/test_tasks.py
```diff
@@ -59,6 +59,35 @@ def setUp(self):
...
+                RuntimeError(5),
+                RuntimeError(4),
+                RuntimeError(3),
+                RuntimeError(2),
+                RuntimeError(1),
+                Warning,
+            ]
+        )
+        with self.assertRaises(Warning):
+            task._retry()
```
davisagli: I think it would be clearer what this was testing if you used the same class of exception for the last one (and assert that the one with the correct `args` was raised). `Warning` in particular made me go do a minute of research because I hadn't seen it before and wanted to double check that it is in fact a subclass of Exception!

cumulusci/tests/test_utils.py
```diff
@@ -332,7 +332,7 @@ def test_package_xml_from_dict(self):
         )
 
 
-class TestTask(BaseTask):
+class FunTestTask(BaseTask):
```
davisagli: Is this just so it doesn't look like a test case? Sounds good to me. (And FUN!)

cdcarter: Yup, I was playing with pytest and the warning was driving me nuts.

cumulusci/core/tests/test_tasks.py
cdcarter: cool! i was hoping this was more obvious, but upon reflection I think your suggestion of just checking the args is good.

cumulusci/core/tests/test_tasks.py
cdcarter: i feel like I chose to do this, to make it more clear that the retry logic just uses the compiled task options dict and doesn't rely on any of that setup.

PR #776, opened 2018-09-21
Feature/bell

### LINKED ISSUE #325: terminal bell on task or flow complete
optionally (presumably set by the user in ~/.cumulusci.yml) print a "\a" character when task or flow completes successfully to alert the terminal

this is useful for long running tasks...

cdcarter:
# Changes
- Play a terminal "bell" sound when a task or flow completes execution. To silence this, you can set dev_config__no_alert in your local cumulusci.yml file.
- Attempt to display a notification on macOS also.
# Issues Closed
#325: terminal bell on task or flow complete

---
sample config:
*~/.cumulusci/cumulusci.yml*
```
dev_config:
    admin_email: christian.carter@salesforce.com
    no_alert: True
```

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1004,6 +1004,8 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
         # Unexpected exception; log to sentry and raise
         handle_exception_debug(config, debug, no_prompt=no_prompt)
 
+    config.alert("Task complete")
+
 
 # Add the task commands to the task group
 task.add_command(task_doc)
@@ -1115,6 +1117,8 @@ def flow_run(config, flow_name, org, delete_org, debug, o, skip, no_prompt):
                 "Scratch org deletion failed.  Ignoring the error below to complete the flow:"
             )
             click.echo(e.message)
+
+    config.alert("Flow Complete")
 
 
 flow.add_command(flow_list)
--- a/cumulusci/cli/config.py
+++ b/cumulusci/cli/config.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import click
+from subprocess import call
 
 import pkg_resources
 
@@ -60,13 +61,13 @@ class CliConfig(object):
             message = "We need your attention!"
         click.echo("\a")
         try:
-            os.system(
+            call(
                 """osascript -e 'display notification "{}" with title "{}"'""".format(
                     message, "CumulusCI"
                 )
             )
-        except:
-            pass
+        except OSError:
+            pass # we don't have oascript, probably.
 
     def get_org(self, org_name=None, fail_if_missing=True):
         if org_name:
--- a/cumulusci/cli/config.py
+++ b/cumulusci/cli/config.py
@@ -50,6 +50,24 @@ class CliConfig(object):
             self.keychain = self.keychain_class(self.project_config, self.keychain_key)
             self.project_config.set_keychain(self.keychain)
 
+    def alert(self, message=None):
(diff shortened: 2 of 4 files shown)


cumulusci/cli/config.py
```diff
@@ -50,6 +50,24 @@ def _load_keychain(self):
...
+        if message is None:
+            message = "We need your attention!"
+        click.echo("\a")
+        try:
+            os.system(
+                """osascript -e 'display notification "{}" with title "{}"'""".format(
+                    message, "CumulusCI"
+                )
+            )
+        except:
```
davisagli: Can we catch a specific exception here, or at least `Exception` so it won't catch KeyboardInterrupt?

cumulusci/cli/config.py
```diff
@@ -50,6 +50,24 @@ def _load_keychain(self):
             self.keychain = self.keychain_class(self.project_config, self.keychain_key)
             self.project_config.set_keychain(self.keychain)
 
+    def alert(self, message=None):
+        try:
+            if self.project_config.dev_config__no_alert:
+                return
+        except AttributeError:
+            pass
```
davisagli: What's the case where there's an AttributeError here? Is there test coverage for that case? Can we avoid it?

cumulusci/cli/cci.py
```diff
@@ -1004,6 +1004,8 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
         # Unexpected exception; log to sentry and raise
         handle_exception_debug(config, debug, no_prompt=no_prompt)
 
+    config.alert("Task complete")
```
davisagli: Is this going to start beeping at me when I run the tests for `task_run` and `flow_run`?

cumulusci/cli/config.py
cdcarter: if dev_config or no_alert are not yet defined. the test where the alert "succeeds" goes down the except & pass.

cumulusci/cli/cci.py
cdcarter: ran the full suite with both py.test and nose, and both suppressed the bell, presumably because they capture stdout.

cumulusci/cli/config.py
cdcarter: it turns out os.system can never throw an exception unless its called improperly. and for that and some other reasons, its a bad idea. I'll replace w/ subprocess.

cumulusci/cli/cci.py
davisagli: I often run nose with the -s flag to un-suppress stdout while working on tests...

cumulusci/cli/config.py
davisagli: > can never throw an exception

wait, isn't that what we want here? what are the other reasons?

cumulusci/cli/cci.py
cdcarter: resolution: after some testing, though we're not sure *why*, even with -s, the task_run and flow_run tests don't cause the bell. so not super alarmed about this. but if we ever need to, let's just wrap the echo in a partial so its mockable.

cumulusci/cli/config.py
davisagli: The except clause is showing as uncovered in coveralls. I think because BaseConfig's `__getattr__` returns None when a key isn't found rather than raising AttributeError. I think we just need to test the case where project_config is not set yet.

The next condition (`if message is None: message = "We need your attention!"`) is also uncovered. I would just make `message` a required arg for the function.

PR #777, opened 2018-09-21
Coverage for more salesforce tasks

### LINKED ISSUE #774: test fails on forks of the repo
https://github.com/SFDO-Tooling/CumulusCI/blob/57e1bcb4d8ebffdca3ddd554ecb3a3306218b272/cumulusci/core/tests/test_config.py#L337

This test requires to be running in a checkout of SFDO-Tooling/CumulusCI, which isn't the case for an external fork.

davisagli:
Also fixes #774 and refactors SalesforceTaskTestCase like we discussed, @cdcarter

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/UninstallPackagedIncremental.py
+++ b/cumulusci/tasks/salesforce/UninstallPackagedIncremental.py
@@ -1,11 +1,11 @@
 import os
-import shutil
-import tempfile
 
 import xmltodict
 
+from cumulusci.core.utils import process_bool_arg
 from cumulusci.tasks.salesforce import UninstallPackaged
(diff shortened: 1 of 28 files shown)


cumulusci/tasks/salesforce/UninstallPackagedIncremental.py
```diff
@@ -1,11 +1,11 @@
 import os
-import shutil
-import tempfile
 
 import xmltodict
 
+from cumulusci.core.utils import process_bool_arg
```
cdcarter: not today...but...we should move this into cumulusci.utils, not core.utils

PR #787, opened 2018-09-26
Feature/safe

cdcarter:
# Critical Changes

# Changes
- Use the `safe_load()` method when loading YAML. If you relied on executing code in your CCI YAML file parsing, that will no longer work.
# Issues Closed


jstvz:
Oh, forgot to ask: is this a critical change?

cdcarter:
I don't think so! YAML tags that end up running python code are rare/probably intentionally evil.

PR #790, opened 2018-09-27
Feature/deps

cdcarter:
Some stuff in here that has opinions. Other solutions welcomed.

# Critical Changes

# Changes
- Remove our runtime dependency on Arrow, and replace its usage in the Batch Apex task.
- Parse  DateTime fields as `datetime`s when listing metadata components from the org.
- Update `click` to the latest version.
- Move `responses` into test dependencies.

# Issues Closed

### DIFF
--- a/cumulusci/salesforce_api/metadata.py
+++ b/cumulusci/salesforce_api/metadata.py
@@ -5,14 +5,12 @@ based on mrbelvedere/mpinstaller/mdapi.py
 
 # TO DO
 #   - add docstrings
-#   - parse dates from SOAP response
 #   - use format() instead of %
 #   - look at https://github.com/rholder/retrying
 
 from __future__ import unicode_literals
 import base64
 
-# import dateutil.parser
 import httplib
 import re
 import time
@@ -26,7 +24,7 @@ import requests
 
 from cumulusci.salesforce_api import soap_envelopes
 from cumulusci.core.exceptions import ApexTestException
-from cumulusci.utils import zip_subfolder
+from cumulusci.utils import zip_subfolder, parse_api_datetime
 from cumulusci.salesforce_api.exceptions import MetadataComponentFailure
 from cumulusci.salesforce_api.exceptions import MetadataApiError
 
@@ -562,10 +560,16 @@ class ApiListMetadata(BaseMetadataApiCall):
             for tag in tags:
                 result_data[tag] = self._get_element_value(result, tag)
             # Parse dates
-            # FIXME: This was breaking things
-            # for key in parse_dates:
-            #    if result_data[key]:
-            #        result_data[key] = dateutil.parser.parse(result_data[key])
+            for key in parse_dates:
+                if result_data[key]:
+                    try:
+                        result_data[key] = parse_api_datetime(result_data[key])
+                    except Exception: 
+                        # parsing the datetime with the dateutil library was "causing problems"
+                        # according to an old comment without much detail, so i'm being 
+                        # unnecessarily defensive here. 
+                        # TODO: determine if this is needed.
+                        pass
             metadata.append(result_data)
         self.metadata[self.metadata_type].extend(metadata)
         return self.metadata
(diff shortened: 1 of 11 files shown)


cumulusci/utils.py
```diff
@@ -21,6 +22,23 @@
...
+def parse_api_datetime(value):
+    """ parse a datetime returned from the salesforce API.
+
+    in python 3 we should just use a strptime %z, but until then we're just going
+    to assert that its a fixed offset of +0000 since thats the observed behavior. getting
+    python 2 to support fixed offset parsing is too complicated for what we need imo."""
+    dt = datetime.strptime(value[0:DATETIME_LEN], API_DATE_FORMAT)
+    offset_str = value[DATETIME_LEN:]
+    assert offset_str == "+0000", "The Salesforce API returned a weird timezone."
+    return dt
```
davisagli: I take it you checked that the API only returns +0000 in practice?

cumulusci/salesforce_api/metadata.py
```diff
@@ -562,10 +560,16 @@ def _process_response(self, response):
...
+            for key in parse_dates:
+                if result_data[key]:
+                    try:
+                        result_data[key] = parse_api_datetime(result_data[key])
+                    except Exception:
+                        # parsing the datetime with the dateutil library was "causing problems"
+                        # according to an old comment without much detail, so i'm being
+                        # unnecessarily defensive here.
+                        # TODO: determine if this is needed.
+                        pass
```
davisagli: I think I would lean toward not being defensive, but raising an exception including the value that couldn't be parsed. That way we'll actually find out what kind of values we aren't handling properly rather than silently pretending that data doesn't exist.

cumulusci/tasks/apex/tests/test_apex_tasks.py
```diff
@@ -459,3 +459,11 @@ def test_run_batch_apex_status_ok(self):
...
         responses.add(responses.GET, url, json=response)
         task()
+
+    @responses.activate
+    def test_run_batch_apex_calc_delta(self):
+        task, url = self._get_url_and_task()
+        response = self._get_query_resp()
+        responses.add(responses.GET, url, json=response)
+        task()
+        self.assertAlmostEquals(task.delta, 60, delta=2)
```
davisagli: Why can't assertEqual here?

cumulusci/utils.py
```diff
@@ -21,6 +22,23 @@
...
+
+def parse_api_datetime(value):
+    """ parse a datetime returned from the salesforce API.
+
+    in python 3 we should just use a strptime %z, but until then we're just going
+    to assert that its a fixed offset of +0000 since thats the observed behavior. getting
+    python 2 to support fixed offset parsing is too complicated for what we need imo."""
+    dt = datetime.strptime(value[0:DATETIME_LEN], API_DATE_FORMAT)
+    offset_str = value[DATETIME_LEN:]
+    assert offset_str == "+0000", "The Salesforce API returned a weird timezone."
```
davisagli: I've heard it recommended to not use `assert` in non-test code because the Python compiler removes assert statements when run with -O. Not that that's likely though...

cumulusci/utils.py
```diff
@@ -21,6 +22,23 @@
     os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
 )
 META_XML_CLEAN_DIRS = ("classes/", "triggers/", "pages/", "aura/", "components/")
+ZERO = timedelta(0)
```
davisagli: This looks unused

cumulusci/utils.py
```diff
@@ -21,6 +22,23 @@
...
+API_DATE_FORMAT = "%Y-%m-%dT%H:%M:%S.%f"
+DATETIME_LEN = (
+    4 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 2 + 1 + 3
+)  # the actual width of a datetime string formatted with above format
```
davisagli: I think this would be clearer and not need the comment if written like this:

```DATETIME_LEN = len("2018-08-07T16:00:56.000+0000")```

cumulusci/utils.py
cdcarter: I guess that makes this just a silent failure if you're running with asserts off. Idk, I got the practice from DRF but I'm not tied to it.

cumulusci/tasks/apex/tests/test_apex_tasks.py
cdcarter: no reason :)

cumulusci/utils.py
cdcarter: and Z, patch incoming

PR #794, opened 2018-10-02
Skip prereleases and postreleases in version check

davisagli:
# Critical Changes

# Changes
* Skip beta releases when checking for a newer cumulusci version

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -65,13 +65,16 @@ def timestamp_file():
         yield f
 
 
-def get_latest_version():
+def get_latest_final_version():
     """ return the latest version of cumulusci in pypi, be defensive """
     # use the pypi json api https://wiki.python.org/moin/PyPIJSON
     res = requests.get("https://pypi.org/pypi/cumulusci/json", timeout=5).json()
     with timestamp_file() as f:
         f.write(bytes(time.time()))
-    return pkg_resources.parse_version(res["info"]["version"])
+    versions = [pkg_resources.parse_version(v) for v in res["releases"].keys()]
+    versions = [v for v in versions if not v.is_prerelease and not v.is_postrelease]
+    versions.sort(reverse=True)
+    return versions[0]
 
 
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cci.py
```diff
@@ -65,13 +65,16 @@ def timestamp_file():
         yield f
 
 
-def get_latest_version():
+def get_latest_final_version():
```
cdcarter: is "final" a PyPI term? nothing about this is...final...per se...

cumulusci/cli/cci.py
```diff
@@ -65,13 +65,16 @@ def timestamp_file():
...
     """ return the latest version of cumulusci in pypi, be defensive """
```
cdcarter: during the refactor the "be defensive" got moved into the calling method. update docstring plz

cumulusci/cli/cci.py
```diff
@@ -65,13 +65,16 @@ def timestamp_file():
...
     # use the pypi json api https://wiki.python.org/moin/PyPIJSON
     res = requests.get("https://pypi.org/pypi/cumulusci/json", timeout=5).json()
     with timestamp_file() as f:
         f.write(bytes(time.time()))
-    return pkg_resources.parse_version(res["info"]["version"])
+    versions = [pkg_resources.parse_version(v) for v in res["releases"].keys()]
```
cdcarter: we already have a LOT of cci releases and more coming, is there any benefit to only parsing and checking the last N? or is this all so fast we dn't care

cumulusci/cli/cci.py
davisagli: It's from PEP 440: https://www.python.org/dev/peps/pep-0440/#final-releases

cumulusci/cli/cci.py
davisagli: Since they seem to be sorted alphabetically and we can have many betas, I'm not sure that's safe.

PR #796, opened 2018-10-03
Ignore push errors that aren't problematic by default

davisagli:
# Critical Changes

# Changes

* In the push failure report, ignore "Package Uninstalled" and "Salesforce Subscription Expired" errors by default.

# Issues Closed


cdcarter:
Smart. Love that it's YAML only, good thinkin ahead.

davisagli:
I meant to do it when I added the option but it got lost in the shuffle.

PR #797, opened 2018-10-03
Big refactor of bulk data tasks

davisagli:
All right, this needs a few more things done but it's kinda ginormous (sorry) so not too early for review to start. Note that this branch is based on `feature/python3`.

Includes:

* Fixed connection resets by downloading an entire result file before processing
* Performance improvements:
  * For postgresql, bulk insert using COPY FROM
  * For other dialects, bulk insert using raw SQL instead of ORM
  * Process batches in parallel
  * Store inserted ids in a separate table to avoid slow updates
  * Perform lookups using SQL joins instead of a separate query for each row
  * For deletes, use the maximum batch size (10000)
* Python 3 compatibility
* Added `operation` option for DeleteData which can be "delete" (default) or "hardDelete"
* Added `start_step` option for LoadData which can be used to resume loading after an error

Any thoughts about how the above should be summarized in the Changes section below?

Things that definitely need attention before calling it done:
* Docs

Things that would be nice to have:
* Multi-tenant data loads (add an org id column to the tables holding inserted SF ids)
* Probably add these tasks to the default cumulusci.yml

Maybe someday:
* Autogenerate a mapping from describe calls

# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/bulkdata.py
```diff
@@ -1,63 +1,119 @@
...
+            else:  # Queued, InProgress
+                pending += 1
+        if pending:
+            return "InProgress"
+        elif failed:
+            return "Failed"
+        else:
+            return "Completed"
+
+    def _wait_for_job(self, job_id):
```
cdcarter: should this use the _poller from basetask? if its a "yes but it doesnt do X" can we capture X in an issue? we know we're gonna need to update the poller a bit so if we can capture requests thats nice

cumulusci/tasks/bulkdata.py
```diff
@@ -1,63 +1,119 @@
...
+            time.sleep(10)
+        self.logger.info("Job {} finished with result: {}".format(job_id, result))
+        return result
+
+    def _sql_bulk_insert_from_csv(self, conn, table, columns, data_file):
+        if conn.dialect.name == "psycopg2":
+            # psycopg2 (the postgres driver) supports COPY FROM
+            # to efficiently bulk insert rows in CSV format
+            with conn.connection.cursor() as cursor:
+                cursor.copy_expert(
```
cdcarter: this is an...amusingly named method in psycopg2?

cumulusci/tasks/bulkdata.py
```diff
@@ -1,63 +1,119 @@
...
+                    "COPY {} ({}) FROM STDIN WITH (FORMAT CSV)".format(
+                        table, ",".join(columns)
```
cdcarter: is there any reason we should be using a csvwriter instead of a naive string.join?

cumulusci/tasks/bulkdata.py
```diff
@@ -1,63 +1,119 @@
...
+                    ),
+                    data_file,
+                )
+        else:
+            # For other db drivers we need to use standard SQL
+            # -- this is optimized for ease of implementation
+            # rather than performance and may need more work.
+            reader = unicodecsv.DictReader(data_file, columns)
+            table = self.metadata.tables[table]
+            rows = list(reader)
```
cdcarter: damn guido for not making iterator.has_next. yes, this is naive but probably fine for the 100row world.

cumulusci/tasks/bulkdata.py
```diff
@@ -1,63 +1,119 @@
...
-class DeleteData(BaseSalesforceApiTask):
+class DeleteData(BaseSalesforceApiTask, BulkJobTaskMixin):
 
     task_options = {
         "objects": {
             "description": "A list of objects to delete records from in order of deletion.  If passed via command line, use a comma separated string",
             "required": True,
-        }
+        },
+        "operation": {
```
cdcarter: I'd rather see this as a soft_delete boolean option that defaults to false. is there a reason to allow a string operation name?

cumulusci/tasks/bulkdata.py
```diff
@@ -69,59 +125,60 @@ def _init_options(self, kwargs):
...
+        if not delete_rows:
+            self.logger.info("  No {} objects found, skipping delete".format(obj))
+            return
+
+        # Upload all the batches
+        delete_job = self.bulk.create_job(obj, self.options["operation"])
+        self.logger.info("  Deleting {} {} records".format(len(delete_rows), obj))
+        batch_num = 1
+        for batch in self._upload_batches(delete_job, delete_rows):
+            self.logger.info("    Uploaded batch {}".format(batch))
```
cdcarter: i'd love to have a better way of doing these logger indents but...

cumulusci/tasks/bulkdata.py
```diff
@@ -69,59 +125,60 @@ def _init_options(self, kwargs):
                 obj.strip() for obj in self.options["objects"].split(",")
             ]
 
+        self.options["operation"] = self.options.get("operation") or "delete"
+
     def _run_task(self):
```
cdcarter: thanks for pulling most of this out of run_task, it reads much nicer.

cumulusci/tasks/bulkdata.py
```diff
@@ -131,7 +188,7 @@ def _upload_batch(self, job, data):
             yield batch_id
 
 
-class LoadData(BaseSalesforceApiTask):
+class LoadData(BaseSalesforceApiTask, BulkJobTaskMixin):
```
cdcarter: I would have expected the mixin to come before the base, is this for a reason?

cumulusci/tasks/bulkdata.py
```diff
@@ -142,107 +199,55 @@ class LoadData(BaseSalesforceApiTask):
             "description": "The path to a yaml file containing mappings of the database fields to Salesforce object fields",
             "required": True,
         },
+        "start_step": {
```
cdcarter: doesn't block merge: as we develop more tasks that have their internal control flow I'd like to make sure we do it consistently

cumulusci/tasks/bulkdata.py
```diff
@@ -142,107 +199,55 @@ class LoadData(BaseSalesforceApiTask):
...
+                break
 
-    def _sobject_api_upload_batches(self, mapping, batches):
-        for batch, batch_rows in batches:
-            pass
+    def _load_mapping(self, mapping):
+        """Load data for a single step."""
+        job_id, local_ids_for_batch = self._create_job(mapping)
+        result = self._wait_for_job(job_id)
+        if result == "Completed":
```
cdcarter: i feel like we might wanna be EAFPy here?

cumulusci/tasks/bulkdata.py
```diff
@@ -254,87 +259,179 @@ def _get_batches(self, mapping, batch_size=None):
...
+            with _download_file(results_url, self.bulk) as f:
+                self.logger.info("  Downloaded results for batch {}".format(batch_id))
+                self._store_inserted_ids_for_batch(f, local_ids, id_table_name, conn)
+            self.logger.info(
+                "  Updated {} for batch {}".format(id_table_name, batch_id)
+            )
+        self.session.commit()
+
+    def _reset_id_table(self, mapping):
+        """Create an empty table to hold the inserted SF Ids"""
```
cdcarter: create a temp table to hold results? why is it called "reset"?

cumulusci/tasks/bulkdata.py
```diff
@@ -501,12 +582,54 @@ def _create_table(self, mapping):
...
+    if record_type and not isinstance(record_type, bytes):
+        record_type = record_type.encode("utf-8")
+    for line in f:
+        if record_type:
+            yield line + b"," + record_type + b"\n"
+        else:
+            yield line
+
+
+def log_progress(iterable, logger, batch_size=10000):
```
cdcarter: i think there miiight be a click helper for this?

cumulusci/tasks/bulkdata.py
davisagli: Oh, probably -- well, I'd like to turn the poller into a separate utility function or class.

cumulusci/tasks/bulkdata.py
davisagli: It's the version of postgres' COPY command that lets you spell out the full command, as opposed to `copy_from` and `copy_to` which don't expose the FORMAT CSV option.

cumulusci/tasks/bulkdata.py
davisagli: I don't really want task code to depend on click though.

cumulusci/tasks/bulkdata.py
cdcarter: > Sure sounds like tasks are a kind of flow ;-)

Sure does. This...update dependencies... they both have substeps that we want to give progress on....

cumulusci/tasks/bulkdata.py
cdcarter: throw this in tasks.utils then

PR #799, opened 2018-10-03
Avoid using is_prerelease to check version since it's missing in old setuptools

### LINKED ISSUE #798: AttributeError: 'tuple' object has no attribute 'is_prerelease'
We had a report of this error. It's from cumulusci.cci.get_latest_final_version and I think it is due to an old setuptools

davisagli:
# Critical Changes

# Changes

# Issues Closed

#798

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -70,6 +70,11 @@ FINAL_VERSION_RE = re.compile(r"^[\d\.]+$")
 
 
 def is_final_release(version):
+    """Returns bool whether version string should be considered a final release.
+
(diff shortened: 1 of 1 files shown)


cumulusci/cli/cci.py
```diff
@@ -65,14 +66,24 @@ def timestamp_file():
         yield f
 
 
+FINAL_VERSION_RE = re.compile(r"^[\d\.]+$")
+
+
+def is_final_release(version):
```
cdcarter: docstring that this is based on PEP440 please

PR #805, opened 2018-10-05
Prepare 2.1.0b1

davisagli:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,22 @@
 History
 =======
 
+2.1.0b1 (2018-10-05)
+--------------------
+
+* It's happening! Hot on the heels of the last release, CumulusCI is making the jump to the modern era by adding **support for Python 3**! (Specifically, Python 3.6 and 3.7.) Don't worry, we'll also continue to support Python 2 until it reaches end-of-life at the start of 2020. Because this is a bit more wide-reaching change than normal, we're releasing a beta first. To install the beta you'll need to explicitly request its version: ``pip install cumulusci==2.1.0b1`.
+  If you want to *replace* an existing Python 2-based installation of CumulusCI, we recommend deleting your virtualenv and starting over with the instructions in the `tutorial <https://cumulusci.readthedocs.io/en/latest/tutorial.html>`_.  If you want to keep your Python 2-based virtualenv around just in case, follow those instructions but name the new virtualenv ``cci-py3`` instead of ``cci``.
+* There are also some big changes to the **bulk data** tasks. Did you know CumulusCI has bulk data tasks? They are not configured by default, because we need to finish documenting them. But we'll list the changes in case someone is already relying on them:
+  * Fixed connection resets by downloading an entire result file before processing.
+  * Improved performance by processing batches in parallel, avoiding the SQLAlchemy ORM, storing inserted Ids in separate tables, and doing lookups using SQL joins rather than a separate query for each row.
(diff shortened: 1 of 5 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,22 @@
 History
 =======
 
+2.1.0b1 (2018-10-05)
+--------------------
+
+* It's happening! Hot on the heels of the last release, CumulusCI is making the jump to the modern era by adding **support for Python 3**! (Specifically, Python 3.6 and 3.7.) Don't worry, we'll also continue to support Python 2 until it reaches end-of-life at the start of 2020. Because this is a bit more wide-reaching change than normal, we're releasing a beta first. To install the beta you'll need to explicitly request its version: ``pip install cumulusci==2.1.0b1``.
```
cdcarter: (Specifically, 3.6+.) ?

cdcarter: Is there a way to `pip install cumulusci>=2.1.0b1` (or something) and the `pip install -U cumulusci` keeps you on the beta track? probably not but...

HISTORY.rst
```diff
@@ -2,6 +2,22 @@
...
+  If you want to *replace* an existing Python 2-based installation of CumulusCI, we recommend deleting your virtualenv and starting over with the instructions in the `tutorial <https://cumulusci.readthedocs.io/en/latest/tutorial.html>`_.  If you want to keep your Python 2-based virtualenv around just in case, follow those instructions but name the new virtualenv ``cci-py3`` instead of ``cci``.
```
cdcarter: "replace an existing python2-based installation" is confusing if you don't really understand how python venvs and installations work. how about

> After the update, CumulusCI will continue to run under your Python 2 interpreter. If you want to switch to the Python 3 interpreter, (which is not yet required), we recommend deleteing...

HISTORY.rst
```diff
@@ -2,6 +2,22 @@
...
+* There are also some big changes to the **bulk data** tasks. Did you know CumulusCI has bulk data tasks? They are not configured by default, because we need to finish documenting them. But we'll list the changes in case someone is already relying on them:
+  * Fixed connection resets by downloading an entire result file before processing.
+  * Improved performance by processing batches in parallel, avoiding the SQLAlchemy ORM, storing inserted Ids in separate tables, and doing lookups using SQL joins rather than a separate query for each row.
```
cdcarter: is the prior bullet worth its own bullet or does it just go here?

HISTORY.rst
```diff
@@ -2,6 +2,22 @@
...
+  * If you're using a postgres database for local storage, performance gets even better by taking advantage of postgres' ``COPY`` command to load CSV directly.
+  * Added a ``hardDelete`` option for bulk deletes.
+  * Added a ``start_step`` option for bulk loads which can be used to resume loading after an error.
+* The ``push_failure_report`` task will now by default hide failures that occurred due to the "Package Uninstalled" or "Salesforce Subscription Expired" errors, which are generally benign.
+* Fixed the check for newer CumulusCI versions to work around an issue with old ``setuptools``.
+* Internal change: We switched CumulusCI's own tests to run using ``pytest``.
```
cdcarter: Contributor change?

cdcarter:
some of these thoughts might just need to be resolved before we really do it.

HISTORY.rst
davisagli: The first bullet was about making it not break, the second bullet is about making it fast.

PR #810, opened 2018-10-08
Add an UpdateDependencies option for which dependencies to install

davisagli:
# Critical Changes

# Changes
* Add a `dependencies` option to the `update_dependencies` task to override the project dependencies.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/UpdateDependencies.py
+++ b/cumulusci/tasks/salesforce/UpdateDependencies.py
@@ -1,5 +1,4 @@
 from builtins import str
-from past.builtins import basestring
 from distutils.version import LooseVersion
 
 from cumulusci.core.utils import process_bool_arg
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/salesforce/UpdateDependencies.py
```diff
@@ -21,6 +20,9 @@ class UpdateDependencies(BaseSalesforceMetadataApiTask):
     api_class = ApiDeploy
     name = "UpdateDependencies"
     task_options = {
+        "dependencies": {
+            "description": "List of dependencies to update. Defaults to project__dependencies"
```
cdcarter: in what form?

davisagli: Same form as `project__dependencies` (a list of dicts specifying either a github url or namespace). I wasn't sure how to explain it clearly without giving an example, which there isn't really room for here.

davisagli: Better now?

PR #820, opened 2018-10-12
Feature/threadlocal

cdcarter:
# Critical Changes

# Changes
- Internal change: use a thread-local variable instead of a global to track the current running task.

# Issues Closed

### DIFF
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -9,11 +9,12 @@ from builtins import object
 from past.utils import old_div
 import logging
 import time
+import threading
 
 from cumulusci.core.exceptions import TaskRequiresSalesforceOrg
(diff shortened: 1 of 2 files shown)


cumulusci/core/tasks.py
```diff
@@ -9,11 +9,13 @@
...
 import logging
 import time
+import threading
 
 from cumulusci.core.exceptions import TaskRequiresSalesforceOrg
 from cumulusci.core.exceptions import TaskOptionsError
 
-CURRENT_TASK = None
+CURRENT_TASK = threading.local()
+CURRENT_TASK.instance = None
```
davisagli: Just to confirm, we don't have any other packages using this with the assumption that it's a task rather than an object with an instance member, right?

PR #825, opened 2018-10-17
Add an 'ignore' option for UninstallPackagedIncremental

davisagli:
I intend to use this to avoid trying to delete quick actions that get implicitly added to the package in DE orgs but can't be deleted because they are referenced from layouts.

# Critical Changes

# Changes
* ``uninstall_packaged_incremental`` task: Added "ignore" option to specify components to skip trying to delete even if they are present in the org but not in the local source.

# Issues Closed


davisagli:
**lurch: attach W-028731

PR #829, opened 2018-10-18
robot keyword fixes and improvements

davisagli:
# Critical Changes

# Changes
* robot keyword fixes and improvements:
  - ``Wait Until Loading Is Complete`` now waits for particular parts of the page to render
  - ``Populate Lookup Field`` now tries several times in case there's an indexing delay
  - new ``Click Header Field Link`` keyword

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -1,10 +1,13 @@
 import logging
 import re
+import time
 from robot.libraries.BuiltIn import BuiltIn
+from selenium.webdriver.common.keys import Keys
 from simple_salesforce import SalesforceMalformedRequest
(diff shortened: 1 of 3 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -166,6 +169,10 @@ def header_field_should_not_have_link(self, label):
         locator = lex_locators["record"]["header"]["field_value_link"].format(label)
         self.selenium.page_should_not_contain_element(locator)
 
+    def click_header_field_link(self, label):
```
jlantz: We should add a test in the ui.robot suite for this keyword

davisagli: Done

PR #833, opened 2018-10-22
Add linux notifications

jstvz:
gnome-notification
gnome-notification-center

# Critical Changes

# Changes
- Add linux notifications upon flow/task completion via `libnotify`.
# Issues Closed


cumulusci/cli/config.py
```diff
@@ -67,18 +67,25 @@ def alert(self, message="We need your attention!"):
...
-                        message.replace('"', r"\"").replace("'", r"\'"), "CumulusCI"
-                    ),
-                ]
-            )
+            call(self._get_platform_alert_cmd(message))
         except OSError:
             pass  # we don't have osascript, probably.
 
+    def _get_platform_alert_cmd(self, message):
+        message = message.replace('"', r"\"").replace("'", r"\'")
```
davisagli: I think we only want this escaping for interpolation into the osascript statement, not for notify-send where it's passed verbatim as an arg (since we're using subprocess.call without shell=True)

cumulusci/cli/config.py
```diff
@@ -67,18 +67,24 @@ def alert(self, message="We need your attention!"):
...
+        if sys.platform == "darwin":
+            return [
+                "osascript",
+                "-e",
+                'display notification "{}" with title "{}"'.format(
+                    message.replace('"', r"\"").replace("'", r"\'"), "CumulusCI"
+                ),
+            ]
+        elif sys.platform.startswith("linux"):
+            return ["notify-send", "--icon=utilities-terminal", "CumulusCI", message]
```
cdcarter: What are our icon choices? 😅

cdcarter:
Yippee!

cumulusci/cli/config.py
jstvz: https://standards.freedesktop.org/icon-naming-spec/icon-naming-spec-latest.html

I was thinking that `face-cool` ( 😎 ), but wanted to keep it profesh 😏

cumulusci/cli/config.py
```diff
@@ -67,18 +67,24 @@ def alert(self, message="We need your attention!"):
...
+        else:
+            raise OSError
```
davisagli: Can you either test or nocover this line?

PR #834, opened 2018-10-22
Explicitly pin transitive dependencies to fix #832

### LINKED ISSUE #832: CumulusCI commands fail with Codeship after upgrading CumulusCI
We have started receiving the attached errors while we try to run CumulusCI commands in Codeship. This is causing all over builds to fail. We need this fixed asap since this is breaking our entire development flow and we have to deploy the files manually to other orgs.

screen shot 2018-10-22 at 12 07 22 pm
screen shot 2018-10-22 at 12 07 43 pm

davisagli:
Also make setup.py load the requirements from the pip requirements files to avoid needing to update 2 places

# Critical Changes

# Changes

# Issues Closed
#832


requirements.txt
```diff
@@ -18,6 +27,10 @@ salesforce-bulk==2.1.0
 sarge==0.1.5.post0
 selenium==3.14.0
 simple-salesforce==0.74.2
+six==1.11.0
```
cdcarter: oh nice, I didn't realize this was in the closure, very handy...

requirements.txt
```diff
@@ -18,6 +27,10 @@ salesforce-bulk==2.1.0
...
 SQLAlchemy==1.2.11
 unicodecsv==0.14.1
+uritemplate==3.0.0
+uritemplate.py==3.0.2
```
cdcarter: both? shrug.

PR #837, opened 2018-10-22
config refactor

cdcarter:
Big ol' refactor of cumulusci.core.config, this is the first hunk I want to get done for MetaDeploy.

## DONE

- Removed dependency on HiYaPyCo, which would choke on a file that exists but was empty. Unfortunately, the implementation of HiYaPyCo made it very difficult to report to the end user *which* file had a parse error.
- Moved the BaseConfig descendants that were empty subclasses into cumulusci/core/config/__init__.py
- Removed YamlGlobalConfig/YamlProjectConfig and put the config loading logic into BaseGlobalConfig/BaseProjectConfig respectively.
- Remove BaseConfig's support for multiple dictionaries in the search_path, which turned out to not work intuitively with the merged dicts that CCI actually uses.

## TODO

- [x] Replace MergedConfig with a simple function
- [x] don't parse the global config YAMLs a second time when creating a BaseProjectConfig
- [x] clean up YAML ordered loader
- [x] use YAML ordered loader for all config loading (at least for Py < 3.6)
- [x] Remove MergedYamlConfig

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/config/BaseGlobalConfig.py
```diff
@@ -23,3 +32,37 @@ def get_project_config(self):
...
+        return config_path
+
+    @property
+    def config_global_path(self):
+        return os.path.abspath(
+            os.path.join(__location__, "..", "..", self.config_filename)
+        )
+
+    def _load_config(self):
+        """ Loads the local configuration """
```
cdcarter: short circuit if there's already config, like we do in BaseProjectConfig

cumulusci/core/config/BaseGlobalConfig.py
```diff
@@ -23,3 +32,37 @@ def get_project_config(self):
...
+        # load the global config
+        with open(self.config_global_path, "r") as f_config:
+            config = yaml.safe_load(f_config)
+        self.config_global = config
+
+        # Load the local config
+        if self.config_global_local_path:
+            config = yaml.safe_load(open(self.config_global_local_path, "r"))
+            self.config_global_local = config
```
cdcarter: this little stanza gets repeated a lot, try and simplify

cumulusci/core/config/BaseProjectConfig.py
```diff
@@ -3,32 +3,110 @@
...
+
+    def _load_config(self):
+        """ Loads the configuration from YAML, if no override config was passed in initially. """
+
+        if (
+            self.config
+        ):  # any config being pre-set at init will short circuit out, but not a plain {}
+            return
+
+        # Verify that we're in a project #TODO: this broke in refactor?
```
cdcarter: remove todo

cumulusci/core/config/BaseProjectConfig.py
```diff
@@ -3,32 +3,110 @@
...
+
+        # merge in any additional yaml that was passed along
+        if self.additional_yaml:
+            additional_yaml_config = yaml.safe_load(self.additional_yaml)
+            if additional_yaml_config:
+                self.config_additional_yaml.update(additional_yaml_config)
+                merge_yaml.append(self.additional_yaml)
+
+        self.config = hiyapyco.load(
+            *merge_yaml, method=hiyapyco.METHOD_MERGE, loglevel="INFO"
```
cdcarter: wait, did i not push the commit that fixes this?! whats my point even.

cumulusci/core/config/MergedConfig.py
```diff
@@ -0,0 +1,96 @@
...
+    """ A merged config that takes strings of YAML instead of python dicts. """
+
+    def __init__(self, **configs):
+        parsed_configs = {}
+        for name, config in configs.items():
+            try:
+                parsed_configs[name] = yaml.safe_load(config)
+            except yaml.YAMLError as err:
+                raise ConfigError(
+                    f"Error parsing YAML for {name}: {err}", config_name=name
```
cdcarter: dont use F strings

cumulusci/core/config/MergedConfig.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class MergedConfig(BaseConfig):
+    """ A merged config takes a collection of dicts and merges them in priority order."""
+
+    def __init__(self, **configs):
+        """ MergedConfig(user_config={}, base_config={},) """
+        self.configs = (
+            configs
+        )  # NOTE: This relies on Py3.6+ behavior, kwarg order is preserved.
```
cdcarter: figure out what to do about this.

cumulusci/core/config/BaseGlobalConfig.py
cdcarter: and update all the tests

cumulusci/core/config/MergedConfig.py
davisagli: Either pass item tuples or a whole ordereddict without kwargifying it

cumulusci/core/config/MergedConfig.py
```diff
@@ -0,0 +1,96 @@
...
+
+from builtins import bytes, int, str
+from collections import OrderedDict
+
+from cumulusci.core.config import BaseConfig
+from cumulusci.core.exceptions import ConfigMergeError, ConfigError
+
+
+class MergedConfig(BaseConfig):
+    """ A merged config takes a collection of dicts and merges them in priority order."""
```
davisagli: I don't really understand why this is its own object instead of a function used to populate the config of another BaseConfig subclass.

cumulusci/core/config/__init__.py
```diff
@@ -6,22 +6,61 @@
...
+    "ConnectedAppOAuthConfig",
+    "FlowConfig",
+    "OrgConfig",
+    "TaskConfig",
+    "ServiceConfig",
+    "BaseTaskFlowConfig",
+    "BaseGlobalConfig",
+    "BaseProjectConfig",
+    "ScratchOrgConfig",
+]
```
davisagli: Is there a problem you're solving by listing these explicitly?

cumulusci/core/exceptions.py
```diff
@@ -70,6 +70,14 @@ class DependencyResolutionError(CumulusCIException):
 class ConfigError(CumulusCIException):
     """ Raised when a configuration enounters an error """
 
+    def __init__(self, message=None, **kwargs):
+        super(ConfigError, self).__init__(message)
+        self.config_name = kwargs.get("config_name", "")
```
davisagli: Why not put config_name=None in the method signature? Also is there a point to this if we're not including it in the stringification of the exception?

cumulusci/core/config/__init__.py
cdcarter: I thought there was but it turned out to be irrelevant.

cumulusci/core/config/MergedConfig.py
```diff
@@ -0,0 +1,96 @@
...
+        new_config = {}
+
+        for name in reversed(OrderedDict(self.configs)):
+            new_config = dictmerge(new_config, self.configs[name], name)
+
+        self.config = new_config
+
+
+class MergedYamlConfig(MergedConfig):
+    """ A merged config that takes strings of YAML instead of python dicts. """
```
davisagli: looks like it takes YAML filenames, not strings of YAML

davisagli: also it looks like this is only used in a test?

cumulusci/core/config/BaseGlobalConfig.py
```diff
@@ -23,3 +32,37 @@ def get_project_config(self):
...
+
+        self.config = MergedConfig(
+            global_local=self.config_global_local, global_config=self.config_global
+        ).config
```
davisagli: I'd still like to see this as function instead of a class. You're extracting the `config` attribute as soon as it's constructed and it doesn't look like MergedConfig being a subclass of BaseConfig is really giving us anything except for more class hierarchy to figure out.

cumulusci/core/config/BaseProjectConfig.py
```diff
@@ -3,32 +3,106 @@
...
+        if not self.config_project_path:
+            raise ProjectConfigNotFound(
+                "The file {} was not found in the repo root: {}. Are you in a CumulusCI Project directory?".format(
+                    self.config_filename, repo_root
+                )
+            )
+
+        # Load the project's yaml config file
+        with open(self.config_project_path, "r") as f_config:
+            project_config = yaml.safe_load(f_config)
```
davisagli: I wonder if we should use the ordered load in here too, to be safely backwards compatible in Python 2. It's conceivable that someone could be using a yaml dict option for a task and relying on its order being preserved.

cdcarter:
@davisagli - agreed with the notes, updated the checklist in the PR with what I believe is remaining.

I'd still like to figure out a way to abstract the YAML-file-loading, ideally to the point where the Config subclasses declare which files they depend upon. I don't have a good design in mind for this yet.

cdcarter:
I left them in the order they are in (highest priority first) for now, because I didn't want to go reverse all the callers of it. Once we're on Py3.6+ and not manually ordering our dicts we can revisit this, but it made sense to me.

cdcarter:
I did at least add a docstring.

davisagli:
> I left them in the order they are in (highest priority first) for now, because I didn't want to go reverse all the callers of it. Once we're on Py3.6+ and not manually ordering our dicts we can revisit this, but it made sense to me.

All 4 callers? I can do that. I think the reason it bugs me is because it's the opposite order of Object.assign and spread properties in Javascript, not to mention dict1.update(dict2) in Python. And I think it'll be less painful to get right from the start than change later.

cumulusci/core/utils.py
```diff
@@ -93,8 +93,8 @@ def merge_config(configs):
     """ recursively deep-merge the configs into one another (highest priority comes first) """
     new_config = {}
 
-    for name in reversed(configs):
-        new_config = dictmerge(new_config, configs[name], name)
+    for name, config in configs.items():
+        new_config = dictmerge(new_config, config, name)
```
davisagli: This change is not just about aesthetics. I figure it must be faster to grab both key and value while iterating rather than do a separate hash lookup with the key.

cdcarter: nah aesthetically its nicer too.

PR #860, opened 2018-10-25
(WIP) New source tracking tasks

davisagli:
# Critical Changes

# Changes
* Added `list_changes` and `retrieve_changes` tasks for retrieving metadata from a scratch org.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/sourcetracking.py
+++ b/cumulusci/tasks/salesforce/sourcetracking.py
@@ -10,6 +10,18 @@ from cumulusci.tasks.metadata.package import __location__
 
 
 class ListChanges(BaseSalesforceApiTask):
+
+    task_options = {
+        "include": {
+            "description": "Include changed components matching this string.",
+            "required": True,
+        }
+    }
+
+    def _init_options(self, kwargs):
+        super(ListChanges, self)._init_options(kwargs)
+        self.options["include"] = process_list_arg(self.options["include"])
+
     def _run_task(self):
         changes = self.tooling.query(
             "SELECT MemberName, MemberType FROM SourceMember WHERE IsNameObsolete=false"
@@ -21,9 +33,11 @@ class ListChanges(BaseSalesforceApiTask):
                 )
             )
             for change in changes["records"]:
-                self.logger.info(
-                    "  {}: {}".format(change["MemberType"], change["MemberName"])
-                )
+                mdtype = change["MemberType"]
+                name = change["MemberName"]
+                if not any(s in mdtype or s in name for s in self.options["include"]):
+                    continue
+                self.logger.info("  {}: {}".format(mdtype, name))
         else:
             self.logger.info("Found no changes.")
 
--- a/cumulusci/tasks/salesforce/sourcetracking.py
+++ b/cumulusci/tasks/salesforce/sourcetracking.py
@@ -1,9 +1,7 @@
 from collections import defaultdict
 import os
 
-import yaml
-
-from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.core.utils import process_list_arg
 from cumulusci.salesforce_api.metadata import ApiRetrieveUnpackaged
 from cumulusci.tasks.salesforce import BaseRetrieveMetadata
 from cumulusci.tasks.salesforce import BaseSalesforceApiTask
@@ -58,7 +56,7 @@ class RetrieveChanges(BaseRetrieveMetadata, BaseSalesforceApiTask):
                 "api_version"
             ] = self.project_config.project__package__api_version
 
-        self.options["include"] = self.options["include"].split(" ")
+        self.options["include"] = process_list_arg(self.options["include"])
 
     def _get_api(self):
         self.logger.info("Querying Salesforce for changed source members")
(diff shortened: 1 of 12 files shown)


cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -0,0 +1,109 @@
+from collections import defaultdict
+import os
+
+import yaml
```
cdcarter: unused import?

cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -0,0 +1,109 @@
...
+from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.salesforce_api.metadata import ApiRetrieveUnpackaged
+from cumulusci.tasks.salesforce import BaseRetrieveMetadata
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.tasks.metadata.package import PackageXmlGenerator
+from cumulusci.tasks.metadata.package import __location__
+
+
+class ListChanges(BaseSalesforceApiTask):
+    def _run_task(self):
```
cdcarter: consider putting the list of changes by membertype into ListChanges.return_values ?

cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _init_options(self, kwargs):
+        super(RetrieveChanges, self)._init_options(kwargs)
+
+        if "api_version" not in self.options:
+            self.options[
+                "api_version"
+            ] = self.project_config.project__package__api_version
+
+        self.options["include"] = self.options["include"].split(" ")
```
cdcarter: `process_list_arg` ?

cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -0,0 +1,109 @@
...
+        type_members = defaultdict(list)
+        for change in changes["records"]:
+            mdtype = change["MemberType"]
+            name = change["MemberName"]
+            if any(s in mdtype or s in name for s in self.options["include"]):
+                self.logger.info("Including {} ({})".format(name, mdtype))
+                type_members[mdtype].append(name)
+
+        types = []
+        for name, members in type_members.items():
```
cdcarter: `types = [MetadataType(name, members) for name, members in type_members.items()]` ?

cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -0,0 +1,109 @@
...
+            types.append(MetadataType(name, members))
+        package_xml = PackageXmlGenerator(self.options["api_version"], types=types)()
+
+        return self.api_class(self, package_xml, self.options.get("api_version"))
+
+    def _run_task(self):
+        super(RetrieveChanges, self)._run_task()
+
+        # update package.xml
```
cdcarter: hmm, feels like we should just use normal update_package_xml task as the next step in a flow, instead of duplicating it?

davisagli:
> did you think at all about tracking the revision number as well?

@cdcarter you mean include it in the printed list, or something more complex?

cumulusci/tasks/salesforce/sourcetracking.py
davisagli: I did consider it, but I figured I'd rather wait for a use case then guess at what data structure is going to be most helpful there.

cumulusci/tasks/salesforce/sourcetracking.py
davisagli: Personally I find the non-comprehension version more readable.

cumulusci/tasks/salesforce/sourcetracking.py
davisagli: Hmm. Is there a use case where you would want to retrieve changes but not update package.xml? I have been thinking no, and thus thinking that we should avoid having a task which doesn't do what you want unless you realize it's meant to be used in a particular flow.

cumulusci/tasks/salesforce/sourcetracking.py
cdcarter: at some point i'm going to just add return_values to every task we have, because if we don't start adding them we'll never iterate on them or use them :) but I hear your point, so I won't make you do it

cumulusci/tasks/salesforce/sourcetracking.py
cdcarter: de gustibus

cumulusci/tasks/salesforce/sourcetracking.py
cdcarter: 👍

cumulusci/tasks/salesforce/sourcetracking.py
davisagli: Working along with black in a pre-commit hook, I suppose one could have an auto-formatter that formats things to one's personal tastes just for editing!

davisagli:
This probably needs to be smarter. If you retrieve a single custom field into a metadata directory that already has that object, it overwrites the object file instead of adding to it.

jlantz:
@davisagli can you add a doc section with some examples of how this can be used?  This seems like something significant enough that we want to provide some docs.  Also, it'd help me in reviewing it :)

davisagli:
closing this, we need to rethink how to do this without clobbering existing package contents

PR #869, opened 2018-10-31
[WIP] CCI Entry Points

cdcarter:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -770,8 +770,8 @@ def org_list(config):
         "config_name",
         "username",
     ]
-    for org in config.project_config.list_orgs():
-        org_config = config.project_config.get_org(org)
+    for org in config.project_config.keychain.list_orgs():
+        org_config = config.project_config.keychain.get_org(org)
         row = [org]
         row.append("*" if org_config.default else "")
         row.append("*" if org_config.scratch else "")
--- a/cumulusci/cli/tests/test_cci.py
+++ b/cumulusci/cli/tests/test_cci.py
@@ -456,8 +456,8 @@ test     Test Service  *""",
     @mock.patch("click.echo")
     def test_org_list(self, echo):
         config = mock.Mock()
-        config.project_config.list_orgs.return_value = ["test1", "test2"]
-        config.project_config.get_org.side_effect = [
+        config.project_config.keychain.list_orgs.return_value = ["test1", "test2"]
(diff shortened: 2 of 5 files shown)


cdcarter:
@davisagli - after bashing at the files for a few hours trying to untangle the dependencies of BaseProjectConfig, I think I have a plan for how to handle the population of repo_info, and the various properties that derive from it.

### Current Behavior

First, a bit of a rambly explanation of whats happening now. Sorry if this is foggy, but this is a weird order of events..

As a reminder, repo_info is a memoized computed property, that gathers repository info from known CI environment variables and CCI named environment variables. Currently, assuming a caller doing:
```
project_config = BaseGlobalConfig().get_project_config()
```
repo_info() is first called partway through BaseProjectConfig's _load_config(), when getting the repo_name to load the project local config. 

Once its called, if we detect we're in "a CI environment" based on `CUMULUSCI_AUTO_DETECT` env var, we populate the `_repo_info` dict with root, name, etc. etc. We also validate that the full set of expected keys is now there, either through Heroku env vars or CUMULUSCI_* ones.

Once that is done, we continue trying to load the config. We either use the repo name we got from the environment, of if we didn't autodetect, we try to pull the repo name from the `origin` git remote in the .git/config on filesystem. Pretty much all the other repo values are figured out in the same way. These values are not memoized, so if we didn't detect CI, we hit the fs for every access.

### Simplification

There's an argument to hoist this completely out of BaseProjectConfig and create some sort of CCIContext. We can talk about if that work is worth doing now, but here's a plan assuming we want to just keep it in BaseProjectConfig for now but be less gnarly.

The main goal is to just allow us to pass in repo_info at construction, which is really easy. It's a one line change. I pushed that. But I want to clean up the accessors. They should all memoize their value into a key on _repo_info if calculated from the fs. There should also be one central source of truth for determining if we *can* check the fs or not. Then, the accessors use the following priority:
- passed in repo_info at constrction
- HerokuCI variables IF CUMULUSCI_AUTO_DETECT is set
- CUMULUSCI_* environment variables
- the filesystem
- Raise configerror with message about env var name. consider making this a warning or tolerable error for the values that arent name/root

Instead of detecting the CI and env vars in the first repo_info call, do it in a new private method called by init.

No longer require everything set, because sometimes they don't make sense.

cdcarter:
you dont need the read the above, David, I just needed to write it all somewhere :)

PR #870, opened 2018-11-01
Create new task to support assigning Permission Sets through CumulusC…

koenwesselman:
# Critical Changes

# Changes

- Added a new task "assign_permission_set" to support assigning Permission Sets through CumulusCI

# Issues Closed


koenwesselman:
@jlantz , perhaps you'd be able to have a look at this pull request? :-) Let me know if there's something I missed.

cumulusci/cumulusci.yml
```diff
@@ -2,6 +2,11 @@ cumulusci:
     keychain: cumulusci.core.keychain.EncryptedFileProjectKeychain
 
 tasks:
+    assign_permission_set:
+        description: Assign Permission Set specified in the "extra" option using [--permsetname "Permission_Set_DeveloperName"]
+        class_path: cumulusci.tasks.sfdx.SFDXOrgTask
+        options:
+            command: 'force:user:permset:assign'
```
jlantz: Since --permsetname is always required (I think), can you change the command option's value to 'force:user:permset:assign --permsetname ' so the only thing you have to do in the extra option is specify the permset's name?  That would make the cci command much nicer: `cci task run assign_permission_set -o extra Permission_Set_DeveloperName` and I think would help avoid issues with cci trying to parse the --permsetname as an option to the task run command.

koenwesselman: Thanks, @jlantz ! I've changed the command now - also added the 'Salesforce' group so that it shows up nicely in the task list command.

docs/tasks.rst
```diff
@@ -2,6 +2,19 @@
...
+
+**Description:** Assign Permission Set specified in the "extra" option using [--permsetname "Permission_Set_DeveloperName"]
+
+**Class::** cumulusci.tasks.sfdx.SFDXBaseTask
+
+Options:
+------------------------------------------
+
+* **command** *(required)*: The full command to run with the sfdx cli. **Default: force:user:permset:assign**
+* **extra**: Append additional options to the command. **Specify Permission Set: --permsetname "Permission_Set_DeveloperName"**
```
jlantz: OK, one last minor change and this should be ready to merge.  The instructions here need to have `--permsetname` removed to reflect the changes in the cumulusci.yml.

koenwesselman: Thanks @jlantz , I've made your suggested change now! 👍

davisagli:
We added a generic `dx` task which can be used to run an sfdx command against any org (including orgs which are in the CumulusCI keychain but not the sfdx keychain).

Issue #772, opened 2018-09-20
Task Groupings


cdcarter:
from @davisagli:
> it'd be nice if we had a way to group tasks in the list, and then maybe a way to hide a group unless you have a flag in your personal settings. In other words I want a way to say "show me all the releng tasks" rather than needing to configure each one

cdcarter:
Humble proposal:

a task definition looks like:
```
tasks:
    blah_task_name:
        class_path: blah
        description: blah
        options:
            name: option_is_blah
```
[JSON Schema definition for the tasks section of cumulusci.yml](https://gist.github.com/cdcarter/d3a10e03f12265f1f49af2aa3d9c8c40#file-schema-json-L102-L129)

I suggest we add a fourth, optional, key to the task definition object:

```
tasks:
    blah_task_name:
        group: releng
        class_path: blah
        description: blah
        options:
            name: option_is_blah
```
defined as a comma seperated list of strings .

If a task defines a group (or more than one), by default, it will *not* be returned by BaseTaskFlowConfig.list_tasks.

We'll also add a new property to the cumulusci section, which contains a comma severated list of groups:
```
cumulusci:
    keychain: MyKeyChainClass
    groups: releng, wlma
```

If a group is added to the cumulusci.groups key, then BaseTaskFlowConfig.list_tasks will now list tasks from that group. This can be overriden on the project yaml, or on an (e.g.) release engineers loal machine.

Still to figure out:
- are groups pre-registered besides just being referenced in a task?
- the behavior of defining a task with the same name as a globally hidden task in a project and not specifying group would be surprising. perhaps all tasks must have a group.
- at the CLI you should be able to list *all* tasks, and list all tasks for a specified list of groups. this is not the default. there's also a way to force running a task in a group, if its not your default group. oh, and see them.
- but unless you've specified a group in YAML or at runtime, the task is invisible.
- docs should list all tasks all groups.

davisagli:
Hmm, I'm having second thoughts about the idea of hiding groups of tasks because:
a. It'll be a source of confusion
b. It sort of goes against our ethos of making our internal tools widely available whenever they aren't truly SFDO Product Delivery-specific.
c. When there are tasks that fit the "SFDO Product Delivery-specific" clause, I'm inclined to say they should go in a separate package and we should have a way for installed Python packages to register tasks that they provide via setuptools entry points.

cdcarter:
[...]

PR #872, opened 2018-11-02
Display task list in groups

davisagli:
Let the bikeshedding over group assignments begin!

# Critical Changes

# Changes
* Group the results of `cci task list`. Set the `group` key of a task in cumulusci.yml to assign it to a group.

# Issues Closed

* #772: Task Groupings

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -908,13 +908,13 @@ def task_list(config):
     headers = ["task", "description"]
     task_groups = OrderedDict()
     for task in config.project_config.list_tasks():
-        group = task['group']
+        group = task["group"]
         if group not in task_groups:
             task_groups[group] = []
         task_groups[group].append(task)
     for group, tasks in task_groups.items():
-        data.append(('', ''))
-        data.append(('-- {} --'.format(group), ''))
+        data.append(("", ""))
+        data.append(("-- {} --".format(group), ""))
         for task in tasks:
             data.append((task["name"], task["description"]))
     table = Table(data, headers)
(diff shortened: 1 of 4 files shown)


cumulusci/core/config/BaseTaskFlowConfig.py
```diff
@@ -12,7 +12,13 @@ def list_infos(infos):
         info = infos[info_name]
         if not info:
             info = {}
-        rv.append({"name": info_name, "description": info.get("description", "")})
+        rv.append(
+            {
+                "name": info_name,
+                "description": info.get("description", ""),
+                "group": info.get("group") or "Other",
```
cdcarter: `"group": info.get("group", "")` and let the frontend (cli) decide what the default/unlabelled group name is.

cumulusci/cli/cci.py
```diff
@@ -905,8 +906,17 @@ def org_scratch_delete(config, org_name):
...
     for task in config.project_config.list_tasks():
-        data.append((task["name"], task["description"]))
+        group = task["group"]
+        if group not in task_groups:
+            task_groups[group] = []
+        task_groups[group].append(task)
+    for group, tasks in task_groups.items():
+        data.append(("", ""))
+        data.append(("-- {} --".format(group), ""))
+        for task in tasks:
```
cdcarter: `tasks = sorted(tasks, lambda x: x['name'])`

i think would really help, now that output isn't purely in YAML order.

cdcarter:
Groups instead of tags, huh? ;)

PR #873, opened 2018-11-04
Feature/ccientrypoint

cdcarter:
I would like the diffstat to take docs as a separate count please... Pushing this to watch CI run.  ;)

I'm going to do a granular description of changes below, and we can simplify them for the release note.

# Critical Changes

# Changes
- BaseProjectConfig has shrugged off some methods that just delegated to the keychain.
- BaseGlobalConfig has shrugged off some unimplemented methods, and `BaseGlobalConfig.get_project_config` is now deprecated in favor of using a runtime.
- Introducing... 🥁CumulusCIRuntime! In order to alleviate the complexities of getting CumulusCI tasks/flows running from within a Python application, CumulusCIRuntime encapsulates a lot of the details and wiring between Keychain, GlobalConfig, and ProjectConfig. Usage docs are barely included.
- CliConfig has been renamed to CliRuntime and now inherits from CumulusCIRuntime. It is still accessible as CliConfig. 

# Issues Closed


docs/usage.rst
```diff
@@ -1,7 +1,28 @@
...
+While most users of CumulusCI use the included command-line interface, ``cci``, it is also a library that can be invoked from any Python application. We take advantage of this to provide apps like MetaCI and MetaDeploy that run CumulusCI automation inside Django apps on Heroku. 
 
-    import cumulusci
+The Python API is primarily for internal users, but rough documentation is provided for experimentation and understanding of the code.
+
+
+Runtime
+-------
+
+The main entry point to CumulusCI is through a Runtime object. ``cumulusci.core.runtime.BaseRuntime`` is a concrete class and sufficient to create a simple CCI, and ``cumulusci.cci.config.CliRuntime`` contains examples of a runtime configured to dynamically provide the Keychain implementation, as well as handle errors differently.
```
davisagli: Maybe we should call it `CumulusCI` instead of `Runtime`? Runtime is fine, it just seems pretty generic if you run across it out of context.

cdcarter:
@jstvz - can you review when you get a chance? CumulusCIRuntime is (mostly incidentally) test covered, but I anticipate a test module popping up soon...

docs/usage.rst
jstvz: I think these need to be renamed, no?

cdcarter:
@jstvz good catch, updated.

PR #880, opened 2018-11-15
Misc documentation updates

davisagli:
# Critical Changes

# Changes
* Documentation updates:
  * Link to the full documentation from the readme.
  * Document installation using venv rather than pyenv since the former is included with Python. (It's still fine to use pyenv if you want.)
  * Give more useful links for how to set up SFDX.
  * Updated robot library docs.

# Issues Closed


docs/cookbook.rst
```diff
@@ -26,27 +26,6 @@ You can override the `path` option by adding the following to your project's cum
             options:
                 path: some_other_dir
 
-Swap out an Ant target for a task
```
cdcarter: Not in this PR if you don't want to, but I'd like to replace this with a cookbook entry on using Command task

docs/robot/CumulusCI_Library.html
```diff
@@ -5,8 +5,8 @@
 <meta http-equiv="Pragma" content="no-cache">
```
cdcarter: i have to admit i'm not fully sure why these files need to be committed, is there a way for readthedocs to generate them with sphinx? ... again, not in this PR ;)

docs/tutorial.rst
```diff
@@ -55,18 +42,23 @@ macOS
 
 .. code-block:: console
 
-    $ pyenv install 3.7.0
-    $ pyenv virtualenv 3.7.0 cci
-    $ # Copy the following line to ~/.bash_profile to automatically activate the virtual environment in all new shells.
-    $ pyenv activate cci
+    $ python3 -m venv cci
```
cdcarter: hmm, have we tested with system python3? I know there were OpenSSL issues with system python2. we may still want them to brew in their python.

docs/tutorial.rst
```diff
@@ -281,7 +272,13 @@ Configuring the Connected App is a one time operation. Once configured, you can
 Using Salesforce DX Scratch Orgs
 --------------------------------
 
-While it is possible to use `cci org connect <org_name>` to connect to a Developer Edition org, the real fun is using CumulusCI along with Salesforce DX.  If you already have the `sfdx` command installed, have connected to your devhub, and have set the `defaultdevhubusername` config settting (use `sfdx force:config:list` to verify), you're ready to start using `cci` with `sfdx`.  If you haven't already set up Salesforce DX, you can learn how at https://developer.salesforce.com/platform/dx.
+While it is possible to use `cci org connect <org_name>` to connect to a Developer Edition org, the real fun is using CumulusCI along with Salesforce DX.  If you already have the `sfdx` command installed, have connected to your devhub, and have set the `defaultdevhubusername` config setting (use `sfdx force:config:list` to verify), you're ready to start using `cci` with `sfdx`.  If you haven't already set up Salesforce DX, you need to take care of a few steps:
```
cdcarter: its still a little bit obscured, so I think we should make a sentence along the lines of:

> SFDX supports multiple DevHubs, so CumulusCI will use the one set as `defaultdevhubusername` when creating scratch orgs.

docs/tutorial.rst
davisagli: oh, good question. I'll experiment a bit

davisagli: Seems to work fine with the Python installer from python.org

docs/robot/CumulusCI_Library.html
jlantz: Ideally we should have a process that puts them in the gh-pages branch so they can just be linked to from the docs.  I just realized we could host static html there recently.

davisagli: Sounds good, can we handle that separately?

Issue #877, opened 2018-11-10
CumulusCI should be able to connect to any DX alias and/or understand dx auth files


Szandor72:
We use vanilla DX tooling mostly and CumulusCI for special tasks like cascading deletes, nsp manipulation or browsertests.

One blocker is the inability to connect a Scratch org to CCI on the fly during a CI process to run tasks against it. 

Ideally, we could connect CCI with any DX alias or at least feed a DX auth url file into the CCI keychain.

davisagli:
cumulusci 2.2.0 added a `cci org import` command to import an org from SFDX.

Issue #674, opened 2018-07-17
cci org scratch_import_sfdx <org_name> <username>


jlantz:
Import a scratch org from the sfdx keychain into the cci keychain with a given name.  This would allow a workaround for cases where CumulusCI wrapping the sfdx command fails to create a scratch org but calling sfdx directly is successful by allowing scratch orgs created directly in sfdx to be imported into cci.

The command needs to construct a ScratchOrgConfig instance for an enabled scratch org using the username provided and attempt to get the org info.  If successful, record the org in the keychain with set_org.

jstvz:
**lurch: add

jstvz:
**lurch: add

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fN7gAAE" target="_blank">W-027939</a>

LurchTheButler:
Tracking <a href="https://foundation.my.salesforce.com/a2x80000007fN7pAAE" target="_blank">W-027942</a>

jstvz:
cf #877

PR #881, opened 2018-11-15
Add scratch org import

jstvz:
# Critical Changes

# Changes
Add the `org import` command. When given a sfdx username (or an alias) and an org name `import` will import a scratch org into the cumulusci keychain.
# Issues Closed
- #877 
- #674


davisagli:
I'm not sure where, but this should probably get mentioned in the docs too.

PR #919, opened 2018-12-27
Fix overriding a flow to None

davisagli:
# Critical Changes

# Changes
* Fixed an issue where overriding a flow step to set flow: None did not work.

# Issues Closed

### DIFF
--- a/cumulusci/core/flows.py
+++ b/cumulusci/core/flows.py
@@ -117,6 +117,8 @@ class BaseFlow(object):
         for step in steps.values():
             if "flow" in step:
                 flow = step["flow"]
+                if flow == "None":
+                    continue
                 if flow in flows:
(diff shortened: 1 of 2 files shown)


cumulusci/core/flows.py
```diff
@@ -117,6 +117,8 @@ def _check_infinite_flows(self, steps, flows=None):
         for step in steps.values():
             if "flow" in step:
                 flow = step["flow"]
+                if flow == "None":
```
cdcarter: should probably at the very least check case insensitively, and i'd consider also supporting blank string and yaml !!null.

davisagli: Fair, but I was just aiming to make it compatible with the existing conditional on line 151

PR #923, opened 2018-12-28
First, probably naive, attempt to handle the error.

### LINKED ISSUE #752: Better error when service connect is called with an undefined service name
Current behavior:
```
$ cci service connect foo
Checking the version!
An update to CumulusCI is available. Use pip install --upgrade cumulusci to update.
  File "/Users/dglick/.pyenv/versions/cci-dev-py27/bin/cci", line 11, in <module>
  File "/Users/dglick/.pyenv/versions/2.7.15/envs/cci-dev-py27/lib/python2.7/site-packages/click/core.py", line 722, in __call__
  File "/Users/dglick/.pyenv/versions/2.7.15/envs/cci-dev-py27/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/dglick/.pyenv/versions/2.7.15/envs/cci-dev-py27/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
  File "/Users/dglick/.pyenv/versions/2.7.15/envs/cci-dev-py27/lib/python2.7/site-packages/click/core.py", line 1066, in invoke
  File "/Users/dglick/.pyenv/versions/2.7.15/envs/cci-dev-py27/lib/python2.7/site-packages/click/core.py", line 1061, in invoke
    cmd_name, cmd, args = self.resolve_command(ctx, args)
  File "/Users/dglick/.pyenv/versions/2.7.15/envs/cci-dev-py27/lib/python2.7/site-packages/click/core.py", line 1100, in resolve_command
    cmd = self.get_command(ctx, cmd_name)
  File "/Users/dglick/Work/CumulusCI/cumulusci/cli/cci.py", line 574, in get_command
    config.project_config, "services__{0}__attributes".format(name)
AttributeError: 'NoneType' object has no attribute 'items'
```

**lurch: add

ceiroa:
I didn't raise a ServiceNotValid exception because if I do the stack trace gets printed, which didn't seem very user-friendly... Happy to get feedback.

# Critical Changes

# Changes
Fixes #752 

# Issues Closed


ceiroa:
screen shot 2018-12-28 at 3 06 25 pm

cumulusci/cli/cci.py
```diff
@@ -594,28 +594,32 @@ def _build_param(self, attribute, details):
...
+                if project:
+                    click.echo("{0} is now configured for this project".format(name))
+                else:
+                    click.echo("{0} is now configured for global use".format(name))
+
+            ret = click.Command(name, params=params, callback=callback)
+            return ret
+
+        except AttributeError:
+            click.echo(name + " doesn't exist")
```
davisagli: Could you move the try/except so that it only catches AttributeError around the lines that try to get the `"services__{0}__attributes".format(name)` attribute? It's best to be as targeted as possible with exception handling, especially for AttributeError which can be raised pretty much anywhere if an attribute is misspelled (we don't have static type checking to catch that at compile time over here in Python land, at least not by default).

davisagli: I would do this instead of the click.echo:

    raise click.UsageError("Sorry, I don't know about the '{}' service.".format(name))

That results in a bit cleaner output:

    % cci service connect asdf
    Checking the version!
    Error: Sorry, I don't know about the 'asdf' service.

The difference between using `name + " string"` and `"{} string".format(name)` is subtle and isn't critical here. I generally tend to use the format method because it works even when you're interpolating a variable that is not a string, whereas using addition would raise a TypeError if given a different type. And the format method stays more consistent with the rest of the codebase.

davisagli: I checked the test coverage using `coverage run $(which pytest); coverage report -m` and it looks like we could use a test for the new except clause. Would you prefer to have a go at adding one on your own or have me do it and let you review? The test can go below `test_service_connect` in `test_cci.py`.

ceiroa: Thanks for the detailed feedback, @davisagli! I'll be happy to continue working on this myself and implement the suggested changes, if there's no hurry getting them in.

davisagli: @ceiroa great! There's no hurry.

ceiroa:
This is ready for a second look, @davisagli. I must confess that I don't understand the syntax in the tests very well yet, so I wouldn't be surprised if my test isn't right. Please feel free to suggest changes. Thanks!

ceiroa:
Saw the errors. Will try to fix.

cumulusci/cli/tests/test_cci.py
```diff
@@ -329,6 +329,15 @@ def test_service_connect(self):
...
         run_click_command(cmd, project=False)
 
+    def test_service_connect_invalid_org(self):
+        multi_cmd = cci.ConnectServiceCommand()
+        ctx = mock.Mock()
+        config = mock.Mock()
+        config.project_config.services__test__attributes = {"foo": {}}
+
+        with self.assertRaises(click.UsageError):
+            multi_cmd.get_command(ctx, config)
```
davisagli: `get_command` takes two params: ctx (the context object used by the `click` library -- in this case we're passing a mock object because it's not actually used and that's easier than setting up a real context) and a string name of the command to get. So you want to pass a string here rather than `config` ... either "test" if you want it to succeed or something else if you want it to fail.

So how do you tell the system to use the mock config object you just created? That's what the `with mock.patch("cumulusci.cli.cci.TEST_CONFIG", config):` block is for in other tests. It temporarily sets the TEST_CONFIG variable in the cci module to our config object, and if that variable is set then the load_config function will use it instead of looking for config files on the hard disk (which is where your test is currently breaking).

cumulusci/cli/tests/test_cci.py
```diff
@@ -329,6 +329,15 @@ def test_service_connect(self):
 
         run_click_command(cmd, project=False)
 
+    def test_service_connect_invalid_org(self):
```
davisagli:
```suggestion
    def test_service_connect_invalid_service(self):
```

PR #924, opened 2018-12-28
[WIP] FlowRunner

cdcarter:
# Critical Changes

# Changes
- Introduces a `cumulusci.core.flowrunner` module that has not yet been integrated into the rest of the library. I will update this with more. I will also align docstrings w/ this, at some point...
    - FlowCoordinator replaces the public BaseFlow API for creating and running a flow. This includes override hooks (like `_pre_task`). 
    - FlowCoordinator recursively validates the entire flow on instantiation, and computes the actual order of tasks (including skips) to run in the form of a list of `StepSpec`s.
    - FlowCoordinator merges options from task definitions, flow definitions, specific overrides, and CLI/runtime overrides.
    - FlowCoordinator delegates to one or more `TaskRunner`s which takes in a StepSpec and context, imports and instantiates the task object, runs the task, and collects the result or exception into an immutable StepResult.
    - FlowCoordinator collects those StepResults through execution, computes result based options, and handles failure/ignore_failure flags.
- Flows now get a logger at `cumulusci.flows.CLASS_NAME` and pass that on to each task to get a child logger at `cumulusci.flows.CLASS_NAME.TASK_CLASS_NAME`. This allows alternate flowrunner implementations to more directly customize tasks run only under that class.
- When refreshing an OAuth token for a persistent org with a refresh token, OrgConfig now (propagates?) raises an exception directly, instead of failing on first use in the flow.

# Issues Closed


cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,277 @@
...
+    def __repr__(self):
+        return "<StepSpec {num}:{name} {cfg}>".format(
+            num=self.step_num, name=self.task_name, cfg=self.task_options
+        )
+
+
+StepResult = namedtuple("StepResult", ["step_num", "task_name", "task"])
+
+
+class NoOpStep(object):
```
cdcarter: Inclined to replace this with another flag on StepSpec

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,277 @@
...
+            org_config=self.org_config,
+            name=step.task_name,
+            stepnum=step.step_num,
+            flow=self,
+        )
+        task()
+        return StepResult(step.step_num, step.task_name, task)
+
+    def run(self):
+        for step in self.steps:
```
cdcarter: I think most of the logging and callback hooks can/should get relatively centralized to this method.

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,428 @@
...
+        :param step: StepSpec
+        :return: StepResult
+        """
+        # get the base task_config from the project config, as a dict for easier manipulation.
+        task_config = copy.deepcopy(self.project_config.get_task(step.task_name).config)
+        if "options" not in task_config:
+            task_config["options"] = {}
+
+        # update the copied task_config with override options from the flow definition
+        task_config["options"].update(step.task_options)
```
davisagli: Shouldn't this part happen prior to running the flow?

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,428 @@
...
+                    self.logger.info("Skipping task: {}".format(step.task_name))
+                    self._rule(fill="*", new_line=True)
+                    continue
+
+                self._rule(fill="-")
+                self.logger.into("Running task: {}".format(step.task_name))
+                self._rule(fill="-", new_line=True)
+
+                self._pre_task(step)
+                runner = TaskRunner(self.project_config, org_config=org_config)
```
davisagli: Once I work on embedding steps from other projects I think the project_config will need to be referenced from the StepSpec.

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,428 @@
...
+        # get the base task_config from the project config, as a dict for easier manipulation.
+        task_config = copy.deepcopy(self.project_config.get_task(step.task_name).config)
+        if "options" not in task_config:
+            task_config["options"] = {}
+
+        # update the copied task_config with override options from the flow definition
+        task_config["options"].update(step.task_options)
+
+        # get task implementation class
+        task_class = import_class(task_config["class_path"])
```
davisagli: It would be nice to find out about a misspelled class path at initialization time rather than runtime.

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,428 @@
...
+        self.runtime = runtime
+        self.org_config = org_config
+
+    def run_step(self, step):
+        """
+        Run a step.
+
+        :param step: StepSpec
+        :return: StepResult
+        """
```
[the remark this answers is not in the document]
cdcarter: agreed.

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,428 @@
...
+                org_config=self.org_config,
+                name=step.task_name,
+                stepnum=step.step_num,
+                flow=self,  # not actually passing the flow, but that doesn't matter, this is more a run_in_flow.
+                # TODO: fix that!
+            )
+            task()
+        except Exception as exc:
+            # David, I don't know if this makes any sense.
+            task.logger.exception("Exception in task {}".format(step.task_name))
```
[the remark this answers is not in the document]
cdcarter: I need to actually test what happens here, yea.

I also don't love the taskrunner reaching into the tasks logger, and probably should use the flowrunners logger?

cumulusci/core/flowrunner.py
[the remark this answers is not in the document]
cdcarter: The lifecycle of the TaskRunner is a little weird right now. you just create it with an orgconfig and then can run it with any number of steps.

In order to make that happen, I think the "correct" thing is to pass `TaskRunner` the StepSpec on initialization. That makes it a single purpose TaskRunner, but I think that's okay, we don't need a flyweight here.

I have dreams of switching cci task run to use TaskRunner too.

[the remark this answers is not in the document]
davisagli: In that case maybe it should just be a run method on StepSpec instead of a separate object?

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,428 @@
...
+                    from_flow=from_flow,
+                )
+            )
+            return visited_steps
+
+        if "flow" in step_config:
+            name = step_config["flow"]
+            step_options = step_config.get("options", {})
+            flow_config = self.project_config.get_flow(name)
+            for sub_number, sub_stepconf in flow_config.steps.items():
```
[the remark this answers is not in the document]
davisagli: Actually we need to sort by the looseversion here because there's nothing in dictmerge that does that for us. You've lost the sort() from the old _get_steps_ordered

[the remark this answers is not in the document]
cdcarter: great point. I think we can wait til we have the full set of steps to sort it? that was my idea, even if I forgot to code it

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,465 @@
...
+        self.state.pre_flow()
+
+    def _post_flow(self):
+        self.state.post_flow()
+
+    def _pre_task(self, spec):
+        self.state.pre_task(spec)
+
+    def _post_task(self, spec, result):
+        self.state.post_tsak(spec, result)
```
davisagli:
```suggestion
        self.state.post_task(spec, result)
```

cumulusci/core/flowrunner.py
```diff
@@ -0,0 +1,465 @@
...
+
+StepResult = namedtuple(
+    "StepResult", ["step_num", "task_name", "result", "return_values", "exception"]
+)
+
+
+class FlowState(object):
+    """ FlowState - carries context to something else, like a CI job, that needs flow hooks run on it. """
+
+    # state is probably not a great name, but who has time to read design patterns...jk...will rename later tho.
```
davisagli: FlowHooks?

cumulusci/cli/cci.py
```diff
@@ -1131,26 +1131,19 @@ def flow_run(config, flow_name, org, delete_org, debug, o, skip, no_prompt):
     if delete_org and not org_config.scratch:
         raise click.UsageError("--delete-org can only be used with a scratch org")
 
-    flow_config = config.project_config.get_flow(flow_name)
+    coord = config.get_flow(flow_name)
```
davisagli: No benefit from abbreviating, call it coordinator so it's obviously not coordinate

cumulusci/core/runtime.py
```diff
@@ -53,6 +55,14 @@ def keychain_cls(self):
     def get_keychain_class(self):
         return None
 
+    @property
+    def callback_cls(self):
+        klass = self.get_callback_class()
+        return klass or self.callback_class
+
+    def get_callback_class(self):
+        return None
```
davisagli: Do we have an existing use case where we need this callable rather than just using self.callback_class?

cumulusci/core/runtime.py
```diff
@@ -75,3 +85,12 @@ def _load_project_config(self, *args, **kwargs):
     def _load_keychain(self):
         self.keychain = self.keychain_cls(self.project_config, self.keychain_key)
         self.project_config.set_keychain(self.keychain)  # never understood this but ok.
+
+    def get_flow(self, name, ctx=None):
+        """ Get a primed and readytogo flow coordinator. """
+        config = self.project_config.get_flow(name)
+        state = self.callback_cls(ctx)
+        coordinator = FlowCoordinator(
+            self, config, name=name, options={}, skip=None, state=state
```
davisagli: The FlowCoordinator arg is called callbacks, not state

cumulusci/core/runtime.py
cdcarter: I suspect preflight and real-installation-jobs will need different callbacks in MetaDeploy. In metadeploy at present, they're two sibling subclasses of BaseFlow in order to do their thing.

cumulusci/core/flowrunner.py
```diff
@@ -115,7 +115,7 @@ def for_display(self):
         skip = ""
         if self.skip:
             skip = " [SKIP]"
-        return "{step_num:8}: {task_name}{skip}".format(
+        return "{step_num}: {task_name}{skip}".format(
```
cdcarter: that was to add some whitespace and keep the task names aligned... does it look weird?

davisagli: It wasn't a valid format and I didn't feel like tracking down what it's supposed to be. Feel free to fix it as long as it doesn't break the tests.

cdcarter: oh i just checked in ipython and we just need to str(self.step_num)

cumulusci/core/flowrunner.py
```diff
@@ -133,11 +133,8 @@ class FlowCallback(object):
 
      This is how you can carry a reference to the Django model a flow run is related to in MetaCI or MetaDeploy."""
 
-    def __init__(self, ctx=None, _factory=None, *args, **kwargs):
+    def __init__(self, ctx=None):
         self.context = ctx
-        if ctx is None:
-            if callable(_factory):
-                self.context = _factory(*args, **kwargs)
```
cdcarter: how'd i know you were gonna delete this? ;) fine fine

cumulusci/core/flowrunner.py
```diff
@@ -337,9 +328,6 @@ def _visit_step(
         """
         number = LooseVersion(str(number))
 
-        if visited_steps is None:
```
cdcarter: i'd kinda prefer to leave this in and remove the explicit passed `[]` in _init_steps, it feels like the intent is clearer, and we'd still get coverage. but fine.

cumulusci/core/flowrunner.py
davisagli: Not opposed to it in principle, but I get annoyed when I have to write test coverage for code that isn't used anywhere else.

cumulusci/core/flowrunner.py
davisagli: Sure, I don't feel strongly about it.

cdcarter:
since it's my PR i can't approve it. but. looks nice.

cumulusci/core/flowrunner.py
cdcarter: i predict something like it will get used in the metadeploy subclass, but i don't have the exact shape yet, this was definitely just a guess.

cumulusci/core/flowrunner.py
```diff
@@ -218,10 +234,11 @@ def __init__(
             callbacks = FlowCallback()
         self.callbacks = callbacks
 
-        # TODO: Support CLI/Runtime Options
-        if not options:
-            options = {}
-        self.runtime_options = options
+        self.runtime_options = defaultdict(dict)
+        if options:
+            for key, value in options.items():
```
cdcarter: i sorta feel like this should be in the CLI and flowrunner just takes a nested dict... thoughts?

cumulusci/core/flowrunner.py
```diff
@@ -244,6 +261,7 @@ def run(self, org_config):
             line = "{} ({})".format(line, self.name)
         self._rule()
         self.logger.info(line)
+        self.logger.info("Flow Description: {}".format(self.flow_config.description))
```
cdcarter: kinda would prefer just a ** or something instead of "Flow Description:" to reduce line text?

cumulusci/core/flowrunner.py
```diff
@@ -374,16 +392,17 @@ def _visit_step(
             step_overrides.update(step_config.get("options", {}))
             task_config["options"].update(step_overrides)
 
+            # merge runtime options
+            if name in self.runtime_options:
+                task_config["options"].update(self.runtime_options[name])
```
cdcarter: as i mentioned in a chat at one point, these are all updates instead of dictmerges which is what we do at the yaml level. should we switch?

cumulusci/core/flowrunner.py
```diff
@@ -398,6 +417,10 @@ def _visit_step(
 
         if "flow" in step_config:
             name = step_config["flow"]
+            if from_flow:
```
cdcarter: oh nice

cumulusci/core/flowrunner.py
```diff
@@ -465,3 +488,17 @@ def _init_org(self):
         if self.org_config.config != orig_config:
             self.logger.info("Org info has changed, updating org in keychain")
             self.project_config.keychain.set_org(self.org_config)
+
+    def resolve_return_value_options(self, options):
+        """Handle dynamic option value lookups in the format ^^task_name.attr"""
+        for key, value in options.items():
+            if isinstance(value, str) and value.startswith(RETURN_VALUE_OPTION_PREFIX):
+                path, name = value[len(RETURN_VALUE_OPTION_PREFIX) :].rsplit(".", 1)
```
cdcarter: not for this PR, but i've longed to write an actual grammar for options values that handles ^^ and $ more centrally

cumulusci/core/flowrunner.py
davisagli: That sounds nice, but I wasn't confident we don't have yaml with options like this.

cumulusci/core/flowrunner.py
davisagli: I think update is both more backwards-compatible and less surprising. We can re-evaluate if a specific use case presents itself.

cumulusci/core/flowrunner.py
davisagli: It'd also be nice to catch invalid paths during init rather than during run, but it looks nontrivial (needs a second validation pass after the steps have been initted I think) and isn't a regression, so let's not do it now.

cumulusci/core/flowrunner.py
cdcarter: I don't think this codepath was taken for yaml options in baseflow... hmm... eh its fine by me.

davisagli: I think you're right. I moved the __ handling to the CLI

cumulusci/core/flowrunner.py
```diff
@@ -261,17 +260,24 @@ def run(self, org_config):
...
         self._rule(fill="-", new_line=True)
 
+        self._init_org()
+        self._rule(fill="-")
+        self.logger.info("Organization:")
+        self.logger.info("  {}: {}".format("Username", org_config.username))
+        self.logger.info("  {}: {}".format("  Org Id", org_config.org_id))
+        self._rule(fill="-", new_line=True)
+
+        self.callbacks.pre_flow()
```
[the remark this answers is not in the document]
davisagli: Oh good point -- I was aiming to show the steps sooner before connecting to the org, but it sounds nice to have a hook to alter the steps based on the state of the org, so I'll change it back.

cumulusci/core/flowrunner.py
```diff
@@ -285,6 +291,10 @@ def run(self, org_config):
 
                 self._rule(fill="-")
                 self.logger.info("Running task: {}".format(step.task_name))
+                if step.task_config["options"]:
+                    self.logger.info("Options:")
+                    for key, value in step.task_config["options"].items():
```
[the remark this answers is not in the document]
davisagli: We need an instantiated task to log the fully initialized/validated options. I moved the logging into TaskRunner so we can do that similarly to how we did before.

It'd be nice if we could turn _init_options into a class method that can be called during flow init, but I saw at least one or two tasks that rely on the org_config there.

PR #938, opened 2019-01-07
Let minidom decode xml instead of requests guessing the charset

### LINKED ISSUE #911: Special characters break deploy task when running user is in French
image
Set my scratch org user to French to test translated labels and got this error when I tried to run `cci task run deploy`.

davisagli:
We've seen a couple reports of UnicodeDecodeErrors which I think are happening when requests makes a bad guess about the charset of the MD API response. It should be safer to pass bytes to the xml parser which can honor the charset in the xml declaration.

# Critical Changes

# Changes

# Issues Closed
#911


cdcarter:
as a matter of style, i might like to see more comments arond

cdcarter:
im bad at github

davisagli:
I should add a test for the case that was failing.

PR #944, opened 2019-01-16
Clean up UpdateDependencies tests

davisagli:
Add a test demonstrating that it's possible to use UpdateDependencies to install a single metadata bundle.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
+++ b/cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
@@ -58,19 +58,9 @@ class TestUpdateDependencies(unittest.TestCase):
         }
         task.api_class = mock.Mock()
         zf = zipfile.ZipFile(io.BytesIO(), "w")
-        try:
-            UD_globals = UpdateDependencies._install_dependency.__globals__
-        except AttributeError:  # Python 2
-            UD_globals = UpdateDependencies._install_dependency.__func__.__globals__
-        download_extract_zip = UD_globals["download_extract_zip"]
-        download_extract_github = UD_globals["download_extract_github"]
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
```diff
@@ -144,3 +134,37 @@ def test_update_dependency_latest_option_err(self):
...
+                ]
+            },
+            project_config=project_config,
+        )
+        zf = zipfile.ZipFile(io.BytesIO(), "w")
+        task._download_extract_github = mock.Mock(return_value=zf)
+        api = mock.Mock()
+        task.api_class = mock.Mock(return_value=api)
+        task()
+        assert task.install_queue == [
```
cdcarter: this is a unittest.TestCase, so you should use self.assertEquals

cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
```diff
@@ -58,19 +58,9 @@ def test_run_task(self, ApiRetrieveInstalledPackages):
         }
         task.api_class = mock.Mock()
         zf = zipfile.ZipFile(io.BytesIO(), "w")
-        try:
-            UD_globals = UpdateDependencies._install_dependency.__globals__
```
cdcarter: THANKYOUUUU

cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
davisagli: But we're running it with pytest... eh, I'll change it.

PR #947, opened 2019-01-22
(wip) Sketch out the metadeploy publish task

davisagli:
Here's a start on the metadeploy_publish task.

To use this you need to connect the metadeploy service with `cci service connect metadeploy`. base_url should be something like http://localhost:8080/admin/rest and the token can be generated in the MetaDeploy admin.

It also assumes you've configured a task in cumulusci.yml that supplies the necessary options. For Cumulus this could look something like this:

```
tasks:
    publish_installer:
        class_path: cumulusci.tasks.metadeploy.Publish
        options:
            product_id: eM3aQ3Q
            flow: install_prod
            title: Install NPSP
            slug: install
            tier: primary
            preflight_message: This is the preflight message.
            post_install_message: This is the post-install message.
```

To do:
- [x] freeze dependencies as decomposed steps
- [ ] don't break if version already exists
- [ ] log link to MD frontend for plan
- [x] sort out what we set as the `kind` for MD steps
- [x] in MD: make version description optional
- [ ] tests
- [ ] docs

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -56,7 +56,6 @@ except ImportError:  # pragma: no cover
     pass
 
 import copy
-import json
 import logging
 from collections import defaultdict
 from collections import namedtuple
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -213,7 +213,7 @@ class TaskRunner(object):
             self._log_options(task)
             task()
         except Exception as e:
-            task.logger.exception("Exception in task {}".format(self.step.task_name))
+            self.flow.logger.exception("Exception in task {}".format(self.step.task_name))
             exc = e
         return StepResult(
             self.step.step_num,
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -56,6 +56,7 @@ except ImportError:  # pragma: no cover
     pass
 
 import copy
+import json
 import logging
 from collections import defaultdict
 from collections import namedtuple
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -213,7 +213,9 @@ class TaskRunner(object):
             self._log_options(task)
             task()
         except Exception as e:
-            self.flow.logger.exception("Exception in task {}".format(self.step.task_name))
+            self.flow.logger.exception(
+                "Exception in task {}".format(self.step.task_name)
+            )
             exc = e
         return StepResult(
             self.step.step_num,
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -8,6 +8,7 @@ from __future__ import unicode_literals
 from builtins import object
 from past.utils import old_div
 import contextlib
+import json
 import logging
 import time
 import threading
@@ -72,17 +73,9 @@ class BaseTask(object):
         # the tasks stepnumber in the flow
         self.stepnum = stepnum
 
-        if self.salesforce_task and not self.org_config:
-            raise TaskRequiresSalesforceOrg(
-                "This task requires a salesforce org. "
-                "Use org default <name> to set a default org "
-                "or pass the org name with the --org option"
-            )
         self._init_logger()
         self._init_options(kwargs)
         self._validate_options()
-        self._update_credentials()
-        self._init_task()
 
     def _init_logger(self):
         """ Initializes self.logger """
@@ -133,6 +126,15 @@ class BaseTask(object):
     def __call__(self):
         # If sentry is configured, initialize sentry for error capture
         self.project_config.init_sentry()
+
+        if self.salesforce_task and not self.org_config:
+            raise TaskRequiresSalesforceOrg(
+                "This task requires a salesforce org. "
+                "Use org default <name> to set a default org "
+                "or pass the org name with the --org option"
+            )
+        self._update_credentials()
+        self._init_task()
 
         with stacked_task(self):
             try:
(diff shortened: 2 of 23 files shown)


cumulusci/tasks/metadeploy.py
```diff
@@ -0,0 +1,123 @@
...
+            "description": "Name of the installation plan to publish",
+            "required": True,
+        },
+        "tag": {"description": "Name of git tag to publish"},
+    }
+
+    def _init_task(self):
+        metadeploy_service = self.project_config.keychain.get_service("metadeploy")
+        self.base_url = metadeploy_service.url
+        self.api = requests.Session()
```
cdcarter: if we end up having more tasks that touch metadeploy, you can do something like BaseSalesforceTask, where `_init_api()` inits an api client at the api member. this works for now, but it is kinda nice to have the logic for constructing an api client chunked out

cumulusci/tasks/metadeploy.py
```diff
@@ -0,0 +1,123 @@
...
+        self.api.headers["Authorization"] = "token {}".format(metadeploy_service.token)
+        self.metadeploy_config = self.project_config.metadeploy
+        if self.metadeploy_config is None:
+            raise PlanNotFound(
+                "No metadeploy configuration found. "
+                "Did you forget to configure it in cumulusci.yml?"
+            )
+        plan_config = getattr(
```
cdcarter: i've tried to encourage us to get away from this habit of reaching deep into project_config in task classes. instead, you can make each thing from the metadeploy_config that you actually need as a task_option, and then give them an option default using the $project_config option syntax.

cumulusci/tasks/metadeploy.py
```diff
@@ -0,0 +1,123 @@
...
+            self.project_config, "metadeploy__plans__{}".format(self.options["plan"])
+        )
+        if plan_config is None:
+            raise PlanNotFound("{} plan not found.".format(self.options["plan"]))
+        self.plan_config = BaseConfig(plan_config)
+
+    def _call_api(self, method, url, **kwargs):
```
cdcarter: again, only really matters if you have to write more MetaDeploy tasks or a MetaDeploy api client for other purposes.


... but i'd add an arg (default false) that causes this to only get the first page instead of running the full cursor

cumulusci/tasks/metadeploy.py
davisagli: At first glance that sounds quite reasonable, I'll see how it plays out.

davisagli:
@cdcarter updated based on your comments

cumulusci/core/flowrunner.py
```diff
@@ -56,6 +56,7 @@
     pass
 
 import copy
+import json
```
jlantz: Unused import?

cumulusci/core/tasks.py
```diff
@@ -8,6 +8,7 @@
 from builtins import object
 from past.utils import old_div
 import contextlib
+import json
```
jlantz: Unused import?

cumulusci/cumulusci.yml
```diff
@@ -151,6 +151,10 @@ tasks:
         description: Publishes a release to the mrbelvedere web installer
         class_path: cumulusci.tasks.mrbelvedere.MrbelvederePublish
         group: Release Management
+    metadeploy_publish:
```
jlantz: I wonder if we should leave this out of the global cumulusci.yml for now.  It really needs to be configured in the project to work at all right?  We might be able to have a default one that uses the new install_prod_no_config flow in a future release, but it seems like we might want to keep this project specific for now.

cumulusci/tasks/metadeploy.py
```diff
@@ -0,0 +1,131 @@
...
+        return results
+
+
+class Publish(BaseMetaDeployTask):
+    """Publishes an installation plan to MetaDeploy.
+    """
+
+    task_options = {
+        "flow": {"description": "Name of flow to publish", "required": True},
+        "product_id": {
```
jlantz: Could we use the github repo url on the Product model to look this up by default and only require this if there's more than one Product for the repo?

cumulusci/tasks/metadeploy.py
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    def _run_task(self):
+        tag = self.options["tag"]
+        label = self.project_config.get_version_for_tag(tag)
+        # @@@ handle checkout/publish from a release tag instead of current commit
+        steps = self._freeze_steps()
+        import pdb
+
+        pdb.set_trace()
```
jlantz: pdb

cumulusci/tasks/salesforce/InstallPackageVersion.py
```diff
@@ -1,3 +1,5 @@
+import json
```
jlantz: Unused import?

cumulusci/tasks/salesforce/UpdateDependencies.py
```diff
@@ -282,3 +282,43 @@ def _uninstall_dependency(self, dependency):
...
+        )
+        steps = []
+        for i, dependency in enumerate(_flatten(dependencies), start=1):
+            if "namespace" in dependency:
+                kind = "managed"
+                # @@@ we want the package name, not namespace
+                name = "Install {}".format(dependency["namespace"])
+            else:
+                kind = "metadata"
+                # @@@ we want a friendly name from...where?
```
jlantz: Good point.  We don't really have a place for that at the moment so the generated string is fine for now

cumulusci/tasks/metadeploy.py
davisagli: Yeah, we just don't have a way to filter the API yet so I figured hardcoding this was simpler.

davisagli:
@jlantz I don't want to actually merge this until there are tests, but it should be okay to start using it for testing the installer

PR #952, opened 2019-01-25
Fixes for MetaDeploy publishing

jlantz:
# Critical Changes

# Changes
* Added boolean arg `tag` to `get_latest_version` to allow returning the tag instead of the version
* `update_admin_profile` task now properly handles the default value for `package_xml` during freeze
* `metadeploy.Publish` task now raises a more helpful exception on HTTP400 that includes the response content showing what the issue was

# Issues Closed

### DIFF
--- a/cumulusci/core/config/BaseProjectConfig.py
+++ b/cumulusci/core/config/BaseProjectConfig.py
@@ -389,12 +389,14 @@ class BaseProjectConfig(BaseTaskFlowConfig):
         gh = get_github_api(github_config.username, github_config.password)
         return gh
 
-    def get_latest_version(self, beta=False):
+    def get_latest_version(self, beta=False, tag=False):
         """ Query Github Releases to find the latest production or beta release """
         gh = self.get_github_api()
         repo = gh.repository(self.repo_owner, self.repo_name)
         if not beta:
             release = repo.latest_release()
+            if tag:
+                return release.tag_name
             version = self.get_version_for_tag(release.tag_name)
             if version is not None:
                 return LooseVersion(version)
@@ -402,6 +404,8 @@ class BaseProjectConfig(BaseTaskFlowConfig):
             for release in repo.releases():
                 if not release.tag_name.startswith(self.project__git__prefix_beta):
                     continue
+                if tag:
+                    return release.tag_name
                 version = self.get_version_for_tag(release.tag_name)
                 if version is None:
                     continue
--- a/cumulusci/tasks/metadeploy.py
+++ b/cumulusci/tasks/metadeploy.py
@@ -19,6 +19,8 @@ class BaseMetaDeployTask(BaseTask):
         results = []
         while next_url is not None:
             response = self.api.request(method, next_url, **kwargs)
+            if response.status_code == 400:
+                raise requests.exceptions.HTTPError(response.content)
             response.raise_for_status()
             response = response.json()
             if "links" in response and collect_pages:
--- a/cumulusci/tasks/salesforce/UpdateAdminProfile.py
+++ b/cumulusci/tasks/salesforce/UpdateAdminProfile.py
@@ -32,10 +32,16 @@ class UpdateAdminProfile(Deploy):
 
     namespaces = {"sf": "http://soap.sforce.com/2006/04/metadata"}
 
+    def __init__(self, *args, **kwargs):
+        self._default_package_xml = False
+        self._package_xml_content = None
+        super(UpdateAdminProfile, self).__init__(*args, **kwargs)
+
     def _init_options(self, kwargs):
         super(UpdateAdminProfile, self)._init_options(kwargs)
 
         if "package_xml" not in self.options:
+            self._default_package_xml = True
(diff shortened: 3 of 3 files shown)


cumulusci/core/config/BaseProjectConfig.py
```diff
@@ -389,19 +389,23 @@ def get_github_api(self):  # pragma: nocover
         gh = get_github_api(github_config.username, github_config.password)
         return gh
 
-    def get_latest_version(self, beta=False):
+    def get_latest_version(self, beta=False, tag=False):
```
davisagli: Using an arg to control what kind of data is returned is a bit of a code smell. I'd rather see two functions, `get_latest_release_version` and `get_latest_release_tag`, and factor their common logic into a `get_latest_release`

cumulusci/tasks/metadeploy.py
```diff
@@ -19,6 +19,8 @@ def _call_api(self, method, path, collect_pages=False, **kwargs):
         results = []
         while next_url is not None:
             response = self.api.request(method, next_url, **kwargs)
+            if response.status_code == 400:
+                raise requests.exceptions.HTTPError(response.content)
```
davisagli: Why? `response.raise_for_status` already does this, with the response accessible as err.response

cumulusci/tasks/salesforce/UpdateAdminProfile.py
```diff
@@ -32,10 +32,16 @@ class UpdateAdminProfile(Deploy):
 
     namespaces = {"sf": "http://soap.sforce.com/2006/04/metadata"}
 
+    def __init__(self, *args, **kwargs):
+        self._default_package_xml = False
+        self._package_xml_content = None
+        super(UpdateAdminProfile, self).__init__(*args, **kwargs)
```
davisagli: Since the default values are immutable you can just put

    _default_package_xml = False
    _package_xml_content = None

as class members.

cumulusci/tasks/salesforce/UpdateAdminProfile.py
```diff
@@ -181,3 +184,8 @@ def _deploy_metadata(self):
         self.logger.info("Deploying updated Admin.profile from {}".format(self.tempdir))
         api = self._get_api(path=self.tempdir)
         return api()
+
+    def freeze(self, step):
+        if self._default_package_xml:
+            del self.options["package_xml"]
```
davisagli: `freeze` should not have the side effect of mutating the task instance's options. Which makes it harder to just call super() here. Why not just leave self.options['package_xml'] unset in _init_options, and make _init_task like this:

    path = self.options.get('package_xml') or os.path.join(CUMULUSCI_PATH, "cumulusci", "files", "admin_profile.xml")
    with open(path, "r") as f:
        self._package_xml_content = f.read()

Actually you can put that at the end of _init_options. That way it errors quickly if the filename is wrong, but also avoids the need to track _default_package_xml and use a custom freeze method.

cumulusci/tasks/metadeploy.py
davisagli: Oh, to make sure the response content is shown in the log.

cumulusci/core/config/BaseProjectConfig.py
jlantz: Fixed in latest commit

cumulusci/tasks/salesforce/UpdateAdminProfile.py
jlantz: Fixed using the suggestion below

cumulusci/tasks/salesforce/UpdateAdminProfile.py
jlantz: Done in latest commit

cumulusci/tasks/salesforce/UpdateAdminProfile.py
davisagli: So now there's no reason to define `__init__` here.

cumulusci/core/tests/test_config.py
```diff
@@ -504,6 +551,40 @@ def test_process_github_dependency(self):
...
+                    u"subfolder": u"src",
+                    u"unmanaged": True,
+                    u"namespace_inject": None,
+                    u"namespace_strip": None,
+                    u"namespace_tokenize": None,
+                },
+            ],
+        )
+        CUMULUSCI_TEST_REPO._contents["unpackaged/pre"] = unpackaged_pre
+        CUMULUSCI_TEST_REPO._contents["unpackaged/post"] = unpackaged_post
```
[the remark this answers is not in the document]
jlantz: I was following a pattern from other tests in this file but that makes sense.  I was actually thinking about moving the dummy repo generation into the setUp method so it gets regenerated before each test.  That might eliminate the need to have multiple different test repositories configured at all and instead each test just modifies the repository with what it needs.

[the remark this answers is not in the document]
davisagli: That sounds good to me.

PR #955, opened 2019-01-26
New Task: connected_app

jlantz:
# Critical Changes

# Changes
* New task `connected_app` makes it easier to deploy and configure the Connected App needed for CumulusCI's keychain to work with persistent orgs.  The connected app is deployed using `sfdx` to an org in the `sfdx` keychain and defaults to the `defaultdevhubusername`

# Issues Closed


cumulusci/tasks/tests/test_command.py
```diff
@@ -80,7 +81,10 @@ def test_handle_returncode(self):
         task_config = TaskConfig({"options": {"command": "ls"}})
         task = Command(self.project_config, task_config)
         with self.assertRaises(CommandException):
-            task._handle_returncode(1, "err")
+            with StringIO() as stderr:
+                stderr.write(u"err")
+                stderr.seek(0)
+                task._handle_returncode(1, stderr)
```
davisagli: The docs for subprocess say that `stderr` is "a bytes sequence, or a string if run() was called with an encoding, errors, or text=True." So I think we want to pass bytes for the test. I think you can just do `task._handle_returncode(1, io.BytesIO(b"err"))`

_handle_returncode will then need to decode the bytes

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def _run_task(self):
+        self.tempdir = tempfile.mkdtemp()
+        self._build_package()
+        self.options["command"] += " -d {}".format(self.tempdir)
+
+        try:
+            super(CreateConnectedApp, self)._run_task()
+        finally:
+            shutil.rmtree(self.tempdir)
```
davisagli: You can use `cumulusci.utils.temporary_dir` as a context manager:

    with temporary_dir() as d:
        # ...do things in d, which will get deleted when the context manager exits...

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+            "required": True,
+        },
+        "email": {
+            "description": "The email address to associate with the connected app.  Defaults to email address from the github service if configured.",
+            "required": True,
+        },
+        "username": {
+            "description": "Create the connected app in a different org.  Defaults to the defaultdevhubusername configured in sfdx.",
+            "required": False,
+        },
```
davisagli: This seems a bit odd since for all other CCI tasks we target orgs using the `--org` param. Use that if passed and otherwise use the dev hub?

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+            command="{} force:config:get defaultdevhubusername --json".format(SFDX_CLI),
+            env=self._get_env(),
+            output_handler=self._process_devhub_output,
+        )
+
+    def _process_json_output(self, line):
+        try:
+            data = json.loads(line)
+            return data
+        except Exception:
```
cdcarter: ValueError should be specific enough

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+
+    def _set_default_username(self):
+        self.logger.info("Getting username for the default devhub from sfdx")
+        self._run_command(
+            command="{} force:config:get defaultdevhubusername --json".format(SFDX_CLI),
+            env=self._get_env(),
+            output_handler=self._process_devhub_output,
+        )
+
+    def _process_json_output(self, line):
```
cdcarter: this could be static, or even just a function in utils

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+                "label value must contain only alphanumeric or underscore characters"
+            )
+
+        # Default email to the github service's email if configured
+        if "email" not in self.options:
+            try:
+                github = self.project_config.keychain.get_service("github")
+            except ServiceNotConfigured:
+                raise TaskOptionsError(
+                    "Could not determine a default for option 'email'.  Either configure the github service using 'cci service connect github' or provide a value for the 'email' option"
```
cdcarter: Did you run black?

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+    </oauthConfig>
+</ConnectedApp>"""
+
+PACKAGE_XML = """<?xml version="1.0" encoding="UTF-8"?>
+<Package xmlns="http://soap.sforce.com/2006/04/metadata">
+    <types>
+        <members>*</members>
+        <name>ConnectedApp</name>
+    </types>
+    <version>44.0</version>
```
cdcarter: make this us the project config's specified api version.

cumulusci/tasks/connectedapp.py
cdcarter: its a bootstrapping problem, right? you can't connect your devhub as a cci `--org` without a connected app.

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+        "overwrite": {
+            "description": "If True, any existing connected_app service in the CumulusCI keychain will be overwritten.  Has no effect if the connect option is False.",
+            "required": False,
+        },
+    }
+
+    def _init_options(self, kwargs):
+        self.client_id = None
+        self.client_secret = None
+        kwargs["command"] = "force:mdapi:deploy --wait 5"
```
cdcarter: parameterize the wait, please

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+        super(CreateConnectedApp, self)._init_options(kwargs)
+
+        # Validate label
+        if not re.match(r"^\w+$", self.options["label"]):
+            raise TaskOptionsError(
+                "label value must contain only alphanumeric or underscore characters"
+            )
+
+        # Default email to the github service's email if configured
```
cdcarter: the SAL team uses this little "hack" of adding email to the porject_conf for their scripts: https://github.com/SalesforceFoundation/sal/blob/master/dev_config/readme.md#use-custom-email-in-dev-data.

i like getting the github user email, personally!

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+        connected_app_path = os.path.join(self.tempdir, "connectedApps")
+        os.mkdir(connected_app_path)
+        self._generate_id_and_secret()
+        with open(
+            os.path.join(connected_app_path, self.options["label"] + ".connectedApp"),
+            "w",
+        ) as f:
+            f.write(
+                CONNECTED_APP.format(
+                    **{
```
cdcarter: instead of a splatted dict, you can just pass in kwargs, e.g.

```
CONNECTED_APP.format(label = self.option["label"], email = self.options["email"])
```
etc...

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,174 @@
...
+            f.write(PACKAGE_XML)
+
+    def _connect_service(self):
+        if not self.options["overwrite"]:
+            try:
+                connected_app = self.project_config.keychain.get_service(
+                    "connected_app"
+                )
+                raise TaskOptionsError(
+                    "The CumulusCI keychain already contains a connected_app service.  Set the 'overwrite' option to True to overwrite the existing service"
```
cdcarter: so in this case a connected app still was created and deployed to the devhub, it just doesn't overwrite the cci service? i guess that works.

cumulusci/tasks/connectedapp.py
jlantz: There's a chicken and egg type problem with this task in particular.  You would only want to create the connected app in a persistent org but you can't have a persistent org in the cci keychain until you have the connected app.  That's why this defaults to using an org from the sfdx keychain.

As a future enhancement, I might want to add another option, `prompt_password` which would prompt the user for a password and security token instead of using an org from the sfdx keychain.

cumulusci/tasks/connectedapp.py
jlantz: Nope, in python 3 it's a JSONDecodeError

jlantz: Well, I guess I should check if JSONDecodeError subclasses ValueError :)

jlantz: >>> from json.decoder import JSONDecodeError
>>> isinstance(JSONDecodeError, ValueError)
False

cdcarter: yea, because the object `JSONDecodeError` is not an instance of `ValueError`, its a subclass, so you'd need to check if an INSTANCE of jsondecodeerror isinstance of valuerror...

```
In [12]: try: json.loads('***')
    ...: except ValueError: print('hello')

Out [12]: hello
```

cumulusci/tasks/connectedapp.py
jlantz: I should probably change that to validate that it can write and not deploy if not

cumulusci/tasks/connectedapp.py
jlantz: I guess it could but then you'd have to pass it the instance with the logger to emit a log line before raising the exception.  It seems like this is a particular use case.  There's nothing special it's doing other than simply decoding json and deciding how to handle an exception.

cumulusci/tasks/connectedapp.py
jlantz: yes :)

cumulusci/tasks/connectedapp.py
jlantz: That might be unstable since I'm using a static string to generate the Connected App.  For example, if a future API version changes the format somehow, this would still work.

cumulusci/tasks/connectedapp.py
cdcarter: well thats annoying

cumulusci/tasks/connectedapp.py
jlantz: What I'm really after here is to catch anything, log a message that there was an issue, and then raise the exception that was thrown.  It seems better to cast a broad net in that use case.

cumulusci/tasks/connectedapp.py
jlantz: It seems like having this string at the top of the task file is easier to find the API version it's using to deploy than if I put the API version somewhere in the class and you have to follow through the code to figure out how it's used.  Like, I could make a class attribute called `package_xml_api_version` and then use that value.  It seems just as easy to adjust that in the string though.

davisagli: I'm fine leaving this as is.

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,176 @@
...
+    client_secret_length = 32
+    deploy_wait = 5
+    task_options = {
+        "label": {
+            "description": "The label for the connected app.  Must contain only alphanumeric and underscores",
+            "required": True,
+        },
+        "email": {
+            "description": "The email address to associate with the connected app.  Defaults to email address from the github service if configured.",
+            "required": True,
```
davisagli: This means if you don't specify an email it'll break before it gets to the point where it tries to use the email from the github service, no?

cumulusci/tasks/connectedapp.py
```diff
@@ -0,0 +1,176 @@
...
+            )
+        self.options["username"] = data["result"][0]["value"]
+
+    def _generate_id_and_secret(self):
+        self.client_id = random_alphanumeric_underscore(self.client_id_length)
+        self.client_secret = random_alphanumeric_underscore(self.client_secret_length)
+
+    def _build_package(self):
+        connected_app_path = os.path.join(self.tempdir, "connectedApps")
+        os.mkdir(connected_app_path)
```
davisagli: Note that `temporary_dir()` actually does a cd into the dir, so you don't really have to prepend self.tempdir to the paths here.

cumulusci/tasks/tests/test_connectedapp.py
```diff
@@ -0,0 +1,269 @@
...
+class TestCreateConnectedApp(unittest.TestCase):
+    """ Tests for the CreateConnectedApp task """
+
+    @classmethod
+    def setUpClass(cls):
+        super(TestCreateConnectedApp, cls).setUpClass()
+        logger = logging.getLogger(cumulusci.core.tasks.__name__)
+        logger.setLevel(logging.DEBUG)
+        cls._task_log_handler = MockLoggingHandler(logging.DEBUG)
+        logger.addHandler(cls._task_log_handler)
```
davisagli: This logging stuff should go in a mixin rather than repeating it in multiple testcase classes.

cumulusci/utils.py
```diff
@@ -484,3 +486,21 @@ def log_progress(
         if not i % batch_size:
             logger.info(progress_message.format(i))
     logger.info(done_message.format(i))
+
+
+def random_alphanumeric_underscore(length):
+    if sys.version_info[0] == 3:
+        import secrets
+
+        byte_length = math.ceil((length * 3) / 4)
```
davisagli: I see from the docs why you are doing this, but I think it's worth a comment.

cumulusci/utils.py
```diff
@@ -484,3 +486,21 @@ def log_progress(
         if not i % batch_size:
             logger.info(progress_message.format(i))
     logger.info(done_message.format(i))
+
+
+def random_alphanumeric_underscore(length):
+    if sys.version_info[0] == 3:
```
davisagli: `>= 3`

docs/tutorial.rst
```diff
@@ -385,6 +351,57 @@ There may be times when you need to import an existing scratch org that wasn't c
...
+In order to connect persistent orgs such as a Developer Edition, Enterprise Edition, or Sandbox org to CumulusCI, you need to have a Connected App configured in a persistent Salesforce org.
+
+Create With CumulusCI
+^^^^^^^^^^^^^^^^^^^^^
+
+CumulusCI includes a task to easily deploy the Salesforce Connected App to any org in your sdfx keychain.  By default, this will deploy to the org configured as the defaultdevhubusername.
+
+.. code-block:: console
+
+    $ cci task run connected_app
```
davisagli: What about the required `label` option?

docs/tutorial.rst
```diff
@@ -385,6 +351,57 @@ There may be times when you need to import an existing scratch org that wasn't c
...
+
+This command will also configure CumulusCI's connected_app service in the keychain for you.  If you want to see the information for the connected app, you can view it with:
```
davisagli: Not unless they specify `-o connect True`, right?

docs/tutorial.rst
```diff
@@ -385,6 +351,57 @@ There may be times when you need to import an existing scratch org that wasn't c
 
 You can hop into a browser logged into any org in your keychain with `cci org browser <org_name>`.
 
+Creating a Connected App
+------------------------
+
+In order to connect persistent orgs such as a Developer Edition, Enterprise Edition, or Sandbox org to CumulusCI, you need to have a Connected App configured in a persistent Salesforce org.
+
```
davisagli: Add a sentence like "You have a choice of whether to create the Connected App from the command line or in Salesforce Setup." otherwise most readers will probably go ahead with the next section without realizing there are 2 options.

requirements_dev.txt
```diff
@@ -8,6 +8,7 @@ flake8==3.7.5
 mock==2.0.0
 pre-commit==1.14.3
 pytest==4.2.0
+pytest-cov==2.6.1
```
davisagli: I've had trouble in the past with `pytest-cov` not initializing coverage early enough, so I prefer using `coverage run path/to/pytest`.

`make coverage` _should_ do the right thing, but I think it needs a fix to work when using pyenv.

docs/tutorial.rst
jlantz: The default cumulusci.yml sets label to CumulusCI

docs/tutorial.rst
jlantz: The default cumulusci.yml sets this to True.  The overwrite option is not defaulted to True though so it will avoid overwriting your existing connected_app if it's already configured.

PR #965, opened 2019-02-03
Enhancements for Bulk Data tasks

jlantz:
# Critical Changes

# Changes

## Bulk Data Task Enhancements
* Mapping enhancements for `QueryData` and `LoadData`
  * The mapping yaml file no longer requires using `Id: sf_id` as a field mapping.  If not provided, `QueryData` and `LoadData` will use local database ids instead of Saleforce OIDs for storing lookup relationships.  Previous mappings which specify the `Id: sf_id` mapping will continue to work as before using the Salesforce OID as the mapping value.
  * The mapping yaml file's `lookups:` section now handles defaults to allow simpler lookup mappings.  The only key required is now `table`.  If the `key_field` is provided it will be used for 
* The `sql_path` option on `QueryData` can be used to provide the file path where a SQL script should be written.  If this option is used, a sqlite in-memory database is used and discarded.  This is useful for storing data sets in a Github repository and allowing diffs of the dataset to be visible when reviewing Pull Requests
  * When using this option, it is best to make sure your mapping yaml file does not provide a field mapping for the `Id` field.  This will help avoid merge conflicts if querying data from different orgs such as scratch orgs.
* The `sql_path` option on `LoadData` can be used to provide the file path where a SQL script file should be read and used to load an in-memory sqlite database for the load operation
# Issues Closed


cumulusci/tasks/bulkdata.py
```diff
@@ -602,7 +678,62 @@ def _import_results(self, mapping, result_file, conn):
             process_incoming_rows(result_file, record_type), self.logger
         )
         data_file = IteratorBytesIO(processor)
-        self._sql_bulk_insert_from_csv(conn, mapping["table"], columns, data_file)
+        if mapping["oid_as_pk"]:
+            self._sql_bulk_insert_from_csv(conn, mapping["table"], columns, data_file)
+        else:
```
davisagli: There's a fair amount happening below; can you add some comments and docstrings? (I could improve on that in my work in this module too.)

cumulusci/tasks/bulkdata.py
```diff
@@ -602,7 +678,62 @@ def _import_results(self, mapping, result_file, conn):
...
+            lookup_dict = mapping["lookups"][lookup_key]
+            model = self.models[mapping["table"]]
+            lookup_mapping = self._get_mapping_for_table(lookup_dict["table"])
+            lookup_model = self.models[lookup_mapping["sf_id_table"]]
+            key_field = lookup_dict.get("key_field", lookup_key.lower())
+            key_attr = getattr(model, key_field)
+            try:
+                self.session.query(model).filter(
+                    key_attr.isnot(None), key_attr == lookup_model.sf_id
+                ).update({key_attr: lookup_model.id}, synchronize_session=False)
```
davisagli: I doubt this will perform well at large scale -- a big reason I switched to using oids as pks was so that we could simply insert (which is much faster especially using COPY FROM) without updating. So I'm glad you're still avoiding this when using oids as pks, but we should probably cover best practices for performance/scalability in docs.

cumulusci/tasks/bulkdata.py
```diff
@@ -629,16 +769,47 @@ def _create_table(self, mapping):
         if t.exists():
             raise BulkDataException("Table already exists: {}".format(mapping["table"]))
 
+        if not mapping["oid_as_pk"]:
+            mapping["sf_id_table"] = mapping["table"] + "_sf_id"
+            # If multiple mappings point to the same table, don't recreate the table
+            if not mapping["sf_id_table"] in self.models:
```
davisagli:
```suggestion
            if mapping["sf_id_table"] not in self.models:
```

cumulusci/tasks/bulkdata.py
jlantz: Added note to the docs about this.

davisagli:
@jlantz Rather than replace the old test mapping with the new one, I'd much rather run the tests with both. This will help ensure that we don't break compatibility with the oid_as_pk and old lookup configuration cases, and it'll help with coverage too.

I would also like to see tests for the SQL dump/load options. I feel a little better about waiting and filling that in later though since that part is a new feature rather than a situation involving backwards-compatibility.

PR #967, opened 2019-02-04
Allow cci service commands to be run from outside a project repository

### LINKED ISSUE #963: Allow cci service * commands when not in a git repo
It should be possible to use commands like `cci service connect github` or `cci service list` from just the global config without requiring you to be in a local git repo configured for CumulusCI.  We'd probably want to make sure you couldn't do `cci service connect github --project True` outside a git repo though.

jlantz:
# Critical Changes

# Changes

* The `cci service` commands are now available even if not in a local git repository configured for CumulusCI.  The `--project` flag is not available on `cci service connect` when outside a project.
* `cumulusci.core.runtime.BaseCumulusCI` now supports the `allow_global_keychain` kwarg which will fall back to initializing the keychain with the global config run from outside a project git repository.

# Issues Closed
Fixes #963

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -577,10 +577,10 @@ def service_list(config):
     data = []
     services = (
         config.project_config.services
-        if not config.global_keychain
+        if not config.is_global_keychain
         else config.global_config.services
     )
-    for serv, schema in list(services.items()):
+    for serv, schema in services.items():
         is_configured = ""
         if serv in config.keychain.list_services():
             is_configured = "* "
@@ -595,7 +595,7 @@ class ConnectServiceCommand(click.MultiCommand):
     def _get_services_config(self, config):
         return (
             config.project_config.services
-            if not config.global_keychain
+            if not config.is_global_keychain
             else config.global_config.services
         )
 
@@ -612,13 +612,13 @@ class ConnectServiceCommand(click.MultiCommand):
     def get_command(self, ctx, name):
         config = load_config(**self.load_config_kwargs)
         services = self._get_services_config(config)
-        attributes = iter(list(services.get(name, {}).get("attributes").items()))
+        attributes = services.get(name, {}).get("attributes").items()
         params = [self._build_param(attr, cnfg) for attr, cnfg in attributes]
-        if config.global_keychain is False:
+        if not config.is_global_keychain:
             params.append(click.Option(("--project",), is_flag=True))
 
         def callback(*args, **kwargs):
-            if config.global_keychain:
+            if config.is_global_keychain:
                 project = False
             else:
                 project = kwargs.get("project", False)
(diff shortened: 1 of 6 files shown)


cumulusci/cli/config.py
```diff
@@ -19,15 +19,20 @@ def __init__(self, *args, **kwargs):
         try:
             super(CliRuntime, self).__init__(*args, **kwargs)
         except (ProjectConfigNotFound, NotInProject) as e:
-            raise click.UsageError(str(e))
+            raise click.UsageError("{}: {}".format(e.__class__.__name__, e))
```
davisagli: What's the purpose of adding the class name here? It feels wonky to put it in front of users in cases where we're handling an expected exception that has a decently understandable message.

cumulusci/core/runtime.py
```diff
@@ -18,16 +18,24 @@ def __init__(self, *args, **kwargs):
             "load_project_config", True
         )  # this & below can be added to fn signature in py3
         load_keychain = kwargs.pop("load_keychain", True)
+        allow_global_keychain = kwargs.pop("allow_global_keychain", False)
 
         self.global_config = None
         self.project_config = None
         self.keychain = None
+        self.global_keychain = False
```
davisagli: Consider naming this `is_global_keychain` or `keychain_is_global` so it's more obviously meant to hold a boolean rather than a keychain object.

cumulusci/cli/cci.py
```diff
@@ -581,29 +590,38 @@ def service_list(config):
...
+            config.project_config.services
+            if not config.global_keychain
+            else config.global_config.services
+        )
+
     def list_commands(self, ctx):
         """ list the services that can be configured """
-        config = load_config()
-        return sorted(config.project_config.services.keys())
+        config = load_config(**self.load_config_kwargs)
```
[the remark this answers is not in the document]
jlantz: load_config is called in two places in that class and should always have the same kwargs.  It seemed better to set it in one place and use that dict.

cumulusci/cli/config.py
jlantz: That's a leftover from a prior implementation I was working on.  I'll reset it.

cumulusci/cli/cci.py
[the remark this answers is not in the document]
davisagli: How about define

    def _load_config(self):
        return load_config(allow_global_keychain=True)

and use that in both places?

This is probably mostly a matter of taste and I'll defer to your preference, but I think it makes the code easier to read because you don't have to go find where self.load_config_kwargs is defined or search the codebase to check whether it's getting set from somewhere else.

jlantz:
The error from review in `cci service connect github` should now be fixed.  I had tested calling the command but not actually completing it to save the service to the keychain which is where the error occurred.

PR #974, opened 2019-02-07
Label metadeploy steps with package name & version

davisagli:
This adjusts how the names for metadeploy steps are determined so that we can show the actual package name and version.

# Critical Changes

# Changes

# Issues Closed


davisagli:
@jstvz for the metadata bundles I'd like to come up with a way to configure a more friendly name, but I'm leaving that for a later pull request. Including the package name in the default does sound good though.

jlantz:
@davisagli We need to fix the test failures on Windows before merging.  It appears to just be an issue with assertions.

PR #975, opened 2019-02-07
Fix three typos.

prescod:
#CHANGES 

Fix some minor typos.

Please tell me if I did anything that doesn't fit your processes.


davisagli:
@prescod our process is generally that once the checks are green and someone has reviewed it, anyone may merge including the person who submitted the PR. Want to do the honors?

You can ignore the red `test_release` build. It's failing because we're in the middle of the Spring '19 release window and the packaging org for the CumulusCI-Test package is on a different version of Salesforce than the scratch org where the package is installed for testing. Should resolve itself this weekend.

PR #976, opened 2019-02-08
Update version upgrade message to include brew

jstvz:
@davisagli Should we attempt to detect whether cci has been installed via `brew` to display the appropriate upgrade command?

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -60,7 +60,7 @@ test-all: ## run tests on every Python version with tox
 	tox
 
 coverage: ## check code coverage quickly with the default Python
-	coverage run --source cumulusci pytest
+	coverage run --source cumulusci -m pytest
 	coverage report -m
 	coverage html
 	$(BROWSER) htmlcov/index.html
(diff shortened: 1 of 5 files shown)


davisagli:
@jstvz that would be neat. Just based on the path of the cumulusci module?

jstvz:
Yeah, can look for `Cellar` for darwin... linuxbrew might be trickier. I'll look it up.

Makefile
```diff
@@ -60,7 +60,7 @@ test-all: ## run tests on every Python version with tox
 	tox
 
 coverage: ## check code coverage quickly with the default Python
-	coverage run --source cumulusci pytest
```
davisagli: Oh nice.

cumulusci/cli/cci.py
```diff
@@ -40,7 +40,7 @@
 from cumulusci.core.exceptions import TaskNotFoundError
 from cumulusci.core.utils import import_class
 from cumulusci.cli.config import CliRuntime
-from cumulusci.cli.config import get_installed_version
+from cumulusci.cli.config import get_installed_version, get_cci_upgrade_command
```
davisagli: This should be imported from its real location in cumulusci.utils

PR #978, opened 2019-02-10
Support relative datetime in bulkdata.QueryData and bulkdata.LoadData

jlantz:
NOTE: This branch has the feature/bulkdata branch merged into it since some of that work was needed for testing.  Please merge the feature/bulkdata PR first which should simplify the diff for reviewing this PR.

# Critical Changes

# Changes

* The new `relative_date` key can be used in a mapping to specify that a date or datetime field should be handled as a relative date which stores the value as an integer of seconds from a timedelta.
* The `relative_day_offset` option adjusts the datetime.now() date used to determine relative dates.  This allows capturing a data set now as if it were captured N days ago.

# Issues Closed


cumulusci/tasks/bulkdata.py
```diff
@@ -781,6 +899,7 @@ def _fields_for_mapping(self, mapping):
             )
         return fields
 
+>>>>>>> master
```
davisagli: @jlantz bad merge

cumulusci/tasks/bulkdata.py
```diff
@@ -58,6 +59,22 @@ def setup_epoch(inspector, table, column_info):
...
+        self.now = now
+        super(RelativeDateType, self).__init__(*args, **kwargs)
+
+    def process_bind_param(self, value, dialect):
+        if isinstance(value, str):
+            value = datetime.datetime.fromisoformat(value[:-1])
+        return int((value - self.now).total_seconds())
+
+    def process_result_value(self, value, dialect):
+        return self.now + datetime.timedelta(seconds=value)
```
davisagli: This looks like it is more or less the same thing as EpochType except:
- it allows configuring the epoch date
- it coerces string bind params to datetimes
- it stores the offset in seconds rather than microseconds

Would you be okay switching it to store microseconds so we could merge the two classes? (Actually we could do that anyway if we added a `resolution` attribute.)

cumulusci/tasks/bulkdata.py
```diff
@@ -554,6 +652,12 @@ def _init_options(self, kwargs):
             self.logger.info("Using in-memory sqlite database")
             self.options["database_url"] = "sqlite://"
             self.options["sql_path"] = os_friendly_path(self.options["sql_path"])
+        # Prepare the datetime used to calculate relative_date values
+        self.relative_date_now = datetime.datetime.now()
+        if self.options.get("relative_day_offset"):
+            self.relative_date_now += datetime.timedelta(
+                days=float(self.options["relative_day_offset"])
+            )
```
davisagli: Can we put this in a function in the mixin so we aren't duplicating the logic

cumulusci/tasks/bulkdata.py
```diff
@@ -579,7 +683,7 @@ def _init_db(self):
         self.metadata.bind = self.engine
 
         # Create the tables
-        self._create_tables()
+        self._create_tables(create=True)
```
davisagli: A function named `_create_tables` that takes a boolean `create` param reads really strangely. What's the actual distinction between its two modes of operation and can we give it a better name based on that?

cumulusci/tasks/bulkdata.py
```diff
@@ -807,14 +926,24 @@ def _download_file(uri, bulk_api):
...
         record_type = record_type.encode("utf-8")
     for line in f:
         if record_type:
-            yield line.rstrip() + b"," + record_type + b"\n"
-        else:
-            yield line
+            line = line.rstrip() + b"," + record_type + b"\n"
+        if relative_date_now:
+            line = line.decode("utf8")
+            for isoformat in re.findall(isoformat_regex, line):
```
davisagli: Shouldn't we target only the columns that were marked as dates? If there's a text field that happens to have an ISO format date in it, we wouldn't want to turn that into an integer.

cumulusci/tasks/bulkdata.py
jlantz: I was actually confused about why EpochType uses microseconds anyhow.  The Salesforce DateTime field doesn't support microseconds.

Aside from that, the reason I did these as separate is that they have slightly different use cases:

* EpochType always uses the epoch date as the reference point and it seems clearer to have the class set that than to require it to be passed on init
* RelativeDateType requires you pass a value for now which is used to calculate the differences.  We could have a single class that optionally takes now as a kwarg but that seems a bit obscure.

cumulusci/tasks/bulkdata.py
jlantz: I need to think about how to implement that.  This implementation doesn't require awareness of the position on columns in rows.  Supporting only columns specified as relative_date columns would require passing more args to this method.  I'll look into it though.

cumulusci/tasks/bulkdata.py
jlantz: Oops, apparently I didn't save the conflict resolution before commit

cumulusci/tasks/bulkdata.py
davisagli: I'm not sure why it uses microseconds (that was here before I started working on it) but that is for the SQL db, not Salesforce.

They don't sound like different use cases to me. We're storing as an integer relative to a base date, which should be the standard epoch date by default but could be something else if specified. A single class whose `__init__` takes a `relative_to` param that defaults to the epoch date sounds perfect.

davisagli:
Replaced by #2015

PR #980, opened 2019-02-11
Generate encryption key and store using keyring

davisagli:
This updates where the CLI looks for the keychain key. Rather than requiring the user to configure the CUMULUSCI_KEY environment variable, we now generate a key and store it in the system keychain using the `keyring` library. The environment variable is still given priority for backwards compatibility, and if there's no key in the keyring yet it will be migrated from the environment variable.

Apologies for also mixing some cleanup into the same pull request (updating CliConfig references to CliRuntime and misc flake8 fixes). The primary changes are in cumulusci/cli/config.py

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -4,7 +4,6 @@ from future import standard_library
 standard_library.install_aliases()
 from past.builtins import basestring
 from builtins import str
-from builtins import object
 from collections import OrderedDict
 import functools
 import json
@@ -13,7 +12,6 @@ import os
 import sys
 import webbrowser
 import code
-import yaml
 import time
 
 from contextlib import contextmanager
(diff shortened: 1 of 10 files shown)


cumulusci/cli/tests/test_config.py
```diff
@@ -34,33 +37,62 @@ def test_init(self):
...
+    def test_get_keychain_key__generates_key(self, keyring):
+        del os.environ["CUMULUSCI_KEY"]
+        keyring.get_password.return_value = None
+
+        config = CliRuntime()
+        self.assertNotEqual(self.key, config.keychain.key)
+        self.assertEqual(16, len(config.keychain.key))
+
+    @mock.patch("cumulusci.cli.config.keyring")
+    def test_get_keychain_key__warns_if_generated_key_cannot_be_stored(self, keyring):
```
jstvz: I hadn't noticed this dunder in the function anywhere else. I think it's fine, just curious if it has some different behavior or is purely aesthetic?

docs/tutorial.rst
```diff
@@ -98,17 +98,6 @@ If you want to use our example project, fork our CumulusCI-Test repo:
 
 If you are using the CumulusCI-Test repo with a Developer Edition Salesforce org, you will need to enable Chatter in the org if it is not already enabled.  With Salesforce DX Scratch Orgs, this is handled for you.
 
-Keychain Key
```
jstvz: Do we want to mention the `$CUMULUSCI_KEY` as a potential workaround for people on platforms not as seamlessly supported by `keyring` (headless linux)?

cumulusci/cli/config.py
```diff
@@ -37,7 +39,23 @@ def get_keychain_class(self):
...
+        except Exception:
+            key_from_keyring = None
+            has_functioning_keychain = False
+        # If no key in environment or file, generate one
+        key = key_from_env or key_from_keyring
+        if key is None:
+            if has_functioning_keychain:
+                key = random_alphanumeric_underscore(length=16)
+            else:
+                raise KeychainKeyNotFound("Unable to store encryption key.")
```
jstvz: Perhaps mention the environment var as fallback?

cumulusci/cli/tests/test_config.py
davisagli: I tend to follow a pattern of test_[method name]__[short description of variation being tested] -- the dunder is just to make it clearer which part is the method name.

docs/tutorial.rst
davisagli: Hmm. I think we should put that somewhere, but I don't think the tutorial is the right place unless we can tell people how to know they need to do this.

jstvz: Yeah, a troubleshooting/FAQ page would be a better place for it. If we do want to include it here, we can just start with something like:

> If you see `<KeychainKeyNotFound traceback>` you'll need to `<set env_var/install deps on linux>`

I think the updated exception below might suffice for now, though.

cumulusci/cli/cci.py
```diff
@@ -623,7 +615,7 @@ def callback(*args, **kwargs):
             else:
                 project = kwargs.get("project", False)
             serv_conf = dict(
-                (k, v) for k, v in list(kwargs.items()) if v != None
+                (k, v) for k, v in list(kwargs.items()) if v is not None
```
prescod: Is this just a coding style preference or is there a real value that is equal to None but is not None which we are concerned about?

davisagli: It probably doesn't actually matter here, but I changed it because flake8 suggested the change; I presume the lint rule is there to avoid calling `__eq__` when an identity check will suffice.

PR #986, opened 2019-02-12
Fix update checking timestamp reading.

prescod:
w+ erases the file first. r+ doesn't.

So before, it would erase the file before trying to read the value. The value was therefore always replaced with "0". Now it will read the result or write the result based on the caller's preference.

We don't cache this underlying file object so its safe that we read it and then write to it a few milliseconds later. It's a different file object.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -57,8 +57,14 @@ def timestamp_file():
     if not os.path.exists(config_dir):
         os.mkdir(config_dir)
 
-    with open(os.path.join(config_dir, "cumulus_timestamp"), "r+") as f:
-        yield f
+    timestamp_file = os.path.join(config_dir, "cumulus_timestamp")
+
+    try:
+        with open(timestamp_file, "r+") as f:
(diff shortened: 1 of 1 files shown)


davisagli:
@prescod that makes sense for why it was always doing the check, but with this change it breaks if the file doesn't exist yet (see the failing tests)

prescod:
Embarrassed that I forgot to run the tests.

The tests now run cleanly on my computer and I tested the scenario by hand as well. 

I decided to code the check as an exception instead of a conditional because it should only trigger the exception once per installation and take the happy path hundreds or thousands of times.

davisagli:
@prescod no worries, that's what CI is for.

I expect the tests to fail on Python2 with `NameError: name 'FileNotFoundError' is not defined`. I think catching OSError will work in both Pythons.

davisagli:
@prescod sorry, looks like it's IOError, not OSError.

If you want you can run the tests locally on Python 2 using `tox -e py27` (assuming you have python2.7 installed and in your PATH)

PR #995, opened 2019-02-20
Updated Windows instructions in the tutorial to be accurate and added some additional detail

robertbwatson:
# Critical Changes

# Changes
* Some of the instructions in the tutorial were inaccurate for Windows users. The tutorial has been updated to be accurate and more clear.

# Issues Closed

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -64,20 +64,20 @@ If you would like to automatically activate the virtual environment whenever you
 Windows
 ^^^^^^^
 
-Ensure you are in your home directory (C:\Users\your.username) and execute the following commands:
+Execute the following commands in a Windows shell, such as Git Bash:
 
 .. code-block:: powershell
 
-    python -m venv cci
-    cci\Scripts\activate
+    python -m venv ~/cci
+    source ~/cci/Scripts/activate
 
 .. note::
    If this fails because you don't have the ability to run scripts,
    try running: ``Set-ExecutionPolicy Unrestricted -Scope CurrentUser``
 
-Similar to the macOS commands above, this creates a virtualenv in a new folder named ``cci`` and activates it.
+Similar to the macOS commands above, this creates a virtualenv in a new folder in your home directory (C:\Users\your.username\) named ``cci`` and activates it.
 
-If you would like to automatically activate the virtual environment whenever you open a new terminal, you can add the activate command to your ``~/.bash_profile``.
+If you would like to automatically activate the virtual environment whenever you open a new shell, you can add the activate command to your ``~/.bash_profile``.
 
 Install CumulusCI
 -----------------
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -36,7 +36,9 @@ In the installer, be sure to check the "Add Python to PATH" checkbox.
 Create Virtual Environment
 --------------------------
 
-A Python Virtual Environment (virtualenv) is an isolated Python environment where you can install packages without modifying the system Python.  Using a virtualenv for cumulusci is recommended to avoid issues and conflicts with other applications using your system Python.
+A Python Virtual Environment (virtualenv) is an isolated Python environment where you can install packages without modifying the system Python.  Using a virtualenv for CumulusCI is recommended to avoid issues and conflicts with other applications using your system Python.
+
+For more information about virtual environments in Python, see the Python Packaging User Guide: https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments
 
 macOS
 ^^^^^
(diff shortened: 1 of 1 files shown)


docs/tutorial.rst
```diff
@@ -36,7 +36,9 @@ In the installer, be sure to check the "Add Python to PATH" checkbox.
 Create Virtual Environment
 --------------------------
 
-A Python Virtual Environment (virtualenv) is an isolated Python environment where you can install packages without modifying the system Python.  Using a virtualenv for cumulusci is recommended to avoid issues and conflicts with other applications using your system Python.
+A Python Virtual Environment (virtualenv) is an isolated Python environment where you can install packages without modifying the system Python.  Using a virtualenv for CumulusCI is recommended to avoid issues and conflicts with other applications using your system Python.
+
+For more information about virtual environments in Python, see the Python Packaging User Guide: https://packaging.python.org/tutorials/installing-packages/#creating-virtual-environments
```
robertbwatson: I moved this out of the macOS section because I feel like a Windows user might just skip over this and not see it. It applies to all users.

robertbwatson:
@jstvz Here is the pull request of the changes I messaged you about earlier. Thanks!

I'm going to have Kiet take a look at this as well.

docs/tutorial.rst
```diff
@@ -59,24 +61,28 @@ Your shell prompt should change once you are in the virtual env to show (cci) at
...
 ^^^^^^^
 
+Ensure you are in your home directory (C:\Users\your.username) and execute the following commands:
+
 .. code-block:: powershell
 
-    python -m venv ~\cci
-    ~\cci\Scripts\Activate.ps1
+    python -m venv cci
+    cci\Scripts\activate
```
davisagli: @robertbwatson were you using PowerShell or not? We may need to provide different instructions based on which shell the user is working in. (But, it's also entirely possible that the PowerShell commands were incorrect, since I didn't have access to test this when I wrote it.)

docs/tutorial.rst
```diff
@@ -59,24 +61,28 @@ Your shell prompt should change once you are in the virtual env to show (cci) at
...
 
 .. note::
    If this fails because you don't have the ability to run scripts,
    try running: ``Set-ExecutionPolicy Unrestricted -Scope CurrentUser``
 
+Similar to the macOS commands above, this creates a virtualenv in a new folder named ``cci`` and activates it.
+
+If you would like to automatically activate the virtual environment whenever you open a new terminal, you can add the activate command to your ``~/.bash_profile``.
```
davisagli: This paragraph doesn't apply to Windows unless you're using bash somehow, which is not what we're assuming above.

docs/tutorial.rst
robertbwatson: No, just the command prompt. Maybe that's why it wasn't working. Salesforce employees on Windows are prohibited from running PowerShell.

docs/tutorial.rst
robertbwatson: I'll re-run through the tutorial and confirm these instructions again. I use Git Bash by default, since we aren't allowed to use PowerShell.

davisagli: @robertbwatson Jason wrote up docs for setting up CumulusCI with VSCode and Git Bash for the community sprint, and I think the intent is to turn them into updated installation documentation here. Those have gotten a lot more real world testing.

robertbwatson: @davisagli Yep, I've been using those and helped Jason for the initial walk-through.

I just made another commit with some updates here, but if a larger documentation update is coming soon that's great too.

docs/tutorial.rst
MatthewBlanski: You want to be careful with this as well, as the command will create the folder in whatever directory one happens to be in. For command prompt, I believe you want to use `python -m venv %userprofile%/cci` . I am going to confirm that in short order. That should place cci in something approaching the 'correct' location, rather than just buried in a random folder.

robertbwatson: This is outdated. The new commit I made references Git Bash instead of the command prompt. When you open a new Git Bash I believe you are automatically placed in your home directory by default, but I could be wrong.

That being said, I opened this PR because I wanted to make some quick changes to make it less confusing for the Windows setup but it sounds like there might be grander changes coming, so I don't know if anything will actually happen to this PR.

robertbwatson:
@davisagli @jstvz Do you two think this will ever be merged, or is there a plan in place to completely revamp the online/public-facing tutorial/documentation? If the latter, I'm okay just closing this. (But I would really like to improve it. :))

davisagli:
@robertbwatson we've been holding off because James is investigating a different way of installing on Windows that would avoid the need to activate a virtualenv.

davisagli:
We merged #1045 recommending a different approach for installation on Windows.

PR #996, opened 2019-02-20
Delete date_created on delete

21aslade:
# Critical Changes
None
# Changes
Deletes the "date created" field when deleting a scratch org. This makes the "days" column in `cci org list` display properly after deletion.
# Issues Closed
None


davisagli:
Thanks for the fix, @21aslade! I'm going to close this pull request and open a new one from a branch within our github org so that the CI system will run the builds.

PR #997, opened 2019-02-21
new task robot_documentation

boakley:
This replaces the old robot_libdoc task with a new task that can creates a single html page of documentation for multiple keyword files (both python and robot format). It takes the same arguments as the old task, though the `path` argument has been enhanced to support passing in a list of files. 

In the HEDA repository, you can generate docs like this:

    $ cci task run robot_libdoc -o path tests/HEDA.py,tests/HEDA.robot -o output index.html

Similarly, for NPSP you can do the following:

    $ cci task run robot_libdoc -o path tests/NPSP.py,tests/NPSP.robot -o output index.html

For CumulusCI, I've replace the old tasks `robot_libdoc_cumulusci`, `robot_libdoc_cumulusci_resource`,`robot_libdoc_salesforce`, and `robot_libdoc_salesforce_resource` with a single `robot_libdoc` task.  I've updated the `robot_docs` flow to use the new task to combine documentation of all keywords into a single file. 

I've updated the robotframework.rst file we use to generate the sphinx documentation to point to this new file. 

# Critical Changes

# Changes
* the old robot_libdoc task has been replaced with a new task of the same name that can generate a single HTML document for multiple keyword files by passing a comma-separated list of files to the `path` option. 

# Issues Closed
- W-031412


cumulusci/tasks/robotframework/template.html
```diff
@@ -0,0 +1,66 @@
...
+              search()
+          })
+      });
+    </script>
+    <style>
+      {{stylesheet}}
+    </style>
+  </head>
+  <body>
+    <div class='container'>
```
davisagli: Not a big deal, but I'm distracted by the single-quoted HTML attributes, especially mixed with double quotes in the same file.

cumulusci/cumulusci.yml
```diff
@@ -233,6 +233,12 @@ tasks:
...
             suites: tests
         group: Robot Framework
+    robot_documentation:
+        description: Generates documentation for project keyword files
+        class_path: cumulusci.tasks.robotframework.RobotDoc
+        options:
+            outputdir: docs
+        group: Robot Framework
     robot_testdoc:
         description:
```
davisagli: It seems potentially confusing to provide 2 different tasks for generating robot framework documentation. Is there any reason we should keep the old one? If not, I would be inclined to completely replace it with the new one, but keep the same task name (robot_testdoc) and option names (path and output) to smooth the transition.

cumulusci/tasks/robotframework/robotdoc.py
```diff
@@ -0,0 +1,98 @@
...
+class RobotDoc(BaseTask):
+    task_options = {
+        "outputdir": {
+            "description": "The path to a folder where generated files will be placed",
+            "required": True,
+        },
+        "files": {
+            "description": "A list of keyword definition files (eg: .py, .robot, .resource)",
+            "required": True,
+        },
```
davisagli: I think we talked about giving the task the path to a folder and making it able to discover the keyword libraries contained in the folder. Did you decide that won't work for some reason? It would require less thought when running the task, and we could provide `tests/` as a default path that would generally work for our products without requiring option overrides in the cumulusci.yml for each product.

cumulusci/tasks/robotframework/robotdoc.py
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    def _validate_options(self):
+        super(RobotDoc, self)._validate_options()
+
+        self.options["files"] = process_list_arg(self.options["files"])
+        if not os.path.exists(self.options["outputdir"]):
+            os.mkdir(self.options["outputdir"])
```
davisagli: Use `os.makedirs` to handle if there are multiple levels of directories.

cumulusci/tasks/robotframework/robotdoc.py
```diff
@@ -0,0 +1,98 @@
...
+                # oddly, robot's exception has a traceback embedded in the message, so we'll
+                # only print out the first line to hide most of the noise
+                self.logger.warn("unexpected error: {}".format(str(e).split("\n")[0]))
+
+        stylesheet_path = os.path.join(os.path.dirname(__file__), "stylesheet.css")
+        with open(stylesheet_path) as f:
+            stylesheet = f.read()
+
+        with open(os.path.join(output_dir, "index.html"), "w") as f:
+            title = "{} Keywords".format(self.project_config.project__package__name)
```
davisagli: We generally try to avoid direct references to the project_config in task code because it keeps the tasks less coupled to the rest of CCI and easier to test.

Instead I would recommend:
1. Add a "title" option to the task.
2. In cumulusci.yml where the task is defined, set the title option to `$project_config.project__package__name`

cumulusci/cumulusci.yml
boakley: I agree. The only real problem with keeping the same arguments is that my script takes the path to more than one file. That's a pretty minor thing, though.

davisagli: It's backwards compatible as long as no one was using commas in their path, which seems unlikely.

cumulusci/tasks/robotframework/template.html
boakley: That's a character weakness showing through. I've been terribly inconsistent about that my whole life and can't seem to shake it. I think I have a small tumor right next to the "quote" brain cell that sometimes swells and sometimes shrinks.

cumulusci/tasks/robotframework/robotdoc.py
boakley: Did I decide it didn't work? No, not really. I just chose the simpler route for the time being. I've got code that can detect, but wan't sure if we wanted to stick with "explicit is better than implicit" in this case. I'm open to either. I definitely see value in it working right out of the box without the user having to edit their config file.

It will be pretty easy for me to switch the code to auto-detect.

cumulusci/tasks/robotframework/robotdoc.py
boakley: Good point. I was also not sure if I should auto-create it or not. Autocreate works great in the normal case, but it could be annoying if you typo'd the path in the config file and it created some random directory somewhere that you had to clean up.

cumulusci/tasks/robotframework/robotdoc.py
boakley: ok. Thanks for the suggestion. I wasn't sure if what I was doing was kosher or not.

cumulusci/tasks/robotframework/template.html
davisagli: We can probably solve it by setting up an autoformatter.

cumulusci/tasks/robotframework/robotdoc.py
davisagli: Good point.

cumulusci/tasks/robotframework/robotdoc.py
boakley: I've made this change, though it means the default will be something like "CumulusCI" rather  than "CumulusCI Keywords", but that's not a big deal.

cumulusci/tasks/robotframework/robotdoc.py
boakley: I switched to using os.makedirs.

cumulusci/tasks/robotframework/template.html
boakley: fixed

davisagli:
@boakley note the failing tests on Windows (appveyor build)

cumulusci/cumulusci.yml
boakley: In the code I pushed earlier today I completely replaced the old robot_libdoc task with the new code. The new task takes the same arguments as the old task.

cumulusci/tasks/robotframework/robotdoc.py
boakley: I haven't added the auto-detect code yet, and I think I'd prefer that we save that for a new release. This little project has already seen a some feature creep. I say we leave auto-detect out for the moment, and perhaps create a separate story for some enhancements in the interest of getting this out the door.

cumulusci/cumulusci.yml
```diff
@@ -233,6 +233,13 @@ tasks:
         options:
             suites: tests
         group: Robot Framework
+    robot_documentation:
+        description: Generates documentation for project keyword files
+        class_path: cumulusci.tasks.robotframework.RobotDoc
+        options:
+            outputdir: docs
+            title: $project_config.project__package__name
+        group: Robot Framework
```
davisagli: @boakley this task definition needs to be updated

cumulusci/tasks/robotframework/libdoc.py
```diff
@@ -0,0 +1,106 @@
...
+                files = ", ".join(["'{}'".format(filename) for filename in bad_files])
+                error_message = "Unable to find the following input files: {}".format(
+                    files
+                )
+            raise TaskOptionsError(error_message)
+
+    def _run_task(self):
+        libraries = []
+        processed_files = {}
+        for input_file in sorted(self.options["path"]):
```
davisagli: Why sorted? Seems like it could be nice to keep it in the order specified in the task option, in case there's a primary library that doesn't come first alphabetically.

boakley: I don't have a good answer. I debated that, and do see value in letting the order be specified in the yml. I see your point, so I've removed the sorting.

Thanks for catching this!

cumulusci/cumulusci.yml
boakley: Done

PR #999, opened 2019-02-21
Support deploying lightning web components

spelak-salesforce:
# Critical Changes

# Changes

-  Added `LightningComponentBundleParser` in `cumulusci/tasks/metadata/package.py` to generate members for Lightning Web Component metadata
    -  Generates member for each direct child folder of `src/lwc`
    -  Copied `MetadataFolderParser` but ignores parsing subitems.  Files in `lwc` folders are not required to have the same name as the folder
    -  Tests `LightningComponentBundleParser` with `TestLightningComponentBundleParser` in `cumulusci/tasks/metadata/tests/test_package.py`
-  Added support for `lwc` in `cumulusci/tasks/metadata/metadata_map.yml`

# Issues Closed


spelak-salesforce:
**lurch attach W-031389

cumulusci/tasks/metadata/package.py
```diff
@@ -1,12 +1,12 @@
-from builtins import str
+from builtins import str  # noqa: F401
```
davisagli: You can go ahead and remove unused imports instead of ignoring the warnings; that should have happened already.

cumulusci/tasks/metadata/package.py
```diff
@@ -362,6 +362,30 @@ def _parse_item(self, item):
...
+
+        # ignore subitems
+
+        return members
+
+    def check_delete_excludes(self, item):
+        return False
+
+    def _parse_subitem(self, item, subitem):
+        return []
```
davisagli: What do you think about replacing both `AuraBundleParser` and `LightningComponentBundleParser` with a single `BundleParser` that is used for both?

Skipping non-directories won't cause problems for Aura.

I don't think we need the check for namespaced members. I assume a package can't override an LWC from another namespace.

I think you can leave out `check_delete_excludes` (the inherited implementation looks innocuous and could be handy sometime) and `_parse_subitem` (it's only called from BaseMetadataParser's `_parse_item` which you've replaced)

cumulusci/tasks/metadata/tests/test_package.py
```diff
@@ -120,7 +121,7 @@ def test_parse_items__skips_files(self):
                 "Account.object",
                 "Custom__c.object",
             ):
-                with open(filename, "w") as f:
+                with open(filename, "w") as f:  # noqa: F841
```
davisagli: Again, I'd rather you fix the problem it's warning about instead of suppressing the warning.

```suggestion
                with open(filename, "w"):
```

cumulusci/tasks/metadata/tests/test_package.py
```diff
@@ -195,7 +221,9 @@ def test_parser(self):
 
     def test_parser__missing_item_xpath(self):
         with self.assertRaises(ParserConfigurationError):
-            parser = MetadataXmlElementParser("TestMDT", None, "test", False)
+            parser = MetadataXmlElementParser(  # noqa: F841
+                "TestMDT", None, "test", False
+            )
```
davisagli: Ditto.

cumulusci/tasks/metadata/package.py
spelak-salesforce: I will make the changes

cumulusci/tasks/metadata/package.py
spelak-salesforce: will do

spelak-salesforce: pushed changes

jstvz:
@Auchtor Here's the PR I mentioned this morning.

cumulusci/tasks/metadata/tests/test_package.py
spelak-salesforce: This change is added this in a commit before I read this request.

cumulusci/tasks/metadata/tests/test_package.py
spelak-salesforce: I added a dummy assert that `parser` is not `none` so the variable is used.  I don't know what needs to be tested: test name is `test_parser__missing_item_xpath`

davisagli: There's an implicit assertion that the parser can be instantiated without an error. The test is probably here mostly to ensure coverage of a particular code path during instantiation.

cumulusci/tasks/metadata/package.py
spelak-salesforce: Changes are commited and pushed

cumulusci/tasks/metadata/package.py
```diff
@@ -355,11 +353,22 @@ def check_delete_excludes(self, item):
...
-        return [item]
+        members = []
+        path = self.directory + "/" + item
+
+        # Skip non-directories
+        if not os.path.isdir(path):
+            return members
+
+        # Add the member if it is not namespaced
+        if "__" not in item:
```
davisagli: When would an Aura or LWC bundle contain a namespaced member? I don't think we need this condition.

cumulusci/tasks/metadata/metadata_map.yml
```diff
@@ -108,6 +108,9 @@ layouts:
 letterhead:
     - type: Letterhead
       class: MetadataFilenameParser
+lwc:
+    - type: LightningComponentBundle
+      class: LightningComponentBundleParser
```
davisagli: Should be `BundleParser` now.

cumulusci/tasks/metadata/package.py
spelak-salesforce: All Bundles should be in the local namespace for the repo.  I can't think of a way that another namespace's Bundle would be included in the metadata.

spelak-salesforce: Changed pushed

cumulusci/tasks/metadata/metadata_map.yml
spelak-salesforce: Great catch.  I thought I changed this when (re-)defining `BundleParser`

spelak-salesforce: Changes pushed

spelak-salesforce:
> but I don't really want to merge this until we're actually able to deploy LWCs without errors. That still requires work on the Deploy task to filter out some files, right?

@davisagli , yes, I am working on the **deploy** now by looking at `cumulusci/tasks/salesforce/Deploy.py` and `.../DeployBundle.py`.  I will add these changes to this Pull Request unless advised otherwise.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +42,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
...
+    def _include_lwc_component_file(self, f):
+        return self._lwc_component_file_extension_whitelist.match(f)
+
+    def _get_api(self, path=None):
+        if not path:
+            path = self.task_config.options__path
+
+        package_zip = self._get_package_zip(path)
+        self.logger.info("Payload size: {} bytes".format(len(package_zip)))
+        return self.api_class(self, package_zip, purge_on_delete=False)
```
davisagli: Moving this is not a problem (don't change it back), but generally I'd prefer for code cleanup to happen in separate pull requests from substantive changes because it makes it easier to know what to pay attention to in code review.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +42,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
         },
     }
 
+    # ignore all files in ./lwc and all sub-directories of a lwc component (i.e. ./lwc/[lwc Component Name]/.*)
+    _directory_blacklist = re.compile("^\\.\\/lwc(\\/.*?\\/.*)?$")
```
davisagli: In another comment I'm recommending moving away from regexes anyway, but fwiw in Python you can use a "raw string" to avoid needing to escape backslashes: `r"^\.$"`

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +42,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
...
-                    self._write_zip_file(zipf, root, f)
+                if self._include_directory(root):
+                    if self._is_lwc_component(root):
+                        for f in files:
+                            if self._include_lwc_component_file(f):
+                                self._write_zip_file(zipf, root, f)
+                    else:
+                        # Include all files
+                        for f in files:
+                            self._write_zip_file(zipf, root, f)
```
davisagli: I'd like to have all the filtering logic in a single method because it will make both the directory-walking logic and the filtering logic clearer (and easier to test) if they are separate. I'm anticipating that we may need to add different filters for other component types in the future, and that will be cleanest if things are sufficiently decoupled that we can delegate the logic for each component to a different class.

I also think it's going to be hard to make this work using regular expressions on Windows which uses a backslash as a path separator. How about something like this:

    for root, dirs, files in os.walk("."):
        root_parts = root.split(os.sep)[1:]
        for f in files:
            parts = root_parts + f.split(os.sep)
            if not self._include_file(parts):
                continue
            self._write_zip_file(zipf, root, f)

And then `_include_file` can work with a list of path segments rather than a string.

spelak-salesforce: -  I will decouple the logic into separate methods.
-  I'm surprised using regex is really a compatibility issue (and it makes sense that it is).  I like your solution and will implement

cumulusci/tasks/salesforce/Deploy.py
spelak-salesforce: Applying suggestions in another comment, I will split `root` and do logic  with the parts instead of regex

cumulusci/tasks/salesforce/Deploy.py
davisagli: It's not the regex per se, it's the fact that the patterns aren't accounting for varying path separators. I haven't thought through the details of what making assertions against a list of path elements will look like, so tell me if it turns out to not make sense.

spelak-salesforce: I created methods `_include_directory` and `_include_file`.  The methods don't interact with the classes state at all.  Can they be static methods, or is it preferred I leave them as instance methods?

davisagli: @spelak-salesforce let's leave them as instance methods. That way if we need to refer to the task options at some point to determine what to include, that will be possible.

cumulusci/tasks/salesforce/Deploy.py
spelak-salesforce: Change is done.  directory/file checking is done in `_include_directory`/`_include_file` respectively

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +41,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
...
+            return (
+                lower_f.endswith(".js")
+                or lower_f.endswith(".js-meta.xml")
+                or lower_f.endswith(".html")
+                or lower_f.endswith(".css")
+                or lower_f.endswith(".svg")
+            )
+        return True
+
+    def _get_files_to_package(self, path):
```
spelak-salesforce: Originally, the logic inside of `_get_files_to_package` was contained in `_get_package_zip`, but the asserts in the test give anything helpful if they failed since `_get_files_to_package` returns a base 64 encoded string.   I created `_get_files_to_package` so I could easily assert what files were included in the relative path.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -117,6 +143,3 @@ def _process_meta_xml(self, zipf):
         )
         zipf = zip_clean_metaxml(zipf, logger=self.logger)
         return zipf
-
-    def _write_zip_file(self, zipf, root, path):
```
spelak-salesforce: `_write_zip_file` usage is replaced in two spots in the changes Lines 76 of `_get_files_to_package` and 86 of `_get_package_zip`

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -1,19 +1,20 @@
 import base64
 import io
-import mock
```
spelak-salesforce: import not used

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -1,19 +1,20 @@
...
 from cumulusci.utils import temporary_dir
+from cumulusci.utils import cd
 from .util import create_task
+from future.utils import bytes_to_native_str
 
 
 class TestDeploy(unittest.TestCase):
     def test_get_api(self):
         with temporary_dir() as path:
-            with open(os.path.join(path, "package.xml"), "w") as f:
```
spelak-salesforce: `f` not used

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -31,7 +32,7 @@ def test_get_api(self):
 
     def test_get_api__managed(self):
         with temporary_dir() as path:
-            with open(os.path.join(path, "package.xml"), "w") as f:
```
spelak-salesforce: `f` not used

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -43,10 +44,356 @@ def test_get_api__managed(self):
 
     def test_get_api__skip_clean_meta_xml(self):
         with temporary_dir() as path:
-            with open(os.path.join(path, "package.xml"), "w") as f:
```
spelak-salesforce: `f` not used

davisagli:
@spelak-salesforce did you see the failed Appveyor build? That's the one that runs the tests on Windows. I haven't looked closely yet but probably you have a literal forward-slash somewhere where you should use `os.sep`

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +41,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
...
+                lower_f.endswith(".js")
+                or lower_f.endswith(".js-meta.xml")
+                or lower_f.endswith(".html")
+                or lower_f.endswith(".css")
+                or lower_f.endswith(".svg")
+            )
+        return True
+
+    def _get_files_to_package(self, path):
+        files_to_pacakge = []
```
davisagli: fix spelling please, throughout

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +41,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
...
+        for root, dirs, files in os.walk("."):
+            root_parts = root.split(os.sep)[1:]
+            if self._include_directory(root_parts):
+                for f in files:
+                    if self._include_file(root_parts, f):
+                        files_to_pacakge.append(os.path.join(root, f))
```
davisagli: I like the factoring out of `_get_files_to_package`. Since the method is only used as an iterator, I would suggest turning it into a generator by using the `yield` statement. Just turn this line into `yield os.path.join(root, f)` and remove the first and last lines of the method. This has the benefit of only holding one filename in memory at a time rather than constructing the full list before returning it.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -42,29 +41,56 @@ class Deploy(BaseSalesforceMetadataApiTask):
...
+        if len(root_parts) == 2 and root_parts[0] == "lwc":
+            # is file of lwc component directory
+            lower_f = f.lower()
+            return (
+                lower_f.endswith(".js")
+                or lower_f.endswith(".js-meta.xml")
+                or lower_f.endswith(".html")
+                or lower_f.endswith(".css")
+                or lower_f.endswith(".svg")
+            )
```
davisagli: You can call `endswith` one time passing a tuple of suffixes, if you want. https://docs.python.org/3/library/stdtypes.html#str.endswith

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -43,10 +44,356 @@ def test_get_api__managed(self):
...
+                    "path": path,
+                    "namespace_tokenize": "ns",
+                    "namespace_inject": "ns",
+                    "namespace_strip": "ns",
+                },
+            )
+
+            expected_set = set(expected)
+            actual_set = set(task._get_files_to_package(path))
+            self.assertEquals(expected_set, actual_set)
```
davisagli: You can use `self.assertSetEqual` and avoid needing to explicitly cast to `set` yourself.

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -43,10 +44,356 @@ def test_get_api__managed(self):
...
+                zipf.close()
+
+            zipf_processed = task._process_zip_file(zipfile.ZipFile(zip_bytes))
+            fp = zipf_processed.fp
+            zipf_processed.close()
+            expected = bytes_to_native_str(base64.b64encode(fp.getvalue()))
+
+            actual = task._get_package_zip(path)
+
+            self.assertEquals(expected, actual)
```
davisagli: I'm sure I've messed this up myself elsewhere, but `assertEquals` is actually deprecated and `assertEqual` is preferred

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -43,10 +44,356 @@ def test_get_api__managed(self):
...
+            for lwc_component_file in lwc_component_files:
+                with open(
+                    os.path.join(lwc_component_path, lwc_component_file.get("name")),
+                    "w",
+                ) as f:
+                    if lwc_component_file.get("body") is not None:
+                        f.write(lwc_component_file.get("body"))
+                    expected.append(
+                        os.path.join(
+                            "./lwc/myComponent", lwc_component_file.get("name")
```
davisagli: Separate out the path components and pass them to os.path.join so you get the right separator instead of always /

spelak-salesforce: Yes.  I saw the test failure on a windows build.  This is done.

cumulusci/tasks/salesforce/tests/test_Deploy.py
spelak-salesforce: Changed to `assertEqual`

cumulusci/tasks/salesforce/tests/test_Deploy.py
spelak-salesforce: `assertSetEqual` [requires](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertSetEqual) both inputs to be sets.

Changed to `assertSetEqual`

cumulusci/tasks/salesforce/Deploy.py
spelak-salesforce: Done.

cumulusci/tasks/salesforce/Deploy.py
spelak-salesforce: Done

cumulusci/tasks/salesforce/Deploy.py
spelak-salesforce: Done.

Wow, that's a neat concept that's new to me.  I feel so surprised that the the test method just passed as it's currently written, but I'm guess that when a generator is called outside a `for in`, the system collects all values.

cumulusci/tasks/salesforce/Deploy.py
davisagli: @spelak-salesforce you've got more of these. search for pacakge

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -61,14 +61,13 @@ def _include_file(self, root_parts, f):
         return True
 
     def _get_files_to_package(self, path):
-        files_to_package = []
+        # files_to_package = []
```
davisagli: @spelak-salesforce can you remove this instead of commenting it out?

spelak-salesforce: It's removed in a later commit.  Sorry it was snuck in.  I commented out to see if the test still passed before full removal.

cumulusci/tasks/salesforce/Deploy.py
davisagli: @spelak-salesforce it passed because `set()` iterates over the argument you pass it to construct the set.

cumulusci/tasks/salesforce/Deploy.py
davisagli: oh yeah, sorry. I should look at the full diff rather than individual commits for reviewing.

spelak-salesforce: No, it's still there.  removing...

spelak-salesforce: Done.

cumulusci/tasks/salesforce/Deploy.py
spelak-salesforce: Fixed this case.  I am proof-reading all code

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -43,10 +44,365 @@ def test_get_api__managed(self):
...
+                    "namespace_strip": "ns",
+                },
+            )
+
+            zip_bytes = io.BytesIO()
+            zipf = zipfile.ZipFile(zip_bytes, "w", zipfile.ZIP_DEFLATED)
+
+            with cd(path):
+                for file_to_pacakge in task._get_files_to_package(path):
+                    zipf.write(file_to_pacakge)
```
davisagli: @spelak-salesforce here's the other place

spelak-salesforce: Done + read through `Deploy.py` and `test_Deploy.py` and found no other errors

cumulusci/tasks/salesforce/tests/test_Deploy.py
```diff
@@ -43,10 +44,365 @@ def test_get_api__managed(self):
...
+                zipf.close()
+
+            zipf_processed = task._process_zip_file(zipfile.ZipFile(zip_bytes))
+            fp = zipf_processed.fp
+            zipf_processed.close()
+            expected = bytes_to_native_str(base64.b64encode(fp.getvalue()))
+
+            actual = task._get_package_zip(path)
+
+            self.assertEqual(expected, actual)
```
spelak-salesforce: @davisagli 
I'm getting CI test failures asserting the result of `_get_package_zip` equals the expected.  The tests pass locally.  It's hard to debug, and the test essentially creates test data then uses a copy of the production code.  Any suggestions?

davisagli: I'm trying to figure it out too. It's especially weird since it passed in the build for the 2nd-to-last commit on this branch, and the only thing that's changed since then is fixing spelling.

davisagli: I re-ran the CircleCI build and it passed this time. Which is mildly disturbing; I guess we'll see if the problem continues to show up.

PR #1003, opened 2019-02-25
Run Salesforce library UI tests in Firefox too

davisagli:
Turns out we're already using a CircleCI image that includes firefox.

# Critical Changes

# Changes

# Issues Closed


davisagli:
@boakley hmm, interesting that this passed despite not having your fix for the Populate Field keyword.

boakley:
That is indeed interesting. I was able to easily duplicate the problem.
Perhaps circleci runs on linux, and linux firefox handles the keyboard
shortcuts the same as chrome.

PR #1007, opened 2019-02-27
Feature/clear field cross browser

boakley:
I've gone through a few permutations of fixes to this keyword. While some of the changes made
marginal improvements, I think adding a .25 second sleep immediately before sending keystrokes
to the input element was the critical point. Before that change I was seeing failures at least 10% of
the time. After adding the sleep I ran tests overnight and had over 700 consecutive tests pass 
using both firefox and chrome. 

# Critical Changes

# Changes
* added stability to the populate_field keyword

# Issues Closed


cumulusci/robotframework/Salesforce.py
```diff
@@ -295,10 +297,35 @@ def populate_lookup_field(self, name, value):
         self.selenium.get_webelement(menu_locator).click()
 
     def _populate_field(self, locator, value):
```
davisagli: Should we expose this as keyword in its own right? There is `populate_field` but it only works for inputs located inside a div with class="uiInput". I would guess we'll want to be able to use this more robust approach with any text input even if it's at a different place in the DOM.

boakley: I commented privately to david, but the gist of it was that I thought we should save making `_populate_field` a public keyword for another day. It is probably a good idea, but I want to separate that discussion from the fixing of this bug so as not to delay getting this bug deployed.

PR #1016, opened 2019-03-05
Add flow to install previous release and then upgrade

davisagli:
This still needs a test but I wanted to check that it's what we want at a high level.

### DIFF
--- a/cumulusci/core/config/BaseProjectConfig.py
+++ b/cumulusci/core/config/BaseProjectConfig.py
@@ -409,6 +409,19 @@ class BaseProjectConfig(BaseTaskFlowConfig):
         if version is not None:
             return LooseVersion(version)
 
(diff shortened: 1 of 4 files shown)


cumulusci/core/config/BaseProjectConfig.py
```diff
@@ -525,7 +538,9 @@ def pretty_dependencies(self, dependencies, indent=None):
                 prefix = "{}    ".format(" " * indent)
         return pretty
 
-    def process_github_dependency(self, dependency, indent=None, include_beta=None):
+    def process_github_dependency(  # noqa: C901
```
jstvz: I mean, it really _is_ too complex, isn't it?

davisagli: 😞

PR #1022, opened 2019-03-12
WIP - new BaseLibrary, mixins for Salesforce, plus tests

boakley:
This change includes a proof of concept for the following features:

- Support for keyword mixins in order to organize all of the Salesforce keywords
- A new BaseLibrary class to serve as the base class for keyword libraries
- Bug fixes to our selenium_retry machinery to support mixins, and to support firefox
- A  robot test which tests some of the above

Note: I haven't actually moved or refactored any of the keywords in Salesforce.py. This change makes the changes necessary to support that. The main goal was to be able to use the mixins, which was difficult given the current implementation of the selenium_retry decorator for classes. 

TL;DR: Going forward, keyword libraries should inherit from `BaseLibrary` rather than use the `@selenium_retry` decorator on a class, and Salesforce.py keywords should be organized into a series of mixin classes for maintainability.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/BaseLibrary.py
+++ b/cumulusci/robotframework/BaseLibrary.py
@@ -1,0 +1,42 @@
+import logging
+from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
+from cumulusci.robotframework.utils import RetryingSeleniumLibraryMixin
+
+
+class BaseLibrary(RetryingSeleniumLibraryMixin):
+    ROBOT_LIBRARY_SCOPE = "GLOBAL"
+
+    def __init__(self, debug=False):
+        self.debug = debug
+        self.retry_selenium = True
+        # Turn off info logging of all http requests
+        logging.getLogger("requests.packages.urllib3.connectionpool").setLevel(
+            logging.WARN
+        )
+
+    @property
+    def salesforce_api_version(self):
+        try:
+            client = self.cumulusci.tooling
+            response = client._call_salesforce(
+                "GET", "https://{}/services/data".format(client.sf_instance)
+            )
+            latest_api_version = float(response.json()[-1]["version"])
+            return latest_api_version
+
+        except RobotNotRunningError:
+            # not sure if this should return None, a reasonable default value,
+            # or raise an exception. :-\
+            return None
+
+    @property
+    def builtin(self):
+        return BuiltIn()
+
+    @property
+    def cumulusci(self):
+        return self.builtin.get_library_instance("cumulusci.robotframework.CumulusCI")
+
+    @property
+    def salesforce(self):
+        return self.builtin.get_library_instance("cumulusci.robotframework.Salesforce")
--- a/cumulusci/robotframework/tests/salesforce/retry.robot
+++ b/cumulusci/robotframework/tests/salesforce/retry.robot
@@ -1,0 +1,122 @@
+*** Settings ***
+
+Resource        cumulusci/robotframework/Salesforce.robot
+Library         TestLibrary.py
+Library         Dialogs
+
+Suite setup     Open test page in browser  ${BROWSER}
+Suite teardown  Close all browsers
+
+
+*** Variables ***
+
+${BROWSER}  chrome
+
+*** Keywords ***
+
+Open test page in browser
+    [Documentation]
+    ...  This opens a browser and loads the file 'testpage.html',
+    ...  which is in the same folder as this test case
+    [Arguments]  ${BROWSER}
+
+    ${here}=      evaluate  os.path.dirname($suite_source)        modules=os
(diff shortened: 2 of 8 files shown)


cumulusci/robotframework/BaseLibrary.py
```diff
@@ -0,0 +1,42 @@
...
+    def builtin(self):
+        return BuiltIn()
+
+    @property
+    def cumulusci(self):
+        return self.builtin.get_library_instance("cumulusci.robotframework.CumulusCI")
+
+    @property
+    def salesforce(self):
+        return self.builtin.get_library_instance("cumulusci.robotframework.Salesforce")
```
davisagli: Hmm. The Salesforce library extends BaseLibrary, which means it can get itself via the `salesforce` attribute. I guess that's mostly harmless but it feels a little weird.

Are our product libraries meant to extend BaseLibrary and use the `salesforce` attribute to call salesforce keywords, or are they meant to extend the Salesforce library and inherit its keywords to use directly?

cumulusci/robotframework/BaseLibrary.py
```diff
@@ -0,0 +1,42 @@
...
+            response = client._call_salesforce(
+                "GET", "https://{}/services/data".format(client.sf_instance)
+            )
+            latest_api_version = float(response.json()[-1]["version"])
+            return latest_api_version
+
+        except RobotNotRunningError:
+            # not sure if this should return None, a reasonable default value,
+            # or raise an exception. :-\
+            return None
```
davisagli: I think it's best to return None and let the caller decide whether they can do something reasonable as a default or not.

I think this might belong better in the CumulusCI library since that's the home for keywords that interact with the org via the API rather than selenium.

cumulusci/robotframework/tests/salesforce/retry.robot
```diff
@@ -0,0 +1,122 @@
+*** Settings ***
```
davisagli: Currently our CI for CumulusCI itself only runs specific robot test files, so you would need to add this one to the list in order for CircleCI to run it. https://github.com/SFDO-Tooling/CumulusCI/blob/master/.circleci/config.yml#L98

cumulusci/robotframework/tests/salesforce/retry.robot
```diff
@@ -0,0 +1,122 @@
...
+    ${expected_retry_count}=  evaluate  $testlib.retry_count + 1
+    execute javascript        return setTimeout(raise_green, 500)
+    click element with default retry  id:green-div
+
+    # assert that the retry happened
+    ${duration}=  get duration of previous keyword
+    should be true  $duration > 2.0
+    should be equal as numbers  ${testlib.retry_count}  ${expected_retry_count}
+
+Verify @selenium_retry(True) keyword decorator
```
davisagli: Do you think we should keep the feature of being able to turn selenium retry on/off at the keyword level? It doesn't really seem worth the extra complexity to me.

cumulusci/robotframework/BaseLibrary.py
boakley: yes, it is weird that the salesforce library has a property referring to itself. I'm still pondering the wisdom of that. On the one hand, Salesforce is special in that it has core keywords, but on the other hand, we should drink our own champaign.

As to the other question, at this point they are meant to extend BaseLibrary and use the `salesforce` attribute. I'm open to the idea of inheriting from Salesforce, but I don't think that buys us much at the moment. I'm open to considering it if there's a compelling reason, but it would break a bunch of keywords.

cumulusci/robotframework/tests/salesforce/retry.robot
boakley: That's a good question. I had assumed they were there to fill a need. 

I can't think of a reason why you might want to turn it off. In that sense I don't see a reason to keep it, but I haven't thought particularly hard on the subject. I think it's safe to say nobody is using it, or somebody would have complained about the pdb statement.

cumulusci/robotframework/BaseLibrary.py
davisagli: Okay. I guess we could override the salesforce attribute on the salesforce library to make it None or something like that.

cumulusci/robotframework/tests/salesforce/retry.robot
davisagli: Yeah. We added it because we thought it would be needed, but so far that has turned out not to be the case.

Also, the retry functionality went through a number of revisions so it's also possible that it got added at a point when we were less selective about what got retried, hence more of a need to be able to opt out. I can't really remember though.

boakley:
This is old code that is no longer relevant.

PR #1029, opened 2019-03-18
Prepare 2.4.0

davisagli:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -7,7 +7,7 @@ History
 
 Critical changes:
 
-* There have been some significant changes to the task for publishing plans to MetaDeploy:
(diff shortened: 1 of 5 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,31 @@
 History
 =======
 
+2.4.0 (2019-03-18)
+------------------
+
+Critical changes:
+
+* There have been some significant changes to the task for publishing plans to MetaDeploy:
```
jlantz: Should these be covered in a doc somewhere instead of just here in release notes?

davisagli:
@prescod that's the intended distinction between the Critical Changes section and the Changes section, but you're right, I should be more specific about the circumstances under which someone would need to pay attention to each of the critical changes.

PR #1033, opened 2019-03-21
Feature/populate robusto

boakley:
The main change in this branch is to fix a bug where `populate field` would append rather than overwrite the data in a currency field. I used this as an excuse to do a fair amount of refactoring of the helper functions used by that keyword. 

This has a robot test that exercises this keyword on several different field types. To support the writing of that test I added a new Salesforce.py keyword `get field value`, and tweaked `populate field` to work on a multiline textarea in addition to an input. 

Tested on chrome, firefox, headlesschrome, and headlessfirefox, and I also ran the NPSP tests (branch feature/robot__bge-automation) to make sure I didn't break anything. 

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -160,6 +160,14 @@ class Salesforce(object):
                 return oid_match.group(2)
         raise AssertionError("Could not parse record id from url: {}".format(url))
 
+    def get_field_value(self, label):
+        """Return the current value of a form field based on the field label"""
+        input_element_id = self.selenium.get_element_attribute(
+            "xpath://label[contains(., '{}')]".format(label), "for"
+        )
+        value = self.selenium.get_value(input_element_id)
+        return value
+
     def get_locator(self, path, *args, **kwargs):
         """ Returns a rendered locator string from the Salesforce lex_locators
             dictionary.  This can be useful if you want to use an element in
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -297,35 +297,63 @@ class Salesforce(object):
         self.selenium.get_webelement(menu_locator).click()
 
     def _populate_field(self, locator, value):
-        # clearing the field in a cross-browser, cross-platform
-        # way is surprisingly difficult. .clear() doesn't work in
-        # all browsers, and sending keyboard shortcuts doesn't work
-        # on all platforms. So, we'll take a belt-and-suspenders
-        # apprach and throw the whole arsenal at the problem.
-
         field = self.selenium.get_webelement(locator)
-
-        # first, force it to have keyboard focus
-        actions = ActionChains(self.selenium.driver)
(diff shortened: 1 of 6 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -160,6 +160,14 @@ def get_current_record_id(self):
                 return oid_match.group(2)
         raise AssertionError("Could not parse record id from url: {}".format(url))
 
+    def get_field_value(self, label):
+        """Return the current value of a form field based on the field label"""
+        input_element_id = self.selenium.get_element_attribute(
+            "xpath://label[contains(., '{}')]".format(label), "for"
```
boakley: I suspect someone will ask "why isn't this xpath in locators.py"? I spoke briefly about this with Jason a few days ago. My personal position is that since this xpath is used in exactly one spot, it makes the code easier to understand if the xpath is hard-coded here rather than added to the locators data structure. Many of our locators are single-purpose, so making them sharable doesn't buy us anything but extra overhead.

davisagli: The point of putting the locators in a separate mapping is so that we can swap between mappings based on the Salesforce version if the DOM structure changes.

In this case, it seems pretty unlikely that the platform will stop using the `label` element to label fields. On the other hand, I could imagine the structure changing to add a div inside the label or something like that, which would mean needing to swap between two versions of the locator during the period when that release rolls out. On the third hand, we need to make updates to the package anyway to support that swapping and can move the locator to a mapping at that point if necessary. So I don't have a problem with this.

cumulusci/robotframework/tests/salesforce/populate.robot
```diff
@@ -0,0 +1,89 @@
...
+
+*** Keywords ***
+Create opportunity
+    [Documentation]
+    ...  Creates a test opportunity with a random name and a close
+    ...  date 30 days in the future. A reference to the created
+    ...  opportunity will be stored in the suite variable ${opportunity},
+    ...  and be return as the result of this keyword.
+
+    ${30 days from now}=  Get Current Date  increment=30 days  result_format=%Y-%m-%d
```
davisagli: Looks like you aren't actually using this.

boakley: Hah! Good catch. That's a bug. I added the variable but forgot to use it in the call to `salesforce insert`.

boakley: I've pushed a fix.

davisagli:
@boakley populate.robot is not actually being run by the `test_robot` step in CircleCI. We either need to add it as a separate command in .circleci/config.yml, or refactor how that invokes robot tests to run an entire folder instead of individual test files.

boakley:
Hah. I added a test and coverage went down. :-\

boakley:
The fact that those other lines aren't covered surprises me since they run when I run the tests locally on my box.

PR #1037, opened 2019-03-26
Update scratch org definition files to use orgPreferenceSettings

davidmreed:
# Critical Changes

# Changes

 - Reformat scratch org definition files to use [`orgPreferenceSettings` instead of `orgPreferences`](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file_config_values.htm), which is now deprecated.
 - Parameterize generation of scratch org definition files based on managed or unmanaged environments.

# Issues Closed

### DIFF
--- a/cumulusci/files/templates/project/scratch_def.json
+++ b/cumulusci/files/templates/project/scratch_def.json
@@ -1,10 +1,11 @@
 {
   "orgName": "{{ package_name }} - {{ org_name }}",
   "edition": "{{ edition }}",
-  "orgPreferences" : {
-    "enabled": [
-      "S1DesktopEnabled",
-      "ChatterEnabled"
-    ]
+  "settings": {
+    "orgPreferenceSettings": {
+      "s1DesktopEnabled": true,
+      "chatterEnabled": true,
+      "translation": true
+    }
   }
 }
--- a/cumulusci/files/templates/project/scratch_def.json
+++ b/cumulusci/files/templates/project/scratch_def.json
@@ -3,9 +3,11 @@
   "edition": "{{ edition }}",
   "settings": {
     "orgPreferenceSettings": {
-      "s1DesktopEnabled": true,
-      "chatterEnabled": true,
-      "translation": true
+      "s1DesktopEnabled": true
+      , "chatterEnabled": true
+      {% if not managed %}
+      , "translation": true 
+      {% endif %}
     }
   }
 }
--- a/orgs/release.json
+++ b/orgs/release.json
@@ -3,9 +3,8 @@
   "edition": "Enterprise",
   "settings": {
     "orgPreferenceSettings": {
-      "s1DesktopEnabled": true,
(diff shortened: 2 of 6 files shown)


orgs/release.json
```diff
@@ -1,11 +1,11 @@
...
-    "enabled": [
-      "S1DesktopEnabled",
-      "Translation",
-      "ChatterEnabled" 
-    ]
+  "settings": {
+    "orgPreferenceSettings": {
+      "s1DesktopEnabled": true,
+      "chatterEnabled": true,
+      "translation": true
```
davisagli: While we're working here, let's disable translation for the beta and release scratch orgs. It refers to translation workbench which is only needed for development. We had one bug where code was written that depended on translation workbench being installed (by referring to a permission that only exists in that case) and we didn't catch it until installing in an org with translation workbench turned off.

cumulusci/files/templates/project/scratch_def.json
```diff
@@ -1,10 +1,11 @@
...
-  "orgPreferences" : {
-    "enabled": [
-      "S1DesktopEnabled",
-      "ChatterEnabled"
-    ]
+  "settings": {
+    "orgPreferenceSettings": {
+      "s1DesktopEnabled": true,
+      "chatterEnabled": true,
+      "translation": true
```
davisagli: See my other comment -- this means that we want translation on for some scratch orgs but not others. Which means we either need two different template files, or a jinja2 conditional within the template based on a variable that we pass when rendering it.

jlantz: I'd say the variable should be to signify an unmanaged vs a managed environment.  The dev and feature org shapes are unmanaged, beta and release are managed.

davidmreed:
@davisagli / @jlantz If the updates look good to you I will re-create this pull request on a branch within the main repo, rather than my fork.

cumulusci/files/templates/project/scratch_def.json
```diff
@@ -1,10 +1,13 @@
...
-    "enabled": [
-      "S1DesktopEnabled",
-      "ChatterEnabled"
-    ]
+  "settings": {
+    "orgPreferenceSettings": {
+      "s1DesktopEnabled": true
+      , "chatterEnabled": true
+      {% if not managed %}
+      , "translation": true
```
davisagli: I'm not a fan of the commas at the start of the line, but I see how it helps with the conditional and I'm willing to put up with it.

davisagli:
@davidmreed let's get this opened as a PR from our github org so we can see what the builds say.

cumulusci/files/templates/project/scratch_def.json
davidmreed: I don't love it either but it seemed less messy than the alternative conditional style.

PR #1039, opened 2019-03-28
Update default Metadata API version

davidmreed:
# Critical Changes

# Changes

 - Update the default Metadata API version in `cumulusci.yml` to 45.0.
 - Revise unit tests to break dependency on default Metadata API version.
 - <s>Remove hard-coded API 43.0 in `InstallPackageVersion` / `InstallPackageZipBuilder`, and add `api_version` option for the task (defaulting to project-level configuration).</s>
 - Ensure that `ApiRetrieveInstalledPackages` respects any configured API version.
 - Fix an inconsistency in `RetrieveReportsAndDashboards` so that the correct API version is used throughout.
 - Correct various flake8 warnings.

# Issues Closed

### DIFF
--- a/cumulusci/salesforce_api/metadata.py
+++ b/cumulusci/salesforce_api/metadata.py
@@ -263,7 +263,7 @@ class ApiRetrieveUnpackaged(BaseMetadataApiCall):
         self._clean_package_xml()
 
     def _clean_package_xml(self):
-        self.package_xml = re.sub("<\?xml.*\?>", "", self.package_xml)
+        self.package_xml = re.sub("<\\?xml.*\\?>", "", self.package_xml)
         self.package_xml = re.sub("<Package.*>", "", self.package_xml, 1)
         self.package_xml = re.sub("</Package>", "", self.package_xml, 1)
         self.package_xml = re.sub("\n", "", self.package_xml)
@@ -364,7 +364,7 @@ class ApiDeploy(BaseMetadataApiCall):
         self.package_zip = package_zip
 
     def _set_purge_on_delete(self, purge_on_delete):
-        if purge_on_delete == False or purge_on_delete == "false":
+        if purge_on_delete is False or purge_on_delete == "false":
             self.purge_on_delete = "false"
         else:
             self.purge_on_delete = "true"
--- a/cumulusci/salesforce_api/metadata.py
+++ b/cumulusci/salesforce_api/metadata.py
@@ -294,8 +294,8 @@ class ApiRetrieveInstalledPackages(BaseMetadataApiCall):
     soap_action_status = "checkStatus"
     soap_action_result = "checkRetrieveStatus"
 
-    def __init__(self, task):
-        super(ApiRetrieveInstalledPackages, self).__init__(task)
+    def __init__(self, task, api_version=None):
+        super(ApiRetrieveInstalledPackages, self).__init__(task, api_version)
         self.packages = {}
 
     def _process_response(self, response):
--- a/cumulusci/salesforce_api/metadata.py
+++ b/cumulusci/salesforce_api/metadata.py
@@ -263,7 +263,7 @@ class ApiRetrieveUnpackaged(BaseMetadataApiCall):
         self._clean_package_xml()
 
     def _clean_package_xml(self):
-        self.package_xml = re.sub("<\\?xml.*\\?>", "", self.package_xml)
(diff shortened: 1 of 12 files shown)


davidmreed:
@jlantz You raised a concern yesterday about the third point above and whether any projects might depend upon the behavior of that specific task being pinned to a specific API version *other than* the project's own. Included here for discussion.

cumulusci/salesforce_api/metadata.py
```diff
@@ -364,7 +364,7 @@ def __init__(self, task, package_zip, purge_on_delete=None, api_version=None):
         self.package_zip = package_zip
 
     def _set_purge_on_delete(self, purge_on_delete):
-        if purge_on_delete == False or purge_on_delete == "false":
+        if purge_on_delete is False or purge_on_delete == "false":
```
davisagli: There's a subtle change in behavior here:

```
>>> 0 == False
True
>>> 0 is False    
False
```

I doubt it matters here, but `if not purge_on_delete or purge_on_delete == "false"` might be preferable.

cumulusci/salesforce_api/tests/test_metadata.py
```diff
@@ -224,7 +224,7 @@ def test_call_faultcode(self):
         response = '<?xml version="1.0" encoding="UTF-8"?><faultcode>foo</faultcode>'
         self._mock_call_mdapi(api, response)
         with self.assertRaises(MetadataApiError):
-            resp = api()
+            resp = api()  # noqa: F841
```
davisagli: I'd say just don't assign the variable if we're not using it (here and below).

```suggestion
            api()
```

cumulusci/salesforce_api/tests/test_metadata.py
```diff
@@ -405,34 +404,10 @@ def test_get_response_status_error_500(self):
...
         response = '<?xml version="1.0" encoding="UTF-8"?><id>1234567890</id>'
         self._mock_call_mdapi(api, response)
-        response_status = '<?xml version="1.0" encoding="UTF-8"?><foo>status</foo>'
         self._mock_call_mdapi(api, response, status_code)
 
         with self.assertRaises(MetadataApiError):
-            resp = api._get_response()
-
-    @responses.activate
-    def test_get_response_status_error_500(self):
```
davisagli: So this was a duplicate method name? We should keep the 2nd one, which would be the one that was actually getting executed before.

cumulusci/salesforce_api/package_zip.py
```diff
@@ -87,17 +87,20 @@ def _populate_zip(self):
 
 
 class InstallPackageZipBuilder(BasePackageZipBuilder):
-    api_version = "43.0"
-
-    def __init__(self, namespace, version, activateRSS=False, password=None):
+    def __init__(
+        self, namespace, version, api_version, activateRSS=False, password=None
+    ):
```
davisagli: I did a search in both our github orgs to see if there are any other repos using this class, and didn't find any, so I think this change in the arguments should be okay as long as we've updated all the call sites in CumulusCI itself.

davidmreed:
@davisagli , regarding `InstallManagedPackage`: do you think there's a case for pinning a minimum API version of 43.0 and using a *newer* API version specced by the project (if any), or would it be best to just maintain the pin?

davisagli:
@davidmreed I was wondering about that too.

If we allow a newer pin, there's a chance CCI will break until we can get a fix out if a future API version adds a new required element (that's what happened when `activateRSS` was added).

If we hardcode it as 43.0, it _should_ keep working in that scenario until we have a chance to update it to the newer API version (I say "should" because there have been core platform bugs where the API version gating for a new metadata attribute was wrong. That happened with `activateRSS` too for a bit.)

cumulusci/salesforce_api/tests/test_metadata.py
davidmreed: The two methods were byte-for-byte identical - looks like an errant copy and paste.

davidmreed:
@davisagli That makes sense. Given that the project doesn't have any other opportunity to influence the content of the generated `package.xml` from that class, it sounds like the best course is to leave the class as is. I'll revert that specific change.

davisagli:
@davidmreed the test_robot build unfortunately fails fairly often with timeouts. You can re-run it by clicking "Details" to go to CircleCI, then the "build_test_report" link under Workflow to view the entire workflow, then the dropdown next to Rerun and choose "Rerun from failed". Let me know if you run into access problems.

PR #1045, opened 2019-04-09
Support installation via pipx

jstvz:
This pull request adds simplified installation instructions for Windows 10
using `pipx`. This utility automates the creation of a virtualenv, the
installation of `cumulusci`'s dependencies, and linking the `cci.exe`
executable in a known location on the user's PATH.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -2,99 +2,100 @@
 Tutorial
 ========
 
-This tutorial is for macOS. Linux and Windows are not yet officially supported but should work for the most part. We have added some Windows info where possible.
+This tutorial is for macOS. Linux and Windows are not yet officially supported but should work for the most part. We have added some Windows and Linux info where possible.
 
 Part 1: Installing CumulusCI
 ============================
 
-On macOS and Linux, the easiest way to install CumulusCI is using `homebrew <https://docs.brew.sh/>`_  and `linuxbrew <https://linuxbrew.sh/>`_, respectively:
-
-.. code-block:: console
-
-    $ brew tap SFDO-Tooling/homebrew-sfdo
-    $ brew install cumulusci
-
-On Windows 10, the easiest way to install CumulusCI is using ```pipx`` <https://github.com/pipxproject/pipx>`__: 
-
-1. In a new command prompt, run: ``python -m pip install --user pipx``
-2. Add the following paths to your ``PATH`` environment variable:
-
-   1. ``%USERPROFILE%\AppData\Roaming\Python\Python37\Scripts``
-   2. ``%USERPROFILE%\.local\bin``
-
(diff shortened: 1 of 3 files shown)


jstvz:
Supersedes #995

docs/tutorial.rst
```diff
@@ -14,6 +14,17 @@ On macOS and Linux, the easiest way to install CumulusCI is using `homebrew <htt
     $ brew tap SFDO-Tooling/homebrew-sfdo
     $ brew install cumulusci
 
+On Windows 10, the easiest way to install CumulusCI is using ```pipx`` <https://github.com/pipxproject/pipx>`__: 
+
+1. In a new command prompt, run: ``python -m pip install --user pipx``
```
davisagli: This is assuming that the user already has the correct version of Python installed, but we don't tell them how to do that until further down.

docs/tutorial.rst
```diff
@@ -14,6 +14,17 @@ On macOS and Linux, the easiest way to install CumulusCI is using `homebrew <htt
...
+2. Add the following paths to your ``PATH`` environment variable:
```
davisagli: We probably need to give instructions on how to do that.

docs/tutorial.rst
```diff
@@ -14,6 +14,17 @@ On macOS and Linux, the easiest way to install CumulusCI is using `homebrew <htt
...
+
+   1. ``%USERPROFILE%\AppData\Roaming\Python\Python37\Scripts``
+   2. ``%USERPROFILE%\.local\bin``
+
+3. In a new command prompt, run: ``pipx install cumulusci``
+4. In a new command prompt, verify by running: ``cci version``
```
davisagli: We should say what they should expect to see.

docs/tutorial.rst
```diff
@@ -33,56 +24,70 @@ Install Python 3: https://www.python.org/downloads/windows/
...
-    $ python3 -m venv cci
+Install CumulusCI
+-----------------
 
-This creates a virtualenv in a new folder named ``cci``. Whenever you want to work with commands installed in this virtualenv, you must first activate it:
+macOS/Linux
+^^^^^^^^^^^
 
-.. code-block:: console
+On macOS and Linux, the easiest way to install CumulusCI is using `homebrew <https://docs.brew.sh/>`_  and `linuxbrew <https://linuxbrew.sh/>`_, respectively:
```
davisagli: This will use a homebrew-installed Python, right? Should we remove the "Install Python 3" link for macOS above?

jstvz: Ah, good catch. I'll remove that macOS/Linux from that section, then move it under Windows.

Also, noticed that they merged Linuxbrew back in 2.0.0, so I'll remove any of those references.

PR #1048, opened 2019-04-18
Feature/robot page objects

boakley:
This has an initial implementation of page objects. A high level description can be found in the quip doc [Page Object Cheat Sheet](https://salesforce.quip.com/MRyoArtZhtdb).  

Note: at present this repo does not have any documentation changes. I'm working on that now.

The bulk of this change is a new keyword library named PageObjects. It provides the following new keywords:

- Go To Page
- Load Page Object
- Current Page Should Be
- Log Page Object Keywords

It also includes a new decorator, `pageobject` which can be used to decorate a class and make it available as a page object.

This change also has some generic page object classes which can be used as base classes for standard salesforce pages. Right now there are only three, but we'll be adding more ASAP. The goal for this PR is to get the basic plumbing in. 

- ListingPage
- HomePage
- DetailPage

This branch also has a small tweak to .circleci/config.yml to consolidate robot reports even when the tests fail.

# Critical Changes

# Changes

# Issues Closed


cumulusci/robotframework/pageobjects/PageObject.py
```diff
@@ -0,0 +1,125 @@
...
+
+    # These must be defined by subclasses
+    PAGE_TYPE = None
+    OBJECT_API_NAME = None
+
+    """Base class to use for all page objects
+
+    Derived classes should define PAGE_TYPE based on the appropriate
+    pageReference type but with the "standard_" prefix removed (eg:
+    objectPage rather than standard_objectPage)
```
davisagli: What's a pageReference type?

cumulusci/robotframework/pageobjects/PageObject.py
```diff
@@ -0,0 +1,125 @@
...
+    _is_current_page
+
+      called by the keyword 'current page should be'. By default
+      it will get the current page location and compare it to a
+      URL based on the PAGE_TYPE and OBJECT_API_NAME attributes.
+
+    When this library is imported it will be added to PYTHONPATH, which is
+    what python uses to resolve name conflicts. If your page object defines
+    a keyword with the same name as some other page object, you can control
+    which one has precedence based on the order that you import the libraries.
```
davisagli: Wait, I thought we were controlling the precedence using `Set Library Search Order`; what's the Python path adjustment needed for? Monkeying with the Python path sounds scary, and also this seems error-prone if some other test suite already imported some of the page objects in a different order.

cumulusci/robotframework/pageobjects/PageObject.py
```diff
@@ -0,0 +1,125 @@
...
+    pageReference type but with the "standard_" prefix removed (eg:
+    objectPage rather than standard_objectPage)
+
+    This class provides the following properties for accessing
+    keywords in other libraries:
+
+    builtin    - robot's built-in keywords (eg: self.builtin.log("..."))
+    selenium   - SeleniumLibrary keywords
+    cumulusci  - cumulusci.robotframework.Salesforce library
+    salesforce - cumulusci.robotframework.CumulusCI library
```
davisagli: These are reversed.

cumulusci/robotframework/__init__.py
```diff
@@ -0,0 +1 @@
+from .pageobjects import *  # noqa: F401,F403
```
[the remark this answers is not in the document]
boakley: Yeah, you're right. That was just lazy coding.

cumulusci/robotframework/pageobject_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+        po._go_to_tab()
+
+    def set_current_page_object(self, name):
+        """This moves the given page object library to the start of the search order
+
+        """
+        po = self._get_page_object(name)
+        self.builtin.log("setting page object to '{}'".format(str(po)), "DEBUG")
+
+        self.builtin.set_library_search_order(po.__class__.__name__)
```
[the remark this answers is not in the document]
boakley: I've been going back and forth on that. I had code to do that and then I took it out right before pushing because it wasn't doing anything useful for the current demo. 

I'm glad you brought it up. I haven't made a final decision, but it's good to know that the idea occurred to you too.

cumulusci/robotframework/pageobject_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+        For the most part this just removes spaces: "Contact Home Page"
+        becomes "ContactHomePage", etc.
+
+        If you omit "Page", we'll attempt to add some missing
+        parts. For example, we'll always append "Page".  Also, if it
+        doesn't end in "Home Page", "List Page" or "New Page", we'll
+        append "Home Page" instead of just "Page".
+
+        For example, both "Contact" and "Contact Home" will be converted
+        to "ContactHomePage".
```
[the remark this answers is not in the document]
boakley: Good comment. Yes, I want to be able to use it for components as well as whole pages (at least, I think I do). I need to put a bit more thought into this magic name transmogrification. It either needs to be smarter or we take it out and force you to be explicit. 

Hmmm. just now, saying/typing that out loud reminds me of the zen of python. Maybe doing the automatic transmogrification is a dead end. If you want to go to the home page, you should be explicit.

cumulusci/robotframework/pageobject_mixin.py
```diff
@@ -0,0 +1,117 @@
...
+            # enough to cache libraries, so we don't have to worry too much
+            # about performance.
+            self.builtin.import_library(normalized_name)
+            library = self.builtin.get_library_instance(normalized_name)
+            return library
+
+        except RuntimeError:
+            # we'll get this error if we couldn't find a suitable class. In
+            # such a case we'll instantiate a generic class and set its attributes
+            # based on the requested page name
```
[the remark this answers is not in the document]
boakley: I'll admit, this is a part of the design I'm only 90% happy with. Besides holding keywords, each instance also has attributes which define the type of the page (based on the lightning component type) and the object name (Contact, Organization, etc). 

If you look a few lines down, we add those attributes after creating the instance:

```
            library.PAGE_TYPE = "objectPage"
            library.OBJECT_API_NAME = objectname
```

The base class methods use these attributes to construct the URL to go to that page.

cumulusci/robotframework/pageobjects/PageObject.py
boakley: That's me pretending I know what I'm talking about :-)

The idea is that we have a base class for each lightning component type - see [Lightning Aura Components Developer Guide](https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/components_navigation_page_definitions.htm). 

To this point I've focused on the "Object Page Type" (hence the file ObjectPages.py elsewhere in this branch). The idea is to have a base definition for object lists, record views, etc. so that when someone is automating a page, they start by inheriting one of our base classes that most closely resembles the page they are automating.

This is definitely something I need more advice on -- is this a good idea, or am I headed in the wrong direction? I think I stumbled on this page when trying to understand the difference between urls like /lightning/r/whatever and /lightning/o/whatever. When I saw a neat little set of documented page types, it seemed like a good way to organize base classes.

While you're free to call it a genius idea :-), I'm perfect open to being told it's over engineering and we really don't need a ton of base classes based on the page type.

cumulusci/robotframework/pageobjects/PageObject.py
boakley: Your observations are correct.

Part of the reason it took me so long to push code was that I was trying to get to a place where a test should be able to import a page with `Library  SurfingSafariHomePage` in their settings, without having to tell robot explicitly where to look.  This code is a remnant of that, and is no longer necessary. I should have removed it before pushing the code.

For the moment at least, you have to fully qualify the page objects when you import them (see for example the file `SurfingSafari.resource`). 

One solution to avoid having to fully qualify the library when importing is to add some something to cumulusci.yml that tells robot where to look for libraries. For this first cut I didn't want to do that.

I'm curious what you think about the goal of being able to import using just the base name (eg: `SurfingSafariHomePage`) versus a fully qualified path (eg: `robot/SurfingSafari/resources/SurfingSafariHomePage.py`).

cumulusci/robotframework/pageobjects/PageObject.py
boakley: hah! good catch!

cumulusci/robotframework/pageobject_mixin.py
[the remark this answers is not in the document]
davisagli: I don't have a strong feeling about it, but it we don't, we should probably at least document that.

cumulusci/robotframework/pageobject_mixin.py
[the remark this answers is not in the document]
davisagli: Okay right, that makes sense.

cumulusci/robotframework/pageobjects/PageObject.py
davisagli: I'm not sure how it will play out but am willing to go with it for now.

cumulusci/robotframework/pageobjects/PageObject.py
davisagli: My gut feeling is that importing the full path is probably better. That was someone reading a test suite can just go to that file instead of needing to understand what the resolution rules are.

cumulusci/core/tests/BarTestPage.py
```diff
@@ -0,0 +1,18 @@
+"""
+This class is used by test_pageobjects
+"""
+from cumulusci.robotframework.pageobjects import BasePage
+from cumulusci.robotframework.pageobjects import pageobject
+
+
+@pageobject("Test", "Bar")
+class BarTestPage(BasePage):
+    object_name = "Bar"
```
davisagli: From the implementation of the decorator, it looks like it's not required to specify `object_name` here since it will be added from the decorator argument. I like that it's readable but not the duplication. How about recommending using the decorator like this:

    @pageobject(page_type="Test", object_name="Bar")

and not explicitly adding the attribute.

cumulusci/robotframework/pageobjects/__init__.py
```diff
@@ -0,0 +1,17 @@
+from robot.libraries.BuiltIn import BuiltIn
+from .PageObjects import PageObjects  # noqa: F401
+from .baseobjects import BasePage, ListingPage, HomePage, DetailPage  # noqa: F401
+
+
+def pageobject(pageType, object_name=None):
```
davisagli: Can this be `page_type` for consistency with the keywords and with CumulusCI's overall code style?

cumulusci/robotframework/pageobjects/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    """A decorator to designate a class as a page object"""
+    BuiltIn().log("importing page object {} {}".format(pageType, object_name), "DEBUG")
+
+    def wrapper(cls):
+        key = (pageType, object_name)
+        PageObjects.registry[key] = cls
+        if not hasattr(cls, "object_name"):
+            cls.object_name = object_name
```
davisagli: Should probably raise an exception if it has an object_name and it's different.

cumulusci/robotframework/tests/salesforce/ui.robot
```diff
@@ -81,15 +81,19 @@ Go To Setup Object Manager
...
 Go To Object List
+    [Tags]  smoke
     Go To Object List  Contact
 
 Go To Object List With Filter
+    [Tags]  smoke
     Go To Object List  Contact  filter=Recent
 
 Go To Record Home
+    [Tags]  smoke
```
davisagli: 🔥 

(I like this)

cumulusci/robotframework/tests/pageobjects/pageobjects.robot
```diff
@@ -0,0 +1,100 @@
+*** Settings ***
```
davisagli: Can you update the CircleCI configuration to actually run this test suite? That should help with the coverage.

cumulusci/robotframework/pageobjects/__init__.py
boakley: yes, good catch. Thanks.

cumulusci/core/tests/BarTestPage.py
boakley: I think that's probably a good suggestion. I'm on the fence about explicitly using the argument names, but you're definitely right about not setting `object_name`. I left the ability to explicitly set it mainly because I wanted to create some test classes where the name used to reference the page and the actual class were different. Not sure if that's ever going to be a valid use case.

cumulusci/robotframework/pageobjects/__init__.py
boakley: Excellent observation, thanks for making the comment.

I think for now I'm going to leave it as is. My main concern is that this is in a decorator, and decorators usually shouldn't raise exceptions. I think that's debatable, though, and I could be swayed either way with nothing more than a good pint of beer :-)

For now I'm going to leave it as is since users probably won't stumble on it, but I'll make a note to mention in the documentation what happens when you do this. I don't want this issue to hold this code up.

davisagli: I think a decorator should raise an exception if you're trying to use it for something that doesn't make sense. This feels like a check we can easily add to prevent a hard-to-debug mistake. I'm less worried about it if we stop recommending to set object_name both as a class attribute and as an argument to the decorator, though.

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
 * **Salesforce Update**: Updates a record using its type, ID, and field=value syntax
 * **SOQL Query**: Runs a SOQL query and returns a REST API result dictionary
 
+PageObjects Library
+===================
+
+The **PageObjects** library provides support for page objects,
+robotframework-style. Even though robot is a keyword driven framework,
```
```suggestion
Robot Framework-style. Even though robot is a keyword-driven framework,
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+we've implemented a way to dynamically load in keywords that are
+unique to a page or an object on the page.
+
+With this library, you can define classes which represent page
+objects. Each class has keywords that are unique to a page or a
+component. These classes can be imported on demand only for tests
+which use these pages or components.
+
+The pageobject Decorator
```
```suggestion
The ``pageobject`` Decorator
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+------------------------
+
+Page objects are normal python classes which use the :code:`pageobject`
```
```suggestion
Page objects are normal Python classes which use the :code:`pageobject`
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+decorator provided by cumulusci. Unlike traditional robot framework
```
```suggestion
decorator provided by CumulusCI. Unlike traditional Robot Framework
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+
+       def click_contact_link(self, name):
+           self.selenium.click_link('xpath://a[@title="{}"]'.format(name))
+           self.salesforce.wait_until_loading_is_complete()
+
+The :code:`pageobject` decorator takes two arguments: :code:`page_type` and
+:code:`object_name`. These two arguments are used to identify the page
+object (eg: :code:`Go To Page  Listing  Contact`). The values can be
+any arbitrary string, but ordinarily should represent standard page
+types ("Listing", "Detail", "Home"), and standard object names.
```
davisagli: The developer name or the user-facing label for the object? If the former, we need to talk about how to make this work both when objects have a namespace prefix and when they don't.

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+
+
+Importing the library
+---------------------
+
+The **PageObjects** library is somewhat unique in that it is not only a
+keyword library, but also the mechanism by which you can import files
+which contain page object classes. This is done by providing the paths
+to one or more python files which implement page objects. You may also
```
```suggestion
to one or more Python files which implement page objects. You may also
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+import **PageObjects** without passing any files to it in order to take
+advantage of some general purpose page objects.
+
+For example, consider the case where you've created two files that
+each have one or more page object definitions. For example, lets say
+in robot/MyProject/resources you have the files PageObjects.py and
```
```suggestion
in ``robot/MyProject/resources`` you have the files ``PageObjects.py`` and
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+MorePageObjects.py. You can import these page objects into a test
```
```suggestion
``MorePageObjects.py``. You can import these page objects into a test
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+each of the keywords for the current page object.
+
+Load Page Object
+^^^^^^^^^^^^^^^^
+
+Example: :code:`Load page object  Listing  Contact`
+
+This will load the page object for the given **page_type** and
+**object_name_**. It is useful when you want to use the keywords from a
+page object without first navigating to that page (ie: when you are
```
```suggestion
page object without first navigating to that page (i.e. when you are
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+--------------------
+
+You do not need to create a page object in order to take advantage of
+the new page object keywords. If you use one of the page object
+keywords for a page that does not have its own page object, the
+**PageObjects** library will try to find a generic page.
+
+For example, if you use :code:`Current page should be  Home  Event` and
+there is no page object by that name, a generic :code:`Home` page object
+will be loaded, and it's object name will be set to :code:`Event`.
```
```suggestion
will be loaded, and its object name will be set to :code:`Event`.
```

docs/robotframework.rst
```diff
@@ -200,6 +200,180 @@ In addition to browser interactions, the Salesforce Library also provides the fo
...
+
+For example, lets say your project has created a custom object named
```
```suggestion
For example, let's say your project has created a custom object named
```

docs/robotframework.rst
boakley: Yes, I haven't given namespaces any consideration. I haven't yet taken the time to grok how they work.

docs/robotframework.rst
boakley: ugh. I hate it when I make this mistake.

docs/robotframework.rst
davisagli: It's not super complicated (though dealing with the implications can be), and I think it's worth grokking before we introduce a framework that assumes objects are always named the same thing.

Standard objects like Contact always have the same name.

Custom objects which are not in a managed package or a namespaced scratch org have a name with a `__c` suffix like MyObject__c. So far no problem.

Custom objects which are in a managed package add a prefix using the package's globally unique namespace. For example npsp__Address__c is the developer name of the Address object from NPSP when it is installed as a managed package. Objects also get this namespace prefix if they are in an unmanaged package being developed in a scratch org that has its own namespace.

Since our products are typically developed as unmanaged packages and then released as managed packages, that means that the Address object would be called Address__c during development but npsp__Address__c once it is released (or in development in a namespaced scratch org).

Since we want to run our robot tests both during development and with managed packages for regression, we need to figure out how to handle this.

boakley: We probably need to discuss this over chat or a video call, but let me get a couple of basic questions out of the way.

At the time that a developer needs to write a keyword, do they know what the namespace is / will be? Could the solution be as simple as defining the namespace in the class or decorator, and then at runtime we do something like this:

    @pageobject(page_type="Listing", object_name="MyObject__c", namespace="npsp")
    def whatever(ListingPage):
        ...
        object_name = self.namespace + self.object_name if self.is_managed() else self.object_name
        url = ".../lightning/o/{}/list".format(object_name)"
        ...

Or is it a bit more complicated than that? Is it possible to write a `self.is_managed` function or property that can use introspection to get the answer (and/or a `self.get_namespace_prefix()` function)?

boakley: poking about through the NPSP code I see that it does `self.cumulusci.sf.describe()` and then does a bit of string manipulation to get the namespace prefix. Would that be the foundation of `self.is_managed()` or `self.get_namespace_prefix()`?

davisagli: That sounds like the right direction, but we might not even need to add the `namespace` argument because it's typically only objects from the current repository that would vary between namespaced and un-namespaced, so we can look up the current project's namespace from cumulusci.yml. (Underlying packages would be installed as managed packages and always have their namespace.)

There's already a keyword in the CumulusCI library, `Get Namespace Prefix` to get the namespace prefix for a specified package (defaulting to the package for the current repository) or an empty string if there's no prefix.

`object_name` could become a property that encapsulates the logic for adding or not-adding the prefix (and memoizes it).

boakley: Will namespaces always only apply to custom objects? That is, if I set `object_name` to `Contact`, I can determine it will never have a namespace associated with it, but if the object is named `MyObject__c`, then I need to do the dance to maybe prepend the namespace?

I (think I) want to support the ability for a teams to create page objects for standard objects, until we reach the point that we have full coverage. For example, NPSP or GEM or whomever might create a `Contact` page object or `Opportunity` page object to get them over a hump, which we can eventually fold into the core. In such a case, we wouldn't want to use the namespace logic for those page objects.

So, the logic for building a URL might look like:

```
    if object_name.endswith("__c") and is_managed():
        prepend the namespace
    else
        don't prepend the namespace
```

davisagli: That's right that it's only for custom objects. It's also only for custom objects that don't already have a different namespace. I think the logic could be something like this:

    @property
    def object_name(self):
        object_name = self._object_name # assuming we change the decorator to set _object_name instead of object_name
        parts = object_name.split("__")
        if len(parts) == 2 and parts[-1] == "c":  # the length check is to skip objects from a different namespace like foobar__otherpackageobject__c
            object_name = self.cumulusci.get_namespace_prefix() + object_name # get_namespace_prefix already takes care of returning an actual prefix or an empty string depending on whether the package is managed
        return object_name

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -150,6 +150,7 @@ def load_page_object(self, page_type, object_name=None):
         """
         pobj = self._get_page_object(page_type, object_name)
         self._set_current_page_object(pobj)
+        return pobj
```
boakley: fwiw, I added this statement to make this method easier to test.

davisagli:
@boakley is this ready for final review?

boakley:
Yes, I think so. I'm working on a tutorial which is going to be part of the
sphinx docs, but I was talking with Jason and he felt it was OK to make
that part of a separate PR.

PR #1051, opened 2019-04-23
Modified 'cci project init' to generate new robot folder structure

boakley:
This modifies `cci project init` to generate the new robot folder structure. It also adds a 
definition for the robot task in the generated cumulusci.yml file to point to the new location
of tests and results.

This includes an update to an existing python tests, and adds one new one to validate the contents 
of the generated cumulusci.yml file. I only added a test for the robot section, but will add some 
additional tests once I get some feedback on whether or not there's a better way to do the test -- this
is the first time I've written a test against the cci config file.  

# Critical Changes

# Changes

# Issues Closed


cumulusci/cli/tests/test_cci.py
```diff
@@ -246,15 +246,55 @@ def test_project_init(self, click):
...
+                "testns",  # package_namespace
+                "43.0",  # api_version
+                "3",  # extend other URL
+                "https://github.com/SalesforceFoundation/Cumulus",  # github_url
+                "default",  # git_default_branch
+                "work/",  # git_prefix_feature
+                "uat/",  # git_prefix_beta
+                "rel/",  # git_prefix_release
+                "%_TEST%",  # test_name_match
+            )
```
davisagli: I should probably refactor it to separate the code that collects user input from the code that actually generates the files. But this is okay for now.

PR #1052, opened 2019-04-24
Allow configuration of scratch org user email address

davidmreed:
# Critical Changes

# Changes

- Allow configuration of the email address assigned to scratch org users, with the order of priority being (1) any `adminEmail` key in the scratch org definition; (2) the `email_address` property on the scratch org configuration in `cumulusci.yml`; (3) the `user.email` configuration property in Git.

# Issues Closed

### DIFF
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -137,6 +137,27 @@ class ScratchOrgConfig(OrgConfig):
         return password
 
     @property
+    def email_address(self):
+        email_address = self.config.get("email_address")
+        if not email_address:
+            # Obtain a default email address from Git.
+            p = sarge.Command(
+                "git config --get user.email",
+                stderr=sarge.Capture(buffer_size=-1),
+                stdout=sarge.Capture(buffer_size=-1),
+                shell=True,
+            )
+            p.run()
+            email_address = io.TextIOWrapper(p.stdout).read().strip()
+
+            if p.returncode or not email_address:
+                email_address = None
+            else:
+                self.config["email_address"] = email_address
+
+        return email_address
+
+    @property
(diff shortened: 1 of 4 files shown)


cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -136,6 +136,27 @@ def password(self):
             password = self.scratch_info["password"]
         return password
 
+    @property
+    def email_address(self):
+        email_address = self.config.get("email_address")
+        if not email_address:
+            # Obtain a default email address from Git.
```
davisagli: How about we factor this out into a utility function in cumulusci.utils that takes the name of the desired git config key as an argument. I suspect this isn't the only place we'll want to do something like that.

cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -136,6 +136,27 @@ def password(self):
             password = self.scratch_info["password"]
         return password
 
+    @property
+    def email_address(self):
+        email_address = self.config.get("email_address")
```
davisagli: No strong opinion but maybe we should call it `admin_email` to better match the name in scratch org definition files.

cumulusci/core/config/ScratchOrgConfig.py
davidmreed: 👍 Will do.

cumulusci/core/config/ScratchOrgConfig.py
davidmreed: I find the SFDX-introduced nomenclature rather counterintuitive. On the theory that most of the time the value will come from Git, and secondly from `cumulusci.yml`, I'd rather call it what the user would understand it to be ("my user's email address").

PR #1056, opened 2019-04-26
Build static resource bundles while deploying

davisagli:
# Critical Changes

# Changes
CumulusCI can now handle building static resource bundles (*.resource) while deploying using the Metadata API. To use this option, specify the `static_resource_path` option for the deploy task (defaults to `static-resources`). Any subdirectory in this path will be turned into a resource file and added to the package during deployment. There must be a corresponding *.resource-meta.xml file for each static resource bundle.

# Issues Closed


cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -60,23 +68,31 @@ def _include_file(self, root_parts, f):
...
             if self._include_directory(root_parts):
                 for f in files:
                     if self._include_file(root_parts, f):
                         yield os.path.join(root, f)
 
+    def _get_static_resource_files(self):
+        for root, dirs, files in os.walk("."):
+            root_parts = root.split(os.sep)[1:]
+            for f in files:
+                if self._include_file(root_parts, f):
```
davidmreed: I don't think I understand calling `_include_file()` here. Shouldn't we zip the static resource bundle content-blind?

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -60,23 +68,31 @@ def _include_file(self, root_parts, f):
             return lower_f.endswith((".js", ".js-meta.xml", ".html", ".css", ".svg"))
         return True
 
-    def _get_files_to_package(self, path):
+    def _get_files_to_package(self):
         for root, dirs, files in os.walk("."):
             root_parts = root.split(os.sep)[1:]
             if self._include_directory(root_parts):
```
davidmreed: Should `_include_directory()` filter out the `static_resources_path` here? Or is there a different mechanism for excluding the un-archived static resources from the final .zip?

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -136,6 +153,68 @@ def _process_meta_xml(self, zipf):
...
+                zip_dest.writestr(name, content)
+
+        # Build static resource bundles and add to package zip
+        with temporary_dir():
+            os.mkdir("staticresources")
+            bundles = []
+            for name in os.listdir(path):
+                bundle_relpath = os.path.join(relpath, name)
+                bundle_path = os.path.join(path, name)
+                if not os.path.isdir(bundle_path):
```
davidmreed: Uncompressed single files are legal static resources. What would you think about just copying regular files over from `static-resources` to `staticresources`? Otherwise, a project that had both single-file and directory static resources would end up with a hybrid directory structure like this, which might be confusing.

    src/
        staticresources/
            myresource.resource  (text/json)
            myotherresource.resource (text/csv)
        static-resources/
            my-resource-dir/
                file1

cumulusci/tasks/salesforce/Deploy.py
davisagli: The `static_resources_path` is meant to be a separate directory outside of the `path` being deployed.

We should add a check for that but my immediate goal is unblocking the project that needs this feature, so I'm inclined to follow up and add this in a separate PR soon.

cumulusci/tasks/salesforce/Deploy.py
davisagli: Yeah I think this was copy-paste coding. Will remove.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -136,6 +153,68 @@ def _process_meta_xml(self, zipf):
         zipf = zip_clean_metaxml(zipf, logger=self.logger)
         return zipf
 
+    def _process_static_resources(self, zip_src):
+        relpath = self.options.get("static_resource_path", "static-resources")
```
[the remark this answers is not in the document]
davisagli: I'm not too worried about that edge case, but it's an easy change that makes sense, so I'll do it.

cumulusci/tasks/salesforce/Deploy.py
davisagli: Personally the hybrid doesn't bother me -- there's a clear distinction between static resources that fit the standard metadata format and can be deployed without modification, and uncompressed bundles that need to be transformed first. I'm curious to see what Jason thinks next week.

PR #1060, opened 2019-04-30
Don't install requests with security extra

davisagli:
It's not needed in Python 2.7.6+, and it causes requests to use pyOpenSSL instead of the stdlib ssl module.

# Critical Changes

# Changes

# Issues Closed


davisagli:
never mind, it's still required by simple-salesforce

PR #1061, opened 2019-04-30
Retry failed MDAPI connections before giving up

davisagli:
# Critical Changes

# Changes
* If CumulusCI gets a connection error while trying to call the Salesforce Metadata API, it will now retry several times before giving up.

# Issues Closed


davisagli:
@jstvz that's my thought -- the existing test_release build exercises this code to prevent regression, and I'm pretty confident in the retry mechanism from when I implemented it for github.

Could you re-hit approve? I had to make another commit to rerun the appveyor build.

PR #1063, opened 2019-05-01
String format package_xml to allow for handling of namespace prefixes

jlantz:
# Critical Changes

# Changes

* The `update_admin_profile` task now uses Python string formatting on the package.xml file used for retrieve.  This allows injection of namespace prefixes using `{managed}` and `{namespaced_org}`.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/UpdateAdminProfile.py
+++ b/cumulusci/tasks/salesforce/UpdateAdminProfile.py
@@ -61,6 +61,10 @@ class UpdateAdminProfile(Deploy):
         )
         with open(path, "r") as f:
             self._package_xml_content = f.read()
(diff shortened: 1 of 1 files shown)


davisagli:
@jlantz are we using tokens like this somewhere else? I know we replace tokens like %%%NAMESPACE%%% in package.xml when deploying, so I'm thinking we should be consistent with that rather than using a different token.

Also are we 100% sure it's never valid to have curly brackets in package.xml?

jlantz:
This is just using the same token strings we use in this task already for record type names.  If there was a need to include curly brackets in the package.xml they could be xml escaped which would not get handled by Python's format method.

PR #1064, opened 2019-05-01
Add Ensure Record Types task

davidmreed:
# Critical Changes

# Changes

 - Add a task, Ensure Record Types, to generate a Record Type and optional Business Process for a specific sObject and deploy the metadata.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/EnsureRecordTypes.py
+++ b/cumulusci/tasks/salesforce/EnsureRecordTypes.py
@@ -74,26 +74,35 @@ class EnsureRecordTypes(BaseSalesforceApiTask):
                 "Record Type Developer Name value must contain only alphanumeric or underscore characters"
             )
 
-    def _infer_business_process(self):
+    def _infer_requirements(self):
         # If our sObject is Lead or Opportunity, we need to generate businessProcess
         # metadata to make the record type deployable.
-        sobject = self.options["sobject"]
 
-        if sobject in SOBJECT_MAP:
+        # Regardless of sObject, we don't need to do any work if it already
+        # has record types in the target org.
+        sobject = self.options["sobject"]
+        describe_results = getattr(self.sf, sobject).describe()
+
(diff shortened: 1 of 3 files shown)


cumulusci/tasks/salesforce/EnsureRecordTypes.py
```diff
@@ -0,0 +1,140 @@
...
+                filter(
+                    lambda pl: pl["active"],
+                    list(
+                        filter(
+                            lambda f: f["name"] == SOBJECT_MAP[sobject],
+                            describe_results["fields"],
+                        )
+                    )[0]["picklistValues"],
+                )
+            )[0]["value"]
```
davisagli: This is mostly a matter of taste but I'd like to suggest that the following is more readable:

    stage_field = [f for f in describe_results["fields"] if f["name"] == SOBJECT_MAP[sobject]][0]
    active_picklist_values = [value for value in stage_field['picklistValues'] if value["active"]]
    self.options["stage_name"] = active_picklist_values[0]["value"]

cumulusci/tasks/salesforce/EnsureRecordTypes.py
```diff
@@ -0,0 +1,140 @@
...
+                    record_type_label=self.options["record_type_label"],
+                    business_process_metadata=business_process_metadata,
+                    business_process_link=business_process_link,
+                )
+            )
+        with open("package.xml", "w") as f:
+            f.write(PACKAGE_XML)
+
+    def _run_task(self):
+        self._infer_business_process()
```
davisagli: At least for the NPSP use case, part of it is that we don't want to create the NPSP Default record type if there are already opportunity record types. We also don't want to overwrite the record type metadata if it already exists. So I think we need to start by retrieving the specified sobject to see what record types it already has.

davidmreed: Agreed. Resolving this in a new commit.

PR #1069, opened 2019-05-07
Add prerelease scratch org

davisagli:
# Critical Changes

# Changes

# Issues Closed


orgs/prerelease.json
```diff
@@ -0,0 +1,14 @@
+{
+  "orgName": "NPSP - Dev Workspace",
```
boakley: Would it make sense to add "prerelease" in the org name? Do we use that name anywhere, or does it show up in user-facing popup messages or anything like that?

PR #1074, opened 2019-05-09
Decompress static resources on retrieve

jlantz:
# Critical Changes

# Changes

* Added `static_resource_path` option to BaseRetrieveMetadata task.  If set, will handle decompressing the static resources into a different directory after retrieve

# Issues Closed


jlantz:
This is no longer needed per discussion above

davisagli:
@jlantz discussion where?

PR #1075, opened 2019-05-09
Prepare 2.4.4

davisagli:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,22 @@
 History
 =======
 
+2.4.4 (2019-05-09)
+------------------
+
+New features:
+
+* Added tasks ``list_changes`` and ``retrieve_changes`` which interact with source tracking in scratch orgs to handle retrieving changed metadata as Metadata API format source.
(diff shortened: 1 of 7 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,22 @@
 History
 =======
 
+2.4.4 (2019-05-09)
+------------------
+
+New features:
+
+* Added tasks ``list_changes`` and ``retrieve_changes`` which interact with source tracking in scratch orgs to handle retrieving changed metadata as Metadata API format source.
+* Added task ``EnsureRecordTypes`` to generate a Record Type and optional Business Process for a specific sObject and deploy the metadata, if the record type does not already exist.
```
davidmreed: How about "if the object does not already have Record Types"?

cumulusci/__init__.py
```diff
@@ -2,5 +2,5 @@
 import os
 
 __import__("pkg_resources").declare_namespace("cumulusci")
-__version__ = "2.4.3"
+__version__ = "4"
```
davidmreed: Is this change intended? Shouldn't we have "2.4.4"?

davisagli: Ouch, yeah, this is a typo.

Issue #973, opened 2019-02-07
Delete Session Records keyword misses some records


davisagli:
delete_session_records iterates over self._session_records, but for each item it calls salesforce_delete, which calls remove_session_record, which mutates self._session_records. So every other record is missed.

davisagli:
Fixed in cumulusci 2.5.0

PR #1082, opened 2019-05-13
Feature/fix delete session records

boakley:
Fixes issue #973 

The problem is that our Delete Session Records keyword was iterating over a list and deleting items on that list.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/tests/salesforce/api.robot
+++ b/cumulusci/robotframework/tests/salesforce/api.robot
@@ -58,3 +58,31 @@ SOQL Query
     Should Be Equal  &{result}[totalSize]  ${1}
     Should Be Equal  &{contact}[FirstName]  &{new_contact}[FirstName]
     Should Be Equal  &{contact}[LastName]  &{new_contact}[LastName]
+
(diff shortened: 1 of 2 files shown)


cumulusci/robotframework/tests/salesforce/api.robot
```diff
@@ -58,3 +58,31 @@ SOQL Query
...
+        ...  LastName=${random string}
+    END
+    @{query}=  Salesforce Query    Contact  LastName=${random string}
+    length should be  ${query}  5  Expected the query to return five records, but it returned ${query}
+
+    # Now, call 'Delete Session Records' and verify all five were deleted
+    Delete Session Records
+    @{query}=  Salesforce Query  Contact
+    ...  LastName=${random string}
+    length should be  ${query}  0  Expected the query to return 0 records, but it returned ${query}
```
davisagli: I probably would have written this as a Python test that mocks the API call to Salesforce. Not worth changing now, but for the sake of build times we should probably reserve full integration tests for cases where we need them.

Issue #868, opened 2018-10-30
Make setting the default cci org also set it as default for sfdx


njjc:
In order to facilitate using the VS Code integration with sfdx, I'd like the command
`cci org default foo` 
to run the command
`sfdx force:config:set defaultusername=test-autogeneratedscratchlogin@example.com`

Both commands are project specific. This would allow my scratch org to follow along in VS Code without typing an extra sfdx command every time I switch orgs.

1handclapping:
Hi @cdcarter, I definitely second this enhancement - I made a quick patch that works for me [here](https://github.com/SFDO-Tooling/CumulusCI/compare/master...1handclapping:feature/new-dx-commands). I figure there is some larger refactoring that will go into 2.3 but if this suffices happy to make a pr.

davisagli:
@njjc added in cumulusci 2.5.0

PR #1083, opened 2019-05-15
sfdx.SFDXBaseTask supports formatting extra with org_config; Add dx_set_defaultusername Task

spelak-salesforce:
# Critical Changes

# Changes
-  **sfdx.SFDXBaseTask** supports formatting `extra` with `self.org_config`
-  Added `dx_set_defaultusername` Task to set SFDX Default Username from `org_config`.   Useful for VS Code integration whose commands assume a Default Username is set.
    ```yaml
    class_path: cumulusci.tasks.sfdx.SFDXBaseTask
    options:
        command: 'force:config:set'
        extra: 'defaultusername="{org_config.sfdx_alias}"'
        format_extra_with_org_config: True
    ```

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -82,9 +82,17 @@ tasks:
         group: Salesforce DX
     dx_push:
         description: Uses sfdx to push the force-app directory metadata into a scratch org
-        class_path: cumulusci.tasks.sfdx.SFDXOrgTask
+        class_path: cumulusci.tasks.sfdx.SFDXBaseTask
         options:
             command: 'force:source:push'
+        group: Salesforce DX
+    dx_set_defaultusername:
+        description: Sets sfdx's defaultusername as Org's sfdx alias
+        class_path: cumulusci.tasks.sfdx.SFDXBaseTask
+        options:
+            command: 'force:config:set'
+            extra: 'defaultusername="{org_config.sfdx_alias}"'
+            format_extra_with_org_config: True
         group: Salesforce DX
     execute_anon:
         description: Execute anonymous apex via the tooling api.
--- a/cumulusci/tasks/sfdx.py
+++ b/cumulusci/tasks/sfdx.py
@@ -28,6 +28,9 @@ class SFDXBaseTask(Command):
             "required": True,
         },
         "extra": {"description": "Append additional options to the command"},
+        "format_extra_with_org_config": {
+            "description": "Formats extra with org_config.  Example: {org_config.sfdx_alias}"
+        },
     }
 
     def _init_options(self, kwargs):
@@ -35,7 +38,12 @@ class SFDXBaseTask(Command):
         self.options["command"] = self._get_command()
         # Add extra command args from
         if self.options.get("extra"):
-            self.options["command"] += " {}".format(self.options["extra"])
+            extra = self.options["extra"]
+
+            if self.options.get("format_extra_with_org_config"):
+                extra = extra.format(org_config=self.org_config)
+
+            self.options["command"] += " {}".format(extra)
 
     def _get_command(self):
         command = "{SFDX_CLI} {command}".format(
--- a/cumulusci/tasks/tests/test_sfdx.py
+++ b/cumulusci/tasks/tests/test_sfdx.py
@@ -1,13 +1,11 @@
 """ Tests for the SFDX Command Wrapper"""
 
-import logging
 import mock
 import unittest
 
 from mock import MagicMock
 from mock import patch
 
-import cumulusci.core.tasks
 from cumulusci.core.config import BaseGlobalConfig
 from cumulusci.core.config import BaseProjectConfig
 from cumulusci.core.config import TaskConfig
@@ -51,6 +49,31 @@ class TestSFDXBaseTask(MockLoggerMixin, unittest.TestCase):
             pass
 
         self.assertEqual("sfdx force:org --help", task.options["command"])
+
+    def test_base_task_with_format_extra_with_org_config(self):
+        """ The command is prefixed w/ sfdx """
+
+        self.task_config.config["options"] = {
+            "command": "force:config:set",
+            "extra": 'defaultusername="{org_config.sfdx_alias}"',
+            "format_extra_with_org_config": "True",
+        }
+        sfdx_alias = "Project Name__dev"
+        org_config = ScratchOrgConfig(
+            {"username": "test@example.com", "sfdx_alias": sfdx_alias}, "test"
+        )
+
+        task = SFDXBaseTask(self.project_config, self.task_config, org_config)
+
+        try:
+            task()
+        except CommandException:
+            pass
+
+        self.assertEqual(
+            'sfdx force:config:set defaultusername="' + sfdx_alias + '"',
+            task.options["command"],
+        )
 
     @patch(
         "cumulusci.tasks.sfdx.SFDXOrgTask._update_credentials",


spelak-salesforce:
@davidmreed , the **coninuous-integration/appveyor/branch** failed but for an unrelated test.  Can you help with this?

davisagli:
@spelak-salesforce This is similar to an issue on our backlog, #868, but we were thinking of it as something that should happen automatically when you set your default CCI org using `cci org default [orgname]` rather than a standalone task. Considering that idea, do you think there are still situations where you would want to use this as a separate task? i.e. a case where you want the org to be the default for sfdx but not for cci?

spelak-salesforce:
> @spelak-salesforce This is similar to an issue on our backlog, #868, but we were thinking of it as something that should happen automatically when you set your default CCI org using `cci org default [orgname]` rather than a standalone task. Considering that idea, do you think there are still situations where you would want to use this as a separate task? i.e. a case where you want the org to be the default for sfdx but not for cci?

-  The **Salesforce Extension Pack** VS Code Extension requires a sfdx Default Username to be set.  There are some nice features.  Also, the **Salesforce Extension Pack** displays the sfdx Default Username on screen (bottom middle) which is useful for knowing where what state the project is in
Screen Shot 2019-05-16 at 9 38 36 AM
-  I've written functions/aliases to help me quickly do the following:
   -  ```bash
        # delete the existing dev org and recreate the dev org
        # to guarantee the scratch org has exactly my source 
        alias dev_org='cci org remove dev && cci flow run dev_org --org dev'
        ```
   -  ```bash
        # Execute test shortcuts
        # I like the sfdx test printout a lot more than cci's test printout
        alias sfdx_test='sfdx force:apex:test:run -c -r human -w 10 -y -n $1'
        alias sfdx_tests='sfdx force:apex:test:run -c -r human -w 10 -n $1'
        alias sfdx_tests_all='sfdx force:apex:test:run -c -l RunLocalTests -r human'
       ```
    -  I use my `dev_org` alias often, so I'm deleting and recreating scratch orgs a dozen+ times a day. 
  Plus I want to quickly execute Apex Tests with sfdx, so I want use my `sfdx_test` aliases.    I need `project.name` from `cumulusci.yml` + know which config is used to known programatically what the sfdx Username is for the current state of the project.   This isn't easily accessible from `cci` (to my knowledge).   
        -  I've written functions/aliases to extract this info from `cumulusci.yml` that is hard-coded to use the **dev** config.
        ```bash
        function yaml {
          python -c "import yaml;print(yaml.safe_load(open('$1'))$2)"
        }

        function dev_sfdx {
          PROJECT_NAME=$(yaml cumulusci.yml  "['project']['name']")
          PROJECT_NAME+='__dev'
          sfdx force:config:set defaultusername=$PROJECT_NAME
        }
        ``
        -   But this is annoying + hard-coded to **dev** config.  I want to:
            -  Be able to create a cci task to execute a sfdx command where I fill in the `sfdx_alias` (this is the update to `sfdx.SFDXBaseTask`)
            -  We don't specifically need a cci `dx_set_defaultusername` in cci.  I can create this task locally in a project.
            -  I also don't want to force the other engineers on my team to have to have their sfdx Default Username to be set.  My plan is to create another `dev_org_dx` flow in our team's project.

davisagli:
@spelak-salesforce I'm not trying to question the usefulness of sfdx commands -- I use them too. What I'm saying is we can make it so that `cci org default dev` takes care of setting `dev` as the default scratch org for CCI commands AND ALSO setting PROJECTNAME__dev as the defaultusername for sfdx.

Then your `dev_org` alias would become:
`alias dev_org='cci org remove dev && cci org default dev && cci flow run dev_org'`
This would set both defaults and you would not need to adjust flows to run an additional task.

Would you use this or am I missing something important?

spelak-salesforce:
> @spelak-salesforce I'm not trying to question the usefulness of sfdx commands -- I use them too. What I'm saying is we can make it so that `cci org default dev` takes care of setting `dev` as the default scratch org for CCI commands AND ALSO setting PROJECTNAME__dev as the defaultusername for sfdx.
> 
> Then your `dev_org` alias would become:
> `alias dev_org='cci org remove dev && cci org default dev && cci flow run dev_org'`
> This would set both defaults and you would not need to adjust flows to run an additional task.
> 
> Would you use this or am I missing something important?

@davisagli , all good.  Yes, I could set a default cci org that sets the sfdx Default Usernameto solve my problem.
-  Make sure to wrap the `PROJECTNAME__dev` in double quotes since the Project Name can have spaces.   
-  In VS Code, it is a lot more useful to set the sfdx Default Username with the alias than the username since whatever is set is displayed on the screen (see image above).

A separate question: Is there anything wrong with the change to `sfdx.SFDXBaseTask` to support formatting `extra` with `org_config` attributes?   
-  This allows one to call sfdx getting the cci `sfdx_alias` for an arbitrary sfdx command.
-  `sfdx.SFDXOrgTask` exists and maybe this change isn't needed if setting cci default org sets the sfdx Default Username
-  Since `format()` is used, is there any risk of modifying `org_config`?

davisagli:
@spelak-salesforce thanks, those are good details to keep in mind. I'll try to work on this addition to the `cci org default` command soon.

I don't see any glaring problems with adding the ability to format `extra` with org_config attributes, but if the change I'm proposing for `cci org default` makes it unnecessary, I'd rather wait to add it until there's a clear need.

davisagli:
@spelak-salesforce I've opened a PR with my proposal, #1086

spelak-salesforce:
@davisagli , could we do something simpler?   

Can we change `sfdx.SFDXOrgTask` to append the `org_config.sfdx_alias` instead of the `org_config.username`?

I think I do want to be able to create a task to set the sfdx Default Username from cci config:
-  My team's repo is in Metadata format
-  I am creating a temporary **sfdx** project in a branch so I can use `force:source:pull` to pull in Metadata changes made in the UI
-  I am currently manually calling `sfdx force:config:set defaultusername="Case Management__dev"` and prefer to use a cci command which pulls the info from the config rather than me looking it up

davisagli:
@spelak-salesforce 

> I am currently manually calling sfdx force:config:set defaultusername="Case Management__dev" and prefer to use a cci command which pulls the info from the config rather than me looking it up

That is what the `cci org default [orgname]` command will do in the next release since we merged #1086

davisagli:
@spelak-salesforce I agree with your suggestion that it would be better for SFDXOrgTask to pass the sfdx_alias instead of the username though.

spelak-salesforce:
@davisagli 

> @spelak-salesforce
> 
> > I am currently manually calling sfdx force:config:set defaultusername="Case Management__dev" and prefer to use a cci command which pulls the info from the config rather than me looking it up
> 
> That is what the `cci org default [orgname]` command will do in the next release since we merged #1086

But if I create a sfdx Project inside a Metadata-format Repo, I need to change directory to the project and won't be in a cci directory anymore.   ... and using cci commands would be unavailable anyways and my Use Case wouldn't work anyways.

-------

> @spelak-salesforce I agree with your suggestion that it would be better for SFDXOrgTask to pass the sfdx_alias instead of the username though.

May I change this Pull Request to only change `sfdx.SFDXOrgTask` to append the `org_config.sfdx_alias` instead of the `org_config.username`?  I have test data ready-to-go to test the `sfdx_alias`.   Then you can decide if you want to merge or not.

davisagli:
> But if I create a sfdx Project inside a Metadata-format Repo, I need to change directory to the project and won't be in a cci directory anymore. ... and using cci commands would be unavailable anyways and my Use Case wouldn't work 😅.

@spelak-salesforce it should work if you create your sfdx project inside the root of the repository. cci looks for a containing directory with a .git subdirectory.

> May I change this Pull Request to only change sfdx.SFDXOrgTask to append the org_config.sfdx_alias instead of the org_config.username?

Yeah, go ahead.

davisagli:
> May I change this Pull Request to only change sfdx.SFDXOrgTask to append the org_config.sfdx_alias instead of the org_config.username?

Closing this due to inactivity. Please feel free to open a new PR with this change.

PR #1088, opened 2019-05-21
refactor install_regression flow

davisagli:
# Critical Changes

# Changes
* Refactored the `install_regression` flow to better parallel other flows. The `install_regression` flow now installs the package without configuring it. Use the new `regression_org` flow to both install and configure.

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -453,6 +453,12 @@ flows:
             2:
                 task: update_admin_profile
 
+    config_regression:
+        description: Configure an org for QA regression after the package is installed
+        steps:
+            1:
+                flow: config_managed
+
     dependencies:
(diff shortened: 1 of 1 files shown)


davidmreed:
Echoing James, but I want to hold Cumulus's infrastructure pull request until we get this merged/released so as not to have to do it twice.

davisagli:
@jstvz not sure I understand the exact order of operations you're proposing. I'll hold off on merging this but let me know if we need to discuss.

jstvz:
Sorry if unclear. Because `regression_org` in SAL doesn't include `config_managed` I'd have to add it, but I realized that we _don't_ have to wait to merge this as long as I can update the relevant flows before we cut a release.

davisagli:
@jstvz that makes more sense. I'm hoping to make a release of CumulusCI in the next couple days so I think it should work out.

PR #1093, opened 2019-05-23
Feature/robot open test browser

### LINKED ISSUE #1068: No support for multiple windows in robot
When running a robot test where a user has to open two windows (for example to test a live chat functionality), the keyword "Open Test Browser" fails. You can run the keyword once to open a window but attempting to open a second window using the same keyword fails with this error:

The custom locator 'text' cannot be registered. A locator of that name already exists.

The error is being caused by the "Add Location Strategy" keyword in "Open Test Browser".
Any solution to this that still allows the user to select which browser to use when running the task?

boakley:
# Critical Changes

# Changes

Updates to the keyword 'Open Test Browser':
- allows you to open more than one browser in a single test case. 
- sets the default size for the browser window to 1280x1024
- adds a new keyword argument `size` to override the default size
- adds a new keyword argument `alias` to let you assign an alias to multiple browser windows

# Issues Closed

This fixes #1068

### DIFF
--- a/cumulusci/robotframework/tests/salesforce/browsers.robot
+++ b/cumulusci/robotframework/tests/salesforce/browsers.robot
@@ -17,6 +17,20 @@ Assert active browser count
     Length should be  ${browsers}  ${expected count}
     ...  Expected to find ${expected count} open browsers, found ${actual count}
 
+Assert window size
+    [Documentation]
+    ...  Verify the actual window size is the expected size
+    ...  give or take a pixel or five.
+    [Arguments]  ${expected width}  ${expected height}
+
+    # On circleci, the browser size is sometimes off by a pixel. How rude!
+    # Since we don't so much care about the precise size as we do that
+    # we are able to change the size, we'll allow a tiny bit of wiggle room.
+    ${actual width}  ${actual height}=  Get window size
+    ${xdelta}=  evaluate  abs(int($actual_width)-int($expected_width))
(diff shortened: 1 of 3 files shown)


cumulusci/robotframework/tests/salesforce/browsers.robot
```diff
@@ -0,0 +1,77 @@
...
+
+Browser aliases
+    [Documentation]  Verify that aliases are properly handled in Open Test Browser
+    [Tags]  issue:1068
+    [Setup]  Run keywords
+    ...  Open test browser  alias=browser1
+    ...  AND  Go to  https://metadeploy.herokuapp.com/products
+    ...  AND  Open test browser  alias=browser2
+    ...  AND  Go to  https://mrbelvedereci.herokuapp.com/
+    [Teardown]  Close all browsers
```
davisagli: At first I thought it was neat that this also implicitly tests that MetaDeploy and MetaCI are up, but since those tools are based on CumulusCI, I want to avoid the situation where one of them goes down due to a bug in CumulusCI and we can't merge the fix because this test is failing. Can we switch to sites that don't depend on CumulusCI?

boakley: yeah, I've been second guessing myself on that. Have any recommendations? I want something where I can hard-code the URL for simplicity. I was thinking google.com and bing.com, but that's kinda rude to use someone else's servers in a test. I also though of maybe a couple of static pages on salesforce.com.

PR #1094, opened 2019-05-24
Prepare 2.5.0

davisagli:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,43 @@
 History
 =======
 
+2.5.0 (2019-05-25)
+------------------
+
+Breaking changes:
+
+* We reorganized the flows for setting up a package for regression testing for better symmetry with other flows.
+  If you were running the ``install_regression`` flow before, you now probably want ``regression_org``.
+
+  Details: The ``install_regression`` flow now installs the package _without_ configuring it.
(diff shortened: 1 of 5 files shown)


HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+  * Added a framework for creating "page object" classes to contain keywords related to a particular page or component.
+  * The ``robot`` task now takes a ``name`` option to control the name of the robot suite in output.
+  * Updates to the keyword ``Open Test Browser``:
+
+    * It allows you to open more than one browser in a single test case. (#1068)
+    * It sets the default size for the browser window to 1280x1024.
+    * Added a new keyword argument ``size`` to override the default size.
+    * Added a new keyword argument ``alias`` to let you assign an alias to multiple browser windows.
+
+Issues fixed:
```
boakley: I don't see issue #1068 mentioned under "Issues Fixed", though you do mention it in an earlier section.

davisagli: I was on the fence about where to put that one, but was considering it more of a new feature than a bug fix because we never really tried to support multiple browser windows previously. Also it made it easier to group with the other changes for this keyword.

PR #1106, opened 2019-06-06
Add PreflightFlowCoordinator to evaluate preflight checks

davisagli:
Preflight checks can be defined in the `checks` list at either the flow or step level. The PreflightFlowCoordinator contains the logic to evaluate them and produce `preflight_results`, which MetaDeploy will make use of.

Each check consists of:

- `when`: a jinja2 expression. Evaluated in a context that has access to `project_config`, `org_config`, and `tasks` which is a way of accessing tasks in a way that will cache their results so that multiple checks using the same task will only run the task once (for example if there are multiple checks making use of a task that gets the current installed package versions)
- `action`: a string that will be set as the `status` in the preflight result if the `when` condition evaluates to true
- `message`: a string that will be set as the `message` in the preflight result if the `when` condition evaluates to true

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -202,7 +202,7 @@ class TaskRunner(object):
 
         # Resolve ^^task_name.return_value style option syntax
         task_config = self.step.task_config.copy()
-        task_config["options"] = task_config["options"].copy()
+        task_config["options"] = task_config.get("options", {}).copy()
         self.flow.resolve_return_value_options(task_config["options"])
 
         task_config["options"].update(options)
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -607,7 +607,7 @@ class PreflightFlowCoordinator(FlowCoordinator):
         self._rule(new_line=True)
 
         self.jinja2_context = {
-            "tasks": TaskCache(self, self.project_config, org_config),
+            "tasks": TaskCache(self),
             "project_config": self.project_config,
             "org_config": self.org_config,
         }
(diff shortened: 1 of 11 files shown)


cumulusci/core/flowrunner.py
```diff
@@ -581,3 +586,96 @@ def _find_result_by_path(self, path):
...
+
+class TaskCache(object):
+    """Provides access to named tasks and caches their results.
+
+    This is intended for use in a jinja2 expression context
+    so that multiple expressions evaluated in the same context
+    can avoid running a task more than once with the same options.
+    """
+
+    def __init__(self, flow, project_config, org_config):
```
davidmreed: Is there an intent behind passing these values directly instead of accessing them through `flow`?

cumulusci/core/flowrunner.py
```diff
@@ -581,3 +586,96 @@ def _find_result_by_path(self, path):
...
+        finally:
+            self.callbacks.post_flow(self)
+
+    def evaluate_check(self, check):
+        self.logger.info("Evaluating check: {}".format(check["when"]))
+        expr = jinja2_env.compile_expression(check["when"])
+        value = bool(expr(**self.jinja2_context))
+        self.logger.info("Check result: {}".format(value))
+        if value:
+            return {"status": check["action"], "message": check.get("message")}
```
davidmreed: I often like the explicit `return None` rather than falling off the edge, but my Python style is not necessarily mainstream. Does it matter / is there an imperative or style rec to handle it this way?

davisagli: I don't think I've seen a particular recommendation. Personally I almost never use an explicit `return None` but my reason is mostly "it's more concise and I know what it means" which isn't a great reason.

I think part of the answer is probably to start using type checks so that we'll be forced to review anywhere None is returned inadvertently.

cumulusci/core/flowrunner.py
davisagli: No, I forgot they're available on the flow. Thanks. I'll refactor this.

PR #1116, opened 2019-06-11
Add options to control UpdateDependencies downgrades

davisagli:
# Critical Changes

# Changes
Added two options to the UpdateDependencies task:
- `allow_newer`: If the org already has a newer release, use it. Defaults to True.
- `allow_uninstalls`: Allow uninstalling a beta release or newer final release if needed in order to install the requested version. Defaults to False.

These defaults are a change from prior behavior since uninstalling packages is not commonly needed when working with scratch orgs, and it is potentially destructive.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
+++ b/cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
@@ -14,7 +14,7 @@ class TestUpdateDependencies(unittest.TestCase):
     @mock.patch(
         "cumulusci.salesforce_api.metadata.ApiRetrieveInstalledPackages.__call__"
     )
-    def test_run_task(self, ApiRetrieveInstalledPackages):
+    def test_run_task_with_downgrades(self, ApiRetrieveInstalledPackages):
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
```diff
@@ -116,6 +120,56 @@ def test_run_task(self, ApiRetrieveInstalledPackages):
         )
         self.assertEqual(10, task.api_class.call_count)
 
+    @mock.patch(
+        "cumulusci.salesforce_api.metadata.ApiRetrieveInstalledPackages.__call__"
+    )
+    def test_run_task_without_downgrades(self, ApiRetrieveInstalledPackages):
```
davidmreed: Do we need a test case for a situation where `allow_newer` and `allow_uninstalls` are both `False`?

davisagli: Line coverage for the module was at 100%, but I added tests for all 4 combinations just to be sure.

PR #1117, opened 2019-06-15
Use Click's UsageError when flow/task is not found

### LINKED ISSUE #960: "cci task run invalid_task" should not give a stack trace
If I try to run a task that hasn't been defined (or maybe I spelled it wrong), I get a stack trace on stderr instead of a more human-friendly response.

At the very least, it should just print out something like `cannot find task 'foobar'` or `invalid task: 'foobar'`, possibly followed by a list of valid tasks. Even better might be to try to do a fuzzy match to existing tasks, so we could display something like "cannot find task 'foobor'. Did you mean 'foobar'?".

jstvz:
Before:
```console
$ cci flow info derpendencies
[...]
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/cumulusci/core/config/BaseTaskFlowConfig.py", line 47, in get_flow
    raise FlowNotFoundError("Flow not found: {}".format(name))
cumulusci.core.exceptions.FlowNotFoundError: Flow not found: derpendencies
```

After:
```console
$ cci flow info derpendencies
Checking the version!
Usage: cci flow info [OPTIONS] FLOW_NAME

Error: Flow not found: derpendencies. Did you mean "dependencies"?
```
# Critical Changes

# Changes
- `cci` no longer prints tracebacks when a flow or task is not found. Additionally, it will suggest a name if a close enough match can be found.
# Issues Closed
#960

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -38,6 +38,7 @@ from cumulusci.core.exceptions import CumulusCIUsageError
 from cumulusci.core.exceptions import OrgNotFound
 from cumulusci.core.exceptions import ScratchOrgException
 from cumulusci.core.exceptions import ServiceNotConfigured
+from cumulusci.core.exceptions import FlowNotFoundError
 from cumulusci.core.exceptions import TaskNotFoundError
 from cumulusci.core.utils import import_global
 from cumulusci.cli.config import CliRuntime
@@ -1024,7 +1025,11 @@ def task_doc(config):
 @click.argument("task_name")
 @pass_config(load_keychain=False)
 def task_info(config, task_name):
-    task_config = config.project_config.get_task(task_name)
+    try:
+        task_config = config.project_config.get_task(task_name)
+    except (CumulusCIUsageError, TaskNotFoundError) as e:
+        raise click.UsageError(str(e))
+
     doc = doc_task(task_name, task_config).encode()
     click.echo(rst2ansi(doc))
 
@@ -1066,7 +1071,14 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
     org, org_config = config.get_org(org, fail_if_missing=False)
     task_config = getattr(config.project_config, "tasks__{}".format(task_name))
     if not task_config:
-        raise TaskNotFoundError("Task not found: {}".format(task_name))
+        raise click.UsageError(
+            "Task not found: {}. Did you mean {}".format(
+                task_name,
+                config.project_config.get_suggested_name(
+                    task_name, config.project_config.tasks
+                ),
+            )
+        )
 
     # Get the class to look up options
     class_path = task_config.get("class_path")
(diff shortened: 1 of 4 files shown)


cumulusci/core/config/BaseTaskFlowConfig.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import unicode_literals
+from difflib import get_close_matches
```
davisagli: Cool, I didn't know about this one.

cumulusci/cli/cci.py
```diff
@@ -1104,7 +1116,7 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
 
             pdb.set_trace()
 
-    except CumulusCIUsageError as e:
+    except (CumulusCIUsageError, TaskNotFoundError) as e:
```
davisagli: TaskNotFoundError is a subclass of CumulusCIUsageError; you don't have to catch it explicitly.

cumulusci/cli/cci.py
```diff
@@ -1203,7 +1218,7 @@ def flow_run(config, flow_name, org, delete_org, debug, o, skip, no_prompt):
     try:
         coordinator = config.get_flow(flow_name, options=options)
         coordinator.run(org_config)
-    except CumulusCIUsageError as e:
+    except (CumulusCIUsageError, FlowNotFoundError) as e:
```
davisagli: FlowNotFoundError is a subclass of CumulusCIUsageError; you don't have to catch it explicitly.

cumulusci/cli/cci.py
```diff
@@ -1066,7 +1071,14 @@ def task_run(config, task_name, org, o, debug, debug_before, debug_after, no_pro
...
     if not task_config:
-        raise TaskNotFoundError("Task not found: {}".format(task_name))
+        raise click.UsageError(
+            "Task not found: {}. Did you mean {}".format(
+                task_name,
+                config.project_config.get_suggested_name(
+                    task_name, config.project_config.tasks
+                ),
+            )
+        )
```
davisagli: You can use project_config.get_task here -- it returns a TaskConfig, so you would just to change the code below to use task_config.config instead of task_config, to keep accessing the raw data.

cumulusci/cli/cci.py
```diff
@@ -1024,7 +1025,11 @@ def task_doc(config):
 @click.argument("task_name")
 @pass_config(load_keychain=False)
 def task_info(config, task_name):
-    task_config = config.project_config.get_task(task_name)
+    try:
+        task_config = config.project_config.get_task(task_name)
+    except (CumulusCIUsageError, TaskNotFoundError) as e:
+        raise click.UsageError(str(e))
```
davisagli: TaskNotFoundError is a subclass of CumulusCIUsageError; you don't have to catch it explicitly.

davisagli: Oh and this branch needs a test for coverage.

PR #1118, opened 2019-06-16
Fix color escapes on Windows

### LINKED ISSUE #813: ANSI color escapes broken on Windows
ANSI color codes are not supported natively by cmd.exe or PowerShell:

```
[32m2018-10-05 16:00:38:[0m foo
```

`coloredlogs` states that it includes `colorama` as a dependency on windows, so it should be working. Also, from the docs:

> Note If you want to use ColoredFormatter on Windows then you may need to call colorama.init(). This is done for you when you call coloredlogs.install().

So perhaps call it `if os.name is 'nt'`?

jstvz:
Before:
ps_bad_color
cmd_bad_color

After:
ps_good_color
cmd_good_color

# Critical Changes

# Changes
- Log colors now display correctly on Windows.
# Issues Closed
Fixes #813

### DIFF
--- a/cumulusci/cli/logger.py
+++ b/cumulusci/cli/logger.py
@@ -23,7 +23,7 @@ def init_logger(log_requests=False):
     for handler in logger.handlers:  # pragma: nocover
         logger.removeHandler(handler)
 
-    if os.name == "nt" and "colorama" in sys.modules:
+    if os.name == "nt" and "colorama" in sys.modules:  # pragma: nocover
         colorama.init()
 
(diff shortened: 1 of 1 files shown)


cumulusci/cli/logger.py
```diff
@@ -14,6 +23,9 @@ def init_logger(log_requests=False):
     for handler in logger.handlers:  # pragma: nocover
         logger.removeHandler(handler)
 
+    if os.name == "nt" and "colorama" in sys.modules:
```
davisagli: I think this is another reasonable place to use `# pragma: nocover` to avoid reporting an uncovered line.

Issue #241, opened 2017-03-01
Show more details in org list


cpolcyn:
Currently `cci org list` shows a list of org aliases and identifies which (if any) is default. We should add some more fields such as:

- Global vs local CCI org
- Scratch vs dev
- Which pod is it running on
- etc

cdcarter:
a fair amount of this can be pulled out of userinfo - tho i don't know if ScratchOrgConfig has been tested to actually inherit that implementation.

jlantz:
Two new things we should show:

* Which scratch org config the org was generated from
* Whether the namespace is applied in the org

PR #1120, opened 2019-06-17
[wip]:  refactor UI logic into new UI module

### LINKED ISSUE #1023: cci org info packaging shouldn't give stack trace if no packaging.org file
On a whim I ran `cci org info packaging` in my repo, and got gnarly stack trace rather than a human-friendly  message. The script should not be giving out a stack trace.

Desired output (more-or-less):
```
(cci) bash-3.2 $ cci org info packaging
Org information could not be found. Expected to find encrypted file at /Users/boakley/.cumulusci/Cumulus/packaging.org
```

Actual output:
```
(cci) bash-3.2 $ cci org info packaging
  File "/usr/local/bin/cci", line 11, in <module>
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "/usr/local/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "/Users/boakley/dev/CumulusCI/cumulusci/cli/cci.py", line 219, in new_func
    func(config, *args, **kw)
  File "/Users/boakley/dev/CumulusCI/cumulusci/cli/cci.py", line 791, in org_info
    org_name, org_config = config.get_org(org_name)
  File "/Users/boakley/dev/CumulusCI/cumulusci/cli/config.py", line 91, in get_org
    org_config = self.keychain.get_org(org_name)
  File "/Users/boakley/dev/CumulusCI/cumulusci/core/keychain/BaseProjectKeychain.py", line 169, in get_org
    self._raise_org_not_found(name)
  File "/Users/boakley/dev/CumulusCI/cumulusci/core/keychain/EncryptedFileProjectKeychain.py", line 102, in _raise_org_not_found
    self.project_local_dir, name
cumulusci.core.exceptions.OrgNotFound: Org information could not be found. Expected to find encrypted file at /Users/boakley/.cumulusci/Cumulus/packaging.org
```

jstvz:
AKA: Add fancy tables for: 
- `flow list`
- `service list`
- `org list`
- `task list`

This PR adds a new module, `cumulusci.cli.ui` that is responsible for formatting and outputting cli data for the user. Its first class, CliTable, wraps the construction of a terminaltables.{SingleTable,AsciiTable} instance and provides utilities for formatting tables and printing them to stdout via click.echo.

In cumulusci.cli.cci, references to plaintable have been replaced with calls to CliTable. In addition, `--json` flags have been added to several commands (`service_list`, `task_list`, `flow_list`), as well as a `--plain` flag to force ascii format tables on those terminals which do not throw a UnicodeEncodeError (e.g. emacs-eshell on Linux, but not Windows, inexplicably). `--plain` also suppresses table column wrapping in org_list to allow copying and pasting usernames.

Also:
Fixes table output in emacs and other dumb terminals. Smoke tested on windows 10.

Only use emoji pictograms on posix platforms since Windows command prompt/powershell can't handle all unicode, yet still report `sys.platform == 'utf-8'`. This commit replaces the default pictogram with CP-1256-compliant ascii strings on non-POSIX platforms.

todos:
- [x] Update boolean pictograms in `org list`
- [x] Allow setting plain output via global config
- [x] Investigate using `click.style`'s `dim` parameter to dim inactive rows on `org/service list`
- [ ] Use tables for `* info` commands
- [ ] Investigate additions from #241
- [ ] Resolve #1023 

`org list`:
screenshot_2019-06-17_13-34-39

`task_list`:
screenshot_2019-06-17_13-35-06

`service list`:
screenshot_2019-06-17_13-35-31

# Critical Changes

# Changes
- We've changed how the output of some commands are displayed in tables. For users that prefer Markdown style tables we've added a `--plain` option to approximate the previous behavior.
# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -8,7 +8,6 @@ from collections import defaultdict
 from collections import OrderedDict
 import functools
 import json
-import operator
 import os
 import sys
 import webbrowser
@@ -21,7 +20,6 @@ from shutil import copyfile
 import click
 import pkg_resources
 import requests
-from plaintable import Table
 from rst2ansi import rst2ansi
 from jinja2 import Environment
 from jinja2 import PackageLoader
@@ -42,6 +40,7 @@ from cumulusci.core.exceptions import TaskNotFoundError
 from cumulusci.core.utils import import_global
 from cumulusci.cli.config import CliRuntime
 from cumulusci.cli.config import get_installed_version
+from cumulusci.cli.ui import CliTable
 from cumulusci.utils import doc_task
 from cumulusci.utils import get_cci_upgrade_command
 from cumulusci.oauth.salesforce import CaptureSalesforceOAuth
@@ -584,22 +583,29 @@ project.add_command(project_dependencies)
 
 
 @click.command(name="list", help="List services available for configuration and use")
+@click.option("--plain", is_flag=True, help="Print the table using plain ascii.")
+@click.option("--json", "print_json", is_flag=True, help="Print a json string")
 @pass_config(allow_global_keychain=True)
-def service_list(config):
-    headers = ["service", "description", "is_configured"]
-    data = []
+def service_list(config, plain, print_json):
     services = (
         config.project_config.services
         if not config.is_global_keychain
         else config.global_config.services
     )
+    configured_services = config.keychain.list_services()
+
+    data = [["Name", "Description", "Configured"]]
     for serv, schema in services.items():
-        is_configured = ""
-        if serv in config.keychain.list_services():
-            is_configured = "* "
-        data.append((serv, schema["description"], is_configured))
-    table = Table(data, headers)
-    click.echo(table)
+        is_configured = serv in configured_services
+        data.append([serv, schema["description"], is_configured])
+
+    if print_json:
+        click.echo(json.dumps(data[1:]))
+        return None
+
+    table = CliTable(data, title="Services", wrap_cols=[1])
+    table.stringify_boolean_cols(2)
+    table.echo(plain)
 
 
 class ConnectServiceCommand(click.MultiCommand):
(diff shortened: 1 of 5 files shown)


davisagli:
Some initial thoughts:
- I'm not wild about the way the horizontal lines double the vertical space that output takes up
- the red x calls more attention to itself than is warranted in most cases
- it looks like it'll make it really hard to copy the username on narrow terminals

jstvz:
1. The additional vertical whitespace is necessary to improve the legibility of the output so I'd argue in favor of keeping it even if we decide not to include horizontal row separators. That said, the horizontal lines improve readability by making it easier to discern which columns go with each row.
2. Fair point about the red "❌". It probably only makes sense as a default in `service list`. For `org list` I think it should be something like (`True/False`):
   - Default: `✔️/`
   - Scratch: `✔️/`
   - Expired: ` /❌`
3. The username is tricky. Most users (those with only sfdx usernames) will only see line wrapping on terminals ≤ ~105 columns wide. So, in order to copy a username, users would have to either maximize their terminal or get the output from `org info`.

Aside: I think this is an instance where some A/B testing with users might help clarify matters.

jlantz:
This looks great!  I agree with the feedback thus far with a few additional comments:

1. I understand the concern about more vertical space being used.  It's a balance between scrolling and readability.  I tend to lean a bit toward having the spacing.  I'm also trying to figure out a use case where that extra spacing presents a functional problem and can't really think of one.

2. I agree about only using the ❌when absolutely needed.  Expired in org list is a good example.  I also wonder if we should have some better representation of the concept of N/A in a column.  For example, Expired for a non-scratch org doesn't make much sense.

3. We could just remove username from the `cci org list` and replace it with something like a Created column.  I can't think of a use case for only copying only the username without the password.  If you need the password, you need to use `cci org info` anyhow.  Another thing we could consider would be having rows which are "sleeping".  For example, a scratch config that's not yet created a scratch org could be a lighter color text than other org configs that have an actual org.  The same could be used in `cci service list` for services that aren't yet connected.

One last thought: I'd love to see this applied to cci * info commands as well even if it was a simple Key/Value table.

jstvz:
> 3. We could just remove username from the `cci org list` and replace it with something like a Created column.  I can't think of a use case for only copying only the username without the password.  

I agree about the lack of utility for usernames; I can't think of a use case for displaying the username for an expired scratch org. However, one place the username column does come in handy is with persistent or global orgs (e.g. trial, packaging). We _could_ split the difference and add an option (`--usernames`) that controls whether they are displayed, buuut I'm okay with replacing that column entirely and relying on `org info` for that tbh.

davisagli:
@jstvz did you change your mind about wanting this, or are you planning to open a new PR?

PR #1121, opened 2019-06-19
Feature/libdoc for pageobjects

boakley:
This updates our robot_libdoc task to handle page object files. The main goal is for the product teams to be able to generate documentation for their own page object libraries. 

I've fixed the issue with not being able to generate documentation for the PageObjects.py library. This includes a commit of a new Keywords.html file which contains the documentation for that library.

This code will not generate documentation for our base classes. For one, they don't actually have any keywords in them yet, and for another I haven't figured out the best way to do that since the base libraries themselves aren't directly usable from robot tests. I've created a separate work item [#W-033984](https://foundation.lightning.force.com/lightning/r/agf__ADM_Work__c/a2x1E000001C6idQAC/view) related to generating documentation for the base classes. 

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/pageobjects/PageObjects.py
+++ b/cumulusci/robotframework/pageobjects/PageObjects.py
@@ -1,5 +1,5 @@
 from robot.api import logger
-from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
+from robot.libraries.BuiltIn import BuiltIn
 from .baseobjects import BasePage
 import inspect
 import robot.utils
@@ -38,10 +38,7 @@ class PageObjects(object):
 
         # Start with this library at the front of the library search order;
         # that may change as page objects are loaded.
-        try:
-            BuiltIn().set_library_search_order("PageObjects")
-        except RobotNotRunningError:
-            pass
+        BuiltIn().set_library_search_order("PageObjects")
 
     @classmethod
     def _reset(cls):
@@ -52,16 +49,12 @@ class PageObjects(object):
         for pobj in cls.registry.values():
             if pobj.__module__ in sys.modules:
                 del sys.modules[pobj.__module__]
-            del pobj
         cls.registry = {}
         cls.cache = {}
 
     @property
     def selenium(self):
-        try:
-            return BuiltIn().get_library_instance("SeleniumLibrary")
-        except RobotNotRunningError:
-            return None
+        return BuiltIn().get_library_instance("SeleniumLibrary")
 
     def __getattr__(self, name):
         """Return the keyword from the current page object
--- a/cumulusci/robotframework/pageobjects/PageObjects.py
+++ b/cumulusci/robotframework/pageobjects/PageObjects.py
@@ -44,7 +44,7 @@ class PageObjects(object):
             pass
 
     @classmethod
-    def reset(cls):
+    def _reset(cls):
         """Reset the internal data structures used to manage page objects
 
         This is to aid testing. It probably shouldn't be used at any other time.
--- a/cumulusci/robotframework/pageobjects/PageObjects.py
+++ b/cumulusci/robotframework/pageobjects/PageObjects.py
@@ -25,7 +25,7 @@ class PageObjects(object):
     cache = {}
 
     def __init__(self, *args):
-        logger.info("initializing PageObjects...", "DEBUG")
+        logger.debug("initializing PageObjects...")
         importer = robot.utils.Importer()
 
         for file_path in args:
--- a/cumulusci/robotframework/pageobjects/PageObjects.py
+++ b/cumulusci/robotframework/pageobjects/PageObjects.py
@@ -1,8 +1,10 @@
-from robot.libraries.BuiltIn import BuiltIn
+from robot.api import logger
+from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
 from .baseobjects import BasePage
 import inspect
 import robot.utils
 import os
+import sys
 
 
 class PageObjects(object):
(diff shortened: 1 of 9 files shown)


cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -23,24 +25,43 @@ class PageObjects(object):
     cache = {}
 
     def __init__(self, *args):
-        BuiltIn().log("initializing PageObjects...", "DEBUG")
+        logger.info("initializing PageObjects...", "DEBUG")
```
davisagli:
```suggestion
        logger.debug("initializing PageObjects...")
```

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -23,24 +25,43 @@ class PageObjects(object):
...
+    @classmethod
+    def _reset(cls):
+        """Reset the internal data structures used to manage page objects
+
+        This is to aid testing. It probably shouldn't be used at any other time.
+        """
+        for pobj in cls.registry.values():
+            if pobj.__module__ in sys.modules:
+                del sys.modules[pobj.__module__]
+            del pobj
```
davisagli: Since there isn't anything in local scope that uses `pobj` after this, I don't think this line accomplishes anything.

cumulusci/tasks/robotframework/libdoc.py
```diff
@@ -51,31 +55,50 @@ def _validate_options(self):
...
-                # two lines:
-                # libdoc.save(self.options['output_file'], "HTML")
-                # self.logger.info("created {}".format(output_file))
+                if self.is_pageobject_library(input_file):
+                    PageObjects._reset()
+                    Importer().import_class_or_module_by_path(
+                        os.path.abspath(input_file)
+                    )
+
+                    for pobj_name in sorted(PageObjects.registry.keys()):
```
davisagli:
```suggestion
                    for pobj_name, pobj in sorted(PageObjects.registry.items()):
```

and then you don't have to do a separate lookup

davisagli: Also I got pretty confused trying to understand this since pobj_name is a tuple but sounds like it's a string

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -23,24 +25,43 @@ class PageObjects(object):
...
+                logger.warn(str(e))
         self.current_page_object = None
 
         # Start with this library at the front of the library search order;
         # that may change as page objects are loaded.
-        BuiltIn().set_library_search_order("PageObjects")
+        try:
+            BuiltIn().set_library_search_order("PageObjects")
+        except RobotNotRunningError:
+            pass
```
davisagli: This except clause and the one below aren't covered by the tests. Are they really needed?

davisagli:
@boakley I tried using this to generate documentation for the page objects included in CumulusCI, and I must be missing something.

If I add `cumulusci/robotframework/pageobjects/baseobjects.py` to the path for the robot_libdoc task, I get an empty section in the output. If I add `cumulusci/robotframework/pageobjects/PageObjects.py`, I get `robot.errors.DataError: Importing test library '/Users/dglick/Work/CumulusCI/cumulusci/robotframework/pageobjects/PageObjects.py' failed: ImportError: attempted relative import with no known parent package`

cumulusci/robotframework/pageobjects/PageObjects.py
boakley: You are correct. I must've added that before I figured out how to mock the context. Or something. I removed the except blocks and all of the tests still pass.

boakley:
> @boakley I tried using this to generate documentation for the page objects included in CumulusCI, and I must be missing something.

You shouldn't be calling libdoc on baseobjects.py (and by "shouldn't" I mean, "I never thought anybody would try to do that"). Those are base classes designed to be used by actual page object classes. Even though they aren't implemented as such, they are effectively abstract base classes.  

The reason yo ugot a blank section is that none of those classes use the `@pageobject` decorator, which is the clue that `robot_libdoc` uses to know which classes to document.  

When I add a keyword and use the decorator, I end up getting the same import error that you reported on using PageObjects.py. It looks like I need to fix the imports to handle this case, and/or make sure the script handles the error more gracefully.

> 
> If I add `cumulusci/robotframework/pageobjects/baseobjects.py` to the path for the robot_libdoc task, I get an empty section in the output.

>  If I add `cumulusci/robotframework/pageobjects/PageObjects.py`, I get `robot.errors.DataError: Importing test library '/Users/dglick/Work/CumulusCI/cumulusci/robotframework/pageobjects/PageObjects.py' failed: ImportError: attempted relative import with no known parent package`

That's a bug. I'll fix it.

davisagli:
>  Those are base classes designed to be used by actual page object classes. Even though they aren't implemented as such, they are effectively abstract base classes.

Aren't they also used non-abstractly as a fallback if a specific page object wasn't found for the (page_type, object_name) combo that was requested? That's what made me think it might work, but it makes sense that it doesn't due to the lack of the `@pageobject` decorator.

boakley:
> > Those are base classes designed to be used by actual page object classes. Even though they aren't implemented as such, they are effectively abstract base classes.
> 
> Aren't they also used non-abstractly as a fallback if a specific page object wasn't found for the (page_type, object_name) combo that was requested? That's what made me think it might work, but it makes sense that it doesn't due to the lack of the `@pageobject` decorator.

Yes, they are. I need to think a bit more about how to document these base classes. Once I solve the import error I might be able to get away with using the decorator without an object name. I'll fiddle around with it some. 

Thanks for bringing this to my attention.

boakley:
> > @boakley I tried using this to generate documentation for the page objects included in CumulusCI, and I must be missing something.
> 
> You shouldn't be calling libdoc on baseobjects.py (and by "shouldn't" I mean, "I never thought anybody would try to do that"). Those are base classes designed to be used by actual page object classes. Even though they aren't implemented as such, they are effectively abstract base classes.
> 
> The reason yo ugot a blank section is that none of those classes use the `@pageobject` decorator, which is the clue that `robot_libdoc` uses to know which classes to document.
> 
> When I add a keyword and use the decorator, I end up getting the same import error that you reported on using PageObjects.py. It looks like I need to fix the imports to handle this case, and/or make sure the script handles the error more gracefully.
> 
> > If I add `cumulusci/robotframework/pageobjects/baseobjects.py` to the path for the robot_libdoc task, I get an empty section in the output.
> 
> > If I add `cumulusci/robotframework/pageobjects/PageObjects.py`, I get `robot.errors.DataError: Importing test library '/Users/dglick/Work/CumulusCI/cumulusci/robotframework/pageobjects/PageObjects.py' failed: ImportError: attempted relative import with no known parent package`
> 
> That's a bug. I'll fix it.

I pushed changes to allow the robot_libdoc task to work with PageObjects.py. I've also created a work item to separately tackle the issue of documenting out base classes (#W-033984)

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -1,22 +1,36 @@
...
+    """Returns a list of method names for the given object
+
+    This excludes methods that begin with an underscore, and
+    also excludes the special method `get_keyword_names`.
+    """
+    names = [
+        member[0]
+        for member in inspect.getmembers(obj, inspect.isroutine)
+        if (not member[0].startswith("_")) and member[0] != "get_keyword_names"
+    ]
```
davisagli: This is definitely a minor stylistic suggestion, but I would do

    names = [name for name, member in inspect.getmembers(obj, inspect.isroutine) if not name.startswith("_") and name != "get_keyword_names"]

I try to avoid using numeric indexes when it's easy to use a more meaningful name instead.

PR #1124, opened 2019-06-19
Potential listener registration system.

prescod:
Makes it possible to register listeners for Robot like this:

>>> from cumulusci.tasks.robotframework import robotframework
>>> robotframework.Robot.register_default_listener(Foo)

You can also register on Robot instances if you prefer (but only after their options have been initialized).

Draft pull request because pytest isn't working for me right now.

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -47,11 +47,22 @@ class Robot(BaseSalesforceTask):
         if "options" not in self.options:
             self.options["options"] = {}
 
+        self.options["options"]["listener"] = self.default_listeners[:]
         if process_bool_arg(self.options.get("verbose")):
-            self.options["options"]["listener"] = KeywordLogger
+            self.options["options"]["listener"].append(KeywordLogger)
 
         if process_bool_arg(self.options.get("pdb")):
             patch_statusreporter()
+
+    default_listeners = []
+
+    @classmethod
+    def register_default_listener(cls, listener_class):
+        """Register listeners to run on classes"""
+        cls.default_listeners.append(listener_class)
+
+    def register_listener(self, listener):
+        self.options["options"]["listener"].append(listener)
 
     def _run_task(self):
         self.options["vars"].append("org:{}".format(self.org_config.name))
@@ -103,3 +114,7 @@ def patch_statusreporter():
         return orig_exit(self, exc_type, exc_val, exc_tb)
 
     StatusReporter.__exit__ = __exit__
+
+
+# Temp code to demonstrate that the features above work
+Robot.register_default_listener(KeywordLogger)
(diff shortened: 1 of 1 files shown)


davisagli:
There's nothing wrong with this code per se (aside from needing tests), but help me understand how it will fit into the big picture. How will your performance listener code get loaded in order to register itself?

prescod:
It would be code like this:

cumulusci/robotframework/CumulusCI.py would import a new module called (perhaps) RobotPerfUtils.py

At the top or bottom of that file there would be a declaration just like this one:

https://github.com/SFDO-Tooling/CumulusCI/pull/1124/files#diff-ed1b857297d75417075e518bc9444024R117

Arguably, it could be a decorator, analogous to how we handle the registration of RQ jobs.

boakley:
> It would be code like this:
> 
> cumulusci/robotframework/CumulusCI.py would import a new module called (perhaps) RobotPerfUtils.py
> 
> At the top or bottom of that file there would be a declaration just like this one:
> 
> https://github.com/SFDO-Tooling/CumulusCI/pull/1124/files#diff-ed1b857297d75417075e518bc9444024R117
> 
> Arguably, it could be a decorator, analogous to how we handle the registration of RQ jobs.

I'm still struggling to grock this. Why does some other file need to register a listener? You showed how we might register a default listener but you haven't showed how that solves your problem with capturing performance data. Where would we put the code that registers the performance listener? 

I feel like I'm one of the blind men trying to describe an elephant.

davisagli:
@prescod I don't like how that couples the CumulusCI library tightly to the robot task at import time. `cci task run robot` is not the only way to run Robot Framework test suites! And it means you can't import the library for unit tests without changing the default listeners. I think it would be better if tasks/robotframework/robotframework.py imports RobotPerfUtils and specifies its own listeners.

boakley:
> @prescod I don't like how that couples the CumulusCI library tightly to the robot task at import time. `cci task run robot` is not the only way to run Robot Framework test suites! And it means you can't import the library for unit tests without changing the default listeners. I think it would be better if tasks/robotframework/robotframework.py imports RobotPerfUtils and specifies its own listeners.

I agree with this statement. This PR seems to add unneeded complexity. If we want the perf stuff to run unconditionally, we should just hard-code the listener in the task.

prescod:
I don't agree but I also don't think it is worth arguing about.

As a purely factual matter I don't understand under what circumstances the CumulusCI library could be used without the robot task (whether run from a command line or a YML or MetaCI).

But I can just hard-code the listener into the _init_options code if that's what you're asking for.

boakley:
> But I can just hard-code the listener into the _init_options code if that's what you're asking for.

I think there's a bug in the current code related to listeners. If the user defines a listener in their cumulusci.yml file and also specifies the verbose option, we overwrite their listener option with `KeywordLogger`. 

If we're going to touch that block of code, we probably should fix that bug. I doubt many teams will specify listeners in their cumulusci.yml file, but I think it's good to leave that as a possibility.

davisagli:
I think we want something like this in `_init_options`:

    listeners = process_list_arg(self.options['options'].get('listener'), [])
    listeners.append(RobotPerfListener)
    if process_bool_arg(self.options.get("verbose")):
        listeners.append(KeywordLogger)
    self.options['options']['listener'] = listeners

boakley:
yep, something like that would work just fine.

PR #1128, opened 2019-06-25
Bulk creation API.

prescod:
I split this aspect of my work out to make it easier to review because it is mostly unrelated to performance stuff.

If you are looking at the files view of github I would start from the bottom and read the functions backwards.

api.robot then salesforce_init_objects then salesforce_collection_insert

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -510,7 +510,9 @@ class Salesforce(object):
         )
 
     def salesforce_init_objects(self, obj_name, number_to_create, **fields):
-        """Create an array of dictionaries with template-formatted arguments"""
+        """Create an array of dictionaries with template-formatted arguments.
+           Use '{number}' to represent the unique index of the row in the list of rows
+           and '{random_str} to represent a random string."""
         self.builtin.log("Inserting {} with values {}".format(obj_name, fields))
         objs = []
         for i in range(int(number_to_create)):
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -494,6 +494,8 @@ class Salesforce(object):
         return res["id"]
 
     def salesforce_collection_insert(self, objects):
+        """Inserts up to 200 records that were created with Salesforce Init Objects"""
+
         def dict_to_insertable(d):
             insertable = {"attributes": {"type": d["type"]}}
             for key, value in d.items():
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,4 @@
 [run]
 omit = */tests/*
+    cumulusci/files/*
 source=cumulusci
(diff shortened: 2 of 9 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -492,6 +493,38 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+        insertables = [dict_to_insertable(o) for o in objects]
+
+        return self.cumulusci.sf.restful(
+            "composite/sobjects",
+            method="POST",
+            json={"allOrNone": False, "records": insertables},
+        )
+
+    def salesforce_init_objects(self, obj_name, number_to_create, **fields):
+        """Create an array of dictionaries with template-formatted arguments"""
```
davisagli: Why not expand the docstring to explain which names will be formatted and show an example? This will be included in the documentation generated by the robot_libdoc task.

cumulusci/robotframework/Salesforce.py
```diff
@@ -492,6 +493,38 @@ def salesforce_insert(self, obj_name, **kwargs):
         self.store_session_record(obj_name, res["id"])
         return res["id"]
 
+    def salesforce_collection_insert(self, objects):
```
davisagli: Docstring, please

cumulusci/robotframework/Salesforce.py
```diff
@@ -492,6 +493,38 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+        def dict_to_insertable(d):
```
davisagli: There's no need for a closure here. So I'd rather this be a class method or module global to avoid defining it each time `salesforce_collection_insert` is called.

cumulusci/robotframework/Salesforce.py
prescod: c1a48814c7aec5a330829373abc517ee6f0ea1ca

cumulusci/robotframework/Salesforce.py
prescod: 1c2b6dbbf9805ba031be9d3fcd0b4849264da35b

cumulusci/robotframework/Salesforce.py
prescod: I'd call that a premature optimization, considering that the code is more or less a bulk insert which takes (...checking...) 1/3 of a second not including any networking time. I'd prefer to keep the other namespaces uncluttered.

davisagli: Okay, you got me, I was using the performance argument against something that simply isn't to my taste. Given that it has full test coverage, I can live with it.

cumulusci/robotframework/Salesforce.py
```diff
@@ -492,6 +493,42 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+        objs = []
+        for i in range(int(number_to_create)):
+            obj = {"type": obj_name}  # Object type to create
+            for name, value in fields.items():
+                if hasattr(value, "format"):  # Duck-check for if it is string-like
+                    obj[name] = value.format(
+                        number=i, random_str=String().generate_random_string()
+                    )
+                else:
+                    obj[name] = value
```
davisagli: This line is missing test coverage though.

PR #1135, opened 2019-06-28
Performance listener and Robot keywords

prescod:
Add a new performance listener.

Add new custom metrics keywords.

Tests will be added before merging.

### DIFF
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -12,7 +12,6 @@ from cumulusci.core.tasks import CURRENT_TASK
 from cumulusci.core.utils import import_global
 from cumulusci.robotframework.utils import set_pdb_trace, PerfJSONConverter
 from cumulusci.tasks.robotframework.robotframework import Robot
-from contextlib import contextmanager
 
 
 class CumulusCI(object):
@@ -48,6 +47,10 @@ class CumulusCI(object):
         logging.getLogger("requests.packages.urllib3.connectionpool").setLevel(
             logging.WARN
         )
+        if self.perf_listener:
+            # we never uninstall because the objects we are hooking are
+            # scoped to the lifetime of this task anyhow
+            self._install_perf_hook()
 
     @property
     def robot_task(self):
@@ -234,43 +237,27 @@ class CumulusCI(object):
         if hook not in self.sf.session.hooks["response"]:
             self.sf.session.hooks["response"].append(hook)
 
-    def _unregister_response_hook(self, hook):
-        if hook in self.sf.session.hooks["response"]:
-            self.sf.session.hooks["response"].remove(hook)
-
-    @contextmanager
-    def _perf_wrapper(self):
+    def _response_callback(self, response, **kwargs):
         self._last_performance_metrics = {}
-        if not self.perf_listener:
-            yield  # just ignore me if I'm not relevant.
+        if "perfmetrics" in response.headers.keys():
+            metric_str = response.headers["perfmetrics"]
+            metadata = {}
+            metadata["url"] = response.request.url
+            metadata["method"] = response.request.method
+            self._last_performance_metrics["_meta"] = metadata
+            include_raw = self.perf_listener.verbosity >= 2
+            perfjson = PerfJSONConverter(metric_str)
+            self._last_performance_metrics.update(
+                perfjson.to_dict(include_raw=include_raw)
+            )
+            self.perf_listener.report(self._last_performance_metrics)
+            # sometimes it is handy to log this stuff
+            # BuiltIn().log(perfjson.to_log_message(metadata))
 
-        def response_callback(response, **kwargs):
-            if "perfmetrics" in response.headers.keys():
-                metric_str = response.headers["perfmetrics"]
-                metadata = {}
-                metadata["url"] = response.request.url
-                metadata["method"] = response.request.method
-                self._last_performance_metrics["_meta"] = metadata
-                include_raw = self.perf_listener.verbosity >= 2
-                perfjson = PerfJSONConverter(metric_str)
-                self._last_performance_metrics.update(
-                    perfjson.to_dict(include_raw=include_raw)
-                )
-
-                # sometimes it is handy to log this stuff
-                # BuiltIn().log(perfjson.to_log_message(metadata))
-
+    def _install_perf_hook(self):
         # https://github.com/forcedotcom/idecore/blob/f107a6cb61ee38cd7f5b24fc9610893f24a33264/config/wsdls/src/main/resources/apex.wsdl#L239
         self.sf.session.headers["Sforce-Call-Options"] = "perfOption=MINIMUM"
-        self._register_response_hook(response_callback)
-        try:
-
-            yield self._last_performance_metrics
-            if self.perf_listener:
-                self.perf_listener.report(self._last_performance_metrics)
-        finally:
-            self.sf.session.headers["Sforce-Call-Options"] = ""
-            self._unregister_response_hook(response_callback)
+        self._register_response_hook(self._response_callback)
 
     def get_performance_metrics(self):
         """Performance metrics keyword: Get the last recorded performance metrics"""
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -3,6 +3,7 @@ import logging
 from robot.api import logger
 from robot.libraries.BuiltIn import BuiltIn
 from simple_salesforce import Salesforce
+from requests import Session
 
 from cumulusci.cli.config import CliRuntime
 from cumulusci.core.config import TaskConfig
@@ -174,14 +175,18 @@ class CumulusCI(object):
     def _init_api(self, base_url=None):
         api_version = self.project_config.project__package__api_version
 
-        rv = Salesforce(
+        session = Session()
+        session.hooks = {"response": []}
+
+        sf = Salesforce(
             instance=self.org.instance_url.replace("https://", ""),
             session_id=self.org.access_token,
             version=api_version,
+            session=session,
         )
         if base_url is not None:
-            rv.base_url += base_url
-        return rv
+            sf.base_url += base_url
+        return sf
 
     def _init_task(self, class_path, options, task_config):
         task_class = import_global(class_path)
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -10,8 +10,9 @@ from cumulusci.core.config import TaskConfig
 from cumulusci.core.exceptions import TaskOptionsError
 from cumulusci.core.tasks import CURRENT_TASK
 from cumulusci.core.utils import import_global
-from cumulusci.robotframework.utils import set_pdb_trace
+from cumulusci.robotframework.utils import set_pdb_trace, perfJSON2Dict
 from cumulusci.tasks.robotframework.robotframework import Robot
+from contextlib import contextmanager
 
 
 class CumulusCI(object):
@@ -226,3 +227,65 @@ class CumulusCI(object):
     def debug(self):
         """Pauses execution and enters the Python debugger."""
         set_pdb_trace()
+
+    def _register_response_hook(self, hook):
+        if hook not in self.sf.session.hooks["response"]:
+            self.sf.session.hooks["response"].append(hook)
+
+    def _unregister_response_hook(self, hook):
+        if hook in self.sf.session.hooks["response"]:
+            self.sf.session.hooks["response"].remove(hook)
+
+    @contextmanager
+    def _perf_wrapper(self):
+        self._last_performance_metrics = {}
+        if not self.perf_listener:
+            yield  # just ignore me if I'm not relevant.
+
+        def response_callback(response, **kwargs):
+            if "perfmetrics" in response.headers.keys():
+                metric_str = response.headers["perfmetrics"]
+                metadata = {}
+                metadata["url"] = response.request.url
+                metadata["method"] = response.request.method
+                self._last_performance_metrics["_meta"] = metadata
+                include_raw = self.perf_listener.verbosity >= 2
+                self._last_performance_metrics.update(
+                    perfJSON2Dict(metric_str, include_raw=include_raw)
+                )
+
+                # sometimes it is handy to log this stuff
+                # BuiltIn().log(perfJSON2LogMessage())
+
+        # https://github.com/forcedotcom/idecore/blob/f107a6cb61ee38cd7f5b24fc9610893f24a33264/config/wsdls/src/main/resources/apex.wsdl#L239
+        self.sf.session.headers["Sforce-Call-Options"] = "perfOption=MINIMUM"
+        self._register_response_hook(response_callback)
+        try:
+
+            yield self._last_performance_metrics
+            if self.perf_listener:
+                self.perf_listener.report(self._last_performance_metrics)
+        finally:
+            self.sf.session.headers["Sforce-Call-Options"] = ""
+            self._unregister_response_hook(response_callback)
+
+    def get_performance_metrics(self):
+        """Performance metrics keyword: Get the last recorded performance metrics"""
+        return self._last_performance_metrics
+
+    def create_duration_metric(self, name):
+        """Custom metric keyword: Start measuring the time it takes to do something: EXPERIMENTAL"""
+        self.perf_listener.create_duration_metric(name)
+
+    def end_duration_metric(self, name):
+        """Custom metric keyword: Finish measuring a duration: EXPERIMENTAL"""
+        self.perf_listener.end_duration_metric(name)
+
+    def create_aggregate_metric(self, name, aggregation):
+        """Custom metric keyword: Start averaging/summing/... something: EXPERIMENTAL
+            aggregation should be "average", or "sum"  """
+        self.perf_listener.create_aggregate_metric(name, aggregation)
+
+    def store_metric_value(self, name, value):
+        """Custom metric keyword: Add this number to the average or sum. EXPERIMENTAL"""
+        self.perf_listener.store_metric_value(name, value)
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -252,13 +252,13 @@ class CumulusCI(object):
                 metadata["method"] = response.request.method
                 self._last_performance_metrics["_meta"] = metadata
                 include_raw = self.perf_listener.verbosity >= 2
-                converter = PerfJSONConverter(metric_str)
+                perfjson = PerfJSONConverter(metric_str)
                 self._last_performance_metrics.update(
-                    converter.perfJSON2Dict(include_raw=include_raw)
+                    perfjson.to_dict(include_raw=include_raw)
                 )
 
                 # sometimes it is handy to log this stuff
-                # BuiltIn().log(perfJSON2LogMessage(metadata))
+                # BuiltIn().log(perfjson.to_log_message(metadata))
 
         # https://github.com/forcedotcom/idecore/blob/f107a6cb61ee38cd7f5b24fc9610893f24a33264/config/wsdls/src/main/resources/apex.wsdl#L239
         self.sf.session.headers["Sforce-Call-Options"] = "perfOption=MINIMUM"
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -10,7 +10,7 @@ from cumulusci.core.config import TaskConfig
 from cumulusci.core.exceptions import TaskOptionsError
 from cumulusci.core.tasks import CURRENT_TASK
 from cumulusci.core.utils import import_global
-from cumulusci.robotframework.utils import set_pdb_trace, perfJSON2Dict
+from cumulusci.robotframework.utils import set_pdb_trace, PerfJSONConverter
 from cumulusci.tasks.robotframework.robotframework import Robot
 from contextlib import contextmanager
 
@@ -252,12 +252,13 @@ class CumulusCI(object):
                 metadata["method"] = response.request.method
                 self._last_performance_metrics["_meta"] = metadata
                 include_raw = self.perf_listener.verbosity >= 2
+                converter = PerfJSONConverter(metric_str)
                 self._last_performance_metrics.update(
-                    perfJSON2Dict(metric_str, include_raw=include_raw)
+                    converter.perfJSON2Dict(include_raw=include_raw)
                 )
 
                 # sometimes it is handy to log this stuff
-                # BuiltIn().log(perfJSON2LogMessage())
+                # BuiltIn().log(perfJSON2LogMessage(metadata))
 
         # https://github.com/forcedotcom/idecore/blob/f107a6cb61ee38cd7f5b24fc9610893f24a33264/config/wsdls/src/main/resources/apex.wsdl#L239
         self.sf.session.headers["Sforce-Call-Options"] = "perfOption=MINIMUM"
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -53,6 +53,8 @@ class CumulusCI(object):
     def robot_task(self):
         if CURRENT_TASK.stack and isinstance(CURRENT_TASK.stack[0], Robot):
             return CURRENT_TASK.stack[0]
+        else:
+            return None
 
     @property
     def project_config(self):
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -67,6 +67,11 @@ class CumulusCI(object):
         return self.project_config.keychain
 
     @property
+    def perf_listener(self):
+        if self.robot_task:
+            return self.robot_task.robot_perf_listener
+
+    @property
     def org(self):
         if self._org is None:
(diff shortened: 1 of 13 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -47,6 +51,15 @@ def _init_options(self, kwargs):
         if "options" not in self.options:
             self.options["options"] = {}
 
+        listeners = process_list_arg(self.options["options"].get("listeners")) or []
+
+        verbosity = int(self.options.get("perf_trace_level", 0))
+        self.robot_perf_listener = RobotPerfListener(self, verbosity)
+        listeners.append(self.robot_perf_listener)
+        if process_bool_arg(self.options.get("verbose")):
+            listeners.append(KeywordLogger)
```
boakley: I haven't run the code so maybe it just magically works, but it seems a bit odd to have a list that includes an instance `RobotPerfListener(self, verbosity)` and a class `KeywordLogger`.

cumulusci/tasks/robotframework/robotperflistener.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def create_duration_metric(self, name):
+        self._current_test.custom_metrics[name] = DurationAggregator()
+
+    def end_duration_metric(self, name):
+        metric_obj = self._current_test.custom_metrics[name]
+        metric_obj.finish()
+
+
+def summarize(test_metrics):
```
boakley: Is there a reason why this is a standalone function rather than a method on `TestMetrics`?

cumulusci/tasks/robotframework/robotperflistener.py
```diff
@@ -0,0 +1,142 @@
...
+    def close(self):
+        filename = os.path.join(self.outputdir, "perf.json")
+        with open(filename, "w") as f:
+            json.dump(self._metrics, f, indent=2)
+
+        # cleanup so we won't be called again and also remove GC loop
+        self.task.robot_perf_listener = None
+        self.task = None
+
+    def report(self, metrics):
```
boakley: I spent several minutes trying to figure out what was going on here. I was thinking this was a typo and meant to be the `report_file` method described in the API. I eventually realized it's an additional method called outside of robot. 

You might want to consider somehow segregating this method, or renaming it (perhaps with _ prefix) and the other non-listener-api methods so it's a bit more clear which are api methods and which are not. For example, simply moving it after `output_file` and adding a comment saying something like "these methods are not used by the listener interface, but are used by other parts of our code" or something similar would make the code a bit easier to grok. 

Another option might be to move the non-listener methods into a class mixin so that it's clear you are adding additional functions to a standard interface.

cumulusci/robotframework/CumulusCI.py
```diff
@@ -47,12 +49,17 @@ def __init__(self, org_name=None):
             logging.WARN
         )
 
+    @property
```
boakley: I like that you added this property 👍 

Though, personally I'd like to see an explicit `return None` just so that it's clear that's what you intend to happen and not that you forgot to handle that case.

cumulusci/robotframework/Salesforce.py
```diff
@@ -475,23 +477,62 @@ def salesforce_delete(self, obj_name, obj_id):
...
-        res = obj_class.create(kwargs)
-        self.store_session_record(obj_name, res["id"])
+
+        with self.cumulusci._perf_wrapper():
+            res = obj_class.create(kwargs)
+            self.store_session_record(obj_name, res["id"])
+
         return res["id"]
 
+    def salesforce_collection_insert(self, objects):
```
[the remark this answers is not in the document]
prescod: My memory must be going bad. I distinctly remember writing that docstring and the one below at some point. It must have got git stashed somewhere!

cumulusci/robotframework/utils.py
```diff
@@ -199,3 +200,34 @@ def run_with_retry(self, *args, **kwargs):
     else:
         # Decorator was used without arguments
         return decorate(target)
+
+
+def perfJSON2Dict(jsondata, include_raw):
```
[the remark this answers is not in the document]
prescod: Strictly speaking there is no reason to instantiate an object because there is no state, but sure I could wrap them in an object for a namespace...

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -32,6 +33,9 @@ class Robot(BaseSalesforceTask):
         "name": {"description": "Sets the name of the top level test suite"},
         "pdb": {"description": "If true, run the Python debugger when tests fail."},
         "verbose": {"description": "If true, log each keyword as it runs."},
+        "perf_trace_level": {
```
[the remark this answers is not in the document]
prescod: I thought about it but normal, verbose, really_verbose seemed kind of weird.

Numbers establish the relationship between the values (2 is noisier than 1 which is noisier than 0) which is harder to convey with just words.

It is a convention I've seen elsewhere:

https://buck.build/command/common_parameters.html

A lot of tools also use -v -vv -vvv which is basically the same.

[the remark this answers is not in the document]
prescod: I'm not religious about it, I could try "normal", "API", "everything" but I actually think it is harder to remember 3 different words than to remember "higher numbers give more data and I should experiment with them until I get the information I'm looking for" (which is how one typically uses these traces).

cumulusci/tasks/robotframework/robotframework.py
prescod: Yeah it is a bit weird and it through me off at first but if you look at KeywordLogger you'll note that it doesn't have "self" params. So the class is used "as" just a namespace.

cumulusci/tasks/robotframework/robotperflistener.py
prescod: I guess only historical reasons. It predates the class. But I could make it a method. I don't care much one way or the other.

cumulusci/tasks/robotframework/robotperflistener.py
prescod: Okay I'll look at something like that.

cumulusci/robotframework/Salesforce.py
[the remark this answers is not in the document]
prescod: Oh, those docstrings are on the branch called Bulk Creation API:

https://github.com/SFDO-Tooling/CumulusCI/pull/1128

I did that one first to theoretically make reviewing easier for you and David but not all of the improvements got merged back.

cumulusci/robotframework/CumulusCI.py
prescod: I literally had the return None in my code for a week and then deleted it just before the checkin because I guessed that someone was going to complain that it was redundant and unnecessarily verbose...

That's the problem with trying to mind-read!

cumulusci/robotframework/Salesforce.py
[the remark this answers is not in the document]
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1135/commits/931368601909654a1d5686bd80bd057a472b4fcd

cumulusci/robotframework/Salesforce.py
```diff
@@ -475,23 +477,62 @@ def salesforce_delete(self, obj_name, obj_id):
...
+        with self.cumulusci._perf_wrapper():
+            res = self.cumulusci.sf.restful(
+                "composite/sobjects",
+                method="POST",
+                json={"allOrNone": False, "records": insertables},
+            )
+            return res
+
+    def salesforce_init_objects(self, obj_name, number, **kwargs):
+        """Create an array of dictionaries with formatted arguments"""
```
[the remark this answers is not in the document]
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1135/commits/931368601909654a1d5686bd80bd057a472b4fcd

cumulusci/robotframework/CumulusCI.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/commit/7d0c5e80800f53ac79ea9b4d4f76d12406df123f

cumulusci/robotframework/utils.py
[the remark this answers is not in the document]
prescod: f468c120ca1d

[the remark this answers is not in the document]
boakley: The state would be the actual performance data, wouldn't it? Instead of having a `jsondata` object that you pass in, the instance would be the `jsondata` object. In other words, instead of `perfJSON2Dict(jsondata, True)` you would do `jsondata.to_dict(True)`.

cumulusci/tasks/robotframework/robotperflistener.py
prescod: 167a3215527ac69b89236be0445247b434bcfce5

cumulusci/tasks/robotframework/robotperflistener.py
prescod: 167a3215527ac69b89236be0445247b434bcfce5

cumulusci/tasks/robotframework/robotframework.py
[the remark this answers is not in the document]
boakley: It's not clear that you could have more than two values. I've seen the convention elsewhere, and it always sucks if there's only two levels. If there's more than two levels then the numbers start to make sense.

cumulusci/tasks/robotframework/robotframework.py
boakley: I don't quite follow what you're saying. Robot will instantiate it the listener when it starts up.  I guess my point was, we might want to either pass in the name of each listener or the instance, but not one of each.

cumulusci/robotframework/CumulusCI.py
boakley: > I literally had the return None in my code for a week and then deleted it just before the checkin because I guessed that someone was going to complain that it was redundant and unnecessarily verbose...

Hah! I do that too sometimes. I write code, I'm happy with it, I use it, and then at the last second before a commit I think "ahh, I should change this".

cumulusci/tasks/robotframework/robotframework.py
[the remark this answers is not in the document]
prescod: Okay you're right. The real problem is that the default of "0" is not documented.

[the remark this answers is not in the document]
prescod: ea5d48408f9d7e970e1e4147367ac9a3e0aed1c0

cumulusci/tasks/robotframework/robotframework.py
prescod: "Robot will instantiate it the listener when it starts up."

It doesn't. It just uses the class as-is.

But sure, I can change the old code to match the new if you prefer.

412352a9d49ebf4b47e6ab2da60b874d1ee3e989

boakley: I stand corrected. That's interesting. I've never seen a listener used this way. The only way I've ever used them was as classes that get instantiated.

cumulusci/robotframework/utils.py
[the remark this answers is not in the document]
prescod: 0aba6230f511ce49a0d7e0d67208c1d73f1dc63f

cumulusci/tasks/robotframework/robotperflistener.py
```diff
@@ -0,0 +1,149 @@
...
+        self.api_metrics = []
+        self.custom_metrics = {}
+
+    def summarize(self):
+        collections = {}
+        for metricset in self.api_metrics:
+            assert isinstance(metricset, dict), metricset
+            for name, value in metricset.items():
+                if not name.startswith("_"):
+                    collections.setdefault(name, []).append(float(value))
```
davisagli: You could initialize collections as a defaultdict(list) and then this would simply be `collections[name].append(float(value))`

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -85,7 +96,7 @@ def _run_task(self):
 class KeywordLogger(object):
     ROBOT_LISTENER_API_VERSION = 2
 
-    def start_keyword(name, attrs):
+    def start_keyword(self, name, attrs):
```
davisagli: Hmm, oops. But why didn't this break before?

cumulusci/tasks/robotframework/robotperflistener.py
```diff
@@ -0,0 +1,149 @@
...
+    @property
+    def value(self):
+        return self.duration
+
+
+Aggregations = {
+    "duration": DurationAggregator,
+    "average": AverageAggregator,
+    "sum": TotalAggregator,
+}
```
davisagli: I like how this is structured but it's usually discouraged to name a Python identifier the same thing as as a builtin (reduce). I also always have found the programming term "reduce" to be pretty jargony. How about something like `aggregate` or `process`

cumulusci/tasks/robotframework/robotperflistener.py
```diff
@@ -0,0 +1,149 @@
...
+class AverageAggregator:
+    """Takes a stream of numbers delivered to the "reduce" function and averages them"""
+
+    def __init__(self):
+        self.value = 0
+        self.count = 0
+
+    def reduce(self, value):
+        self.value = (self.value * self.count + float(value)) / (self.count + 1)
+        self.count += 1
```
davisagli: Might make more sense to store a running sum, and then implement value as a property which does the division on demand?

cumulusci/robotframework/Salesforce.py
```diff
@@ -239,7 +240,8 @@ def get_record_type_id(self, obj_type, developer_name):
         soql = "SELECT Id FROM RecordType WHERE SObjectType='{}' and DeveloperName='{}'".format(
             obj_type, developer_name
         )
-        res = self.cumulusci.sf.query_all(soql)
+        with self.cumulusci._perf_wrapper():
+            res = self.cumulusci.sf.query_all(soql)
```
davisagli: I thought the goal was to have a listener that could collect performance metrics automatically without needing to remember to activate it for each call to Salesforce. Are there simple_salesforce API calls that we don't want performance metrics for?

cumulusci/robotframework/utils.py
```diff
@@ -199,3 +200,35 @@ def run_with_retry(self, *args, **kwargs):
...
+            metric["metrics"] + "-" + metricType: metric[metricType]
+            for metricType in ("totalTime", "totalCalls")
+            for metric in self.data["summary"]
+        }
+        if include_raw:
+            rc["_raw"] = self.data
+        return rc
+
+    def to_csv(self):
+        """Helper function to generate CSV-like data from performance metrics JSON"""
```
davisagli: But not actually CSV? What's the use case for this output?

cumulusci/tasks/robotframework/robotperflistener.py
prescod: True, and I did think about that. But to me they are potato/potato and the other way requires an extra import.

But then from my point of view they are both equally idiomatic and perhaps newer Python programmers consider setdefault archaic for this use-case.

cumulusci/tasks/robotframework/robotframework.py
prescod: It wasn't broken, actually, just confusing and non-idiomatic. 

You passed a class and it treated the class as an object. The self pointer was never needed and the class was never instantiated into an object.

I left it the way it was at first but it confused Bryan. And now it confused you (the author). And it had confused me (briefly) at first too. So now it is more idiomatic Python even though the old way worked okay.

So do you think that it was just an accident that it worked? Did you intend for it to be used as a traditional class instead of just a namespace with a function in it?

cumulusci/tasks/robotframework/robotperflistener.py
prescod: Sure, Process is fine. I did struggle with that name.

prescod: 56f2557da12f9fefd02d4421308ab58c022e3651

cumulusci/tasks/robotframework/robotperflistener.py
prescod: Sure.

335f58a8eb4aea73c8d6d7b2261e8ddfa06a375e

cumulusci/robotframework/Salesforce.py
prescod: Well that was the goal from the point of view of the Robot programmer. It wasn't explicitly a goal for the Python programmer.

I inherited the hooking bits of the code and didn't really question it. I guess I could hook sf.session in listener.start_suite() and then unhook in close() or something like that? Knowing when to unhook is clearer now than it would have been at the start of the project when I didn't have a Listener object around listening for that event.

cumulusci/robotframework/utils.py
prescod: Well at the risk of drawing your eyes to some other questionable code I'll link here:

https://github.com/SFDO-Tooling/CumulusCI/pull/1135/files#diff-d58bc12615af74abeefd97ffb7ccf87fR261

It may be used as a Perf debugging tool and it MAY end up being the way I get the Perf data into MetaCI. Whether to have MetaCI read CSV-like lines out of output.xml OR try to weave together information from output.xml and perf.json is an open question right now. There are trade-offs either way.

In the April version of this prototype I put the perf data right into output.xml.

Regardless of how MetaCI reads it, I do think that having the JSON option is good, especially for the more detailed output modes.

cumulusci/robotframework/Salesforce.py
prescod: Okay, fixed in the last two checkins: 3bd6399 and 
0e47c29

prescod:
@boakley @davisagli When you are back from your holidays, one thing you might want to look at with this PR is that I've added some extra command lines to CircleCI to give them some testing while we are otherwise testing.

https://github.com/SFDO-Tooling/CumulusCI/pull/1135/files#diff-1d37e48f9ceff6d8030570cd36286a61

cumulusci/robotframework/Salesforce.py
```diff
@@ -488,10 +489,58 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+
+            Which would generate:
+
+                | [{'FirstName': '"User 0"', 'LastName': '"u4PNlGUD"', 'type': 'Contact'},
+                |  {'FirstName': '"User 1"', 'LastName': '"Kyd4PC5x"', 'type': 'Contact'},
+                |  {'FirstName': '"User 2"', 'LastName': '"n4EGs4Vp"', 'type': 'Contact'}]
+           """
+        objs = []
+        for i in range(int(number_to_create)):
+            obj = {"type": obj_name}  # Object type to create
```
cdcarter: Opportunity has a standard field named `Type` so there may be some issues if someone tries to set that. (hi Paul! 👋)

prescod: Welcome back!

Good point. Custom objects could have a "type" as well. Fixed here:

a18b92f4fa36997e31f490f5a940cdf19df040e7

Your feedback is always welcome!

cumulusci/robotframework/Salesforce.py
```diff
@@ -488,10 +490,93 @@ def salesforce_insert(self, obj_name, **kwargs):
...
         self.store_session_record(obj_name, res["id"])
+
         return res["id"]
 
+    def salesforce_init_object(self, obj_name, **fields):
+        obj = {"attributes": {"type": obj_name}}  # Object type to create
+        obj.update(fields)
+        return obj
+
+    def salesforce_init_objects(self, obj_name, number_to_create, **fields):
```
boakley: See the quip doc [Creating test data in robot tests](https://salesforce.quip.com/jCyVA1a0DT9i) for my comments about this keyword as a whole.

cumulusci/robotframework/Salesforce.py
```diff
@@ -488,10 +490,93 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+            formatted_fields = {
+                name: format_str(value, i) for name, value in fields.items()
+            }
+            newobj = self.salesforce_init_object(obj_name, **formatted_fields)
+            objs.append(newobj)
+
+        return objs
+
+    def salesforce_collection_insert(self, objects):
+        """Inserts up to 200 records that were created with Salesforce Init Objects"""
```
boakley: This says "up to 200 records" but that doesn't seem to be enforced anywhere. If there's a hard limit on the number of objects passed in, that should be checked for and an appropriate error thrown.

cumulusci/robotframework/utils.py
boakley: I think if this function returns something csv-like but not actually csv, maybe a name change is in order. Or perhaps even better, why not use a csv library to generate real csv?

cumulusci/robotframework/utils.py
```diff
@@ -199,3 +200,35 @@ def run_with_retry(self, *args, **kwargs):
...
+
+        def row(values):
+            return ",".join(values) + "\n"
```
boakley: is there a danger that `values` itself may have newlines or commas in it? This is where using a csv library would help.

cumulusci/tasks/robotframework/robotframework.py
davisagli: Just an accident, I think. I may have copied code from some example of RF listeners.

cumulusci/robotframework/Salesforce.py
davisagli: Clever solution!

cumulusci/robotframework/Salesforce.py
```diff
@@ -488,10 +489,58 @@ def salesforce_insert(self, obj_name, **kwargs):
...
         self.builtin.log("Inserting {} with values {}".format(obj_name, kwargs))
         obj_class = getattr(self.cumulusci.sf, obj_name)
+
         res = obj_class.create(kwargs)
         self.store_session_record(obj_name, res["id"])
+
         return res["id"]
 
+    def salesforce_collection_insert(self, objects):
+        """Inserts up to 200 records that were created with Salesforce Init Objects"""
```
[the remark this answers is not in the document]
davisagli: It's a limit of the "SObjects Collection" form of calling the REST API: https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections.htm

cumulusci/robotframework/CumulusCI.py
```diff
@@ -211,3 +232,50 @@ def _run_task(self, task_class, task_config):
...
         """Pauses execution and enters the Python debugger."""
         set_pdb_trace()
+
+    def _register_response_hook(self, hook):
+        if hook not in self.sf.session.hooks["response"]:
+            self.sf.session.hooks["response"].append(hook)
+
+    def _response_callback(self, response, **kwargs):
+        self._last_performance_metrics = {}
+        if "perfmetrics" in response.headers.keys():
```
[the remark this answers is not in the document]
prescod: This header comes all of the way from the server.

https://i.imgur.com/FbKCvqV.png

Which is part of:

https://cdcarter.github.io/sfdx/2018/05/22/charles-proxy-sfdx

cumulusci/robotframework/Salesforce.py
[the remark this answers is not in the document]
prescod: 5f41df24afe7378c446b914a84efe1747ee93087

cumulusci/robotframework/Salesforce.py
```diff
@@ -488,10 +489,58 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+            json={"allOrNone": False, "records": insertables},
+        )
+
+    def salesforce_init_objects(self, obj_name, number_to_create, **fields):
+        """Create an array of dictionaries with template-formatted arguments appropriate for a Collection Insert.
+            Use ``{number}`` to represent the unique index of the row in the list of rows and ``{random_str}`` to represent a random string.
+            For example:
+
+                | @{objects} =  Salesforce Init Objects  Contact  3
+                | ...  FirstName="User {number}"
```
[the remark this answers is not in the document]
prescod: de093df0

cumulusci/robotframework/utils.py
prescod: No, its just a list of numbers.

And the sense in which it is CSV-like is that it has a header line in Python comment syntax (because there is no CSV comment syntax) for human readability.

That's what this line does:

        res += "#" + row(metrics[0].keys())

If you strip off that line the rest of the lines can be interpreted as CSV safely.

cumulusci/robotframework/CumulusCI.py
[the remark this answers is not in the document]
boakley: Ok, now I see. You're saying "if the server sent us perfmetrics, then...". Makes sense.

cumulusci/robotframework/Salesforce.py
boakley: I think if this 200 is a hard limit, and this is a user-facing keyword, we should probably be prepared to catch errors if we exceed that limit and return a useful error rather than whatever error the server throws.

cumulusci/robotframework/utils.py
boakley: It's been years since I've done much with csv files, but IIRC it's a fairly well established convention that the first row can (should?) be column names. With or without the leading `#`, the line should probably be able to be parsed as csv so that automated tools won't crash and burn trying to parse the first row. 

I'm not going to get overly passionate about this, but I think separating the keys by commas would be worth doing. If there's a chance some of the keys have commas in them, they would need to be properly escaped.

PR #1137, opened 2019-06-28
Fix coverage checking and document how to do it.

prescod:
Coverage checking did not work for me so I fixed it.

And I added documentation about how to do it.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,3 +1,4 @@
 [run]
 omit = */tests/*
+    cumulusci/files/*
 source=cumulusci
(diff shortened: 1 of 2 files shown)


.coveragerc
```diff
@@ -1,3 +1,4 @@
 [run]
 omit = */tests/*
+    cumulusci/files/*
```
jstvz: Now I remember why I don't like `.ini` files.

prescod:
I have no idea what AppVeyor is complaining about. Tools complaining about checkins intended to make Tools happy...

PR #1138, opened 2019-07-01
Add cci org shell command

davidmreed:
# Critical Changes

# Changes

 - Add the `cci org shell` command, which opens a Python shell pre-populated with a `simple_salesforce` session on the selected org (as `sf`).

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -38,19 +38,18 @@ from cumulusci.core.exceptions import CumulusCIUsageError
 from cumulusci.core.exceptions import OrgNotFound
 from cumulusci.core.exceptions import ScratchOrgException
 from cumulusci.core.exceptions import ServiceNotConfigured
-from cumulusci.core.exceptions import ServiceNotValid
 from cumulusci.core.exceptions import FlowNotFoundError
 
 from cumulusci.core.utils import import_global
 from cumulusci.cli.config import CliRuntime
 from cumulusci.cli.config import get_installed_version
+from cumulusci.salesforce_api.utils import get_simple_salesforce_connection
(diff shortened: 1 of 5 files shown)


davidmreed:
I prefer (2). In `cumulusci.utils`, would you say?

davisagli:
@davidmreed how about a new utils.py or rest.py module in cumulusci.salesforce_api. I always feel bad adding more things to cumulusci.utils.

cumulusci/cli/cci.py
```diff
@@ -1009,6 +1012,27 @@ def org_scratch_delete(config, org_name):
...
+@pass_config
+def org_shell(config, org_name):
+    org_name, org_config = config.get_org(org_name)
+    org_config.refresh_oauth_token(config.keychain)
+
+    sf = get_simple_salesforce_connection(config.project_config, org_config)
+
+    code.interact(
+        banner="Use `sf` to access org `{}` via simple_salesforce".format(org_name),
+        local=dict(globals(), **locals()),
```
davisagli: Why include the `cci` module's globals here?

Should we include a direct reference to `project_config` too?

davidmreed: I assumed (without a good reason!) that there was a design decision behind doing so in the `shell` command that would also apply here - but if that command does so more with an eye towards facilitating CCI debugging, there's no need to do so here, and exposing just `sf` and `project_config` (perhaps also a `Salesforce_Bulk` instance) would be enough.

davisagli: I'd suggest exposing org_config too.

davidmreed: Done!

PR #1143, opened 2019-07-15
Add Create Community task

davidmreed:
# Critical Changes

# Changes

 - Added the `CreateCommunity` task allowing for API-based Community creation.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/CreateCommunity.py
+++ b/cumulusci/tasks/salesforce/CreateCommunity.py
@@ -50,7 +50,7 @@ class CreateCommunity(BaseSalesforceApiTask):
         # Let's simulate that without actually using a browser.
         self.logger.info("Preparing org for Communities")
         s = requests.Session()
-        s.get(self.org_config.start_url)
+        s.get(self.org_config.start_url).raise_for_status()
         r = s.get(
             "{}/sites/servlet.SitePrerequisiteServlet".format(
                 self.org_config.instance_url
--- a/cumulusci/tasks/salesforce/CreateCommunity.py
+++ b/cumulusci/tasks/salesforce/CreateCommunity.py
@@ -7,6 +7,11 @@ from cumulusci.core.exceptions import SalesforceException
 
 class CreateCommunity(BaseSalesforceApiTask):
     api_version = "46.0"
+    task_docs = """
+    Create a Salesforce Community via the Connect API.
+    Specify the `template` "VF Template" for Visualforce Tabs community,
+    or the name for a specific desired template
+    """
     task_options = {
         "template": {
             "description": "Name of the template for the community.",
@@ -45,11 +50,7 @@ class CreateCommunity(BaseSalesforceApiTask):
         # Let's simulate that without actually using a browser.
         self.logger.info("Preparing org for Communities")
         s = requests.Session()
-        s.get(
-            "{}/secur/frontdoor.jsp?sid={}".format(
-                self.org_config.instance_url, self.org_config.access_token
-            )
-        )
+        s.get(self.org_config.start_url)
         r = s.get(
             "{}/sites/servlet.SitePrerequisiteServlet".format(
                 self.org_config.instance_url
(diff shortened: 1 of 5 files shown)


cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -0,0 +1,83 @@
+import json
+import requests
+from datetime import datetime
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.core.exceptions import SalesforceException
+
+
+class CreateCommunity(BaseSalesforceApiTask):
```
davisagli: Please set `task_docs` and run `cci task doc > docs/tasks.rst` to update the generated task docs.

cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -0,0 +1,83 @@
...
+        # Before we can create a Community, we have to click through the "New Community"
+        # button in the All Communities setup page. (This does some unknown behind-the-scenes setup).
+        # Let's simulate that without actually using a browser.
+        self.logger.info("Preparing org for Communities")
+        s = requests.Session()
+        s.get(
+            "{}/secur/frontdoor.jsp?sid={}".format(
+                self.org_config.instance_url, self.org_config.access_token
+            )
+        )
```
davisagli: Use `self.org_config.start_url` to get this URL without duplicating the logic.

Nice that it's possible to do this without resorting to Robot Framework!

cumulusci/tasks/salesforce/tests/test_CreateCommunity.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class test_CreateCommunity(unittest.TestCase):
+    @responses.activate
+    def test_creates_community(self):
+        cc_task = create_task(CreateCommunity, task_options)
+        cc_task.sf = mock.Mock()
+        cc_task.sf.restful.return_value = {
+            "communities": [{"name": "Test Community", "id": "000000000000000"}]
+        }
```
davisagli: I generally prefer to use `responses` to provide mock responses, rather than mocking the library call which makes the request. That way if the library ever changes what request it's producing or how it processes the response, our tests will let us know.

Open to pushback if that's hard in this particular case for some reason.

cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -0,0 +1,84 @@
...
+            "templateName": self.options["template"],
+            "urlPathPrefix": self.options["url_path_prefix"],
+        }
+
+        # Before we can create a Community, we have to click through the "New Community"
+        # button in the All Communities setup page. (This does some unknown behind-the-scenes setup).
+        # Let's simulate that without actually using a browser.
+        self.logger.info("Preparing org for Communities")
+        s = requests.Session()
+        s.get(self.org_config.start_url)
```
davisagli: We should probably get the response here and call response.raise_for_status() so that we'll error out if authentication fails.

cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -0,0 +1,84 @@
...
+        r = s.get(
+            "{}/sites/servlet.SitePrerequisiteServlet".format(
+                self.org_config.instance_url
+            )
+        )
+        if r.status_code != 200:
+            raise SalesforceException("Unable to prepare org for Communities")
+
+        self.logger.info("Sending request to create Community")
+        self.sf.restful("connect/communities", method="POST", data=json.dumps(payload))
```
davisagli: If this returns a non-Ok response we should report it and abort.

davidmreed: `simple_salesforce` will invoke its exception handlers if this call fails with an HTTP status. That particular endpoint doesn't return anything but a plain-text message saying that it's in progress - the failures seem to all be asynchronous.

cumulusci/tasks/salesforce/CreateCommunity.py
davidmreed: 👍

cumulusci/tasks/salesforce/CreateCommunity.py
davisagli: Ah right, I forgot simple_salesforce takes care of that for us. Thanks for checking.

PR #1144, opened 2019-07-22
Add Bulk Mapping Generator

davidmreed:
# Critical Changes

# Changes

 - Add task to generate a Bulk Data mapping file for a package.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata.py
+++ b/cumulusci/tasks/bulkdata.py
@@ -11,7 +11,6 @@ import os
 import time
 import tempfile
 import xml.etree.ElementTree as ET
-import yaml
 
 from salesforce_bulk.util import IteratorBytesIO
 from sqlalchemy.ext.automap import automap_base
@@ -31,7 +30,12 @@ from sqlalchemy import event
 import requests
 import unicodecsv
 
-from cumulusci.core.utils import process_bool_arg, ordered_yaml_load
+from cumulusci.core.utils import (
+    process_bool_arg,
+    process_list_arg,
+    ordered_yaml_load,
+    ordered_yaml_dump,
+)
 from cumulusci.core.exceptions import BulkDataException
 from cumulusci.core.exceptions import TaskOptionsError
 from cumulusci.tasks.salesforce import BaseSalesforceApiTask
@@ -146,11 +150,7 @@ class DeleteData(BaseSalesforceApiTask, BulkJobTaskMixin):
         super(DeleteData, self)._init_options(kwargs)
 
         # Split and trim objects string into a list if not already a list
-        if not isinstance(self.options["objects"], list):
-            self.options["objects"] = [
-                obj.strip() for obj in self.options["objects"].split(",")
-            ]
-
+        self.options["objects"] = process_list_arg(self.options["objects"])
         self.options["hardDelete"] = process_bool_arg(self.options.get("hardDelete"))
 
     def _run_task(self):
@@ -800,21 +800,6 @@ class ExtractData(BulkJobTaskMixin, BaseSalesforceApiTask):
                 f.write(line + "\n")
 
 
-def represent_ordereddict(dumper, data):
-    value = []
-
-    for item_key, item_value in data.items():
-        node_key = dumper.represent_data(item_key)
-        node_value = dumper.represent_data(item_value)
-
-        value.append((node_key, node_value))
-
-    return yaml.nodes.MappingNode(u"tag:yaml.org,2002:map", value)
-
-
-yaml.SafeDumper.add_representer(OrderedDict, represent_ordereddict)
-
-
 class GenerateMapping(BaseSalesforceApiTask):
     task_docs = """
     Generate a mapping file for use with the `extract_dataset` and `load_dataset` tasks.
@@ -826,7 +811,7 @@ class GenerateMapping(BaseSalesforceApiTask):
     and B also refers to A. Mapping generation will fail for such data models; to
     resolve the issue, specify the `ignore` option with the name of one of the
     involved lookup fields to suppress it. `ignore` can be specified as a list in
-    `cumulusci.yml` or as a semicolon-separated string at the command line.
+    `cumulusci.yml` or as a comma-separated string at the command line.
 
     In most cases, the mapping generated will need minor tweaking by the user. Note
     that the mapping omits features that are not currently well supported by the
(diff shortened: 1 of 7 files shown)


cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+    for item_key, item_value in data.items():
+        node_key = dumper.represent_data(item_key)
+        node_value = dumper.represent_data(item_value)
+
+        value.append((node_key, node_value))
+
+    return yaml.nodes.MappingNode(u"tag:yaml.org,2002:map", value)
+
+
+yaml.SafeDumper.add_representer(OrderedDict, represent_ordereddict)
```
davisagli: I don't like that importing this module has the side effect of changing the global yaml dumper.

For the inverse operation (making sure that yaml is loading into OrderedDicts in python 2) we already have `cumulusci.core.utils.ordered_yaml_load` which uses its own subclass of yaml.SafeLoader. I think I would prefer adding an `ordered_yaml_dump` function there which similarly subclasses yaml.SafeDumper.

(Soon when we drop Python 2 support we should be able to get rid of all this and just use normal dicts internally since they are ordered in Python 3.6+)

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+    and B also refers to A. Mapping generation will fail for such data models; to
+    resolve the issue, specify the `ignore` option with the name of one of the
+    involved lookup fields to suppress it. `ignore` can be specified as a list in
+    `cumulusci.yml` or as a semicolon-separated string at the command line.
+
+    In most cases, the mapping generated will need minor tweaking by the user. Note
+    that the mapping omits features that are not currently well supported by the
+    `extract_dataset` and `load_dataset` tasks, including self-lookups and references to
+    the `User` object.
+    """
```
davisagli: Don't forget to run `cci task doc > docs/tasks.rst` to update the task documentation.

davisagli: (We'll also want some updates to the narrative documentation in https://cumulusci.readthedocs.io/en/latest/bulk_data.html -- that can happen in another pull request.)

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+        ].endswith("__"):
+            self.options["namespace_prefix"] += "__"
+
+        if "ignore" in self.options:
+            if type(self.options["ignore"]) is str:
+                self.options["ignore"] = [
+                    f.strip() for f in self.options["ignore"].split(";")
+                ]
+        else:
+            self.options["ignore"] = []
```
davisagli: Use cumulusci.core.utils.process_list_arg for this. (It wants comma-separated rather than semicolon-separated, which I think will work fine here and be more consistent with other task options that accept lists.)

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+                if field["type"] == "reference":
+                    if field["relationshipOrder"] == 1 or self._is_any_custom_api_name(
+                        field["name"]
+                    ):
+                        self.mapping_objects.extend(
+                            filter(
+                                lambda o: o not in self.mapping_objects
+                                and self._is_object_mappable(self.describes[o]),
+                                field["referenceTo"],
+                            )
```
davisagli: How about:

    self.mapping_objects.extend([obj for obj in field['referenceTo'] if obj not in self.mapping_objects and self._is_object_mappable(self.describes[obj])])

Maybe this is just taste but I almost find comprehensions more readable than filter or map with a lambda expression.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+                        self.logger.warning(
+                            "Field {}.{} is a polymorphic lookup, which is not supported".format(
+                                obj, field
+                            )
+                        )
+
+    def _run_task(self):
+        self._collect_objects()
+        self._build_schema()
+        self._build_mapping()
```
davisagli: I'd suggest organizing this class with _run_task first, then the 3 methods called here, followed by other methods that they end up calling. It makes the module easier to read and understand in one pass. I also wouldn't complain if you feel inclined to add some docstrings.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+    def _split_dependencies(self, objs, dependencies):
+        stack = []
+        objs_remaining = objs.copy()
+
+        # The structure of `dependencies` is:
+        # key = object, value = set of objects it references.
+
+        # Iterate through our list of objects
+        # For each object, if it is not dependent on any other objects, place it at the end of the stack.
+        # Once an object is placed in the stack, remove dependencies to it (they're satisfied)
```
davisagli: So this is basically a topological sort of the object dependency graph?

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+            lookups = []
+            for field in self.schema[obj].values():
+                if field["type"] == "reference":
+                    lookups.append(field["name"])
+                else:
+                    fields.append(field["name"])
+            self.mapping[key]["fields"] = OrderedDict()
+            if fields:
+                if "Id" not in fields:
+                    fields.append("Id")
```
davisagli: This line is missing coverage.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+            if lookups:
+                lookups.sort(key=field_sort)
+                self.mapping[key]["lookups"] = OrderedDict()
+                for field in lookups:
+                    self.mapping[key]["lookups"][field] = {
+                        "table": self.schema[obj][field]["referenceTo"][0].lower()
+                    }
+                    if len(self.schema[obj][field]["referenceTo"]) > 1:
+                        self.logger.warning(
+                            "Field {}.{} is a polymorphic lookup, which is not supported".format(
```
davisagli: This line is missing coverage.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+    def _is_lookup_to_included_object(self, field):
+        return field["type"] == "reference" and self._are_lookup_targets_in_operation(
+            field
+        )
+
+    def _collect_objects(self):
+        self.mapping_objects = []
+
+        # Cache the global describe, which we'll walk.
+        self.global_describe = self.sf.describe()
```
davisagli: We should log something before starting this so it doesn't look like it's doing nothing.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+                self.logger.info("Mapped objects: {}".format(", ".join(stack)))
+                self.logger.info("Remaining objects:")
+                for obj in objs_remaining:
+                    self.logger.info(
+                        "{} (depends on {})".format(
+                            obj,
+                            ", ".join(dependencies[obj]) if obj in dependencies else "",
+                        )
+                    )
+                raise BulkDataException("Cannot complete mapping")
```
davisagli: We should make BulkDataException subclass CumulusCIFailure instead of CumlusCIException. That way the CLI will print out the error message without a traceback.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+                )
+                self.logger.info("Mapped objects: {}".format(", ".join(stack)))
+                self.logger.info("Remaining objects:")
+                for obj in objs_remaining:
+                    self.logger.info(
+                        "{} (depends on {})".format(
+                            obj,
+                            ", ".join(dependencies[obj]) if obj in dependencies else "",
+                        )
+                    )
```
davisagli: For manually resolving cycles, it'd be helpful to also have the information of which field manifests the dependency.

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+                        self._is_lookup_to_included_object(field),
+                    ]
+                ):
+                    if self._is_field_mappable(obj, field):
+                        self.schema[obj][field["name"]] = field
+
+                        if field["type"] == "reference":
+                            for target in field["referenceTo"]:
+                                if (
+                                    obj in self.mapping_objects
```
Br4nd0R: Is there a circumstance where `obj` wouldn't be in `self.mapping_objects`? I could be missing something, but it looks like `obj` would always be contained in `self.mapping_objects`

davisagli:
Couple more thoughts:
- Maybe we should filter out the fields that are included in a scratch org created with `hasSampleData: true`?
- Is there a reason we’re lowercasing the local table and fieldnames? Might be simpler to keep them the same as what we get from the API unless the user chooses to diverge

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +799,313 @@ def _sqlite_dump(self):
...
+
+        value.append((node_key, node_value))
+
+    return yaml.nodes.MappingNode(u"tag:yaml.org,2002:map", value)
+
+
+yaml.SafeDumper.add_representer(OrderedDict, represent_ordereddict)
+
+
+class GenerateMapping(BaseSalesforceApiTask):
```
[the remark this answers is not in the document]
davidmreed: I like this plan - would you be open to doing it in a separate PR?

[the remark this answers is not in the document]
davisagli: Certainly.

cumulusci/tasks/bulkdata.py
davidmreed: Basically, yeah, accounting for the fact that it's not a DAG.

cumulusci/tasks/bulkdata.py
davidmreed: Thanks for tackling this bit!

cumulusci/tasks/bulkdata.py
davidmreed: Both halves of this `if` have been made redundant by checks elsewhere. Thanks!

davidmreed:
> Maybe we should filter out the fields that are included in a scratch org created with hasSampleData: true?

I'd rather extirpate any and all places that `hasSampleData` is used. I don't like the idea of hard-coding metadata not to be pulled - some user could create genuine fields of the same name and the failure would be mysterious.

> Is there a reason we’re lowercasing the local table and fieldnames? Might be simpler to keep them the same as what we get from the API unless the user chooses to diverge

Inertia, I think. I don't see any strong reason to do it save that the user could change the casing on the API name and Salesforce itself would keep working, since it's case-insensitive, while we'd break. But if the user changes the case on an API name I would anticipate most integrations breaking; one oughtn't do that.

> It'd be easy enough to show the object label in addition to developer name which might help clarify

In the cycle warning log?

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +800,324 @@ def _sqlite_dump(self):
...
+
+        if self.options["namespace_prefix"] and not self.options[
+            "namespace_prefix"
+        ].endswith("__"):
+            self.options["namespace_prefix"] += "__"
+
+        if "ignore" in self.options:
+            self.options["ignore"] = process_list_arg(self.options["ignore"])
+        else:
+            self.options["ignore"] = []
```
davisagli: This could be a one-liner:

    self.options["ignore"] = process_list_arg(self.options.get("ignore", []))

cumulusci/tasks/bulkdata.py
```diff
@@ -797,6 +800,324 @@ def _sqlite_dump(self):
...
+        # Iterate through our list of objects
+        # For each object, if it is not dependent on any other objects, place it at the end of the stack.
+        # Once an object is placed in the stack, remove dependencies to it (they're satisfied)
+        while objs_remaining:
+            objs_without_deps = list(
+                filter(
+                    lambda obj: obj not in dependencies or not dependencies[obj],
+                    objs_remaining,
+                )
+            )
```
davisagli: Here's another one I'd suggest converting to a list comprehension.

PR #1150, opened 2019-07-24
Mapping generator: Report which fields create a dependency

davisagli:
Here's my take on adjusting the reporting of unresolved dependencies to include a mention of which field creates the dependency.

I also ran `cci task doc` to update the task docs.

# Critical Changes

# Changes

# Issues Closed


davisagli:
(Sorry, I obviously haven't updated the tests. But I figured this would be the easiest way to explain what I was thinking.)

davidmreed:
Elegant. I like it. I'll fix the tests when I address the remaining review comments.

PR #1154, opened 2019-07-26
show_info displays step details for cci flow info

Br4nd0R:
# Critical Changes
# Changes
new method: `FlowCoordinator.show_summary()` to display sub-flow task info for `cci flow info [flow_name]`
# Issues Closed

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -84,6 +84,7 @@ class StepSpec(object):
 
     __slots__ = (
         "step_num",  # type: str
+        "step_sequence",  # type list(int)
         "task_name",  # type: str
         "task_config",  # type: dict
         "task_class",  # type: str
@@ -96,6 +97,7 @@ class StepSpec(object):
     def __init__(
         self,
         step_num,
+        step_sequence,
         task_name,
         task_config,
         task_class,
@@ -105,6 +107,7 @@ class StepSpec(object):
         when=None,
     ):
         self.step_num = step_num
+        self.step_sequence = step_sequence
         self.task_name = task_name
         self.task_config = task_config
         self.task_class = task_class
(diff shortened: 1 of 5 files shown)


cumulusci/core/flowrunner.py
```diff
@@ -289,6 +289,34 @@ def _rule(self, fill="=", length=60, new_line=False):
...
+                    )
+                    step_num_by_level[indent] += 1
+                    # when a step num increases we need to
+                    # reset step levels below it to 1
+                    idx2 = indent + 1
+                    while idx2 < len(step_num_by_level):
+                        step_num_by_level[idx2] = 1
+                        idx2 += 1
+                indent += 1
+            previous_step = step
```
davisagli: @Br4nd0R nice work. for fun I wrote my own version which uses the numbers from step.step_num instead of re-counting in step_num_by_level:

```python
        lines = []
        previous_parts = []
        for step in self.steps:
            parts = step.path.split('.')
            numbers = str(step.step_num).split('.')
            task_name = parts.pop()
            for i, flow_name in enumerate(parts):
                if len(previous_parts) < i + 1 or previous_parts[i] != flow_name:
                    lines.append("{}{}) flow: {}".format("    " * i, numbers[i], flow_name))
            lines.append("{}{}) task: {}".format("    " * (i + 1), numbers[-1], task_name))
            previous_parts = parts
        return "\n".join(lines)
```

Your approach looks valid to me though, so please don't feel pressure to adopt anything from my version.

cumulusci/core/tests/test_flowrunner.py
```diff
@@ -131,6 +131,21 @@ def test_init(self):
...
+                "steps": {"1": {"flow": "nested_flow_2"}},
+            }
+        )
+        flow = FlowCoordinator(self.project_config, flow_config, name="test_flow")
+        flow.show_summary()
+
+        self.assertEquals(
+            True, "Description: test description" in self.flow_log["info"]
+        )
+        # TODO: We should probably assert on output after description
```
davisagli: This is another good reason to make the function return a string and leave outputting it up to the caller. It makes it easier to test.

cumulusci/core/flowrunner.py
```diff
@@ -289,6 +289,34 @@ def _rule(self, fill="=", length=60, new_line=False):
         if new_line:
             self.logger.info("")
 
+    def show_summary(self):
+        """ Display the description and steps contained within a given FlowCoordinator """
+        self.logger.info(
+            "Description: {}".format(self.flow_config.config["description"])
+        )
```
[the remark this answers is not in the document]
jstvz: 💯

cumulusci/core/flowrunner.py
Br4nd0R: I do like how this version iterates directly over self.steps, and how it differentiates between tasks and flows. One thing that I'm running into with this algorithm is that if there is a step_path that just contains a single task at the top level (i.e. `len(parts) == 1`) then the `for` loop that utilizes `enumerate()` never executes and we get an `UnboundLocalError` when we attempt to reference `i` when appending the task name.

Also, we talked briefly about how `step.step_num` isn't consistent throughout all steps (some start with 1 and others with 0) so we will want to work through how to address that too.

davisagli: Good point about step paths that only contain a top-level task. We could avoid that error by initializing `i = -1` before the loop.

Step numbering is intentionally flexible to make it possible to override a flow and insert steps in the middle. So it's possible to use 0, use non-integers, and leave gaps in the sequence. We should preserve those numbers when we display the flow so that it's easy to refer back to the configuration. To deal with this I think we'll want to modify the StepSpec object to store a second number sequence for each step, stored as a list so that we can distinguish between levels. Then we can use this new set of numbers for display purposes.

Another thing I thought of: flow steps can have conditions attached using the "when" clause -- see the `deploy_unmanaged` flow for an example. We should probably display these to make it clear that there is conditional logic involved to determine what actually runs.

Br4nd0R:
@davisagli - Latest commit now displays `step.when`. I didn't modify `StepSpec` yet; however, I did update the code to display the entire `step.step_num` value for each step (instead of splitting and displaying based on indent). A flow and its' first child task have the same step number, but wouldn't creating an additional list in `StepSpec` also have the same outcome? A flow like `dev_org` would show the first flow as `1` instead of `1.1` and the same for the first child task?

I think I may be missing the benefit of creating the additional number sequence on `StepSpec`, so let me know if I'm off the mark at all.

davisagli:
@Br4nd0R looking at the `dev_org` flow as an example... my thought is that the first step would have num_path = ("1", "1") and would give this output:

    1) flow: dependencies
        1) task: update_dependencies
        ...

The `dx_push` task within the deploy_unmanaged flow would have num_path = ("2", "3.1") and would output like this:

    2) flow: deploy_unmanaged
        ...
        3.1) dx_push

PR #1157, opened 2019-07-30
Bulk Data: fail loudly on load errors

davidmreed:
# Critical Changes

# Changes

- The `LoadData` task will throw a `BulkDataException` if any records fail during the load operation.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata.py
+++ b/cumulusci/tasks/bulkdata.py
@@ -231,10 +231,17 @@ class LoadData(BulkJobTaskMixin, BaseSalesforceApiTask):
         "sql_path": {
             "description": "If specified, a database will be created from an SQL script at the provided path"
         },
+        "ignore_row_errors": {
+            "description": "If True, allow the load to continue even if individual rows fail to load."
+        },
     }
 
     def _init_options(self, kwargs):
         super(LoadData, self)._init_options(kwargs)
+
+        self.options["ignore_row_errors"] = process_bool_arg(
+            self.options.get("ignore_row_errors", False)
+        )
         if self.options.get("sql_path"):
             if self.options.get("database_url"):
                 raise TaskOptionsError(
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/bulkdata.py
```diff
@@ -442,13 +444,18 @@ def _store_inserted_ids(self, mapping, job_id, local_ids_for_batch):
...
+                # We can't get new Ids for some reason, or determine batch status.
+                # Fail the job to preserve integrity of data store.
+                if not isinstance(e, BulkDataException):
+                    raise BulkDataException(
+                        "Failed to download results for batch {} ({})".format(
+                            batch_id, e
+                        )
+                    )
+                else:
+                    raise e
```
davisagli: I'd suggest writing this as follows:

    except BulkDataException:
        raise
    except Exception as e:
        raise BulkDataException("Failed to download results for batch {} ({})".format(batch_id, str(e)))

cumulusci/tasks/bulkdata.py
```diff
@@ -266,14 +266,16 @@ def _run_task(self):
             self.logger.info("Running Job: {}".format(name))
             result = self._load_mapping(mapping)
             if result != "Completed":
-                break
+                raise BulkDataException(
+                    "Job {} did not complete successfully".format(name)
+                )
```
davisagli: When there's a failed batch we should probably make _parse_job_state grab the `stateMessage` and pass it along so we can show a more informative message here.

cumulusci/tasks/bulkdata.py
davidmreed: 👍 Also removing the old nocover pragma and adding a test case.

cumulusci/tasks/bulkdata.py
```diff
@@ -485,7 +487,7 @@ def produce_csv():
                     sf_id = row[0]
                     yield "{},{}\n".format(local_id, sf_id).encode("utf-8")
                 else:
-                    self.logger.warning("      Error on row {}: {}".format(i, row[3]))
+                    raise BulkDataException("Error on row {}: {}".format(i, row[3]))
```
davisagli: I'm concerned that this may be too sensitive. If a handful of records fail due to lock timeouts in the middle of an 8-hour LDV bulk load, do we really want to have to start over?

davidmreed: On the other hand, on non-volume-oriented loads, lack of some specific record invalidates the state of the org.

How about having fail-on-row-error be the default with a switch to allow row-level errors to be silently ignored?

davidmreed: (I also have some improvements to the tests yet to push; I realized one of them was missing a bit of coverage.)

davisagli: Sure, an option to ignore row-level errors sounds like the right escape valve for that scenario.

PR #1159, opened 2019-07-31
Feature/robot lint

boakley:
This adds a new robot_lint task. It includes several new tests, and one new package added to requirements.txt (robotframework-lint). 

I have updated the documentation to mention the new task. I also tweaked another section of the documentation to mention that tasks should throw exceptions to signal failure, as that didn't seem to be mentioned anywhere.

Finally, I also added docs/_build to .gitignore.

# Critical Changes

# Changes

# Issues Closed


docs/cookbook.rst
```diff
@@ -124,7 +124,12 @@ Quick background about CumulusCI tasks
 
 All tasks in CumulusCI are python classes that subclass `cumulusci.core.tasks.BaseTask`.  The general usage of a task is two step: initialize an instance then call it to run the task.
 
-For most tasks, you'll want to override the `_run_task` method in your subclass to provide implementation.
+For most tasks, you'll want to override the `_run_task` method in your subclass to provide the implementation. The return value of this function is ignored. Exceptions from `cumulus.core.exceptions` should be raised to communicate task status to the user or flow. If no exceptions are thrown, the task is considered to have completed successfully.
```
davisagli: The return value is not necessarily ignored; it's stored as part of the StepResult when a task is running as part of a flow, where it could be looked up by tools that build on top of CumulusCI.

cumulusci/tasks/robotframework/robot_lint.py
```diff
@@ -0,0 +1,219 @@
...
+
+        + example.robot
+        W: 2, 0: No suite documentation (RequireSuiteDocumentation)
+        E: 30, 0: No testcase documentation (RequireTestDocumentation)
+
+    To see a list of all configured options, set the 'list' option to True:
+
+        cci task run robot_list -o list True
+
+    """
```
davisagli: Looks like you need to run `cci task doc > docs/tasks.rst` to update the file which becomes the source of https://cumulusci.readthedocs.io/en/latest/tasks.html

cumulusci/cumulusci.yml
```diff
@@ -844,7 +848,7 @@ services:
         description: The Saucelabs username to use for tasks.
         required: True
       api_key:
-        description: The Saucelabs api key to use for tasks  
+        description: The Saucelabs api key to use for tasks
         required: True
   sentry:
     description: Configure connection to sentry.io for error tracking
```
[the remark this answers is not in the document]
boakley: I recently changed my emacs to automatically trim trailing whitespace.

cumulusci/tasks/robotframework/robot_lint.py
```diff
@@ -0,0 +1,219 @@
+from cumulusci.core.tasks import BaseTask
```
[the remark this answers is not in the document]
boakley: Yeah, I was wondering about that. I like the idea of naming the file after the task rather than partly after the task, but you're right that the leading robot_ is redundant, and leaving it off is more consistent with the libdoc task which also leaves it off. 

I'll change it.

cumulusci/tasks/robotframework/robot_lint.py
```diff
@@ -0,0 +1,219 @@
...
+            try:
+                files = self._get_files()
+                args = self._get_args()
+
+                # the result is a count of the number of errors,
+                # though I don't think the caller cares.
+                result = linter.run(args + sorted(files))
+
+            except Exception as e:
+                raise CumulusCIUsageError(str(e))
```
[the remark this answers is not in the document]
boakley: Yeah, that makes sense.

cumulusci/tasks/robotframework/robot_lint.py
boakley: Cool. I didn' t know `cci task doc` was a thing.

cumulusci/tasks/robotframework/robot_lint.py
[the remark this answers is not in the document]
davisagli: I kind of like that idea too -- although there's potential to have multiple named task configs that use the same class, and it would be a big refactor to standardize on that across all the tasks. It would help avoid the current problem where we have some task classes named the same as their modules though, which can make importing the module difficult if the package __init__.py imports the class.

cumulusci/tasks/robotframework/robot_lint.py
davisagli: I should probably figure out how to make it run in a precommit hook.

cumulusci/tasks/robotframework/robot_lint.py
```diff
@@ -0,0 +1,219 @@
...
+
+        all_files = []
+        for path in expanded_paths:
+            if os.path.isdir(path):
+                for root, dirs, files in os.walk(path):
+                    for filename in files:
+                        if filename.endswith(".robot") or filename.endswith(
+                            ".resource"
+                        ):
+                            all_files.append(os.path.join(root, filename))
```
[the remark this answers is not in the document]
boakley: good catch.

docs/cookbook.rst
boakley: I changed the wording from "The return value of this function is ignored" to "The return value of this function is saved as part of the StepResult.

I'm open to other suggestions if you have them. From the perspective of someone who had not written a brand new task from scratch before, the requirements of a task with respect to result codes, output, and exceptions weren't entirely clear.

docs/cookbook.rst
```diff
@@ -124,7 +124,12 @@ Quick background about CumulusCI tasks
 
 All tasks in CumulusCI are python classes that subclass `cumulusci.core.tasks.BaseTask`.  The general usage of a task is two step: initialize an instance then call it to run the task.
 
-For most tasks, you'll want to override the `_run_task` method in your subclass to provide implementation.
+For most tasks, you'll want to override the `_run_task` method in your subclass to provide the implementation. The return value of this function is saved as part of the StepResult. Exceptions from `cumulus.core.exceptions` should be raised to communicate task status to the user or flow. If no exceptions are thrown, the task is considered to have completed successfully.
+
+Task Exceptions
+---------------
+
+If the task has an error that should be considered a build failure (e.g. a metadata deployment failure, test failure, etc) it can raise the exception `cumulus.core.exceptions.CumulusCIFailure`. If you want to flag a usage error (e.g. the task receives an invalid set of options) it should raise the exception `cumulus.core.exceptions.CumulusCiUsageError`.
```
davisagli:
```suggestion
If the task has an error that should be considered a build failure (e.g. a metadata deployment failure, test failure, etc) it can raise the exception `cumulus.core.exceptions.CumulusCIFailure`. If you want to flag a usage error (e.g. the task receives an invalid set of options) it should raise the exception `cumulus.core.exceptions.CumulusCIUsageError`.
```

cumulusci/tasks/robotframework/lint.py
```diff
@@ -0,0 +1,207 @@
...
+
+    Some rules are configurable, and can be configured with the
+    `configure` option. This option takes a list of values in the form
+    *<rule>*:*<value>*,*<rule>*:*<value>*,etc.  For example, to set
+    the line length for the LineTooLong rule you can use '-o configure
+    LineTooLong:80'. If a rule is configurable, it will show the
+    configuration options in the documentation for that rule
+
+    The filename will be printed once before any errors or warnings
+    for that file. The filename is preceeded by `+ `
```
davisagli:
```suggestion
    for that file. The filename is preceeded by `+`
```

Without this change, running `cci task info robot_lint` prints a warning: `<string>:36: (WARNING/2) Inline interpreted text or phrase reference start-string without end-string.`

boakley: Interesting. I should get in the habit of writing a test that verifies the task info command doesn't barf. That, or we create a test that iterates over all known tasks and runs the info command to make sure it doesn't barf so we don't have to remember to do it for each new task.

docs/cookbook.rst
```diff
@@ -312,7 +317,7 @@ If you want to run the full packaging flow where feature branches build unmanage
 
 * CUMULUSCI_ORG_packaging: The output from `cci org info packaging`, assuming you've already connected your packaging org to your local toolbelt.
 * CUMULUSCI_ORG_beta: The output from `cci org info beta`, assuming you've already connected your beta org to your local toolbelt.
-* CUMULUSCI_SERVICE_github: The output from `cci project show_github`, assuming you've already configured github locally via `cci project connect_github` 
+* CUMULUSCI_SERVICE_github: The output from `cci project show_github`, assuming you've already configured github locally via `cci project connect_github`
```
jstvz: Now that I think about it, is this line still valid? We can address in another PR.

docs/robotframework.rst
```diff
@@ -447,8 +447,10 @@ CumulusCI includes two tasks for working with Robot Framework tests and keyword
...
 * **robot_testdoc**: Generates html documentation of your whole robot test suite and writes to **robot/<project name>/doc/<project_name>.html**.
+* **robot_lint**: Performs static analysis of robot files (files with
+  .robot and .resource), flagging issues that may reduce the quality of the code.
 
-Additionally, the RobotLibDoc task class can be wired up to generate library documentation if you choose to create a library of robot keywords for your project using. For example, if you have defined a robot resource file named MyProject.resource and placed it in the **resources** folder, you would add the following added to the cumulusci.yml file:
+Additionally, the **robot_libdoc** task can be wired up to generate library documentation if you choose to create a library of robot keywords for your project using. For example, if you have defined a robot resource file named MyProject.resource and placed it in the **resources** folder, you would add the following added to the cumulusci.yml file:
```
jstvz:
```suggestion
Additionally, the **robot_libdoc** task can be wired up to generate library documentation if you choose to create a library of robot keywords for your project. For example, if you have defined a robot resource file named MyProject.resource and placed it in the **resources** folder, you would add the following to the cumulusci.yml file:
```

cumulusci/tasks/robotframework/lint.py
```diff
@@ -0,0 +1,207 @@
...
+        # result is the number of errors detected
+        if result > 0:
+            message = (
+                "1 error was detected"
+                if result == 1
+                else "{} errors were detected".format(result)
+            )
+            raise CumulusCIFailure(message)
+
+    def _get_files(self):
```
jstvz: Another approach might be:

```python
from pathlib import Path
def _get_files(self):
    cwd = Path('.')
    robot_files = set(cwd.glob('**/*.robot'))
    resource_files = set(cwd.glob('**/*.resource'))
    return {str(file) for file in robot_files.union(resource_files) if file.is_file()}
```

boakley: Yeah, I thought about using pathlib, but it's only supported on 3.x. Don't we still require support for 2.7?

davisagli: Correct.

docs/cookbook.rst
jstvz: Logged in #1173.

cumulusci/tasks/robotframework/lint.py
jstvz: I guess I thought we had `pathlib2` in the requirements for some reason. 🤷‍♂

PR #1162, opened 2019-07-31
Feature/bulk data package

davidmreed:
This PR breaks out `bulkdata.py` into a subpackage, with minimal changes otherwise.

# Critical Changes

# Changes

# Issues Closed


davidmreed:
n.b. Coveralls marks a number of lines as "new uncovereds", but they're actually just moved from a different file; coverage should be the same.

PR #1169, opened 2019-08-07
Feature/debug log tasks

prescod:
Create new CCI tasks for enabling and downloading Debug Logs.

### DIFF
--- a/cumulusci/tasks/debug_logs.py
+++ b/cumulusci/tasks/debug_logs.py
@@ -90,10 +90,7 @@ class DownloadDebugLogs(BaseSalesforceApiTask):
     """
 
     task_options = {
-        "debug_log_dir": {
-            "description": "What directory to put the logs in",
-            "required": True,
-        }
+        "debug_log_dir": {"description": "Where to put the logs", "required": True}
     }
 
     def _run_task(self):
--- a/debug_logs.py
+++ b/debug_logs.py
@@ -1,0 +1,138 @@
+import datetime
+import os
+import csv
+
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+
+class TurnOnDebugLogs(BaseSalesforceApiTask):
+    task_docs = """
+    Turn on debugging information in Salesforce
+    """
+
+    debug_level_id = None  # Class variable
+
+    def _run_task(self):
+        """Create a TraceFlag for a given user."""
+        self._delete_debug_levels()
+        self._delete_trace_flags()
+        self.logger.info("Creating DebugLevel object")
+        DebugLevel = self._get_tooling_object("DebugLevel")
+        result = DebugLevel.create(
+            {
+                "ApexCode": "Info",
+                "ApexProfiling": "Debug",
+                "Callout": "Info",
+                "Database": "Info",
+                "DeveloperName": "CumulusCI",
+                "MasterLabel": "CumulusCI",
+                "System": "Info",
+                "Validation": "Info",
+                "Visualforce": "Info",
+                "Workflow": "Info",
+            }
+        )
+        if TurnOnDebugLogs.debug_level_id:
+            self.logger.info(
+                "Debug logs are already turned on ({TurnOnDebugLogs.debug_level_id})"
+            )
+            return
+        TurnOnDebugLogs.debug_level_id = result["id"]
+        self.logger.info("Setting up trace flag to capture debug logs")
+        # New TraceFlag expires 12 hours from now
+        expiration_date = datetime.datetime.utcnow() + datetime.timedelta(
+            seconds=60 * 60 * 12
+        )
+        TraceFlag = self._get_tooling_object("TraceFlag")
+        result = TraceFlag.create(
+            {
+                "DebugLevelId": result["id"],
+                "ExpirationDate": expiration_date.isoformat(),
+                "LogType": "USER_DEBUG",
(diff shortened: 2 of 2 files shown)


cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
...
+            for record in result["records"]:
+                TraceFlag.delete(str(record["Id"]))
+
+    def _delete_debug_levels(self):
+        """
+        Delete existing DebugLevel objects.
+        This will automatically delete associated TraceFlags as well.
+        """
+        self.logger.info("Deleting existing DebugLevel objects")
+        result = self.tooling.query("Select Id from DebugLevel")
```
davidmreed: We should delete only our own DebugLevel by DeveloperName. Users may have created their own Debug Levels that they don't want smashed if they run this task locally.

cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
...
+            )
+            return
+        TurnOnDebugLogs.debug_level_id = result["id"]
+        self.logger.info("Setting up trace flag to capture debug logs")
+        # New TraceFlag expires 12 hours from now
+        expiration_date = datetime.datetime.utcnow() + datetime.timedelta(
+            seconds=60 * 60 * 12
+        )
+        TraceFlag = self._get_tooling_object("TraceFlag")
+        result = TraceFlag.create(
```
Br4nd0R: Do we want to capture `result` if it's not going to be referenced?

cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
...
+        self.logger.info("Created TraceFlag for user")
+
+    def _delete_trace_flags(self):
+        """
+        Delete existing DebugLevel objects.
+        This will automatically delete associated TraceFlags as well.
+        """
+        self.logger.info("Deleting existing TraceFlag objects")
+        result = self.tooling.query(
+            "Select Id from TraceFlag Where TracedEntityId = '{}'".format(
```
Br4nd0R: Do we have any formatting best practices when writing SQL/SOQL queries as string in our code? (e.g. all caps for keywords like `SELECT`, `FROM`, `WHERE`)

cumulusci/tasks/debug_logs.py
Br4nd0R: Do we run a similar risk inside of `_delete_trace_flag()` where the query is selecting all flags where the `TracedentityId == self.org_config.user_id`? Is there the potential that some of those flags were not created by this task?

cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
+import datetime
```
davisagli: Let's move this module into the `cumulusci.tasks.apex` package.

cumulusci/tasks/debug_logs.py
davisagli: No, we should not. flake8 should warn about this but maybe it's confused by the reference on line 49 to the previous value of `result`.

cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
...
+import os
+import csv
+
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+
+class TurnOnDebugLogs(BaseSalesforceApiTask):
+    task_docs = """
+    Turn on debugging information in Salesforce
+    """
```
davisagli: Please configure the new tasks in cumulusci.yml and run `cci task doc > docs/tasks.rst` to update the generated task reference docs.

cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
...
+                "Workflow": "Info",
+            }
+        )
+        if TurnOnDebugLogs.debug_level_id:
+            self.logger.info(
+                "Debug logs are already turned on ({TurnOnDebugLogs.debug_level_id})"
+            )
+            return
+        TurnOnDebugLogs.debug_level_id = result["id"]
+        self.logger.info("Setting up trace flag to capture debug logs")
```
davisagli: It would be nice to log when it will expire. It also might be nice to provide a task option for how long it should last.

cumulusci/tasks/debug_logs.py
```diff
@@ -0,0 +1,135 @@
...
+                "Validation": "Info",
+                "Visualforce": "Info",
+                "Workflow": "Info",
+            }
+        )
+        if TurnOnDebugLogs.debug_level_id:
+            self.logger.info(
+                "Debug logs are already turned on ({TurnOnDebugLogs.debug_level_id})"
+            )
+            return
```
davisagli: I don't quite understand what this is trying to do. If the debug level is already set up, shouldn't we return up at the top of this method before deleting and recreating it?

But a class attribute is not a great place to keep track of that since it'll be lost between multiple runs of cci or multiple users running cci on different machines. I think querying the DebugLevel object in the org would be the thing to do.

cumulusci/tasks/debug_logs.py
davisagli: Yeah I think we need to be more careful to not destroy debug levels / trace flags that came from something else.

prescod:
Closing for now because I don't have a need for this functionality coming up soon and it sounds like the improvements needed are largeish.

PR #1170, opened 2019-08-07
Include option for including links to PRs with no release notes

Br4nd0R:
# Changes
* The `github_release_notes` task now has an `include_empty` option to for the ability to include links to PRs that have no release notes.

### DIFF
--- a/cumulusci/tasks/release_notes/generator.py
+++ b/cumulusci/tasks/release_notes/generator.py
@@ -12,6 +12,7 @@ from cumulusci.tasks.release_notes.provider import GithubChangeNotesProvider
 class BaseReleaseNotesGenerator(object):
     def __init__(self):
         self.change_notes = []
+        self.empty_pull_requests = []
         self.init_parsers()
         self.init_change_notes()
 
@@ -45,9 +46,16 @@ class BaseReleaseNotesGenerator(object):
 
     def _parse_change_note(self, change_note):
         """ Parses an individual change note through all parsers in
-        self.parsers """
+        self.parsers. If no lines were added then appends the change
+        note to the list of empty PRs"""
+        line_added_by_parsers = False
         for parser in self.parsers:
-            parser.parse(change_note)
+            line_added = parser.parse(change_note)
+            if not line_added_by_parsers:
+                line_added_by_parsers = line_added
+
+        if not line_added_by_parsers:
+            self.empty_pull_requests.append(change_note)
 
     def render(self):
         """ Returns the rendered release notes from all parsers as a string """
@@ -98,6 +106,7 @@ class GithubReleaseNotesGenerator(BaseReleaseNotesGenerator):
         link_pr=False,
         publish=False,
         has_issues=True,
+        include_empty=False,
     ):
         self.github = github
         self.github_info = github_info
@@ -107,6 +116,7 @@ class GithubReleaseNotesGenerator(BaseReleaseNotesGenerator):
         self.link_pr = link_pr
         self.do_publish = publish
         self.has_issues = has_issues
+        self.include_empty_pull_requests = include_empty
         self.lines_parser_class = None
         self.issues_parser_class = None
         super(GithubReleaseNotesGenerator, self).__init__()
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/release_notes/generator.py
```diff
@@ -12,6 +12,7 @@
 class BaseReleaseNotesGenerator(object):
     def __init__(self):
         self.change_notes = []
+        self.empty_pull_requests = []
```
davisagli: Thinking about the naming here, we should probably call this `empty_change_notes` for symmetry with the existing `change_notes` list. Pull requests are the source of the change notes in the case of the github release notes generator, but this base class doesn't assume that's where they come from.

cumulusci/tasks/release_notes/generator.py
```diff
@@ -136,6 +146,21 @@ def _get_release(self):
...
+            section_lines.append("\n# Pull requests with no release notes")
+            for content in self.empty_pull_requests:
+                section_lines.append(
+                    "\n* {}".format(self._mark_down_link_to_pr(content))
+                )
+
+        return section_lines
+
+    def _mark_down_link_to_pr(self, pull_request):
+        if pull_request:
```
davisagli: When would this line be false?

cumulusci/tasks/release_notes/generator.py
```diff
@@ -136,6 +146,21 @@ def _get_release(self):
...
+            return "[[PR{}]({})]".format(pull_request.number, pull_request.html_url)
```
davisagli: Do you think we should include the PR title?

cumulusci/tasks/release_notes/generator.py
```diff
@@ -186,8 +211,10 @@ def _update_release_content(self, release, content):
                 if parser_content and not parser.replaced:
                     new_body.append(parser_content + "\r\n")
 
-            content = u"\r\n".join(new_body)
-
+        # add empty PRs section
+        if self.include_empty_pull_requests:
+            new_body.extend(self._render_empty_pr_section())
```
davisagli: If the task is run twice we'll end up with 2 copies of this list in the release notes. That's probably not a huge problem since typically this task is only run once by automation. But worth being aware of that difference from how the other parser sections are handled.

cumulusci/tasks/release_notes/task.py
```diff
@@ -25,6 +25,9 @@ class GithubReleaseNotes(BaseGithubTask):
             )
         },
         "publish": {"description": "Publish to GitHub release if True (default=False)"},
+        "include_empty": {
+            "description": "If True, include links to PRs that have no release notes (default=False)"
+        },
```
davisagli: Don't forget to adjust the release_beta flow in cumulusci.yml to set this to True.

cumulusci/tasks/release_notes/tests/test_generator.py
```diff
@@ -122,6 +122,18 @@ def test_init_with_last_tag(self):
...
+    def test_mark_down_link_to_pr(self):
+        pr = mock.Mock(number=1, html_url="http://pr", body="# Changes\r\n\r\nfoo")
+        github_info = self.github_info.copy()
+        self.mock_util.mock_get_repo()
+        generator = GithubReleaseNotesGenerator(
+            self.gh, github_info, PARSER_CONFIG, self.current_tag, self.last_tag
+        )
+        actual_link = generator._mark_down_link_to_pr(pr)
+        expected_link = "[[PR{}]({})]".format(pr.number, pr.html_url)
+        self.assertEquals(expected_link, actual_link)
```
davisagli: We should probably also add a test that is more of an integration test (but still using faked github data) showing that the generator will create the empty PRs section if there are PRs with no other change notes.

davisagli: You can see the missing test coverage here: https://coveralls.io/builds/25062620/source?filename=cumulusci%2Ftasks%2Frelease_notes%2Fgenerator.py#L150

cumulusci/tasks/release_notes/generator.py
Br4nd0R: I don't think it ever would be. Removing :)

cumulusci/tasks/release_notes/generator.py
Br4nd0R: I think it would help, especially since it would indicate to a QA/QE what exactly the scope of work was.

Br4nd0R:
@davisagli - The added tests now provide full coverage for the new code; however, I'm not sure I've satisfied the "integration" test that you mentioned earlier. I was attempting to have a test where I can setup mocked PR data and utilize the `GithubReleaseNotesGenerator.__call__` method but we don't currently have the endpoint that `GithubChangeNotesProvider._get_pull_requests()` calls out to mocked under `cumulusci/tasks/release_notes/tests/utils.py --> MockUtil`. 

I can definitely make a mock for the endpoint but I wasn't quite sure how to tie-together pull requests generated by `MockUtil.mock_pull_request` to be returned by the newly mocked endpoint.

Would love to hear your thoughts, and definitely happy to implement this as it would be nice to test all of the parts as a whole.

davisagli:
@Br4nd0R I think you can add the mock response for _get_pull_requests something like this:

        responses.add("GET", "{}/pulls".format(mock_util.repo_url), json=[pr1, pr2])

But I think your existing tests do a good job of testing the pieces and a full integration test isn't super important (especially since the piece that that would add of interacting with github is mostly mocked out anyway).

PR #1174, opened 2019-08-09
Bulk Data: sObject Updates

davidmreed:
# Critical Changes

# Changes

 - `LoadData` now supports the key `action: update` to perform a Bulk API update job
 - `LoadData` now supports an `after: <step name>` on a lookup entry to defer updating that lookup until a dependent sObject step is completed.
 - `GenerateMapping` now handles self-lookups and reference cycles by generating `after:` markers wherever needed. 
- A small issue in `QueryData` affecting mappings using `oid_as_pk: False` has been fixed.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata/generate.py
+++ b/cumulusci/tasks/bulkdata/generate.py
@@ -13,7 +13,7 @@ class GenerateMapping(BaseSalesforceApiTask):
     mapping suitable for extracting data in packaged and custom objects as well as
     customized standard objects.
 
-    Mappings must be serializable, and hence must resolvereference cycles - situations
+    Mappings must be serializable, and hence must resolve reference cycles - situations
     where Object A refers to B, and B also refers to A. Mapping generation will stop
     and request user input to resolve such cycles by identifying the correct load order.
     Alternately, specify the `ignore` option with the name of one of the
@@ -172,7 +172,7 @@ class GenerateMapping(BaseSalesforceApiTask):
                     # First, determine what manner of lookup we have here.
                     referenceTo = self.schema[obj][field]["referenceTo"]
 
-                    if len(referenceTo) > 1:  # Polymorphiy lookup
+                    if len(referenceTo) > 1:  # Polymorphic lookup
                         self.logger.warning(
                             "Field {}.{} is a polymorphic lookup, which is not supported".format(
                                 obj, field
--- a/cumulusci/tasks/bulkdata/generate.py
+++ b/cumulusci/tasks/bulkdata/generate.py
@@ -1,6 +1,7 @@
 from collections import defaultdict, OrderedDict
 
-from cumulusci.core.exceptions import BulkDataException
+import click
+
 from cumulusci.core.utils import ordered_yaml_dump, process_list_arg
 from cumulusci.tasks.salesforce import BaseSalesforceApiTask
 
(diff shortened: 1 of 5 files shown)


cumulusci/tasks/bulkdata/generate.py
```diff
@@ -209,7 +227,12 @@ def _split_dependencies(self, objs, dependencies):
                                 other_obj, ", ".join(dependencies[obj][other_obj])
                             )
                         )
-                raise BulkDataException("Cannot complete mapping")
+                choice = click.prompt(
+                    "Which object should we load first?",
+                    type=click.Choice(list(objs_remaining)),
+                    show_choices=True,
+                )
+                objs_without_deps = [choice]
```
davisagli: Generally, tasks should not assume they're running in the CLI, so should avoid using imports from `click`. That said, we don't (yet?) have a generic framework for tasks to prompt for input in a runtime-environment-agnostic way, and we don't particularly need for this task to be run other than in the CLI, so I'm inclined to let it stand for now.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -82,19 +84,38 @@ def _run_task(self):
...
                     "Job {} did not complete successfully".format(name)
                 )
+            if name in self.after_steps:
+                for after_name, after_step in self.after_steps[name].items():
+                    self.logger.info("Running post-load step: {}".format(after_name))
+                    result = self._load_mapping(after_step)
+                    if result != "Completed":
+                        raise BulkDataException(
+                            "Job {} did not complete successfully".format(after_name)
+                        )
```
davisagli: This line is missing test coverage.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -82,19 +84,38 @@ def _run_task(self):
...
+        action = mapping["action"]
+
+        if action == "insert":
+            job_id = self.bulk.create_insert_job(
+                mapping["sf_object"], contentType="CSV"
+            )
+        else:
+            job_id = self.bulk.create_update_job(
+                mapping["sf_object"], contentType="CSV"
+            )
```
davisagli: This line is missing test coverage.

cumulusci/tasks/bulkdata/generate.py
```diff
@@ -166,19 +169,34 @@ def _build_mapping(self):
...
                         self.logger.warning(
                             "Field {}.{} is a polymorphic lookup, which is not supported".format(
                                 obj, field
                             )
                         )
+                    elif referenceTo[0] == obj:  # Self-lookup
+                        self.mapping[key]["lookups"][field] = {
+                            "table": referenceTo[0].lower(),
+                            "after": key,
+                        }
```
davisagli: This line is missing test coverage.

cumulusci/tasks/bulkdata/generate.py
```diff
@@ -166,19 +169,34 @@ def _build_mapping(self):
...
+                    elif stack.index(referenceTo[0]) > stack.index(
+                        obj
+                    ):  # Dependent lookup
+                        self.mapping[key]["lookups"][field] = {
+                            "table": referenceTo[0].lower(),
+                            "after": "Insert {}".format(referenceTo[0]),
+                        }
```
davisagli: This line is missing test coverage.

cumulusci/tasks/bulkdata/generate.py
```diff
@@ -196,8 +214,8 @@ def _split_dependencies(self, objs, dependencies):
             ]
 
             if not objs_without_deps:
-                self.logger.error(
-                    "Unable to complete mapping; the schema contains reference cycles or unresolved dependencies."
+                self.logger.info(
+                    "CCI needs help to complete the mapping; the schema contains reference cycles and unresolved dependencies."
```
davisagli:
```suggestion
                    "CumulusCI needs help to complete the mapping; the schema contains reference cycles and unresolved dependencies."
```

cumulusci/tasks/bulkdata/generate.py
davidmreed: Do we have any established convention for saying "Task X requires interaction"? It's a really important point, and it might be good to call out this expectation.

cumulusci/tasks/bulkdata/load.py
davidmreed: Sorry about these! I will add coverage.

cumulusci/tasks/bulkdata/generate.py
davisagli: No...it's probably a good idea, because then we could at least make the metadeploy publish task throw an error if you try to publish a flow that includes one of them. But I think the ideal end state is one where inputs like this are included in the list of task options so there's a chance of specifying them completely enough in cumulusci.yml that interaction isn't required.

PR #1187, opened 2019-08-21
Automatically retry UNABLE_TO_LOCK_ROW Apex test failures

davidmreed:
To test this PR against an org, make sure it is generated _without_ the org setting `DisableParallelApexTesting`.

# Critical Changes

# Changes

- The `run_tests` task now supports a `retry_errors` parameter. When all unit test failures' message or stack trace match one of the regular expressions specified there, the failing tests will be retried serially. By default, this will happen only when all unit test failures are due to row locking errors.
- Existing retry options have been removed from `run_tests`; these options controlled retrying result queries.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/apex/testrunner.py
+++ b/cumulusci/tasks/apex/testrunner.py
@@ -9,8 +9,6 @@ import io
 import json
 import re
 
-from collections import defaultdict
-
 from cumulusci.tasks.salesforce import BaseSalesforceApiTask
 from cumulusci.core.exceptions import TaskOptionsError, ApexTestException
 from cumulusci.core.utils import process_bool_arg, process_list_arg, decode_to_unicode
@@ -114,15 +112,8 @@ class RunApexTests(BaseSalesforceApiTask):
         },
         "poll_interval": {
             "description": (
-                "Seconds to wait between polling for Apex test "
-                + "results.  Defaults to 3"
+                "Seconds to wait between polling for Apex test " + "results."
             )
-        },
-        "retry_interval": {
-            "description": "Number of seconds to wait before the next retry (default=5),"
-        },
-        "retry_interval_add": {
-            "description": "Number of seconds to add before each retry (default=5),"
         },
         "junit_output": {
             "description": "File name for JUnit output.  Defaults to test_results.xml"
@@ -132,8 +123,12 @@ class RunApexTests(BaseSalesforceApiTask):
         },
         "retry_errors": {
             "description": "A list of regular expression patterns to match against "
-            "test failures. If all failures match, the failing tests are retried in "
-            "serial mode"
+            "test failures. If failures match, the failing tests are retried in "
+            "serial mode."
+        },
+        "retry_always": {
+            "description": "By default, all failures must match retry_errors to perform "
+            "a retry. Set retry_always to True to retry all failed tests if any failure matches."
         },
     }
 
@@ -165,13 +160,21 @@ class RunApexTests(BaseSalesforceApiTask):
 
         self.options["managed"] = process_bool_arg(self.options.get("managed", False))
 
-        self.options["retries"] = int(self.options.get("retries", 3))
         self.options["retry_errors"] = process_list_arg(
             self.options.get("retry_errors", [])
         )
-        self.options["retry_errors"] = [
-            re.compile(r) for r in self.options["retry_errors"]
-        ]
+        try:
+            self.options["retry_errors"] = [
+                re.compile(r) for r in self.options["retry_errors"]
+            ]
+        except re.error as e:
+            raise TaskOptionsError(
+                "An invalid regular expression was provided ({})".format(e)
+            )
+        self.options["retry_always"] = process_bool_arg(
+            self.options.get("retry_always", False)
+        )
+
         self.counts = {}
 
     # pylint: disable=W0201
@@ -181,6 +184,7 @@ class RunApexTests(BaseSalesforceApiTask):
         self.job_id = None
         self.results_by_class_name = {}
         self.result = None
+        self.retries = None
 
     def _get_namespace_filter(self):
         if self.options["managed"]:
@@ -238,7 +242,9 @@ class RunApexTests(BaseSalesforceApiTask):
     def _get_test_results(self, allow_retries=True):
         result = self.tooling.query_all(TEST_RESULT_QUERY.format(self.job_id))
 
-        retry_details = defaultdict(lambda: list())
+        if allow_retries:
+            self.retry_details = {}
+
         for test_result in result["records"]:
             class_name = self.classes_by_id[test_result["ApexClassId"]]
             self.results_by_class_name[class_name][
@@ -249,20 +255,9 @@ class RunApexTests(BaseSalesforceApiTask):
             # Determine whether this failure is retriable.
             if allow_retries and self._is_retriable_failure(test_result):
                 self.counts["Retriable"] += 1
-                retry_details[test_result["ApexClassId"]].append(
+                self.retry_details.setdefault(test_result["ApexClassId"], []).append(
                     test_result["MethodName"]
                 )
-
-        # Once we've evaluated all of the results, if all failures are retriable,
-        # return the details for retry.
-        if (
-            self.counts["Fail"] > 0
-            and self.counts["CompileFail"] == 0
-            and self.counts["Retriable"] == self.counts["Fail"]
-        ):
-            return retry_details
-
-        return None
 
     def _process_test_results(self):
         test_results = []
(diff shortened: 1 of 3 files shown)


davidmreed:
Open question: in an attempt to guarantee no negative performance impact on builds, I defined the default behavior to be "retry tests if and only if all test failures match given regexes".

However, the use of exception handlers often masks some of the row locking errors. For example, a build will fail with three test failures: two row locks, and one with a custom exception message. The third is _actually_ a row lock, but the underlying failure message is hidden from us.

I'm increasingly leaning towards the idea that `retry_always` should be set to `True` by default, i.e., retry all failures if _any_ failure is identified as matching the defined patterns. I think the potential performance impact is minimal - we'd spuriously retry tests only if a package presented with both row locks and genuine test failures in the same build, and even in this case I think the net outcome would be positive in eliding the row locks and presenting only the genuine failures.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -115,23 +112,24 @@ class RunApexTests(BaseSalesforceApiTask):
         },
         "poll_interval": {
             "description": (
-                "Seconds to wait between polling for Apex test "
-                + "results.  Defaults to 3"
+                "Seconds to wait between polling for Apex test " + "results."
```
davisagli:
```suggestion
                "Seconds to wait between polling for Apex test results."
```

cumulusci/tasks/apex/testrunner.py
```diff
@@ -115,23 +112,24 @@ class RunApexTests(BaseSalesforceApiTask):
...
             )
         },
-        "retries": {"description": "Number of retries (default=10)"},
-        "retry_interval": {
-            "description": "Number of seconds to wait before the next retry (default=5),"
-        },
-        "retry_interval_add": {
-            "description": "Number of seconds to add before each retry (default=5),"
-        },
```
davisagli: Are we sure that no one is using the old retry mechanism?

cumulusci/tasks/apex/testrunner.py
```diff
@@ -168,6 +160,21 @@ def _init_options(self, kwargs):
...
+        self.options["retry_errors"] = process_list_arg(
+            self.options.get("retry_errors", [])
+        )
+        try:
+            self.options["retry_errors"] = [
+                re.compile(r) for r in self.options["retry_errors"]
+            ]
+        except re.error as e:
+            raise TaskOptionsError(
+                "An invalid regular expression was provided ({})".format(e)
```
davisagli: Does this error show _which_ expression failed to compile?

cumulusci/tasks/apex/testrunner.py
```diff
@@ -222,15 +230,36 @@ def _get_test_classes(self):
         self.logger.info("Found {} test classes".format(result["totalSize"]))
         return result
 
-    def _get_test_results(self):
+    def _is_retriable_failure(self, test_result):
```
davisagli: https://english.stackexchange.com/questions/305273/retriable-or-retryable convinced me to let this stand :)

cumulusci/tasks/apex/testrunner.py
```diff
@@ -222,15 +230,36 @@ def _get_test_classes(self):
...
                 test_result["MethodName"]
             ] = test_result
             self.counts[test_result["Outcome"]] += 1
+
+            # Determine whether this failure is retriable.
+            if allow_retries and self._is_retriable_failure(test_result):
+                self.counts["Retriable"] += 1
+                self.retry_details.setdefault(test_result["ApexClassId"], []).append(
+                    test_result["MethodName"]
+                )
```
davisagli: Consider initializing retry_details as defaultdict(list) and then this can just be `self.retry_details[test_result["ApexClassId"]].append(test_result["MethodName"])`

cumulusci/tasks/apex/testrunner.py
```diff
@@ -115,23 +112,24 @@ class RunApexTests(BaseSalesforceApiTask):
...
-        "retry_interval_add": {
-            "description": "Number of seconds to add before each retry (default=5),"
-        },
         "junit_output": {
             "description": "File name for JUnit output.  Defaults to test_results.xml"
         },
         "json_output": {
             "description": "File name for json output.  Defaults to test_results.json"
         },
+        "retry_errors": {
```
davisagli: Should this be `retry_failures`?

davidmreed:
Thanks, @davisagli! Regarding the philosophical question, which is a good one:

Row locking errors are, as far as I know, always non-meaningful _in a unit test_. Row locks that occur in production code are definitely meaningful, but because we cannot control parallelism in a unit test context, it's impossible to test code against production row locks.

Rather, unit test row locks are difficult to predict, impossible to produce intentionally, and represent in essence an artifact of the underlying test runner architecture rather than behavior of the code itself. 

Row locks in unit tests can indicate the pseudo-problem that unit test setup is being written in a way that encourages the unit tests to row lock. I think we'd still surface that in the sense that we do show in the logs that retries have taken place, but it's true that that's not very visible.

@jlantz suggested that perhaps we could ship with two lists, `retry` and `retry_serial`, and allow projects to designate which tests they want retried and how, and more generally that we should make projects be explicit about the desired behavior.

To address both concerns, I think I'd say let's ship with no retries by default and let projects opt in. We can document the patterns used to detect row locks to make that opt-in easy, and projects can decide for themselves if they want `retry_always` turned on.

I'm less enthusiastic about providing support for parallel retries, since that veers away from the territory of fixing row locks and more into the territory of enabling flaky Apex unit tests that should be fixed at the source.

cumulusci/tasks/apex/testrunner.py
davisagli: Some quick searching suggests no at least within our products. We should probably mention the removal of this in the changenotes though.

davidmreed: Good point. I removed it on the basis that it was retrying something that should never fail in a retryable manner - a single Tooling API query. That query should either return no results, or it should fail with some kind of authentication problem that's not retryable (retriable?).

davisagli: Yeah, I agree with removing it.

davisagli:
@davidmreed okay I feel better about this if we're just adding the feature for projects to opt in to but not including a recommended retry policy.

cumulusci/tasks/apex/testrunner.py
davidmreed: Done - thank you.

cumulusci/tasks/apex/testrunner.py
davidmreed: Added to change list.

cumulusci/tasks/apex/testrunner.py
davidmreed: It shows which token failed, but doesn't identify the source expressed.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -314,16 +360,61 @@ def _run_task(self):
...
+        # Did we get back retriable test results? Check our retry policy,
+        # then enqueue new runs individually, until either (a) all retriable
+        # tests succeed or (b) a test fails.
+        able_to_retry = (self.counts["Retriable"] and self.options["retry_always"]) or (
+            self.counts["Retriable"] and self.counts["Retriable"] == self.counts["Fail"]
+        )
+        if not able_to_retry:
+            self.counts["Retriable"] = 0
+        else:
+            self.logger.warning(
```
Br4nd0R: What do you think about extracting this else clause into a method like `_attempt_retries()`?

cumulusci/tasks/apex/testrunner.py
davisagli: Can we compile these one at a time instead of in a comprehension so we can include the failed expression in our own error message?

cumulusci/tasks/apex/testrunner.py
davidmreed: Sold!

cumulusci/tasks/apex/testrunner.py
davisagli: @davidmreed

davidmreed: @davisagli  I'm probably missing something obvious - what's the impact of the terminology?

davisagli: @davidmreed my thinking is just that the key we're looking for in the test results is `Fail`, not `Error`, so calling this errors feels like making up our own terminology.

davidmreed: Oh, I see. `failures` probably makes more sense. I will update it.

PR #1189, opened 2019-08-27
Feature/roll up child branch prs

Br4nd0R:
# Changes
* Child branches of feature branches now have their change notes aggregated in the parent pull request description.

### DIFF
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -77,6 +77,8 @@ def validate_service(options):
 
 def get_pull_requests_with_base_branch(repo, base_branch_name, head=None):
     """Returns a list of pull requests with the given base branch"""
+    if head:
+        head = repo.owner.login + ":" + head
     return list(repo.pull_requests(base=base_branch_name, head=head))
 
 
@@ -109,6 +111,5 @@ def add_labels_to_pull_request(repo, pull_request, *labels):
 def is_label_on_pull_request(repo, pull_request, label_name):
     """Returns True if the given label is on the pull request with the given
     pull request number. False otherwise."""
-    issue = repo.issue(pull_request.number)
-    labels = issue.labels()
+    labels = list(repo.issue(pull_request.number).labels())
     return any(label_name == issue_label.name for issue_label in labels)
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -108,8 +108,8 @@ def create_pull_request(repo, branch_name, base=None, title=None):
 def add_labels_to_pull_request(repo, pull_request, labels):
     """Adds a label to a pull request via the issue object
         Args:
-            repo: github repository object
-            pull_request: ShortPullRequest object
+            repo: Repository object
+            pull_request: ShortPullRequest object that exists in repo
             labels: list(str) of labels to add to the pull request"""
     issue = repo.issue(pull_request.number)
     issue.add_label(labels)
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -97,7 +97,7 @@ def get_pull_request_by_branch_name(repo, branch_name):
 def create_pull_request(repo, branch_name, base=None, title=None):
     """Creates a pull request for the given branch"""
     base = base or "master"
-    title = title or "Auto-Generate Pull Request"
+    title = title or "Auto-Generated Pull Request"
     pull_request = repo.create_pull(title, base, branch_name)
     return pull_request
 
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -80,21 +80,12 @@ def get_pull_requests_with_base_branch(repo, base_branch_name, head=None):
     return list(repo.pull_requests(base=base_branch_name, head=head))
 
 
-def get_pull_request_by_branch_name(repo, branch_name):
-    """Returns a single pull request if found, or None if nothing is returned.
-    Will throw an error if more than one pull request is returned"""
+def get_pull_request_by_head(repo, branch_name):
+    """Returns all pull requests with head equal to the given branch name."""
     if branch_name == repo.default_branch:
         return None
 
-    pull_requests = list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
-    if len(pull_requests) == 0:
-        return None
-    elif len(pull_requests) == 1:
-        return pull_requests[0]
-    else:
-        raise GithubException(
-            "Expected one pull request but received {}".format(len(pull_requests))
-        )
+    return list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
 
 
 def create_pull_request(repo, branch_name, base=None, title=None):
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -120,4 +120,4 @@ def is_label_on_pull_request(repo, pull_request, label_name):
     pull request number. False otherwise."""
     issue = repo.issue(pull_request.number)
     labels = issue.labels()
-    return any(label_name in issue_label.name for issue_label in labels)
+    return any(label_name == issue_label.name for issue_label in labels)
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -80,7 +80,7 @@ def get_pull_requests_with_base_branch(repo, base_branch_name, head=None):
     return list(repo.pull_requests(base=base_branch_name, head=head))
 
 
-def get_pull_request_by_head(repo, branch_name):
+def get_pull_requests_by_head(repo, branch_name):
     """Returns all pull requests with head equal to the given branch name."""
     if branch_name == repo.default_branch:
         return None
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -98,10 +98,7 @@ def create_pull_request(repo, branch_name, base=None, title=None):
     """Creates a pull request for the given branch"""
     base = base or "master"
     title = title or "Auto-Generate Pull Request"
-    try:
-        pull_request = repo.create_pull(title, base, branch_name)
-    except Exception as e:
-        raise GithubException("Error creating pull request:\n{}".format(e))
+    pull_request = repo.create_pull(title, base, branch_name)
     return pull_request
 
 
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -75,9 +75,9 @@ def validate_service(options):
         )
 
 
-def get_pull_requests_with_base_branch(repo, base_branch_name):
+def get_pull_requests_with_base_branch(repo, base_branch_name, head=None):
     """Returns a list of pull requests with the given base branch"""
-    return list(repo.pull_requests(base=base_branch_name))
+    return list(repo.pull_requests(base=base_branch_name, head=head))
 
 
 def get_pull_request_by_branch_name(repo, branch_name):
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -83,7 +83,7 @@ def get_pull_requests_with_base_branch(repo, base_branch_name):
 def get_pull_request_by_branch_name(repo, branch_name):
     """Returns a single pull request if found, or None if nothing is returned.
     Will throw an error if more than one pull request is returned"""
-    if branch_name == "master":
+    if branch_name == repo.default_branch:
         return None
 
     pull_requests = list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -83,6 +83,9 @@ def get_pull_requests_with_base_branch(repo, base_branch_name):
 def get_pull_request_by_branch_name(repo, branch_name):
     """Returns a single pull request if found, or None if nothing is returned.
     Will throw an error if more than one pull request is returned"""
+    if branch_name == "master":
+        return None
+
     pull_requests = list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
     if len(pull_requests) == 0:
         return None
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -82,7 +82,7 @@ def get_pull_requests_with_base_branch(repo, base_branch_name):
 
 def get_pull_request_by_branch_name(repo, branch_name):
     """Returns a single pull request if found, or None if nothing is returned.
-    Will throw an error is more than one pull request is returned"""
+    Will throw an error if more than one pull request is returned"""
     pull_requests = list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
     if len(pull_requests) == 0:
         return None
@@ -102,20 +102,19 @@ def create_pull_request(repo, branch_name, base=None, title=None):
     return pull_request
 
 
-def add_labels_to_pull_request(repo, pull_request, labels):
+def add_labels_to_pull_request(repo, pull_request, *labels):
     """Adds a label to a pull request via the issue object
         Args:
             repo: Repository object
             pull_request: ShortPullRequest object that exists in repo
             labels: list(str) of labels to add to the pull request"""
     issue = repo.issue(pull_request.number)
-    issue.add_label(labels)
+    issue.add_labels(*labels)
 
 
 def is_label_on_pull_request(repo, pull_request, label_name):
     """Returns True if the given label is on the pull request with the given
     pull request number. False otherwise."""
-    return any(
-        label_name in pr_label.name
-        for pr_label in repo.issue(pull_request.number).labels()
-    )
+    issue = repo.issue(pull_request.number)
+    labels = issue.labels()
+    return any(label_name in issue_label.name for issue_label in labels)
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -73,3 +73,52 @@ def validate_service(options):
         raise GithubException(
             "Could not confirm access to the GitHub API: {}".format(str(e))
         )
+
+
+def get_pull_requests_with_base_branch(repo, base_branch_name):
+    """Returns a list of pull requests with the given base branch"""
+    return list(repo.pull_requests(base=base_branch_name))
+
+
+def get_pull_request_by_branch_name(repo, branch_name):
+    """Returns a single pull request if found, or None if nothing is returned.
+    Will throw an error is more than one pull request is returned"""
+    pull_requests = list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
+    if len(pull_requests) == 0:
+        return None
+    elif len(pull_requests) == 1:
+        return pull_requests[0]
+    else:
+        raise GithubException(
+            "Expected one pull request but received {}".format(len(pull_requests))
+        )
+
+
+def create_pull_request(repo, branch_name, base=None, title=None):
+    """Creates a pull request for the given branch"""
+    base = base or "master"
+    title = title or "Auto-Generate Pull Request"
+    try:
+        pull_request = repo.create_pull(title, base, branch_name)
+    except Exception as e:
+        raise GithubException("Error creating pull request:\n{}".format(e))
+    return pull_request
+
+
+def add_labels_to_pull_request(repo, pull_request, labels):
+    """Adds a label to a pull request via the issue object
+        Args:
+            repo: github repository object
+            pull_request: ShortPullRequest object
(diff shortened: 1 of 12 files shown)


cumulusci/core/tests/test_github.py
```diff
@@ -31,8 +40,9 @@ def isclosed(self):
         return True
 
 
-class TestGithub(unittest.TestCase):
-    def tearDown(self):
+class TestGithub:
+    @classmethod
+    def teardown(cls):
```
davisagli: Can you confirm that this is being called after each test? I think it might need to be defined like this:

    def teardown_method(self, method):

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,83 @@ def _init_change_notes(self):
         return DirectoryChangeNotesProvider(self, self.directory)
 
 
+class ParentPullRequestNotesGenerator(BaseReleaseNotesGenerator):
+    """Aggregates notes from child pull requests in to a parent pull request
+    Attributes:
+        UNAGGREGATED_SECTION_HEADER: Str header for section on pull requests where
+        unaggregated chile pull requests are linked to.
```
davisagli:
```suggestion
        unaggregated child pull requests are linked to.
```

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,83 @@ def _init_change_notes(self):
...
+
+    def _init_parsers(self):
+        """Invoked from Super Class"""
+        for cfg in self.parser_config:
+            parser_class = import_global(cfg["class_path"])
+            self.parsers.append(parser_class(self, cfg["title"]))
+
+        # New parser to collect developer notes above tracked headers
+        self.parsers.append(GithubLinesParser(self, "Notes From Child PRs"))
+        self.parsers[-1]._in_section = True
```
davisagli: With this here, is https://github.com/SFDO-Tooling/CumulusCI/pull/1189/files#diff-df899217f555b1934861052e5487ca3dR61 still necessary?

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,83 @@ def _init_change_notes(self):
...
 
+class ParentPullRequestNotesGenerator(BaseReleaseNotesGenerator):
+    """Aggregates notes from child pull requests in to a parent pull request
+    Attributes:
+        UNAGGREGATED_SECTION_HEADER: Str header for section on pull requests where
+        unaggregated chile pull requests are linked to.
+
+    Methods:
+        aggregate_child_change_notes: Aggregates change notes from child prs into parent
+        update_unaggregated_pr_header: Adds a new child pr to the section
```
davisagli: I'd leave the Attributes and Methods sections out of the class docstring and instead make sure the attribute has a comment next to it and the methods have their own docstrings.

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,83 @@ def _init_change_notes(self):
...
+        for cfg in self.parser_config:
+            parser_class = import_global(cfg["class_path"])
+            self.parsers.append(parser_class(self, cfg["title"]))
+
+        # New parser to collect developer notes above tracked headers
+        self.parsers.append(GithubLinesParser(self, "Notes From Child PRs"))
+        self.parsers[-1]._in_section = True
+
+    def aggregate_child_change_notes(self, pull_request):
+        """Aggregates all change notes from child pull reqeusts.
```
davisagli:
```suggestion
        """Aggregates all change notes from child pull requests.
```

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,83 @@ def _init_change_notes(self):
...
+        Child pull reqeusts are pull requests that have a base branch
+        equal to the the given pull requests head."""
```
davisagli:
```suggestion
        equal to the the given pull request's head."""
```

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,83 @@ def _init_change_notes(self):
...
+        self.change_notes = get_pull_requests_with_base_branch(
+            self.repo, pull_request.head.label.split(":")[1]
+        )
+        if len(self.change_notes) == 0:
+            raise CumulusCIException(
+                "No pull requests with base branch {} found.".format(pull_request.head)
+            )
```
davisagli: It's pretty normal if there are no inbound pull requests yet, so I think we want to silently return in this case (maybe log something) rather than raising an exception.

cumulusci/cumulusci.yml
```diff
@@ -110,6 +110,10 @@ tasks:
         description: Retrieves a list of the currently installed managed package namespaces and their versions
         class_path: cumulusci.tasks.salesforce.GetInstalledPackages
         group: Salesforce Metadata
+    github_parent_pr_notes:
+        description: Merges the description of a child PRs to the respective Parent's PR (if one exists).
+        class_path: cumulusci.tasks.release_notes.task.ParentPullRequestNotes
+        group: GitHub
```
davisagli: We need to talk through how this will fit into our MetaCI Plans. It probably makes sense to add it to the ci_feature flow which is triggered on every commit to a feature branch. That has some implications: The task can assume it's being called from a feature branch, but the flow won't be able to work out the parent & child branch names to pass to the task. The task will need to look at self.project_config.repo_branch to know what the current branch is, and determine whether it's a parent or child (based on whether it has a double-underscore or not) and behave accordingly.

davisagli: Okay, that sounds good and I think I was just a bit confused.

You should be able to add a step like this to the ci_feature flow yaml to pass in the branch name:

    task: github_parent_pr_notes
    options:
        branch_name: $project_config.repo_branch

cumulusci/tasks/release_notes/generator.py
```diff
@@ -95,6 +103,78 @@ def _init_change_notes(self):
...
+        body = pull_request_to_update.body
+        if self.UNAGGREGATED_SECTION_HEADER not in body:
+            body += self.UNAGGREGATED_SECTION_HEADER
+
+        pull_request = get_pull_request_by_branch_name(self.repo, branch_name_to_add)
+        if pull_request:
+            # TODO: Should we alert user if PR isn't found?
+            pull_request_link = markdown_link_to_pr(pull_request)
+            if pull_request_link not in body:
+                body += "\r\n* " + markdown_link_to_pr(pull_request)
```
davisagli:
```suggestion
                body += "\r\n* " + pull_request_link
```

cumulusci/tasks/release_notes/task.py
```diff
@@ -1,6 +1,24 @@
+# -*- coding: utf-8 -*-
+"""Release Note Tasks
+
+Classes:
+    GithubReleaseNotes
+    ParentPullRequestNotes
```
davisagli: I'd leave this out. It's liable to become out of date compared to something like the VSCode Outline panel that is actually aware of the code.

cumulusci/tasks/release_notes/task.py
```diff
@@ -55,3 +73,127 @@ def _run_task(self):
...
+    isn't found, one is created and the BUILD_NOTES_LABEL is applied to it.
+
+    If the BUILD_NOTES_LABEL is present on the pull request, then all notes from the 
+    child pull request are aggregated into the parent pull request. If the BUILD_NOTES_LABEL
+    is not detected on the parent pull request then a link to the child pull request
+    is placed under the "Unaggregated Pull Reqeusts" header.
+
+    When given the parent_branch_name option, this task will query for a corresponding pull request.
+    If a pull request is not found, the task exits. If a pull request is found, then all notes
+    from child pull requests are re-aggregated and the body of the parent is replace entirely.
```
davisagli:
```suggestion
    from child pull requests are re-aggregated and the body of the parent is replaced entirely.
```

cumulusci/tasks/release_notes/task.py
```diff
@@ -55,3 +73,127 @@ def _run_task(self):
...
+    def _run_task(self):
+        branch_name = self.options.get("branch_name")
+        parent_branch_name = self.options.get("parent_branch_name")
+
+        if (not branch_name and not parent_branch_name) or (
+            branch_name and parent_branch_name
+        ):
+            raise TaskOptionsError(
+                "You must specify either branch_name or (exclusive) parent_branch_name."
+            )
```
davisagli: Option validation should go in the task's _init_options method. (That way it's raised when the task is constructed instead of when it's running.)

cumulusci/tasks/release_notes/task.py
```diff
@@ -55,3 +73,127 @@ def _run_task(self):
...
+            self.logger.info(
+                "Pull request not found for branch: {}.\nExiting...".format(branch_name)
+            )
+            return
+
+        base_branch = pull_request.base.ref
+        feature_prefix = self.project_config.project__git__prefix_feature
+        if not base_branch.startswith(feature_prefix):
+            self.logger.info("Pull request's base is not a feature branch.\nExiting...")
+            return
```
davisagli: We don't need the child pull request in order to find the parent pull request. We can do it from the parent/child branch naming convention:

1. If branch_name doesn't have a double underscore, we can exit early (it's not a child)
2. Split branch_name on double underscore and take the first part as the parent branch name (i.e. starting with feature/parent__child we expect the parent branch name to be feature/parent)
3. Look for a pull request where head = the parent branch name and base is repo.default_branch. This should be the parent pull request.

cumulusci/tasks/release_notes/task.py
```diff
@@ -55,3 +73,127 @@ def _run_task(self):
...
+    child pull request are aggregated into the parent pull request. If the BUILD_NOTES_LABEL
+    is not detected on the parent pull request then a link to the child pull request
+    is placed under the "Unaggregated Pull Reqeusts" header.
+
+    When given the parent_branch_name option, this task will query for a corresponding pull request.
+    If a pull request is not found, the task exits. If a pull request is found, then all notes
+    from child pull requests are re-aggregated and the body of the parent is replace entirely.
+    """
+
+    BUILD_NOTES_LABEL = "Build Change Notes"
```
davisagli: Why don't we turn this into a task option. Then if a project wants to use a different lable they can configure it in cumulusci.yml

cumulusci/tasks/release_notes/task.py
```diff
@@ -55,3 +73,127 @@ def _run_task(self):
...
+        parent_pull_request = self._get_parent_pull_request(base_branch)
+        if is_label_on_pull_request(
+            self.repo, parent_pull_request, self.BUILD_NOTES_LABEL
+        ):
+            generator.aggregate_child_change_notes(parent_pull_request)
+        else:
+            generator.update_unaggregated_pr_header(parent_pull_request, branch_name)
+
+    def _handle_parent_branch_name_option(self, generator, parent_branch_name):
+        pull_request = get_pull_request_by_branch_name(self.repo, parent_branch_name)
```
davisagli: As I mentioned above, it's quite likely that there will be multiple pull requests where head is the parent branch. We should look through them for one where base is the repo.default_branch

cumulusci/tasks/release_notes/task.py
```diff
@@ -55,3 +73,127 @@ def _run_task(self):
...
+        elif is_label_on_pull_request(self.repo, pull_request, self.BUILD_NOTES_LABEL):
+            # We can only aggregate child change notes when given the parent_branch option
+            #
+            # We aren't able to append to the 'Unaggregated Pull Reqeusts' header.
+            # We don't know at what time the label was applied to the pull request;
+            # and therefore, we cannot determine which child pull requests are already
+            # aggregated into the parent pull request, and which ones should be
+            # included in the 'Unaggregated Pull Requests' section.
+            generator.aggregate_child_change_notes(pull_request)
+        else:
```
davisagli: Since running the task with parent_branch_name specified is intended as a manual task only, I'm sort of inclined to say we should go ahead and aggregate the child change notes whether or not the label is there. That way if a project doesn't want to use the label and automatic updates at all but still wants to aggregate manually from time to time, they can do so.

cumulusci/tasks/release_notes/tests/test_task.py
```diff
@@ -23,9 +25,292 @@ def test_run_task(self, GithubReleaseNotesGenerator):
...
         generator.assert_called_once()
+
+
+class TestParentPullRequestNotes(GithubApiTestMixin):
+
+    BRANCH_NAME = "test-branch"
+
+    @pytest.fixture
+    def gh_api(self):
+        return get_github_api("TestOwner", "TestRepo")
```
davisagli: We can put reused pytest fixtures in a conftest.py module (either in one particular test folder or at the root of cumulusci) and then they'll be available to all tests without needing to redefine them.

cumulusci/tasks/release_notes/tests/test_task.py
```diff
@@ -23,9 +25,292 @@ def test_run_task(self, GithubReleaseNotesGenerator):
...
+    def task_provider(self, project_config):
+        def _task_provider(options):
+            task_config = TaskConfig(options)
+            task = ParentPullRequestNotes(project_config, task_config)
+            task.repo = mock.Mock()
+            task.logger = mock.Mock()
+            task.github = mock.Mock()
+            return task
+
+        return _task_provider
```
davisagli: This is clever. I might call it a factory instead of a provider.

cumulusci/core/github.py
```diff
@@ -73,3 +73,51 @@ def validate_service(options):
...
+            "Expected one pull request but received {}".format(len(pull_requests))
+        )
+
+
+def create_pull_request(repo, branch_name, base=None, title=None):
+    """Creates a pull request for the given branch"""
+    base = base or "master"
+    title = title or "Auto-Generated Pull Request"
+    pull_request = repo.create_pull(title, base, branch_name)
+    return pull_request
```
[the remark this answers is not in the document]
davisagli: (Following up later now that I'm looking at the tests, I see you're testing these separately and then mocking them out for some other tests. Testability is a pretty legit reason to keep these separate.)

cumulusci/core/tests/test_github.py
```diff
@@ -2,17 +2,31 @@
...
 import unittest
 
-from github3.session import AppInstallationTokenAuth
 import mock
 import responses
 
+from github3.repos.repo import Repository
+from github3.pulls import ShortPullRequest
+from github3.exceptions import ConnectionError
+from github3.session import AppInstallationTokenAuth
```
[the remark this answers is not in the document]
Br4nd0R: I like it. I can rework the files I touched to adhere to that format.
.

cumulusci/core/tests/test_github.py
```diff
@@ -101,11 +129,91 @@ def test_get_github_api_for_repo__not_installed(self, create_token):
...
+            pulls=self.get_expected_pull_requests(1),
+            head=repo.owner.login + ":" + "some-other-branch",
+        )
+        pull_request = get_pull_request_by_branch_name(repo, "some-other-branch")
+        assert pull_request is not None
+
+        # ConnectionError present when we reachout with
+        # a branch name (url parameter) that we aren't expecting
+        with pytest.raises(ConnectionError):
+            get_pull_request_by_branch_name(repo, "does-not-exist")
```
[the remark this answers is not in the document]
Br4nd0R: @davisagli - I think my comment can be improved here. In this case the `ConnectionError` is coming from the `responses` module. In this case, we add a response that contains the `head=` url parameter. The first time we call `get_pull_request_by_branch_name` the given branch name matches the expected parameter value.

In the next case we use a value for `head` that doesn't exist and so the endpoint doesn't fully match the response that we added so it fails with a `ConnectionError`. 

[The docs](https://github.com/getsentry/responses) state, 'If you attempt to fetch a url which doesn't hit a match, responses will raise a ConnectionError'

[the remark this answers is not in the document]
davisagli: Oh that makes more sense -- but then I think you want to import ConnectionError from requests.exceptions instead of github3.exceptions? (The latter must coincidentally import the former, since the test works. But it was the github3 import that confused me about where the ConnectionError comes from.)

cumulusci/core/github.py
```diff
@@ -73,3 +73,51 @@ def validate_service(options):
...
+    issue = repo.issue(pull_request.number)
+    issue.add_labels(*labels)
+
+
+def is_label_on_pull_request(repo, pull_request, label_name):
+    """Returns True if the given label is on the pull request with the given
+    pull request number. False otherwise."""
+    issue = repo.issue(pull_request.number)
+    labels = issue.labels()
+    return any(label_name in issue_label.name for issue_label in labels)
```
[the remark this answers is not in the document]
Br4nd0R: good catch. I think we want `any(label_name == issue_label.name for ...`

cumulusci/core/tests/test_github.py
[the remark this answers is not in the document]
Br4nd0R: ahh ok, didn't even think of that. I'll switch the import and rerun tests to ensure all is well.

cumulusci/tasks/release_notes/task.py
Br4nd0R: I might be missing something here, but wouldn't that mean the following scenario would always be possible:
(1) Doc writer begins to make manual changes to the notes on a long running feature branch.
(2) After those manual changes occur a new commit comes through on a child branch
(3) The commit triggers this task to execute with the 'branch_name' option which no longer checks for the presence of the label, and automatically re-aggregates all of the change notes, which overwrites all of the manual changes that the writer made.

Also, there wouldn't be the need for the 'Unaggregated Pull Requests' section of the change notes.

cumulusci/tasks/github/tests/util_github_api.py
```diff
@@ -361,7 +361,10 @@ def _get_expected_pull_request(
         pr.update(kw)
         return pr
 
-    def _get_expected_issue(self, issue_number, owner=None, repo=None):
+    def get_expected_pull_requests(self, num_pull_requests):
```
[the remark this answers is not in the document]
Br4nd0R: Yes we can!

cumulusci/tasks/release_notes/tests/test_task.py
Br4nd0R: Done.

cumulusci/core/tests/test_github.py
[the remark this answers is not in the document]
davisagli: Oh I see, github3 is catching the requests ConnectionError and reraising it as its own exception class. (in models.py in github3)

cumulusci/tasks/release_notes/task.py
davisagli: This comment is on the _handle_parent_branch_name_option method -- I was only proposing a change for when parent_branch_name is specified, not branch_name.

cumulusci/tasks/release_notes/generator.py
```diff
@@ -51,6 +57,8 @@ def _parse_change_note(self, change_note):
         line_added_by_parsers = False
         for parser in self.parsers:
             line_added = parser.parse(change_note)
+            if parser.title == "Notes From Child PRs":
+                parser._in_section = True
```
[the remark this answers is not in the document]
Br4nd0R: @davisagli - When we init parser (in the link you provided) we set `_in_section` to `True` 
so we can capture any developer notes in the first change note processed. For any subsequent change_notes being parsed in the transaction we need to reset `_in_section` back to `True` in order to it to pick up dev notes again.

cumulusci/tasks/release_notes/task.py
Br4nd0R: done!

cumulusci/core/github.py
```diff
@@ -73,3 +73,51 @@ def validate_service(options):
...
+
+    pull_requests = list(repo.pull_requests(head=repo.owner.login + ":" + branch_name))
+    if len(pull_requests) == 0:
+        return None
+    elif len(pull_requests) == 1:
+        return pull_requests[0]
+    else:
+        raise GithubException(
+            "Expected one pull request but received {}".format(len(pull_requests))
+        )
```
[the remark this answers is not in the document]
Br4nd0R: Done.

cumulusci/tasks/release_notes/task.py
Br4nd0R: done.

PR #1191, opened 2019-08-29
Updates to persistent org connections

davisagli:
# Critical Changes

# Changes
* There is now a default CumulusCI global connected app that can be used to connect to persistent orgs (assuming you know the credentials) without creating a new connected app. It's still possible to configure a custom connected app using `cci service connect connected_app` if more control over the connected app settings is needed.
* When CumulusCI is being run in a non-interactive context it can now obtain an access token for a persistent org using a JWT instead of a refresh token. This approach is used if the SFDX_CLIENT_ID and SFDX_HUB_KEY environment variables are set. This makes it easier to manage persistent org connections in the context of a hosted service because it's possible to replace the connected app's certificate without needing to obtain new refresh tokens for each org.

# Issues Closed

### DIFF
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -9,8 +9,6 @@ from cumulusci.core.exceptions import SalesforceCredentialsException
 from cumulusci.oauth.salesforce import SalesforceOAuth2
 from cumulusci.oauth.salesforce import jwt_session
 
-SFDX_CLIENT_ID = os.environ.get("SFDX_CLIENT_ID")
-SFDX_HUB_KEY = os.environ.get("SFDX_HUB_KEY")
 
 class OrgConfig(BaseConfig):
     """ Salesforce org configuration (i.e. org credentials) """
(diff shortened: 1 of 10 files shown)


cumulusci/core/config/OrgConfig.py
```diff
@@ -19,6 +22,23 @@ def __init__(self, config, name):
         super(OrgConfig, self).__init__(config)
 
     def refresh_oauth_token(self, keychain, connected_app=None):
+        SFDX_CLIENT_ID = os.environ.get("SFDX_CLIENT_ID")
+        SFDX_HUB_KEY = os.environ.get("SFDX_HUB_KEY")
+        if SFDX_CLIENT_ID and SFDX_HUB_KEY:
+            login_url = (
+                "https://test.salesforce.com"
```
Br4nd0R: Would it make sense to have `SBX_LOGIN_URL` and `PROD_LOGIN_URL` attributes on `BaseConfig`?

davisagli: I guess we could turn them into constants somewhere, and that could be helpful if someone wants to use CumulusCI with one of our internal test environments or when running the core application locally. I'd rather work on that in a separate PR when we can get input from someone who actually needs to do that, though.

cumulusci/core/keychain/BaseProjectKeychain.py
```diff
@@ -11,6 +12,15 @@
 from cumulusci.core.sfdx import sfdx
 
 
+DEFAULT_CONNECTED_APP = ConnectedAppOAuthConfig(
+    {
+        "client_id": "3MVG9i1HRpGLXp.or6OVlWVWyn8DXi9xueKNM4npq_AWh.yqswojK9sE5WY7f.biP0w7bNJIENfXc7JMDZGO1",
```
Br4nd0R: Hardcoding this is ok, as we don't expect to have to ever switch from the default connected app, correct?

davisagli: That's the idea. If we did need to delete the existing connected app, users would need to upgrade to a new version of CumulusCI or configure their own connected app (the way they've needed to until now).

PR #1192, opened 2019-08-29
Fix returning cached preflight task results

davisagli:
@davidmreed This should fix the problem we saw with cached preflight task results. Do you see any easy way to improve the test so that we're actually testing for that?

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -704,7 +704,7 @@ class CachedTaskRunner(object):
     def __call__(self, **options):
         cache_key = (self.task_name, tuple(options.items()))
         if cache_key in self.cache.results:
-            return self.cache.results[cache_key]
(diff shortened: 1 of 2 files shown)


davidmreed:
What about replicating the structure in V4S's installer that first triggered the bug? Two consecutive references to the same preflight check, with the first expected to pass based on the test configuration, and the second expected to fail? I think that could be asserted against based on how the test is laid out now.

davisagli:
@davidmreed yeah, sounds good. I adjusted the test and confirmed that it only passes with the fix in place.

davidmreed:
Looks like py27 is failing, but only on Windows.

davisagli:
@davidmreed thanks for calling my attention to that. I added a sort to make sure the dict items end up in a stable order for the cache key.

PR #1196, opened 2019-09-03
Feature/factory task base

prescod:
A base class for Factory-boy derived data generators.

Look at tests/dummy_data_factory.py to see what the API looks like from a consumer's point of view.

The gist is that you have one class derived from ModuleDataFactory with a make_records method and a series of FactoryBoy classes inheriting from factory.alchemy.SQLAlchemyModelFactory for individual tables.

The extra classes are picked up with some introspection magic so you don't need to "register" them.

There is also some magic which attaches them to the right session dynamically.

### DIFF
--- a/cumulusci/tasks/bulkdata/tests/dummy_data_factory.py
+++ b/cumulusci/tasks/bulkdata/tests/dummy_data_factory.py
@@ -1,14 +1,5 @@
-from datetime import date, datetime
-
 import factory
 from cumulusci.tasks.bulkdata.factory_utils import ModuleDataFactory, Models
-
-
-def now():
-    return datetime.now().date()
-
-
-START_DATE = date(2019, 1, 1)  # Per https://salesforce.quip.com/gLfGAPtqVzUS
 
 
 class GenerateDummyData(ModuleDataFactory):
--- a/cumulusci/tasks/bulkdata/tests/dummy_data_factory.py
+++ b/cumulusci/tasks/bulkdata/tests/dummy_data_factory.py
@@ -12,7 +12,7 @@ START_DATE = date(2019, 1, 1)  # Per https://salesforce.quip.com/gLfGAPtqVzUS
 
 
 class GenerateDummyData(ModuleDataFactory):
-    """Generate data specific to a particular use case"""
+    """Generate data based on test mapping.yml"""
 
     def make_records(self, num_records, factories):
         factories.create_batch("Contact", 10)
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/bulkdata/tests/dummy_data_factory.py
```diff
@@ -0,0 +1,41 @@
+from datetime import date, datetime
+
+import factory
+from cumulusci.tasks.bulkdata.factory_utils import ModuleDataFactory, Models
+
+
+def now():
+    return datetime.now().date()
```
davisagli: This appears to be unused.

cumulusci/tasks/bulkdata/tests/dummy_data_factory.py
```diff
@@ -0,0 +1,41 @@
...
+
+class Contact(factory.alchemy.SQLAlchemyModelFactory):
+    class Meta:
+        model = Models.contacts
+        exclude = [
+            "household"
+        ]  # don't try to attach the household to the Contact directly
+
+    id = factory.Sequence(lambda n: n + 1)
+    household = factory.SubFactory(Household)  # create households automatically
```
davisagli: This seems a bit backward as an example since Contact has a master-detail relationship to Account (of which Household is a record type). With real data it would be unusual for each contact to have a unique household rather than having some households with multiple contacts.

prescod: I find that things are often created "backwards" in these factories based on the way the specification is organized or based on how the code is more easily written.

This organization of the factories confirms to the constraint that "every contact will have its household_id field filled out" which might be a useful constraint in a test.

According to mapping_v2.yml, "contact" is something that "knows about" household because it has a household_id column. Making the account also "know about" the household seemed overcomplicated to me. Essentially the household would be responsible for setting the household_id of the contact which felt slightly icky.

But it all depends on the requirements. If there were supposed be 3 contacts per household then yes I'd want the households to create the contacts and if some contacts were created without households (through some other entry point) then that presumably would be okay.

I have written the code both ways so if you feel strongly about it I can switch to households creating contacts.

davisagli: Okay, fair enough.

davisagli:
@prescod there are some new lines missing test coverage: https://coveralls.io/builds/25535128/source?filename=cumulusci%2Ftasks%2Fbulkdata%2Ffactory_utils.py#L31

PR #1197, opened 2019-09-04
Feature/delete with WHERE clause

prescod:
Allow WHERE clauses on delete operations so I can clean up records in an org selectively (especially in LDV orgs where I want to clean up new data).


cumulusci/tasks/bulkdata/delete.py
```diff
@@ -51,11 +56,24 @@ def _create_job(self, obj):
         self.bulk.close_job(delete_job)
         return delete_job
 
-    def _query_salesforce_for_records_to_delete(self, obj):
+    def compose_query(self, obj, criteria):
```
davidmreed: Should this be `_compose_query()`?

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -51,11 +56,24 @@ def _create_job(self, obj):
...
+        query = "select Id from {}".format(obj)
```
davidmreed: Can we use typical SOQL query formatting, with keywords in uppercase (`SELECT Id FROM ...`)?

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -14,6 +14,10 @@ class DeleteData(BaseSalesforceApiTask, BulkJobTaskMixin):
             "description": "A list of objects to delete records from in order of deletion.  If passed via command line, use a comma separated string",
             "required": True,
         },
+        "criteria": {
+            "description": "A criteria using the same syntax as a where-clause",
```
davidmreed: It seems atypical that the same criteria can be applied to the queries for all of the sObjects being deleted (if more than one). Do you see the use case as being a single-object run with a WHERE clause, or is there a need to apply WHEREs to multiple sObjects?

cumulusci/tasks/bulkdata/delete.py
prescod: Great question!

I actually decided to do it this way to enable sub-classing so I consider this part of the public interface in case you wanted to do something more sophisticated than a WHERE clause.

"Joins" in particular, I think.

cumulusci/tasks/bulkdata/delete.py
prescod: Sure, I agree that two features of this task are not entirely orthogonal. And I was worried about that at first.

Yes: you're unlikely to use the multi-object and the WHERE features together. 

But trying to make the two features work together nicely very quickly turns into a messy science project. Something like:

objects=account:name LIKE '%foo%',contact:name='foo'

I felt it was better to just let users pick whether they want to take advantage of the multi-object mode or the WHERE mode. If in some obscure case they happen to work together properly for you then that's fine as well.

One can always achieve multi-object deletes like this (using simplified Robot syntax):

Run Task   DeleteData    objects=account   criteria=name LIKE '%foo%
Run Task   DeleteData    objects=contact    criteria=name='foo'

The performance will be almost identical, I think, because the DeleteData task itself is just a loop over the object types.

davidmreed: What would you say to just adding a note to the option description to say "This option is only available when deleting a single object's records", and a check when we init options that it's only populated if the length of the object list is 1?

davisagli: I can understand that argument but I do think we should check and throw an exception if you try to specify criteria and multiple objects at the same time.

Also I wonder if we should just call this `where` instead of `criteria`. Criteria makes me think of Salesforce workflow rule criteria which have a different structure. https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/customfield.htm#filteritem

cumulusci/tasks/bulkdata/delete.py
prescod: be943daa57569c76ca1a48e8b234feade4de71a4

cumulusci/tasks/bulkdata/delete.py
prescod: Check with exception: 85f710dd54e23f4f376a5f229e50bf28b8732a30

Rename 'criteria' to 'where': b0744bd1aa42c51368cd638a8f07ea49e06f2524

PR #1200, opened 2019-09-10
Add Domain column to cci org list

bethbrains:
# Critical Changes

# Changes
-- Adds a column to `cci org list` table to show myDomain URL and pod to quickly compare to browser URL

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -947,7 +947,7 @@ def org_list(config, plain):
     plain = plain or config.global_config.cli__plain_output
     header = ["Name", "Default", "Username"]
     persistent_data = [header]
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cci.py
```diff
@@ -959,7 +959,12 @@ def org_list(config, plain):
         row = [org, org_config.default]
         if isinstance(org_config, ScratchOrgConfig):
             org_days = org_config.format_org_days()
-            row.extend([org_days, org_config.expired, org_config.config_name])
+            instance_url = org_config.instance_url if org_config.expired == False else ''
+            try:
+                instance_url = instance_url.split('//')[-1].rsplit('.',3)[0]
+            except Exception:
```
jstvz: Should we be more explicit about what we're catching here?

PR #1204, opened 2019-09-13
Add method for cleanly asking if an org is alive

prescod:
Clean up scratch org aliveness testing, per your suggestion.

None of these functions are used as often as I would have guessed so it wasn't a big deal yet. Cleaning up the API now will avoid that as the code base grows.

### DIFF
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -142,8 +142,12 @@ class ScratchOrgConfig(OrgConfig):
         return self.config.setdefault("days", 1)
 
     @property
+    def alive(self):
+        return self.date_created and not self.expired
+
+    @property
     def expired(self):
-        return self.expires and self.expires < datetime.datetime.now()
+        return bool(self.expires) and self.expires < datetime.datetime.now()
 
     @property
     def expires(self):
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -142,7 +142,7 @@ class ScratchOrgConfig(OrgConfig):
         return self.config.setdefault("days", 1)
 
     @property
-    def alive(self):
+    def active(self):
         """Check if an org is alive"""
         return self.date_created and not self.expired
 
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -143,10 +143,12 @@ class ScratchOrgConfig(OrgConfig):
 
     @property
     def alive(self):
+        """Check if an org is alive"""
         return self.date_created and not self.expired
 
     @property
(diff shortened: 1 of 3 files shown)


cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -141,9 +141,13 @@ def email_address(self):
     def days(self):
         return self.config.setdefault("days", 1)
 
+    @property
+    def alive(self):
```
davisagli: Let's call it `active`, because that's the same terminology used in platform documentation about scratch orgs.

cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -141,9 +141,13 @@ def email_address(self):
...
+        return self.date_created and not self.expired
+
     @property
     def expired(self):
-        return self.expires and self.expires < datetime.datetime.now()
+        return bool(self.expires) and self.expires < datetime.datetime.now()
```
davisagli: Is there an actual change in behavior here or is it just to be clearer about what we're expecting to get?

cumulusci/core/config/ScratchOrgConfig.py
prescod: That's fine.

Do you also want me to rename the (older) days_alive method or leave it?

davisagli: Leave it.

cumulusci/core/config/ScratchOrgConfig.py
prescod: Previously this method could return None/True/False.

True = Expired
False = Active
None = Never created so not expired

Not it returns True or False.

True = Expired
False = Active or Never created so not expired

Trinary is kind of unusual and confusing.

davisagli: Oh yeah, I see what you mean now.

cumulusci/core/config/ScratchOrgConfig.py
davisagli: @prescod I'm still waiting for you to rename this.

prescod: https://github.com/SFDO-Tooling/CumulusCI/blob/cleanup_org_state_api/cumulusci/core/config/ScratchOrgConfig.py#L145

cumulusci/core/tests/test_config_expensive.py
```diff
@@ -457,11 +457,29 @@ def test_expires(self, Command):
         config.date_created = now
         self.assertEqual(config.expires, now + timedelta(days=1))
 
+        config = ScratchOrgConfig({"days": 1}, "test")
+        now = datetime.now()
```
davisagli: Unused

prescod: 84f0cc48

Issue #1208, opened 2019-09-13
Add Community Publish functionality to CCI


MatthewBlanski:
For the full service on community handling with CCI, adding [community publish](https://developer.salesforce.com/docs/atlas.en-us.chatterapi.meta/chatterapi/connect_resources_communities_publish.htm) from the chatter API would be ideal.

davisagli:
Added in cumulusci 2.5.9

PR #1209, opened 2019-09-13
Publish Communities Using the Chatter API

MatthewBlanski:
# Critical Changes

# Changes
Added a task to publish communities using the chatter api

# Issues Closed
[1208](https://github.com/SFDO-Tooling/CumulusCI/issues/1208)

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -179,6 +179,10 @@ tasks:
         description: Publish a release to the MetaDeploy web installer
         class_path: cumulusci.tasks.metadeploy.Publish
         group: Release Management
+    publish_community:
+        description: Publishes a Community in the target org using the Connect API
+        class_path: cumulusci.tasks.salesforce.PublishCommunity
+        group: Salesforce
     push_all:
         description: Schedules a push upgrade of a package version to all subscribers
         class_path: cumulusci.tasks.push.tasks.SchedulePushOrgQuery
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -182,7 +182,7 @@ tasks:
     publish_community:
         description: Publishes a Community in the target org using the Connect API
         class_path: cumulusci.tasks.salesforce.PublishCommunity
-        group: Salesforce
+        group: Salesforce Communities
     push_all:
         description: Schedules a push upgrade of a package version to all subscribers
         class_path: cumulusci.tasks.push.tasks.SchedulePushOrgQuery
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -110,6 +110,10 @@ tasks:
         description: Retrieves a list of the currently installed managed package namespaces and their versions
         class_path: cumulusci.tasks.salesforce.GetInstalledPackages
         group: Salesforce Metadata
+    github_parent_pr_notes:
+        description: Merges the description of a child pull request to the respective parent's pull request (if one exists).
+        class_path: cumulusci.tasks.release_notes.task.ParentPullRequestNotes
+        group: GitHub
     github_clone_tag:
         description: Clones a github tag under a new name.
         class_path: cumulusci.tasks.github.CloneTag
@@ -417,6 +421,11 @@ flows:
                 task: run_tests
             5:
                 task: github_parent_to_children
+            6:
+                task: github_parent_pr_notes
+                options:
+                    branch_name: $project_config.repo_branch
+                    build_notes_label: "Build Notes"
 
     ci_feature_beta_deps:
         description: Install the latest beta version of dependencies and run apex tests.
--- a/cumulusci/tasks/salesforce/PublishCommunity.py
+++ b/cumulusci/tasks/salesforce/PublishCommunity.py
@@ -19,9 +19,6 @@ class PublishCommunity(BaseSalesforceApiTask):
         },
     }
 
-    def _init_options(self, kwargs):
-        super(PublishCommunity, self)._init_options(kwargs)
-
     def _run_task(self):
         community_id = self.options.get("community_id", None)
         community_name = self.options.get("name", None)
--- a/cumulusci/tasks/salesforce/PublishCommunity.py
+++ b/cumulusci/tasks/salesforce/PublishCommunity.py
@@ -1,0 +1,60 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.core.exceptions import SalesforceException
+
+
+class PublishCommunity(BaseSalesforceApiTask):
+    api_version = "46.0"
+    task_docs = """
+    Publishes a Salesforce Community via the Connect API.
+    """
+    task_options = {
+        "name": {
+            "description": "The name of the community to publish.",
+            "required": True,
+        },
+        "communityid": {
+            "description": "The id of the community to publish.",
+            "required": False,
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super(PublishCommunity, self)._init_options(kwargs)
+
+    def _run_task(self):
+        community_id = self.options.get("communityid", None)
+        community_name = self.options["name"]
+
+        if community_id is None:
+            self.logger.info(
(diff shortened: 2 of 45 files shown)


MatthewBlanski:
I find it odd that the tests on `test_PublishCommunity.py` are flagging that `self` has no property `sf`. My manual tests of the code to see that it would work wouldn't have functioned if that were not the case.

cumulusci/tasks/salesforce/PublishCommunity.py
```diff
@@ -0,0 +1,60 @@
...
+    api_version = "46.0"
+    task_docs = """
+    Publishes a Salesforce Community via the Connect API.
+    """
+    task_options = {
+        "name": {
+            "description": "The name of the community to publish.",
+            "required": True,
+        },
+        "communityid": {
```
davidmreed: Can we do `snake_case` for this parameter?

I'd suggest we make either Id or Name required - i.e., throw a `TaskOptionsError` if neither is provided, but allow the user to pick which one.

cumulusci/tasks/salesforce/PublishCommunity.py
```diff
@@ -0,0 +1,60 @@
...
+            if self.options["name"] in communities:
+                community_id = communities[self.options["name"]]["id"]
+            else:
+                raise SalesforceException(
+                    'Unable to find a community named "{}"'.format(community_id)
+                )
+        else:
+            self.logger.info('Checking name for community "{}"'.format(community_id))
+            community = self.sf.restful("connect/communities/{}".format(community_id))
+            if community_name != community["name"]:
```
davidmreed: I'm not sure I follow the purpose of this check. Why do we need to have the user provide a name, and then validate that name, if we have the Id?

cumulusci/tasks/salesforce/PublishCommunity.py
```diff
@@ -0,0 +1,60 @@
...
+                raise SalesforceException(
+                    'The community name for {} is "{}" and does not match "{}", the name you provided'.format(
+                        community_id, community["name"], community_name
+                    )
+                )
+
+        self.logger.info(
+            'Sending request to publish Community "{}"({})'.format(
```
davidmreed: Nit-pick - when we're referring to the setup entity, can we consistently capitalize Community? We have upper and lower in a couple of places.

cumulusci/tasks/salesforce/__init__.py
```diff
@@ -54,3 +55,33 @@
 from cumulusci.tasks.salesforce.UninstallPackagedIncremental import (
     UninstallPackagedIncremental,
 )
+
```
davidmreed: @davisagli  Do we have a preference for this style of fix or just overriding flake8 with --no-verify when it balks at `__init__.py`?

cumulusci/tasks/salesforce/PublishCommunity.py
```diff
@@ -0,0 +1,60 @@
...
+                community_name, community_id
+            )
+        )
+        response = self.sf.restful(
+            "connect/communities/{}/publish".format(community_id), method="POST"
+        )
+
+        self.logger.info(response["message"])
```
davidmreed: What are your thoughts about polling until the publish operation is completed?

MatthewBlanski: @davidmreed This is not a polling action, but a one-and-done in which the API simply informs you it has 'started' the process and will email you shortly. So it's an immediate response.

cumulusci/tasks/salesforce/PublishCommunity.py
MatthewBlanski: @davidmreed  With either optional, we no longer would need to do this. It was an input validation with name required.

cumulusci/tasks/salesforce/PublishCommunity.py
MatthewBlanski: @davidmreed I applied this throughout the task.

cumulusci/tasks/salesforce/PublishCommunity.py
MatthewBlanski: @davidmreed Incorporated both suggestions.

MatthewBlanski:
> @MatthewBlanski Have you successfully gotten this to publish in an otherwise-untouched scratch org? I'm getting the same behavior I saw when this API came out:
> 
> > simple_salesforce.exceptions.SalesforceResourceNotFound: Resource connect/communities/0DBS000000093OXOAY/publish Not Found. Response content: [{'errorCode': 'NOT_FOUND', 'message': 'The requested resource does not exist'}]
> 
> I have a bug out with the PM but not sure what its status is. Let me know if you've figured out how to work around it!

@davidmreed I have and I just tested it again and it functioned. I first use the create_community task and wait for it to fully resolve. After the community is created, I ran publish_community. The environment has Communities enabled and the following org preferences enabled: Chatter, NotesReservedPref01, Networks, Translation, and S1DesktopEnabled. It is a Developer edition org.

cumulusci/cumulusci.yml
```diff
@@ -179,6 +179,10 @@ tasks:
         description: Publish a release to the MetaDeploy web installer
         class_path: cumulusci.tasks.metadeploy.Publish
         group: Release Management
+    publish_community:
+        description: Publishes a Community in the target org using the Connect API
+        class_path: cumulusci.tasks.salesforce.PublishCommunity
+        group: Salesforce
```
davisagli:
```suggestion
        group: Salesforce Communities
```

cumulusci/tasks/salesforce/PublishCommunity.py
davisagli: Is there any way we can detect when publishing is completed by polling for something? We'll want to use this in flows for setting up an org where we need to know that the community is published before proceeding (to something like running a Robot Framework test that interacts with the community, for example).

cumulusci/tasks/salesforce/PublishCommunity.py
```diff
@@ -0,0 +1,66 @@
...
+            "required": False,
+        },
+        "community_id": {
+            "description": "The id of the Community to publish.",
+            "required": False,
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super(PublishCommunity, self)._init_options(kwargs)
```
davisagli: Don't need this

cumulusci/tasks/salesforce/__init__.py
davisagli: I told Matthew to do it this way, since using `--no-verify` doesn't seem very sustainable. I'd rather configure pyflakes to skip the F401 check just for this file, but I don't think it gives us a way to do that.

cumulusci/tasks/salesforce/tests/test_PublishCommunity.py
```diff
@@ -0,0 +1,175 @@
...
+                }
+            ],
+        )
+
+        with self.assertRaises(SalesforceResourceNotFound):
+            cc_task._init_task()
+            cc_task._run_task()
+
+    @responses.activate
+    def test_throws_exeption_for_unset_name_and_id(self):
```
davidmreed:
```suggestion
    def test_throws_exception_for_unset_name_and_id(self):
```

cumulusci/tasks/salesforce/PublishCommunity.py
```diff
@@ -0,0 +1,63 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.core.exceptions import SalesforceException
+from cumulusci.core.exceptions import TaskOptionsError
+
+
+class PublishCommunity(BaseSalesforceApiTask):
+    api_version = "46.0"
+    task_docs = """
+    Publishes a Salesforce Community via the Connect API. Publish a Salesforce Community via the Connect API. Warning: This does not work with the Community Template 'VF Template' due to an existing bug in the API.
```
davidmreed:
```suggestion
    Publish a Salesforce Community via the Connect API. Warning: This does not work with the Community Template 'VF Template' due to an existing bug in the API.
```

PR #1210, opened 2019-09-15
Add No Url Path Prefix Option to Create Community Task

MatthewBlanski:
# Critical Changes

# Changes
Create Community task now allows a blank url path prefix.

# Issues Closed
[1208](https://github.com/SFDO-Tooling/CumulusCI/issues/1208)

### DIFF
--- a/cumulusci/tasks/salesforce/CreateCommunity.py
+++ b/cumulusci/tasks/salesforce/CreateCommunity.py
@@ -60,11 +60,11 @@ class CreateCommunity(BaseSalesforceApiTask):
             community_list = self.sf.restful("connect/communities")["communities"]
             communities = {c["name"]: c for c in community_list}
 
-            for community in communities.items():
-                if community["url_path_prefix"] is None:
+            for community_name in communities.keys():
+                if communities[community_name].get("urlPathPrefix") is None:
                     raise SalesforceException(
                         "A community without a url path prefix already exists, named {}. Try again with the url path prefix".format(
-                            community["name"]
+                            community_name
                         )
                     )
                 self.logger.info("No community without a url path prefix found.")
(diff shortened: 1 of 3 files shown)


cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -59,6 +55,27 @@ def _run_task(self):
...
+            communities = {c["name"]: c for c in community_list}
+
+            for community_name in communities.keys():
+                if communities[community_name].get("urlPathPrefix") is None:
+                    raise SalesforceException(
+                        "A community without a url path prefix already exists, named {}. Try again with the url path prefix".format(
+                            community_name
+                        )
+                    )
+                self.logger.info("No community without a url path prefix found.")
```
davisagli: I checked and in the case where we try to create a 2nd community without a urlPathPrefix, the POST to connect/communities gets this response: `[{'errorCode': 'INVALID_INPUT', 'message': 'Enter a URL for your community.'}]`

So why don't we check for that response after the fact (and raise our more informative error), rather than doing a separate request beforehand. It'll avoid an unnecessary request, and avoid a potential (though unlikely) race condition if a community without a prefix gets created in between the two requests.

MatthewBlanski: @davisagli I've added in the tests. After looking at the behavior produced by the api, I am conflicted on whether or not we push our own more helpful error message to supplant that of the API's. It shares the same exception type with the one thrown from the same community name being used. Thus we would need to parse that exception message.

In this case, I feel as though logging a bug with Salesforce that though the message from the API is technically accurate (you do need to enter a url path prefix to get the new community to create), it is less helpful as this is only required because a community without a url path prefix exists.

davisagli: @MatthewBlanski I'm okay with simply showing the error from the server. That has the benefit of being more likely to be something that users will find if they google the error message.

Issue #1212, opened 2019-09-15
Add Task to list communities


MatthewBlanski:
Adding a task to list communities and their status is a QOL improvement for working with communities. Allowing the ability to see community id, name, url path prefix, description, and status will enhance the ability to work with a community publish task in organizations that have two communities by the same name.

PR #1214, opened 2019-09-15
Add List Communities Task

MatthewBlanski:
# Critical Changes

# Changes
Added task to list communities in the current org.

# Issues Closed
[1212](https://github.com/SFDO-Tooling/CumulusCI/issues/1212)

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -160,7 +160,7 @@ tasks:
             activateRSS: True
         group: Salesforce Packages
     list_communities:
-        description: Lists communities for the current org using the Connect API.
+        description: Lists Communities for the current org using the Connect API.
         class_path: cumulusci.tasks.salesforce.ListCommunities
     list_metadata_types:
         description: Prints the metadata types in a project
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -162,6 +162,7 @@ tasks:
     list_communities:
         description: Lists Communities for the current org using the Connect API.
         class_path: cumulusci.tasks.salesforce.ListCommunities
+        group: Salesforce Communities
     list_metadata_types:
         description: Prints the metadata types in a project
         class_path: cumulusci.tasks.util.ListMetadataTypes
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -159,6 +159,9 @@ tasks:
             version: latest_beta
             activateRSS: True
         group: Salesforce Packages
+    list_communities:
+        description: Lists communities for the current org using the Connect API.
+        class_path: cumulusci.tasks.salesforce.ListCommunities
     list_metadata_types:
         description: Prints the metadata types in a project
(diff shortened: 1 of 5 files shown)


cumulusci/cumulusci.yml
```diff
@@ -159,6 +159,9 @@ tasks:
             version: latest_beta
             activateRSS: True
         group: Salesforce Packages
+    list_communities:
+        description: Lists Communities for the current org using the Connect API.
+        class_path: cumulusci.tasks.salesforce.ListCommunities
```
davisagli: Let's add `group: Salesforce Communities` to this and the other community-related tasks so that they are grouped together in `cci task list`

cumulusci/tasks/salesforce/ListCommunities.py
```diff
@@ -0,0 +1,42 @@
...
+
+class ListCommunities(BaseSalesforceApiTask):
+    api_version = "46.0"
+    task_docs = """
+    Lists Communities for the current org via the Connect API.
+    """
+    task_options = {}
+
+    def _init_options(self, kwargs):
+        super(ListCommunities, self)._init_options(kwargs)
```
davisagli: There's no reason to include this since you're not changing the behavior from the super class.

cumulusci/tasks/salesforce/ListCommunities.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def _run_task(self):
+        community_list = self.sf.restful("connect/communities")["communities"]
+        communities = [c for c in community_list]
```
davisagli: Just do:

    communities = self.sf.restful("connect/communities")["communities"]

If that doesn't produce something that can be iterated over by `for community in communities:` I'll be surprised.

cumulusci/tasks/salesforce/ListCommunities.py
```diff
@@ -0,0 +1,42 @@
...
+                )
+            )
+            communities_output.append(
+                "* **Template:** {}\n".format(community["templateName"])
+            )
+            communities_output.append(
+                "* **Description:** {}\n".format(community["description"])
+            )
+
+        communities_output2 = "\n".join(communities_output).encode()
```
davisagli: Please test this out with a community that has non-ascii characters in its name -- and it'd be good to include that in the unit test so that it gets run in both Python 2 and 3. I suspect we'll need to do more work on the encoding here.

MatthewBlanski: @davisagli Unfortunately this is where I am out of my depth. Running through the rest of the application to handle encoding appears to be a monumental task. This does seem to only be a Python 2 issue as Python 3 handles it without a problem.

davisagli: @MatthewBlanski I added `from __future__ import unicode_literals` to make sure that string literals are interpreted as unicode in Python 2 (so that interpolated values don't get implicitly encoded as latin1), and specified "utf-8" as the codec to use at the end since it's not the default in Python 2.

PR #1216, opened 2019-09-16
Feature/more forgiving batch wait

prescod:
# Critical Changes

# Changes

Previously, CCI would check that a batch job was completed correctly based on whether it both reported no error message and also had a TotalJobItems=JobItemsProcessed. On extremely large jobs with lots of Ajax, this seems like too conservative of a strategy. Sometimes the Salesforce Platform returns JobItemsProcessed<TotalJobItems but otherwise it seems that job succeeds and there is no serious reason to be concerned. In this case, CCI will now report a warning rather than an error and other processes can continue.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/apex/batch.py
+++ b/cumulusci/tasks/apex/batch.py
@@ -49,7 +49,7 @@ class BatchApexWait(BaseSalesforceApiTask):
             self.logger.info("There have been some batch failures.")
             self.logger.info("Error values:")
             self.logger.info(repr(vals))
-            raise SalesforceException("There were import errors", vals)
+            raise SalesforceException("There were import errors: %s" % repr(vals))
         elif not self.done_for_sure:
             self.logger.info("The final record counts do not add up.")
             self.logger.info("This is probably related to W-1132237")
--- a/cumulusci/tasks/apex/batch.py
+++ b/cumulusci/tasks/apex/batch.py
@@ -36,14 +36,14 @@ class BatchApexWait(BaseSalesforceApiTask):
             key: value
             for key, value in self.batch.items()
             if key
-            in [
+            in {
                 "Id",
                 "Status",
                 "ExtendedStatus",
                 "NumberOfErrors",
                 "JobItemsProcessed",
                 "TotalJobItems",
-            ]
+            }
         }
         if not self.success:
             self.logger.info("There have been some batch failures.")
--- a/cumulusci/tasks/apex/batch.py
+++ b/cumulusci/tasks/apex/batch.py
@@ -1,6 +1,5 @@
 """ a task for waiting on a Batch Apex job to complete """
 
-import datetime
 from cumulusci.utils import parse_api_datetime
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/apex/batch.py
```diff
@@ -33,9 +32,28 @@ def _run_task(self):
...
+            for key, value in self.batch.items()
+            if key
+            in [
+                "Id",
+                "Status",
+                "ExtendedStatus",
+                "NumberOfErrors",
+                "JobItemsProcessed",
+                "TotalJobItems",
+            ]
```
davisagli: Not really consequential, but probably makes more sense to use a set literal here?

cumulusci/tasks/apex/batch.py
```diff
@@ -33,9 +32,28 @@ def _run_task(self):
...
+        }
         if not self.success:
-            self.logger.info("There were some batch failures.")
-            raise SalesforceException(self.batch["ExtendedStatus"])
+            self.logger.info("There have been some batch failures.")
+            self.logger.info("Error values:")
+            self.logger.info(repr(vals))
+            raise SalesforceException("There were import errors", vals)
```
davisagli: This is going to print something like:

    cumulusci.core.exceptions.SalesforceException: ('There were import errors', "{'Id': 'asdf', etc... }")

Did you mean to format this somehow?

cumulusci/tasks/apex/batch.py
```diff
@@ -33,9 +32,28 @@ def _run_task(self):
...
+        elif not self.done_for_sure:
+            self.logger.info("The final record counts do not add up.")
+            self.logger.info("This is probably related to W-1132237")
```
davisagli: We should probably link to the known issue (which is public) rather than an internal work item.

prescod: Google turns up the public page when you search for W-1132237

davisagli: I forgot those are referenced publicly. But why not save someone the googling and link to https://success.salesforce.com/issues_view?id=a1p300000008XvaAAE

cumulusci/tasks/apex/batch.py
prescod: f0fa6cc2775f05b282dd44d730538e39219b11f6

cumulusci/tasks/apex/batch.py
prescod: I didn't feel strongly about this, but PEP 352 advocates for a single string argument, so I've done that in e2368baa9e03caeb68b394c971de85d4836d78a7

cumulusci/tasks/apex/batch.py
prescod: 99f6708a

davisagli:
@prescod can you update this PR description to include a description of the change that's suitable to put in the release notes?

PR #1218, opened 2019-09-19
Feature/collection apis

prescod:
# Changes

Add keywords for generating several fake SObjects according to a template.

The template can include Jinja2 syntax and faker calls, with or without parameters.


cumulusci/robotframework/Salesforce.py
```diff
@@ -24,9 +25,25 @@
...
+
+    def __call__(self, *args, **kwargs):
+        return self.func(*args, **kwargs)
+
+
+class FakerTemplateLibrary:
+    faker = faker.Faker()
+
+    def __getattr__(self, name):
+        # defer loading heavy faker library until actually needed
```
davisagli: Looks like despite this comment it's initialized at import time on line 40. Or do you mean that faker defers loading subsets of itself until they are used?

cumulusci/robotframework/Salesforce.py
```diff
@@ -24,9 +25,25 @@
...
+        return StringGenerator(
+            lambda *args, **kwargs: self.faker.format(name, *args, **kwargs)
+        )
```
davisagli: Can you add docstrings for these classes and move them to a utils module?

cumulusci/robotframework/Salesforce.py
prescod: I think I refactored the laziness away at some point but I've added it back now because Faker is big and I'm not sure how/when it initializes all of its internal databases. It's just a couple of lines of code to make it a lazy load.

cumulusci/robotframework/Salesforce.py
prescod: Done.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+            self.store_session_record(obj["attributes"]["type"], record["id"])
+            obj["id"] = record["id"]
+            obj[STATUS_KEY] = record
+
+        return objects
+
+    def salesforce_collection_update(self, objects):
+        """Updates up to 200 records described as Robot/Python dictionaries"""
+        for obj in objects:
+            assert obj["id"], "Should be a list of objects with Ids"
```
boakley: If you had 199 objects with ids and one without, this error message would be a bit misleading. You might want to make this message more specific, such as `Object number <n> does not have an 'id' property` or something like that.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+
+            Jinja2 Expression Syntax is allowed so computed templates like this are also allowed: {{1000 + number}}
+
+            Templates can also be based on faker patterns like those described here:
+
+            https://faker.readthedocs.io/en/master/providers.html
+
+            Most examples can be pasted into templates verbatim:
+
+        @{objects}=  Salesforce Collection Generate  Contact  200
```
boakley: This example needs the `|` prefix so that it's treated as preformatted code.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+                | ...  LastName={{random_str}}
+                | ...  Age={{number}}
+
+            Which would generate:
+
+                | [{'FirstName': 'User 0', 'LastName': 'u4PNlGUD', 'type': 'Contact', 'Age': '0'},
+                |  {'FirstName': 'User 1', 'LastName': 'Kyd4PC5x', 'type': 'Contact', 'Age': '1'},
+                |  {'FirstName': 'User 2', 'LastName': 'n4EGs4Vp', 'type': 'Contact', 'Age': '2'}]
+
+            Jinja2 Expression Syntax is allowed so computed templates like this are also allowed: {{1000 + number}}
```
boakley: You should probably include a link to jinja2 documentation. There's a good chance many of our testers have never heard of jinja2. Also, `{{1000 + number}}` needs to have backticks added so it appears as code.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+            formatted_fields = {
+                name: format_str(value, i) for name, value in fields.items()
+            }
+            newobj = self._salesforce_generate_object(obj_name, **formatted_fields)
+            objs.append(newobj)
+
+        return objs
+
+    def salesforce_collection_insert(self, objects):
+        """Inserts up to 200 records that were created with Salesforce Collection Generate.
```
boakley: instead of hard-coding 200 in several places, would it make sense to make this a constant defined at the start of the file? What is the chance that the API limit will change in the foreseeable future?

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+        ...  Email={{fake.email(domain="salesforce.com")}}
+
+           """
+        objs = []
+
+        def format_str(value, i):
+            if isinstance(value, str):
+                value = Template(value).render(
+                    number=i,
+                    random_str=self.random_string_generator,
```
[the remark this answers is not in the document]
prescod: I guess its just left over from before...I'll remove it.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
 
+    def _salesforce_generate_object(self, obj_name, **fields):
+        obj = {"attributes": {"type": obj_name}}  # Object type to create
+        obj.update(fields)
+        return obj
+
+    random_string_generator = StringGenerator(String().generate_random_string)
+    faker_template_library = FakerTemplateLibrary()
+
+    def salesforce_collection_generate(self, obj_name, number_to_create, **fields):
```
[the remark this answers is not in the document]
prescod: The dictionary does have a Salesforce type associated with it but otherwise not really.

cumulusci/robotframework/Salesforce.py
prescod: My thinking on this one is that if you have even a single object without an "id" then it almost certainly means you generated the objects in some non-standard way. You're not supposed to be generating these lists "by hand". These are supposed to be the return value from salesforce_collection_insert.

I will clarify the error message but more in the direction I'm describing.

cumulusci/robotframework/Salesforce.py
prescod: Okay I clarified it somewhat.

cumulusci/robotframework/Salesforce.py
prescod: I did this but it isn't really practical to do it in docstrings. (It's possible but the code gets pretty ugly).

cumulusci/robotframework/Salesforce.py
prescod: Done.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +525,115 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+        """Returns an array of dictionaries with template-formatted arguments appropriate for a Collection Insert.
+            Use ``{{number}}`` to represent the unique index of the row in the list of rows, ``{{random_str}}`` to represent a random string
+            and a string that consists of just ``{{number}}`` to generate an actual integer (as opposed to a string-encoded number)
+
+            For example:
+
+                | @{objects} =  Salesforce Collection Generate  Contact  3
+                | ...  FirstName=User {{number}}
+                | ...  LastName={{random_str}}
```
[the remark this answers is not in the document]
prescod: Done.

cumulusci/robotframework/tests/salesforce/perf/collection_perf.robot
```diff
@@ -0,0 +1,98 @@
+*** Settings ***
+Documentation   Tests of collections of records based on Jess Lopez's work here: 
+...             https://salesforce.quip.com/dsUXAOxiKz28  (Salesforce Internal)
```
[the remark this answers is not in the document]
prescod: Removed.

cumulusci/robotframework/Salesforce.py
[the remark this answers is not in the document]
prescod: Renamed.

cumulusci/robotframework/Salesforce.py
prescod: Done.

cumulusci/robotframework/Salesforce.py
boakley: Are you certain this code buys us anything? I would be surprised to see the overhead being more than a few ms.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +506,111 @@ def salesforce_insert(self, obj_name, **kwargs):
...
         return res["id"]
 
+    def _salesforce_generate_object(self, obj_name, **fields):
+        obj = {"attributes": {"type": obj_name}}  # Object type to create
+        obj.update(fields)
+        return obj
+
+    def generate_test_data(self, obj_name, number_to_create, **fields):
+        """Returns an array of dictionaries with template-formatted arguments appropriate for a Collection Insert.
+            Use ``{{number}}`` to represent the unique index of the row in the list of rows.
```
davisagli: According to PEP 257, in multi-line docstrings the indentation after the first line should match the indentation of the quotes that start the docstring.

cumulusci/robotframework/template_utils.py
```diff
@@ -0,0 +1,55 @@
...
+
+    @property
+    def faker(self):
+        """Defer loading heavy faker library until actually needed"""
+        self._faker = self._faker or __import__("faker").Faker()
+        return self._faker
+
+    def __getattr__(self, name):
+        return StringGenerator(
+            lambda *args, **kwargs: self.faker.format(name, *args, **kwargs)
```
davisagli: This might be a good place to use functools.partial

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +506,111 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+
+        records = self.cumulusci.sf.restful(
+            "composite/sobjects",
+            method="POST",
+            json={"allOrNone": True, "records": objects},
+        )
+
+        for record, obj in zip(records, objects):
+            if record["errors"]:
+                raise AssertionError(*record["errors"])
```
davisagli: It would be helpful to provide info on which object had the error.

cumulusci/robotframework/Salesforce.py
```diff
@@ -499,6 +506,111 @@ def salesforce_insert(self, obj_name, **kwargs):
...
+
+        return objects
+
+    def salesforce_collection_update(self, objects):
+        """Updates up to 200 records described as Robot/Python dictionaries"""
+        for obj in objects:
+            assert obj[
+                "id"
+            ], "Should be a list of objects with Ids returned by Salesforce Collection Insert"
+            del obj[STATUS_KEY]
```
davisagli: I think this could be removed since it gets re-set below. Then this can also work with a list of objects that's been constructed in some other way and doesn't have statuses.

cumulusci/robotframework/Salesforce.py
prescod: No, I'm not certain, but I do know that 

a) Faker is about 4 MB of Python code including everything from German names to Dutch company names. It may do lazy loading internally but I'm not sure and I was hoping not to do a lot of investigation just to save the one line of code.

b) its difficult to benchmark because once a module is loaded, its loaded, so running tests is difficult. And even from a fresh Python its very subject to caching effects.

c) CCI seems to have a lot of "technical debt" relating to performance which I am reluctant to add to. It takes 3 seconds just to get a command listing from CCI on my (cold) laptop and 1 second if its hot.

d) its just one line of code. I don't think it is time well spent to do a bunch of investigation when I can be confident that I'm not adding to CCI's performance technical debt with the one line of code.

cumulusci/robotframework/Salesforce.py
prescod: I think I deleted it so that it won't get sent to Salesforce. I'd assume that either the client or the server side complains about it. I can make the deletion conditional though.

davisagli: Oh yeah, makes sense.

cumulusci/robotframework/template_utils.py
prescod: Maybe...I feel like it is approximately the same cognitive load on the reader either way, personally.

davisagli: Yeah...I don't feel strongly about it.

cumulusci/robotframework/Salesforce.py
prescod: I think I'll leave it as it is. The constraint is validated client side (I think you suggested that) so the limit is fixed until we update the code which we'll do if Salesforce changes their limits.

cumulusci/robotframework/Salesforce.py
prescod: Fixed.

https://github.com/SFDO-Tooling/CumulusCI/pull/1218/commits/0dad119559cbd7e98387c4eab57b8b66b9c43816#diff-17aaca465159b9f94d1457f75bf63cf5R514

cumulusci/robotframework/Salesforce.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/blob/8c311e02818bc8b83afbf7c189babfbdbf500fd6/cumulusci/robotframework/Salesforce.py#L587

cumulusci/robotframework/Salesforce.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/blob/feature/collection-apis/cumulusci/robotframework/Salesforce.py#L602

PR #1219, opened 2019-09-23
Feature/salesforce base class

prescod:
# Critical Changes

# Changes

Certain methods that many keyword libraries seem to implement have been refactored into a common base class.

# Issues Closed


cumulusci/robotframework/Salesforce.py
```diff
@@ -15,13 +15,15 @@
...
+from .salesforce_robot_library_base import SalesforceRobotLibraryBase
+
 OID_REGEX = r"^(%2F)?([a-zA-Z0-9]{15,18})$"
 
 lex_locators = {}  # will be initialized when Salesforce is instantiated
 
 
 @selenium_retry
-class Salesforce(object):
+class Salesforce(SalesforceRobotLibraryBase):
```
davisagli: It seems non-ideal for the Salesforce library to have a `salesforce` property that returns an instance of itself.

cumulusci/robotframework/salesforce_robot_library_base.py
```diff
@@ -0,0 +1,42 @@
...
+        subtask_config = TaskConfig({"options": options})
+        return self.cumulusci._run_task(taskclass, subtask_config)
+
+    def _batch_apex_wait(self, class_name):
+        """Helper method for waiting for Salesforce jobs to finish"""
+        return self._run_subtask(BatchApexWait, class_name=class_name)
+
+    def _run_apex(self, code):
+        """Helper method for running Apex code easily"""
+        return self._run_subtask(AnonymousApexTask, apex=code)
```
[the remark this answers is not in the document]
prescod: They are intended for use by Python programmers (e.g. me). 

run_task_class takes a string as an argument, which is appropriate for a Robot programmer. 

_run_subtask takes a class as an argument, which is appropriate for a Python programmer (and helps the programmer detect errors earlier/faster).

The other two could probably be keywords, if we aren't worried about keyword creep. Usually when you might be tempted to put Apex in a Robot script it might be better to make a new keyword in Python, so that's one argument for keeping that one as a helper function instead of a keyword. But on the other hand, maybe there are plentiful legitimate uses for Apex right in robot code.

cumulusci/robotframework/Salesforce.py
prescod: I suppose so, but I guess the root cause of the weirdness is using this unusual dynamic import technique. I don't know why we import code this way.

Anyhow, I can fix it by removing the base class usage from Salesforce.py or by removing this helper here.

cumulusci/robotframework/salesforce_robot_library_base.py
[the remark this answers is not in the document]
davisagli: I agree passing a class is better so that typos can be caught at import time. But then my suggestion is to make run_task_class accept either a string or a class so that we don't need a new method that does subtly different things in a different location in the codebase. That's just going to confuse the next person who's trying to figure out the best way to run a task from Robot Framework.

Avoiding keyword creep is good. I think avoiding function creep is good too. These are one-liners so what is the benefit of factoring them into methods instead of inlining them? I think it just obscures the details of what's happening and makes it harder to read the code that calls these.

cumulusci/robotframework/Salesforce.py
boakley: What do you mean by "this unusual dynamic import technique"? Are you referring to the properties? It's not importing anything, it's simply providing a shortcut for commonly used objects.

cumulusci/robotframework/salesforce_robot_library_base.py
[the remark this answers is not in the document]
boakley: I agree with David - if we're going to make a change like this, I think it makes more sense to define `run_task` to accept either a string or a dictionary.

cumulusci/robotframework/salesforce_robot_library_base.py
```diff
@@ -0,0 +1,34 @@
+import logging
+
+from cumulusci.core.config import TaskConfig
+from robot.libraries.BuiltIn import BuiltIn
+from cumulusci.tasks.apex.anon import AnonymousApexTask
+
+from cumulusci.tasks.apex.batch import BatchApexWait
+
+
+class SalesforceRobotLibraryBase(object):
```
boakley: I recommend renaming this to `BaseLibrary`. Both `Salesforce` and `Robot` are a bit redundant, and the namespace provides the additional context. This will also make documentation a bit cleaner, as we'll be able to say "use `BaseLibrary` when creating a traditional robot library, and `BasePage` when creating page objects."

prescod:
There wasn't much consensus on whether this added value or what to do with it.

PR #1224, opened 2019-09-25
Give feedback on org setup.

prescod:
# Critical Changes

# Changes

Creating orgs now generates feedback.

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1068,6 +1068,8 @@ def org_scratch(config, config_name, org_name, default, devhub, days, no_passwor
     if default:
         config.keychain.set_default_org(org_name)
         click.echo("{} is now the default org".format(org_name))
+    else:
+        click.echo("{} is configured for use".format(org_name))
 
 
 @org.command(


cumulusci/cli/cci.py
```diff
@@ -1068,6 +1068,8 @@ def org_scratch(config, config_name, org_name, default, devhub, days, no_passwor
     if default:
         config.keychain.set_default_org(org_name)
         click.echo("{} is now the default org".format(org_name))
+    else:
+        click.echo("{} is configured for use".format(org_name))
```
```suggestion
        click.echo("{} has been created".format(org_name))
```

It's probably not going to be "configured for use" until they run a flow in it.

prescod: My thinking was the opposite. It isn't "created" until they run a flow in it.

davisagli: That doesn't make any sense to me. We just ran the command called create_scratch_org and its result was creating the org.

davisagli: Oh wait, you mean that it was configured in the CCI keychain, not that we deployed product configuration to the org. Yeah I guess it's fine.

prescod: "Your org is configured and when needed, it will be created, so that you can configure it."

PR #1227, opened 2019-09-25
Tag performance tests

prescod:
Tag performance tests so we can run them in the standard flows and plans configured in MetaCI.

# Critical Changes

# Changes

# Issues Closed


cumulusci.yml
```diff
@@ -37,6 +37,21 @@ flows:
...
             2:
                 task: robot_testdoc
+    test_performance_scratch:
+        description: Test performance of a scratch org
+        steps:
+            1: 
+                task: robot
+                options:
+                    suites: cumulusci/robotframework/tests/salesforce/perf
+                    include: perf
```
davisagli: Specifying this suite selects the same tests as the tag, so what's the tag for?

prescod: Mostly future-proofing because I couldn't pick which strategy is better. 

But I guess we should just pick a strategy about whether we'd rather have a standardized folder structure, tag name or just flow name across repositories.

When this suite grows, we won't necessarily want to run it when we run all of the other tests.

prescod: I'll put together a document with some standards.

prescod: Okay, I switched to a directory-only approach. It doesn't work across projects (directly) but it does make it extremely easy to include or exclude the tests versus functional tests.

All of my APIs are still tested through API.Robot. They just aren't tested on as much data.

PR #1228, opened 2019-09-25
Feature/determine parent branch

Br4nd0R:
* Logic now works for triggering off of commits to parent branches


cumulusci/tasks/release_notes/generator.py
```diff
@@ -134,8 +134,9 @@ def aggregate_child_change_notes(self, pull_request):
         Child pull requests are pull requests that have a base branch
         equal to the the given pull request's head."""
         self.change_notes = get_pull_requests_with_base_branch(
-            self.repo, pull_request.head.label.split(":")[1]
+            self.repo, pull_request.head.ref
         )
+        self.change_notes = list(filter(is_merged, self.change_notes))
```
davisagli: Consider

    self.change_notes = [note for note in self.change_notes if is_merged(note)]

but I don't feel strongly about it

cumulusci/tasks/release_notes/task.py
```diff
@@ -108,96 +109,77 @@ class ParentPullRequestNotes(BaseGithubTask):
...
+        return child_branch_name
+
+    def _is_parent_branch_base_branch(self, pull_request):
+        """Returns True if self.branch_name is the base branch of target pull_request"""
+        True if pull_request.base.ref == self.branch_name else False
+
+
+def is_pull_request_merged(pull_request):
+    """Takes a github3.pulls.ShortPullRequest"""
+    return pull_request.merged_at is not None
```
davisagli: Put this in the github utils module to avoid duplicating it between generator.py and task.py

cumulusci/tasks/release_notes/task.py
```diff
@@ -108,96 +109,77 @@ class ParentPullRequestNotes(BaseGithubTask):
...
+
+        child_branch_name = None
+        # TODO is there a case where more than one would be expected?
+        if len(merged_prs) == 1 and merged_prs[0].base.ref == self.branch_name:
+            return merged_prs[0].head.ref
+        return child_branch_name
+
+    def _is_parent_branch_base_branch(self, pull_request):
+        """Returns True if self.branch_name is the base branch of target pull_request"""
+        True if pull_request.base.ref == self.branch_name else False
```
davisagli: This isn't returning anything so _is_parent_branch_base_branch will always be None

cumulusci/tasks/release_notes/task.py
```diff
@@ -108,96 +109,77 @@ class ParentPullRequestNotes(BaseGithubTask):
...
-                self.repo, parent_pull_request, self.build_notes_label
+                self.repo, parent_pull_request, self.options.get("build_notes_label")
             )
         else:
             parent_pull_request = requests[0]
         return parent_pull_request
+
+    def _get_child_branch_name_from_merge_commit(self):
+        pull_requests = get_pull_requests_by_commit(
+            self.github, self.repo, self.project_config.repo_commit
```
davisagli: Does this work? I expected we would need to search for one of the parent commits, not the merge commit, since the child branch doesn't include the merge commit.

cumulusci/tasks/release_notes/generator.py
Br4nd0R: Yours is definitely more readable

cumulusci/tasks/release_notes/task.py
Br4nd0R: Turns out, I'm not even referencing this method anymore. :)

pytest.ini
```diff
@@ -1,2 +1,4 @@
 [pytest]
 testpaths = cumulusci
+filterwarnings =
+    ignore::DeprecationWarning
```
davisagli: @Br4nd0R I know these are coming from libraries we use instead of from our own code, but I don't like ignoring all deprecation warnings since we could miss something in the future.

Br4nd0R: apologies, forgot to remove this. The warnings were cluttering the debug output in VSCode.

PR #1238, opened 2019-10-01
Feature/robot libdoc enhancements

boakley:
This change lets you use wildcards to define the path option of the robot_libdoc task. There's also a bugfix that catches errors when creating the output, where it throws a proper `TaskOptionsError` rather than just spew a traceback. 

This includes a new function in utils, `process_glob_list_arg` which works like `process_list_arg`, but goes the extra step and expands wildcards.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/core/tests/test_utils.py
+++ b/cumulusci/core/tests/test_utils.py
@@ -68,10 +68,9 @@ class TestUtils(unittest.TestCase):
 
             # Recursive
             os.mkdir("subdir")
-            touch("subdir/baz.resource")
-            self.assertEqual(
-                ["subdir/baz.resource"], utils.process_glob_list_arg("**/*.resource")
-            )
+            filename = os.path.join("subdir", "baz.resource")
+            touch(filename)
+            self.assertEqual([filename], utils.process_glob_list_arg("**/*.resource"))
 
     def test_decode_to_unicode(self):
         self.assertEqual(u"\xfc", utils.decode_to_unicode(b"\xfc"))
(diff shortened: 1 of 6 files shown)


cumulusci/core/tests/test_utils.py
```diff
@@ -34,6 +36,43 @@ def test_process_list_arg(self):
...
+            self.assertEqual(
+                ["*.bar", "x.y.z"], utils.process_glob_list_arg("*.bar, x.y.z")
+            )
+
+            # Recursive
+            os.mkdir("subdir")
+            touch("subdir/baz.resource")
+            self.assertEqual(
+                ["subdir/baz.resource"], utils.process_glob_list_arg("**/*.resource")
+            )
```
davisagli: This test is failing in Windows due to the different path separator. It looks like the glob does the right thing but you need to use os.path.join to build the expected result.

cumulusci/core/utils.py
```diff
@@ -44,6 +45,31 @@ def process_bool_arg(arg):
...
+        return []
+
+    files = []
+    for path in initial_list:
+        more_files = glob.glob(path, recursive=True)
+        if len(more_files):
+            files += sorted(more_files)
+        else:
+            files.append(path)
+    return files
```
davisagli: It's possible for multiple different globs to include the same file. So I think it'd be a good idea to work with `files` as a set and then convert to a list as you return it.

boakley: hmmm. I see your point. Sets are unordered though, so we'll lose the ordering. That's not necessarily a bad thing, but it's a regression.

davisagli: If ordering is important you can use a dict, which preserve insertion order in Python 3.6+

boakley: yeah, that's what I did.

PR #1244, opened 2019-10-03
Feature/remove ordered dicts

Br4nd0R:
* Still need to strip out references in `cumulusci/core/utils.py` for ordered yaml loading. 

# Changes
* No more ordered dicts


cumulusci/tasks/bulkdata/load.py
```diff
@@ -397,7 +397,7 @@ def _init_mapping(self):
 
     def _expand_mapping(self):
         # Expand the mapping to handle dependent lookups
-        self.after_steps = defaultdict(lambda: OrderedDict())
+        self.after_steps = defaultdict(lambda: {})
```
davisagli:
```suggestion
        self.after_steps = defaultdict(dict)
```

PR #1246, opened 2019-10-03
We no longer need the [security] extra for requests

davisagli:
(It installed PyOpenSSL to handle certificate validation in Python < 2.7.9)

# Critical Changes

# Changes

# Issues Closed


davisagli:
We can't remove this at this time, because simple-salesforce still requires requests with the security extra.

PR #1247, opened 2019-10-03
Display failed Apex tests on package upload

Br4nd0R:
* Use Tooling API to query ApexTestResult records based on ApexTestResult.TestTimestamp > (now - time to execute since package upload)
* Significant refactoring of original class into smaller more logical methods

Screen Shot 2019-10-08 at 9 30 32 AM

#Changes
* Now displaying failed Apex tests in a table with columns: Class, Method, Message, Stacktrace on package upload.


cumulusci/tasks/salesforce/PackageUpload.py
```diff
@@ -116,6 +111,35 @@ def _run_task(self):
...
+            self.logger.error("  {}".format(error["message"]))
+            if error["message"] == "ApexTestFailure":
+                apex_test_failures = True
+
+        if apex_test_failures:
+            self.logger.error("Failed Apex Tests:")
+            test_start_datetime = datetime.utcnow() - timedelta(
+                seconds=self.upload_time_seconds
+            )
+            failed_tests_query = f"SELECT ApexClass.Name, MethodName, Message FROM ApexTestResult WHERE Outcome='Fail' AND TestTimestamp > {test_start_datetime.isoformat()+'Z'}"
```
jstvz: I think we'll need `StackTrace` too.

cumulusci/tasks/salesforce/PackageUpload.py
```diff
@@ -116,6 +111,35 @@ def _run_task(self):
...
+            results = self.tooling.query(failed_tests_query)
+            for test in results["records"]:
+                self.logger.error(
+                    f"    {test['ApexClass']['Name']}.{test['MethodName']}"
```
jstvz: Since we're going to need the `StackTrace`, you may want to consider just dumping these results into a table.

cumulusci/tasks/salesforce/PackageUpload.py
```diff
@@ -116,6 +111,35 @@ def _run_task(self):
...
+        for error in self.upload["Errors"]["errors"]:
+            self.logger.error("  {}".format(error["message"]))
+            if error["message"] == "ApexTestFailure":
+                apex_test_failures = True
+
+        if apex_test_failures:
+            self.logger.error("Failed Apex Tests:")
+            test_start_datetime = datetime.utcnow() - timedelta(
+                seconds=self.upload_time_seconds
+            )
```
davisagli: You're calculating the elapsed time using the start time, and then subtracting it from now to get the start time. Why not just store the start time and use it?

cumulusci/tasks/salesforce/package_upload.py
```diff
@@ -0,0 +1,222 @@
...
+
+        exception = self._get_exception_type(error)
+        raise exception("Package upload failed")
+
+    def _log_apex_test_failures(self):
+        self.logger.error("Failed Apex Tests:")
+        soql_query = self._get_failed_tests_soql_query()
+        results = self.tooling.query(soql_query)
+
+        table_header_row = ["Class", "Method", "Stacktrace"]
```
Br4nd0R: Would 'Apex Class' be better? Title of the table is 'Apex Test Failures' though...

davisagli: I think just "Class" is fine, given that context.

cumulusci/tasks/salesforce/package_upload.py
```diff
@@ -0,0 +1,222 @@
...
+            f"AND TestTimestamp > {package_upload_datetime}Z"
+        )
+
+    def _get_package_upload_iso_timestamp(self):
+        """Returns a datetime of approximately when package upload began
+        This assumes a short time between the call to
+        _make_package_upload_request() and this function"""
+        test_start_datetime = datetime.utcnow() - timedelta(
+            seconds=(time.time() - self._upload_start_time)
+        )
```
davisagli: I still don't understand the reason for the date arithmetic. Why not just set `self._upload_start_time = datetime.utcnow()` up above, and use self._upload_start_time.isoformat() for the query?

cumulusci/tasks/salesforce/package_upload.py
```diff
@@ -0,0 +1,222 @@
...
+            f"Created PackageUploadRequest {self.upload_id} for Package {package_id}"
+        )
+        self._poll()
+
+    def _log_package_upload_errors(self):
+        self.logger.error("Package upload failed with the following errors")
+        error = {"message": ""}
+        apex_test_failures = False
+        for error in self.upload["Errors"]["errors"]:
+            self.logger.error("  {}".format(error["message"]))
```
davisagli: Could use an f-string here

cumulusci/tasks/salesforce/package_upload.py
Br4nd0R: Ahhh. Yes...was definitely overcomplicating this.

cumulusci/tasks/salesforce/package_upload.py
Br4nd0R: yes definitely

PR #1253, opened 2019-10-08
Feature/log instead of error

Br4nd0R:
* Added `markdown_link_to_pr()` method to `cumulusci.core.github`
* Moved `_unaggregated_pr_header()` from `cumulusci.tasks.release_notes.generator.py` into `cumulusci.tasks.release_notes.task.py`

# Changes
* We now log an error instead of throwing an exception when a pull request is not found (or when more than one is found) when attempting to aggregate change notes from child branches in flow `github_parent_pr_notes`

### DIFF
--- a/cumulusci/tasks/release_notes/task.py
+++ b/cumulusci/tasks/release_notes/task.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 from cumulusci.core.utils import process_bool_arg
 from cumulusci.tasks.github.base import BaseGithubTask
 from cumulusci.tasks.release_notes.generator import (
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/release_notes/task.py
```diff
@@ -181,3 +182,41 @@ def _get_child_branch_name_from_merge_commit(self):
...
+
+        pull_requests = get_pull_requests_with_base_branch(
+            self.repo,
+            branch_name_to_add.split("__")[0],
+            branch_name_to_add,
+            state="all",
+        )
+
+        if len(pull_requests) == 0:
+            self.logger.error(f"No pull request for branch {branch_name_to_add} found.")
```
davisagli:
```suggestion
            self.logger.info(f"No pull request for branch {branch_name_to_add} found.")
```

I don't think we need to highlight it in red; there are normal situations where this can happen.

PR #1269, opened 2019-10-18
Feature/create contact pageobjects

boakley:
The original goal of this branch was to update our create_contacts.robot test to use page objects.  That required just a few small changes in `create_contact.robot`. However, that small change required some larger changes to the page object implementation. 

I needed a new page object for the "new object" modal, so I created a "NewDialog" class. There are also two new keywords in PageObjects.py -- `wait for dialog` and `wait for page object` which wait for an element to appear on a page rather than waiting for a complete page change. To do this, page objects can now define `_wait_to_appear` which can do whatever it wants to wait for an element to appear on a page. Normally this would be to call the `wait for page to contain element` keyword from SeleniumLibrary. 

Prior to this branch, we defined base page objects simply as classes in baseobjects.py - `DetailPage`, `ListingPage`, etc. I decided it was better to have these base pages be full-fledged page objects so I moved them to a separate module named BasePageObjects.py, and they each now use the `pageobject` decorator on the class.

To support the new keywords I added a new decorator, `capture_screenshot_on_error` which can be used to automatically capture screenshots on keyword failure, much like SeleniumLibrary does.  I have not yet applied this decorator to all of our keywords. That will happen in a separate project. 

Finally, there are a few new tests, and improved documentation. 

# Critical Changes

# Changes

* The create_contact.robot test which is included when initializing a new project with `cci project init` now makes use of page objects.
* PageObjects.py has two new keywords: `wait for dialog` and `wait for page object` which can be used for page objects that represent an overlay or some part of a page rather than a whole page.
* `cumulusci.robotframework.utils` now has a decorator named `capture_screenshot_on_error` which can be used to automatically capture a screenshot when a keyword fails.
* Prior to this change, `Go to page  Detail  Contact` required you to use a keyword argument for the object id (eg: `Go to page  Detail  Contact  object_id=${object_id}`). You can now specify the object id as a positional parameter (eg: `Go to page  Detail  Contact  ${object_id}`)

# Issues Closed


cumulusci/robotframework/pageobjects/BasePageObjects.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _is_current_page(self):
+        self.selenium.location_should_contain(
+            "/lightning/o/{}/list".format(self.object_name)
+        )
+
+
+@pageobject("New")
+class NewDialog(BasePage):
+    """A page object representing the New Object dialog
```
davisagli: These are called Modals in the Lightning Design System (https://www.lightningdesignsystem.com/components/modals/) so I would use that term throughout rather than dialog.

cumulusci/robotframework/utils.py
```diff
@@ -199,3 +200,18 @@ def run_with_retry(self, *args, **kwargs):
     else:
         # Decorator was used without arguments
         return decorate(target)
+
+
+def capture_screenshot_on_error(func):
+    """Decorator for capturing a screenshot if a keyword throws an error"""
```
davisagli: It seems to me that this would be more helpful if we can turn on capturing a screenshot after an error as globally as possible, rather than just decorating specific keywords. With the decorator, it seems inevitable that people will run into situations where the error happened in a keyword that didn't have the decorator, and then extra effort is required to add the decorator and reproduce the issue.

My suggestion would be that we recommend that test authors do this at the top of any test suite that uses selenium:

    Test Teardown | Run Keyword If Test Failed | Capture Page Screenshot

(Or maybe there's even a way to make that be a default test teardown if it's not set?)

cumulusci/core/tests/test_robotframework_utils.py
```diff
@@ -0,0 +1,33 @@
+import unittest
```
davisagli: Please move this file to `cumulusci/robotframework/tests/test_utils.py`. In CumulusCI we aim to put tests for Python modules in the tests directory of the folder where the module-under-test is.

cumulusci/robotframework/pageobjects/BasePageObjects.py
```diff
@@ -0,0 +1,270 @@
...
+
+    def _is_current_page(self):
+        self.selenium.location_should_contain(
+            "/lightning/o/{}/home".format(self.object_name)
+        )
+
+
+@pageobject("Detail")
+class DetailPage(BasePage):
+    """A page object representing the standard Detail page.
```
davisagli: I think this is typically called a Record Page in Salesforce documentation.

cumulusci/robotframework/pageobjects/BasePageObjects.py
```diff
@@ -0,0 +1,270 @@
...
+    | # Using the id:
+    | Go to page  Detail  Contact  ${contact_id}
+    |
+    | # Using lookup parameters
+    | Go to page  Detail  Contact  FirstName=${first_name}  LastName=${last_name}
+
+    """
+
+    def _go_to_page(self, object_id=None, **kwargs):
+        """Go to the detail page for the given object.
```
davisagli:
```suggestion
        """Go to the detail page for the given record.
```

In Salesforce "object" is pretty much a synonym for table. Rows in that table are called records, not objects.

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -164,31 +180,35 @@ def get_page_object(self, page_type, object_name):
...
         """Go to the page of the given page object.
 
         The URL will be computed from the page_type and object_name
-        associated with the object, if possible.
+        associated with the object, plus possibly additional arguments.
 
         Different pages support different additional arguments. For
-        example, a Listing page supports the keyword argument `filter_name`.
+        example, a Listing page supports the keyword argument `filter_name`,
+        and a Detail page can be given an object id, or paramemters for
```
davisagli:
```suggestion
        and a Detail page can be given an object id, or parameters for
```

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -254,6 +267,46 @@ def load_page_object(self, page_type, object_name=None):
...
         return pobj
 
+    @capture_screenshot_on_error
+    def wait_for_dialog(self, page_type, object_name, expected_heading=None, **kwargs):
+        """Wait for the given page object dialog to appear.
+
+        This will both wait for the dialog, and verify that the
+        dialog has an expected heading. The expected heading will be
+        the page type (eg "New") and object name (eg: "Contact")
+        separated by a space (eg: "New Contact").
```
davisagli: This doesn't quite match the implementation--from the docstring it sounds like passing page_type=Edit and object_name=Contact would use an expected heading of "Edit Contact"

davisagli: Also it seems like bad coupling that the base PageObjects class needs to know that a type of page object for dialogs exists. Why do we need a specific `wait_for_dialog` instead of just using the more generic `wait_for_page_object`?

cumulusci/tasks/robotframework/tests/test_robotframework.py
```diff
@@ -349,6 +349,24 @@ def setUp(self):
...
         shutil.rmtree(self.tmpdir)
 
+    def test_file_title(self):
+        """Verify that the TITLE in the file is added to the generated html"""
+        # The page object file has a TITLE attribute and a docstring;
+        # make sure they are picked up.
+        title_element = self.html_body.find(".//div[@class='file-header']/h2")
+        assert title_element.text == "This is the title"
+
+    def test_file_descrption(self):
```
davisagli:
```suggestion
    def test_file_description(self):
```

cumulusci/robotframework/utils.py
boakley: I don't think we want to raise it to a more global level. Not all keywords should capture screenshots. API keywords, for example, don't need to have screenshots on failure. 

Also, the test teardown is problematic, as that code may not run immediately after the error occurs.  Time may pass, or other code may run before the screenshot is taken. It's a good safety net, but it's not as effective as capturing the screenshot immediately after an error has occurred.  Plus, by using the decorator the screenshot shows up in the log immediately before the error message without the user having to drill down into a teardown to see it. 

That being said, automatically taking a screenshot without the user having to do anything at all has merit. We could easily add that to a listener that we automatically enable when we run tests.

cumulusci/robotframework/pageobjects/BasePageObjects.py
boakley: It's had this name since we first introduced page objects. Changing it now would break a lot of tests.

cumulusci/robotframework/pageobjects/PageObjects.py
boakley: I mainly kept the two for readability in tests. I think a test is more readable with `wait for dialog` (or `wait for modal` than `wait for page object` since that's what you would ask a manual tester to do. 

OTOH, there will be times when a page object isn't a modal. A page object could be something else such as a complex lightning component. In that case, `wait for modal` would be wrong since it's not waiting on a modal.

cumulusci/robotframework/utils.py
davisagli: That's a fair point that screenshots are not relevant to all keywords. (Though if a browser was open when an error happened in a non-selenium keyword, I don't think it would hurt to take the screenshot anyway to provide extra context.)

I don't understand the point about a delay before test teardown. Won't an error cause the test to end so that test teardown is the very next thing that happens?

I haven't thought through the details but I like the idea of using a listener. It seems like that might give us the ability to be a bit more selective but also avoid needing to decorate lots of separate keywords.

By the way, there is a previous attempt I made to automatically take screenshots whenever there's an exception in a selenium command. That is here: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/robotframework/utils.py#L88 ...this worked, but I sometimes wished that I would get a screenshot even for exceptions that didn't occur during selenium commands.

cumulusci/robotframework/pageobjects/BasePageObjects.py
davisagli: Can we rename it to RecordPage and use that in docs, but also keep it available as DetailPage as a backward-compatible alias?

cumulusci/robotframework/pageobjects/PageObjects.py
davisagli: That mostly has me convinced. Readability in tests is important. On the other hand, `wait for page object` gives the reader a better clue that the test is switching to a different page object.

cumulusci/robotframework/utils.py
boakley: > I don't understand the point about a delay before test teardown. Won't an error cause the test to end so that test teardown is the very next thing that happens?

Not necessarily. In the normal case it would, but tests could call [Run keyword and continue on failure](http://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Run%20Keyword%20And%20Continue%20On%20Failure) which lets you call additional keywords even after the test has technically failed. We do that in one of our core tests so that we can make multiple assertions rather than stopping at the first failure.

boakley: I'll think a bit more about the listener idea. My gut tells me there's some edge case that I'm forgetting about, but maybe I'm just hungry :-)

davisagli: Okay that makes sense. I didn't know about `Run keyword and continue on failure`.

cumulusci/robotframework/pageobjects/BasePageObjects.py
boakley: Yes, we could. I think I'll save that for a separate work item, since this specific change was just a result of moving code from one file to another.

cumulusci/robotframework/pageobjects/PageObjects.py
boakley: I think in the interest of moving forward I'm going to keep both of these keywords, unless you feel strongly that we should only have one.

docs/robotframework.rst
```diff
@@ -361,6 +361,8 @@ The **PageObjects** library provides the following keywords:
 * Go To Page Object
 * Load Page Object
 * Log Page Object Keywords
+* Wait For Dialog
```
```suggestion
* Wait For Modal
```

docs/robotframework.rst
```diff
@@ -417,6 +419,26 @@ This will load the page object for the given **page_type** and
 page object without first navigating to that page (i.e. when you are
 already on the page and don't want to navigate away).
 
+Wait For Dialog
```
```suggestion
Wait For Modal
```

docs/robotframework.rst
```diff
@@ -417,6 +419,26 @@ This will load the page object for the given **page_type** and
...
+^^^^^^^^^^^^^^^
+
+Example: :code:`Wait for dialog  New  Contact`
```
```suggestion
Example: :code:`Wait for modal  New  Contact`
```

docs/robotframework.rst
```diff
@@ -449,15 +471,18 @@ page objects:
...
 - **Detail** (eg: :code:`Go to page  Detail  Contact  ${contact id}`)
   Detail pages refer to pages with a URL that matches the
   pattern "<host>/lightning/r/<object name>/<object id>/view"
+- **Home** (eg: :code:`Go to page  Home  Contact`)
+  Home pages refer to pages with a URL that matches the pattern
+  "<host>/lightning/o/<object name>/home"
+- **Listing** (eg: :code:`Go to  page  Listing  Contact`)
+  Listing pages refer to pages with a URL that matches the pattern
+  "<host>b/lightning/o/<object name>/list"
+- **New** (eg: :code:`Wait for dialog  New  Contact`)
```
```suggestion
- **New** (eg: :code:`Wait for modal  New  Contact`)
```

cumulusci/robotframework/utils.py
davisagli: Did you think more about the listener idea? I'd rather not add the decorator if we're just going to replace it.

boakley: I think I prefer keeping the decorator. I think it gives more control to our keyword authors.

docs/robotframework.rst
boakley: Ugh. Thanks for catching the robotfamework.rst problems. I grepped .robot and .py files but forgot the documentation files.

docs/robotframework.rst
boakley: fixed

docs/robotframework.rst
boakley: fixed

cumulusci/robotframework/utils.py
boakley: Ugh. I didn't realize we did that. I'll look into it.

boakley: I want to keep the decorator. After doing some investigation I learned that there are many cases where our automatic screenshot code won't capture screenshots, since it only catches exceptions thrown by the selenium module rather than SeleniumLibrary.

Because it doesn't catch everything, I need an easy way to capture screenshots on failure. I can get rid of the decorator and make liberal use of try/except, but I found myself duplicating that pattern over and over which is why I created the decorator in the first place. 

With respect to the concern over duplicate screenshots, at the moment that chance is zero since the page objects aren't using the retrying selenium decorator.  Maybe they should, but they are working fine without it, and do things that that code wouldn't catch.

docs/robotframework.rst
boakley: Thanks for catching this!

cumulusci/robotframework/utils.py
davisagli: @boakley I'm thinking that in the end it would be best if we did all of the following:
1. Add this decorator, as a way to do targeted screenshots when particular keywords fail.
2. Add a listener that takes a screenshot after any keyword fails _if_ a browser is open. This gives an option for users who want easy visibility into failures without needing to track down a specific place to add the decorator and then re-run the tests.
3. Remove the current screenshot capture that's monkeypatched in by the selenium_retry decorator. It's too tightly coupled to the retry decorator, untested, and doesn't catch all the situations where a screenshot is desired.

In the interest of getting this PR merged, I'd suggest we do the first one now and follow up later on 2 and 3. Does that sound okay to you?

boakley: I agree on points 1 and 3, point 2 needs a bit more thought. There are times when we won't want a keyword to generate a screenshot on error, but perhaps we can come up with a solution that addresses that.

PR #1275, opened 2019-10-22
Python 3.8 Updates

Br4nd0R:
* tox.ini updated to include 3.8.0
* We now have three separate jobs for python tests that will run in circleci for py versionse: 3.6, 3.7, and 3.8.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -13,6 +13,7 @@ CumulusCI
 
 CumulusCI is a command line tool belt and set of reusable Python classes useful in the development and release process of building a Salesforce Managed Package application.
 
+
 Key Features
 ------------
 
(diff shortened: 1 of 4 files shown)


README.rst
```diff
@@ -27,6 +28,9 @@ For more information, read the `full documentation`_.
 If you just want a quick intro, watch this screencast demo of using CumulusCI to configure a Salesforce project from a GitHub repository:
 https://asciinema.org/a/91555
 
+Supported Python Versions: 3.6, 3.7, 3.8
```
Br4nd0R: @davisagli - Wasn't sure where to exactly put this. Let me know if there's a standard place to include this, or if you think it would look better under a subheading.

davisagli: This is fine.

PR #1292, opened 2019-10-29
Add a get_latest_api_version method to org_config

prescod:
1. Add a get_latest_api_version method to OrgConfig

2. Use it in places that needed the API version (Salesforce.py in particular)

3. Add it to `cci org info`

4. Refactor the code to get the Salesforce client in OrgConfig and ScratchOrgConfig

5. Fix the tests.

6. Allow black to fix a few Python 2 Unicode-isms (just in one file)


cumulusci/cli/cci.py
```diff
@@ -895,9 +895,12 @@ def org_info(config, org_name, print_json):
         ]
         keys = [key for key in org_config.config.keys() if key in UI_KEYS]
         keys.sort()
+        pairs = [[key, str(org_config.config[key])] for key in keys]
+        pairs.append(["release", str(org_config.latest_api_version)])
```
davisagli: I would call this `api_version`. As it stands API versions are incremented at the same time that releases are made, but I would expect "release" to be either the release number (e.g. 222) or customer-facing release name like Winter '20.

cumulusci/cli/cci.py
```diff
@@ -895,9 +895,12 @@ def org_info(config, org_name, print_json):
         ]
         keys = [key for key in org_config.config.keys() if key in UI_KEYS]
         keys.sort()
```
davisagli: This sort isn't needed since you added the one below.

cumulusci/core/config/OrgConfig.py
```diff
@@ -63,6 +67,26 @@ def _refresh_token(self, keychain, connected_app):
     def lightning_base_url(self):
         return self.instance_url.split(".")[0] + ".lightning.force.com"
 
+    @property
+    def salesforce_client(self):
+        if not self._client:
+            self._client = Salesforce(
```
davisagli: Use cumulusci.salesforce_api.utils.get_simple_salesforce_connection to get the simple-salesforce client, to make sure we're consistent in how it gets set up.

cumulusci/core/config/OrgConfig.py
```diff
@@ -63,6 +67,26 @@ def _refresh_token(self, keychain, connected_app):
...
+            )
+        return self._client
+
+    @property
+    def latest_api_version(self):
+        if not self._latest_api_version:
+            response = self.salesforce_client._call_salesforce(
+                "GET",
+                "https://{}/services/data".format(self.salesforce_client.sf_instance),
+            )
```
davisagli: Use `client.restful("services/data", method="GET")` to avoid using an internal method and to take advantage of the fact that the client already knows how to construct the full URL.

cumulusci/core/config/OrgConfig.py
```diff
@@ -63,6 +67,26 @@ def _refresh_token(self, keychain, connected_app):
...
+            self._latest_api_version = float(response.json()[-1]["version"])
```
davisagli: I think it'd probably be better to just store it as a string; I can't think of a case where we actually want it as a float.

cumulusci/cli/cci.py
prescod: You are just referring to this line, right? The name OrgConfig.latest_api_version is fine?

cumulusci/core/config/OrgConfig.py
prescod: That function wants a project_config. Will we always have it whenever we want to grab a salesforce_client from an org? The OrgConfig object doesn't seem to have a reference to the project_config so it would need to be passed from the client. Or get_simple_salesforce_connection could be refactored to make the project_config optional.

cumulusci/core/config/OrgConfig.py
prescod:
```
>>> from cumulusci.salesforce_api.utils import get_simple_salesforce_connection
>>> from cumulusci.cli import cci
>>> config = cci.load_config()
>>> org_name, org_config = config.get_org("dev")
>>> sf = get_simple_salesforce_connection(config.project_config, org_config)
2019-10-29 16:39:38: Getting scratch org info from Salesforce DX
sf.restful("services/data")
>>> sf.restful("services/data")
  File "<console>", line 1, in <module>
  File "/Users/pprescod/code/envs/cumulusci-py3/lib/python3.7/site-packages/simple_salesforce/api.py", line 305, in restful
    **kwargs)
  File "/Users/pprescod/code/envs/cumulusci-py3/lib/python3.7/site-packages/simple_salesforce/api.py", line 478, in _call_salesforce
    exception_handler(result, name=name)
  File "/Users/pprescod/code/envs/cumulusci-py3/lib/python3.7/site-packages/simple_salesforce/util.py", line 61, in exception_handler
    raise exc_cls(result.url, result.status_code, name, response_content)
simple_salesforce.exceptions.SalesforceResourceNotFound: Resource services/data Not Found. Response content: [{'errorCode': 'NOT_FOUND', 'message': 'The requested resource does not exist'}]

```

With debug logging turned on I can see:

DEBUG:urllib3.connectionpool:https://flow-page-8225-dev-ed.cs40.my.salesforce.com:443 "GET /services/data/v46.0/services/data HTTP/1.1" 404 None

The same thing happens if I add a leading slash to the relative URL. With enough ../../.. stuff I can get it to work but that presumably is not what you were intending.

cumulusci/core/config/OrgConfig.py
prescod:
```
>>> from cumulusci.salesforce_api.utils import get_simple_salesforce_connection
>>> from cumulusci.cli import cci
>>> config = cci.load_config()
>>> org_name, org_config = config.get_org("dev")
>>> sf = get_simple_salesforce_connection(config.project_config, org_config)
2019-10-29 16:39:38: Getting scratch org info from Salesforce DX
sf.restful("services/data")
>>> sf.restful("services/data")
  File "<console>", line 1, in <module>
  File "/Users/pprescod/code/envs/cumulusci-py3/lib/python3.7/site-packages/simple_salesforce/api.py", line 305, in restful
    **kwargs)
  File "/Users/pprescod/code/envs/cumulusci-py3/lib/python3.7/site-packages/simple_salesforce/api.py", line 478, in _call_salesforce
    exception_handler(result, name=name)
  File "/Users/pprescod/code/envs/cumulusci-py3/lib/python3.7/site-packages/simple_salesforce/util.py", line 61, in exception_handler
    raise exc_cls(result.url, result.status_code, name, response_content)
simple_salesforce.exceptions.SalesforceResourceNotFound: Resource services/data Not Found. Response content: [{'errorCode': 'NOT_FOUND', 'message': 'The requested resource does not exist'}]

```

With debug logging turned on I can see:

DEBUG:urllib3.connectionpool:https://flow-page-8225-dev-ed.cs40.my.salesforce.com:443 "GET /services/data/v46.0/services/data HTTP/1.1" 404 None

The same thing happens if I add a leading slash to the relative URL. With enough ../../.. stuff I can get it to work but that presumably is not what you were intending.

cumulusci/core/config/OrgConfig.py
davisagli: @prescod sorry, I was thinking about a different context. I'd like to adjust get_simple_salesforce_connection to do something reasonable when a project_config isn't available, but we can do that separately.

cumulusci/core/config/OrgConfig.py
```diff
@@ -63,6 +67,26 @@ def _refresh_token(self, keychain, connected_app):
     def lightning_base_url(self):
         return self.instance_url.split(".")[0] + ".lightning.force.com"
 
+    @property
+    def salesforce_client(self):
+        if not self._client:
+            self._client = Salesforce(
+                instance=self.instance_url.replace("https://", ""),
+                session_id=self.access_token,
+                version=45,
```
davisagli: Does this actually work? I think it needs to be "45.0"

cumulusci/core/config/OrgConfig.py
davisagli: @prescod sorry, I forgot it adds that portion of the path too. We can keep using _call_salesforce here.

PR #1298, opened 2019-10-30
Feature/connect sandbox

jofsky:
Addressing connect sandbox display issue on keychain

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -33,6 +33,7 @@ from cumulusci.core.exceptions import ScratchOrgException
 from cumulusci.core.exceptions import ServiceNotConfigured
 from cumulusci.core.exceptions import FlowNotFoundError
 
+from cumulusci.core.sfdx import sfdx
 from cumulusci.core.utils import import_global
 from cumulusci.cli.config import CliRuntime
 from cumulusci.cli.config import get_installed_version
@@ -794,8 +795,9 @@ def org_browser(config, org_name):
 @pass_config
 def org_connect(config, org_name, sandbox, login_url, default, global_org):
     config.check_org_overwrite(org_name)
-
+    print(config.project_config)
     connected_app = config.keychain.get_service("connected_app")
+    print(vars(connected_app))
     if sandbox:
         login_url = "https://test.salesforce.com"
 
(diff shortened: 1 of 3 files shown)


cumulusci/cli/cci.py
```diff
@@ -807,15 +807,50 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+        config_value = (
+            io.TextIOWrapper(result.stdout, encoding=sys.stdout.encoding).read()
+        )
+        parsed = config_value.split("\n")
+        # skipping first two header rows
+        for i in range(3,len(parsed)-1):
+            value = parsed[i].split(' ')
+            if value[0].lower() == "expiration" and value[len(value)-1] != '':
+                org_config = ScratchOrgConfig(oauth_dict,org_name)
+                org_config.config.update({"username": username})
```
davisagli: This isn't a good way to get the org info because it will only work for orgs that sfdx already knows about, and the whole reason for this feature is to be able to connect an org that someone else created and neither sfdx nor cci know about yet.

Instead, run `org_config._load_orginfo()` and then look at the values in org_config.organization_sobject which includes the values documented in https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_organization.htm

cumulusci/cli/cci.py
```diff
@@ -807,15 +807,50 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+        for i in range(3,len(parsed)-1):
+            value = parsed[i].split(' ')
+            if value[0].lower() == "expiration" and value[len(value)-1] != '':
+                org_config = ScratchOrgConfig(oauth_dict,org_name)
+                org_config.config.update({"username": username}) 
     if default:
         config.keychain.set_default_org(org_name)
         click.echo(f"{org_name} is now the default org")
 
+    org_config.refresh_oauth_token(config.keychain)
```
davisagli: What's this for? We should already have a fresh access token from the oauth flow above.

cumulusci/cli/cci.py
davisagli: Make sure you add "scratch": True to the config dict when creating the ScratchOrgConfig instance. This is needed so that it knows to use ScratchOrgConfig instead of OrgConfig when the org is loaded back from the keychain in the future.

jofsky: Yeah I was going to change this parsing not proud of this work. Why I didn't want to raise a pr yet. Will implement the change.

cumulusci/cli/cci.py
jofsky: Not sure why I have this if I'm being totally honest

cumulusci/cli/cci.py
davisagli: @jofsky If it's a scratch org I think we'll need to add it to the sfdx keychain too, in addition to the cci keychain. I haven't used it before but the auth:sfdxurl:store sfdx command (https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_force_auth.htm#cli_reference_force_auth) looks like it could help with that. You have to create a temporary file containing `force://<refreshToken>@<instanceUrl>` (using the refresh_token and instance_url from the org_config) and then call the force:auth:sfdxurl command referencing that file. See how this goes and we can talk about it...

cumulusci/cli/cci.py
```diff
@@ -809,13 +809,63 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+                )
+                - datetime.strptime(
+                    org_config.organization_sobject["CreatedDate"], datetime_format
+                )
+            ).days
+            # need to handle this case!
+            if org_config.organization_sobject["NamespacePrefix"]:
+                org_config.config.update(
+                    {"namespaced": org_config.organization_sobject["NamespacePrefix"]}
+                )
```
davisagli: `namespaced` is a boolean based on whether the org has a namespace; you can just do `"namespaced": bool(org_config.organization_sobject["NamespacePrefix"])`

cumulusci/cli/cci.py
```diff
@@ -809,13 +809,63 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+                    ),
+                    "days": days,
+                }
+            )
+            if not org_config.expired:
+                org_config.config.update({"active": True})
+        else:
+            # Unsure of this case below. I believe a sandbox that is
+            # not a scratch org is a trialforce org.
+            click.echo("Connecting as a trialforce org")
```
davisagli: Sandboxes don't have anything to do with Trialforce and they are a type of persistent org. You can do something like this for the overall structure of this code:

    is_scratch = org_config.organization_sobject["IsSandbox"] and org_config.organization_sobject["TrialExpirationDate"]
    if is_scratch:
        # connect as scratch org
    else:
        click.echo("Connecting as a persistent org.")

cumulusci/cli/cci.py
```diff
@@ -809,13 +809,63 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+            if org_config.organization_sobject["NamespacePrefix"]:
+                org_config.config.update(
+                    {"namespaced": org_config.organization_sobject["NamespacePrefix"]}
+                )
+            org_config.config.update(
+                {
+                    "created": True,
+                    "config_file": "",
+                    "config_name": None,
+                    "email_address": org_config.userinfo__email,
```
davisagli: Let's store the email_address and username for persistent orgs too.

davisagli:
Closing since we went in a different direction with #1320

Issue #1134, opened 2019-06-27
Validate format of api_version


davisagli:
If api_version is written as a number `45` instead of a string`'45.0'` in cumulusci.yml, the Salesforce client will generate bad URLs which leads to hard-to-debug errors (see #1131 for an example). We should make CumulusCI validate that it's a valid API version when the project config is loaded.

davisagli:
Fixed in cumulusci 3.1.2

PR #1301, opened 2019-10-31
Feature/brandon/validate api version format

Br4nd0R:
# Changes
* Validate that api_format is of type `str` and not `int` upon load of configs.

# Issues Closed
* This resolves issue #1134


Br4nd0R:
@davisagli - Let me know if you want something more robust like checking for '.0' within the string.

cumulusci/core/config/project_config.py
```diff
@@ -116,6 +116,18 @@ def _load_config(self):
...
+        self._validate_config()
+
+    def _validate_config(self):
+        """Performs validation checks on the configuration"""
+        self._validate_package_api_format()
+
+    def _validate_package_api_format(self):
+        api_version = self.config["project"]["package"]["api_version"]
+        if isinstance(api_version, int):
+            message = "Package API Version must be specified as a string ('45.0') not an int (45)."
```
davisagli: We should probably explicitly validate that it is the format we need (a string) rather than just checking for int as one bad case. I think yaml like `45.0` might actually end up as a float if it's not in quotes. Someone could even accidentally put in a dict or date or something else.

You asked about validating for "45.0" vs "45". Let's decide that based on what actually works when calling the API. I suspect we need the full dotted version but could you check?

cumulusci/core/config/project_config.py
```diff
@@ -123,9 +124,13 @@ def _validate_config(self):
         self._validate_package_api_format()
 
     def _validate_package_api_format(self):
+        api_version_regex = r"^\d\d+\.0$"
```
davisagli: This is probably unnecessary optimization, but I always tend to define regexes as a constant at module scope so they only get compiled once:

    import re
    API_VERSION_RE = re.compile(r"^\d\d+\.0$")

(and then use it like `API_VERSION_RE.match(target_string)`)

cumulusci/core/config/project_config.py
```diff
@@ -123,9 +124,13 @@ def _validate_config(self):
...
         api_version = self.config["project"]["package"]["api_version"]
-        if not isinstance(api_version, str):
-            message = f"Package API Version must be specified as a string (i.e. '45.0'). Found: {type(api_version)}"
+
+        if not re.match(api_version_regex, api_version):
```
davisagli: I think this will throw an exception if api_version is not actually a string. Let's do `api_version = str(api_version)` to make sure it is one, then do the validation against the string and return it.

PR #1313, opened 2019-11-04
Add support for injecting 'ns.' for Apex references

davidmreed:
# Critical Changes

# Changes

- Added the `%%%NAMESPACE_DOT%%%` and `%%%NAMESPACED_ORG_DOT%%%` tokens to inject `ns.` instead of `ns__`, for references to Apex classes.

# Issues Closed


davisagli:
@davidmreed can you show me an example of where you want to use this, just so I can make sure I understand the use case?

davidmreed:
@davisagli  I'm going to close this and put it on the back shelf. Misunderstanding between me and the CaseMan team about a behavior in a namespaced org.

Issue #1101, opened 2019-06-03
`cci org import` should set days based on the existing org's expiration


davisagli:
`cci org import` adds an sfdx org to the CCI keychain. It currently always sets `days: 1`. Instead it should set `days` based on the existing org's expiration date.

davisagli:
Fixed in cumulusci 3.1.2

PR #1314, opened 2019-11-04
Fix `days` on `cci org import`

Br4nd0R:
# Changes
* `cci org import` now sets `days` to calculated value

# Issues Closed
* This closes issue #1101

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -848,6 +848,7 @@ def org_import(config, username_or_alias, org_name):
 
     info = scratch_org_config.scratch_info
     scratch_org_config.config["days"] = calculate_org_days(info)
+    scratch_org_config.config["created_date"] = info["created_date"]
 
     config.keychain.set_org(scratch_org_config)
     click.echo(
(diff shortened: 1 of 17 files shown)


cumulusci/cli/cci.py
```diff
@@ -851,6 +856,15 @@ def org_import(config, username_or_alias, org_name):
     )
 
 
+def calculate_org_days(info):
+    """Returns the difference in days between created_date (ISO 8601),
+    and expiration_date (%Y-%m-%d)"""
+    created_date_str = info["created_date"].split("T")[0]
```
Br4nd0R: Not sure if there's a better way to easily convert the ISO datetieme to a simple date.

jofsky: datetime_format = "%Y-%m-%dT%H:%M:%S.%f+0000"
will work

Br4nd0R: well, I'm already using strptime() to process that date portion, which is all I want. I guess I was looking to see if there's a one-liner that converts the ISO formatted string to a date.

jofsky: Gotcha, not sure about that one.

PR #1316, opened 2019-11-05
More compassionate post-authentication message

prescod:
# Critical Changes

# Changes

Nicer confirmation message after you collect an org on CCI.

### DIFF
--- a/cumulusci/oauth/salesforce.py
+++ b/cumulusci/oauth/salesforce.py
@@ -119,7 +119,7 @@ class OAuthCallbackHandler(BaseHTTPRequestHandler):
             http_body = f"error: {args['error'][0]}\nerror description: {args['error_description'][0]}"
         else:
             http_status = http.client.OK
-            http_body = "OK"
+            http_body = "Congratulations! Your authentication succeeded."
             code = args["code"]
             self.parent.response = self.parent.oauth_api.get_token(code)
             if self.parent.response.status_code >= http.client.BAD_REQUEST:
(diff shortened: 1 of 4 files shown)


cumulusci/oauth/salesforce.py
```diff
@@ -119,7 +119,7 @@ def do_GET(self):
             http_body = f"error: {args['error'][0]}\nerror description: {args['error_description'][0]}"
         else:
             http_status = http.client.OK
-            http_body = "OK"
+            http_body = "Congratulations! Your authentication succeeded."
```
davisagli: Nice! How about:

```suggestion
            http_body = "Congratulations! Your org is now connected to CumulusCI."
```

boakley: I just noticed this yesterday and wrote a note to submit an issue. "OK" was a pretty lame result.

What you have here is definitely better, but if we're going to fix it, couldn't we do a little better than just a plain text message? For example, instead of "Congratulations" couldn't we embed a little astro image or a metaci logo or something?

prescod: @davisagli I was being slightly careful because I don't know if there are any possible failure cases where authentication succeeds but "connection" fails. If you feel confident that they are basically equivalent then I will use your message.

@boakley If you want to submit HTML for a fancy version then I'm enthusiastic about that.

davisagli: @prescod ah, fair enough. Deciding what to do with the authenticated token is outside of this oauth handler's responsibilities. We could make it happen (if we allow for passing in a Python function as a post-auth callback) but that's more invasive than I was thinking.

PR #1317, opened 2019-11-05
Feature/more flexible format str

prescod:
Internal reorganization of this code.

Allow callers to pass arbitrary keyword arguments to this Jinja function.

Make the function more globally visible by moving the file

### DIFF
--- a/cumulusci/core/template_utils.py
+++ b/cumulusci/core/template_utils.py
@@ -1,4 +1,5 @@
 from functools import lru_cache
+from faker import Faker
 
 from jinja2 import Template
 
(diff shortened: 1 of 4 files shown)


prescod:
@davisagli I've double checked that there are no references in Cumulus. All such references go through the Robot library which is updated here.

I also added another feature today, which will be useful in my other project. The ability to generate e.g. Spanish-sounding names.

cumulusci/core/template_utils.py
```diff
@@ -47,9 +48,11 @@ def __getattr__(self, name):
 
 faker_template_library = FakerTemplateLibrary()
 
+Template = lru_cache(512)(Template)
+
 
-def format_str(value, i):
-    if isinstance(value, str):
-        value = Template(value).render(number=i, fake=faker_template_library)
+def format_str(value, variables={}, fake=faker_template_library):
```
davisagli: Don't use mutable defaults! It's a trap!

prescod: Good catch!

prescod: Although it is a trap, in this case the function doesn't (and shouldn't) mutate the dict. In the future I would use a frozenmap as the default to enforce that.

PR #1319, opened 2019-11-05
Data Dictionary Generator

davidmreed:
# Critical Changes

# Changes

- The task `generate_data_dictionary` has been added. This task indexes the fields and objects created in each GitHub release for the project and generates a data dictionary in CSV format.

# Issues Closed


cumulusci/tasks/tests/test_datadictionary.py
```diff
@@ -0,0 +1,559 @@
...
+    <fields>
+        <fullName>Type__c</fullName>
+        <inlineHelpText>Type of field.</inlineHelpText>
+        <label>Type</label>
+        <type>Text</type>
+    </fields>
+</CustomObject>"""
+        task = create_task(GenerateDataDictionary, {})
+        task.project_config.keychain.get_service = Mock()
+        task.project_config.init_sentry = Mock()
```
davidmreed: Is there a less-monkeypatchy, more-recommended way to mock this out?

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,304 @@
...
+            "description": "Path to a CSV file to contain an field-level data dictionary.",
+            "default": "Field Data Dictionary.csv",
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+
+        self.options["object_path"] = (
+            self.options.get("object_path") or "sObject Data Dictionary.csv"
```
Br4nd0R: It looks like the default filenames are used more than once. Could we define the default filenames as constant class members and then refer to them here and where the defaults are defined in `task_options`?

cumulusci/tasks/tests/test_datadictionary.py
davisagli: hmm, init_sentry should exit right away anyway as long as the keychain doesn't have a sentry service configured. So maybe just don't mock keychain.get_service?

I need to redo all the sentry stuff at some point, since we're using the raven lib instead of sentry-sdk which is now preferred.

cumulusci/tasks/datadictionary.py
davisagli: @Br4nd0R the defaults in task_options aren't actually used for anything at this time.

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,304 @@
...
+    task_options = {
+        "object_path": {
+            "description": "Path to a CSV file to contain an sObject-level data dictionary.",
+            "default": "sObject Data Dictionary.csv",
+        },
+        "field_path": {
+            "description": "Path to a CSV file to contain an field-level data dictionary.",
+            "default": "Field Data Dictionary.csv",
+        },
+    }
```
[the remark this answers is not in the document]
davidmreed: I like including the package name. I'll switch that around.

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,304 @@
...
+                for p in zip_name_list
+            ]
+            if "src/objects" in self.name_list:
+                # MDAPI format
+                self._process_mdapi_release(zip_file, version)
+
+            if "force-app/main/default/objects" in self.name_list:
+                # FIXME: check sfdx-project.json for directories to process.
+                # SFDX format
+                self._process_sfdx_release(zip_file, version)
```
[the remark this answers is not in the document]
davidmreed: Modifications for the previous comment make it easy to elide this kludge. Removed!

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,304 @@
...
+                            field_data["type"],
+                            field_data["help_text"],
+                            field_data["picklist_values"],
+                            field_data["version"],
+                        ]
+                    )
+
+    def _version_from_tag_name(self, tag_name):
+        return LooseVersion(
+            tag_name[len(self.project_config.project__git__prefix_release) :]
```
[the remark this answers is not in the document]
davidmreed: Yes, I'll skip releases missing the assigned prefix, and will also stop the task if the project does not have a release tag prefix configured.

cumulusci/tasks/datadictionary.py
Br4nd0R: Even if they aren't used and we have two explicitly defined strings with the same value, wouldn't it make sense to create a variable and reference that instead of defining the same string twice?

davidmreed:
@Br4nd0R:

>One question: What occurs if a field is removed from an object? I probably missed where this case is taken care of, but just wondering for my own clarification.

Currently we don't handle it at all. Removing fields is fairly uncommon. Do you feel not having that is a deal-breaker for this version?

Br4nd0R:
@davidmreed - Not a deal breaker for me at all, and thanks for the explanation. I was assuming that deletion of fields might happen more frequently than they actually do.

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,304 @@
...
+                        sobject_name,
+                        ET.fromstring(
+                            zip_file.read(
+                                os.path.join(
+                                    os.path.sep, self.zip_prefix, obj_file
+                                ).strip(os.path.sep)
+                            )
+                        ),
+                        version,
+                    )
```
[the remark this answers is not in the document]
davidmreed: As well as rewriting with pathlib, I validated the assumption of POSIX-style paths on Windows.

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,281 @@
...
+
+        if self.options.get("field_path") is None:
+            self.options[
+                "field_path"
+            ] = f"{self.project_config.project__name} Field Data Dictionary.csv"
+
+        if not self.project_config.project__git__prefix_release:
+            raise CumulusCIException(
+                "Set the `project__git__prefix_release` property in `cumulusci.yml` to use this task"
+            )
```
davisagli: I'd suggest making this a required task option and specify it as `release_prefix: $project_config.project__git__prefix_release` in the task config in cumulusci.yml.

Reasons:
1. The default from cumulusci.yml will be shown when you run `cci task info`
2. It makes it easier to do a one-off run of the task with a different prefix without editing cumulusci.yml

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,281 @@
...
+
+    def _run_task(self):
+        self.logger.info("Starting data dictionary generation")
+
+        self._init_schema()
+        self._walk_releases()
+        self._write_results()
+
+    def _init_schema(self):
```
davisagli: Could you add some docstrings?

cumulusci/tasks/datadictionary.py
```diff
@@ -0,0 +1,281 @@
...
+            zip_file = download_extract_github_from_repo(repo, ref=release.tag_name)
+            version = self._version_from_tag_name(release.tag_name)
+            self.logger.info(f"Analyzing version {version}")
+
+            if "src/objects/" in zip_file.namelist():
+                # MDAPI format
+                self._process_mdapi_release(zip_file, version)
+
+            if "force-app/main/default/objects/" in zip_file.namelist():
+                # FIXME: check sfdx-project.json for directories to process.
```
davisagli: I've got some code in the retrieve-changes-better branch which does this too. We should probably turn it into a method of BaseProjectConfig.

davidmreed: I'd like to reserve that improvement (along with handling dependencies) for the next version, by which time retrieve-changes-better I assume will be merged.

Issue #1282, opened 2019-10-24
CCI org list falsely displays scratch orgs as persistent orgs 


jofsky:
When executing the cci org connect --sandbox <username> is used, cci org list falsely displays sandbox and scratch orgs as persistent orgs.

<img width="1187" alt="Screen Shot 2019-10-24 at 11 27 13 AM" src="https://user-images.githubusercontent.com/28512644/67514600-d9e5a300-f651-11e9-8849-96576868306c.png">

davisagli:
Sandboxes are persistent orgs; the issue is that we aren't recognizing scratch orgs and setting the org config's `scratch` setting to true.

PR #1320, opened 2019-11-06
pushing adjusted connect method to handle all connected orgs

jofsky:
Implemented a Connected Orgs table instead of Persistent Orgs table, which lists **None** if a persistent and the time of expiration if it is a scratch or demo org.

# Critical Changes

# Changes

# Issues Closed
Fixes [#1282](https://github.com/SFDO-Tooling/CumulusCI/issues/1282)


cumulusci/cli/cci.py
```diff
@@ -809,6 +809,13 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
     oauth_dict = oauth_capture()
     org_config = OrgConfig(oauth_dict, org_name)
     org_config.load_userinfo()
+    org_config._load_orginfo()
+    if not org_config.organization_sobject["TrialExpirationDate"]:
+        org_config.config["expiration"] = None
+    else:
+        org_config.config["expiration"] = org_config.organization_sobject[
+            "TrialExpirationDate"
+        ]
```
davisagli: In the case of ScratchOrgConfig, there is a Python property called `expires` which returns a datetime object (or maybe it's a date? I forget) calculated from the `date_created` and the `days`.

I think we should aim to have both types of org config return something similar when you access `org_config.expires`. So that means we should parse the TrialExpirationDate into a date or datetime and store it in `expires` instead of `expiration`.

cumulusci/cli/cci.py
```diff
@@ -946,6 +951,10 @@ def org_list(config, plain):
                 "username", org_config.userinfo__preferred_username
             )
             row.append(username)
+            if "expiration" in org_config.config.keys():
+                row.append(org_config.config["expiration"])
+            else:
+                row.append("None")
```
davisagli: You could simply do:

    row.append(org_config.expiration or "None")

But I think it might be slightly easier to understand if we use "never" as the text when there's no expiration instead of "none"

cumulusci/cli/cci.py
```diff
@@ -958,10 +967,11 @@ def org_list(config, plain):
     wrap_cols = ["Username"] if not plain else None
     persistent_table = CliTable(
         persistent_data,
-        title="Persistent Orgs",
+        title="Connected Orgs",
         wrap_cols=wrap_cols,
         bool_cols=["Default"],
     )
+    persistent_table.stringify_boolean_col(col_name="Default", true_str=CHECKMARK)
```
davisagli: What does this do?

cumulusci/cli/tests/test_cci.py
```diff
@@ -548,7 +547,7 @@ def test_org_connect(self, oauth):
 
         config.check_org_overwrite.assert_called_once()
         config.keychain.set_org.assert_called_once()
-        config.keychain.set_default_org.assert_called_once_with("test")
+        # config.keychain.set_default_org.assert_called_once_with("test")
```
davisagli: Why is this commented out?

cumulusci/cli/cci.py
jofsky: I was thinking about Persistent what do you think?

cumulusci/cli/tests/test_cci.py
jofsky: Ty for catching this

cumulusci/cli/cci.py
jofsky: vestige of previous code design before rewrite

cumulusci/cli/tests/test_cci.py
jofsky: However it's failing this test although I expect it not to do you know why this would be?

cumulusci/cli/cci.py
jofsky: Screen Shot 2019-11-06 at 12 34 01 PM

Proposed feature functionality, Unknown if its been connected to key chain pre update, Persistent if persistent org and Date time if not persistent org

cumulusci/cli/tests/test_cci.py
```diff
@@ -715,9 +714,12 @@ def test_org_list(self, cli_tbl):
             dim_rows=[0, 1],
         )
         persistent_table_call = mock.call(
-            [["Name", "Default", "Username"], ["test2", False, "test2@example.com"]],
+            [
+                ["Name", "Default", "Username", "Expires"],
+                ["test2", False, "test2@example.com", "Unknown"],
```
jofsky: I don't know about this seems suspect.

cumulusci/cli/cci.py
davisagli: @jofsky I like that. We should probably just show the date and not the time, since expiration doesn't take effect immediately (I think there's a job that runs each night to reclaim expired orgs)

cumulusci/cli/cci.py
```diff
@@ -809,6 +807,14 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
     org_config = OrgConfig(oauth_dict, org_name)
     org_config.load_userinfo()
+    org_config._load_orginfo()
+    if not org_config.organization_sobject["TrialExpirationDate"]:
+        org_config.config["expires"] = "Persistent"
+    else:
+        datetime_format = "%Y-%m-%dT%H:%M:%S.%f+0000"
+        org_config.config["expires"] = datetime.datetime.strptime(
+            org_config.organization_sobject["TrialExpirationDate"], datetime_format
+        )
```
davisagli: Use cumulusci.utils.parse_api_datetime

cumulusci/cli/tests/test_cci.py
davisagli: @jofsky it makes sense to me -- the sample OrgConfig created above doesn't have an `expires` key set, so "Unknown" is what we expect to find here.

It would be good to add a second sample OrgConfig that _does_ have an expiration date set, and add a line here to confirm that it gets displayed with the expected values.

cumulusci/cli/cci.py
jofsky: Ok should I reformat too? I feel like I should.

cumulusci/cli/tests/test_cci.py
jofsky: Ok will add

cumulusci/cli/cci.py
```diff
@@ -753,8 +754,6 @@ def service_info(config, service_name, plain):
 
 
 # Commands for group: org
-
-
```
davisagli: The blank lines here were intentional since the comment is describing the group of several commands that follows; not just org_browser.

cumulusci/cli/cci.py
```diff
@@ -809,6 +807,19 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+    if (
+        org_config.organization_sobject
+        and "TrialExpirationDate" in org_config.organization_sobject.keys()
+    ):
+        if org_config.organization_sobject["TrialExpirationDate"] is None:
+            org_config.config["expires"] = None
+        else:
+            org_config.config["expires"] = parse_api_datetime(
+                org_config.organization_sobject["TrialExpirationDate"]
+            )
```
davisagli: You can see in the coveralls report that lines 815-820 are not covered by the existing tests. It looks like we need to add a test for org_connect in the scenario where the org has a TrialExpirationDate.

cumulusci/cli/cci.py
```diff
@@ -809,6 +807,19 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+    else:
+        org_config.config["expires"] = "Persistent"
```
davisagli: I connected a persistent org but it shows as "Unknown" -- I think it ended up on line 816 because TrialExpirationDate was present but None. I think you have lines 816 and 822 reversed from where they should be.

cumulusci/cli/cci.py
davisagli: @jofsky I'd still like to see this formatted without the time.

cumulusci/cli/cci.py
jofsky: Was going to ask you about this!

cumulusci/cli/cci.py
jofsky: good catch

AUTHORS.rst
```diff
@@ -11,6 +11,7 @@ CumulusCI is developed by the Salesforce.org Release Engineering team:
 * Christian Carter <christian.carter@salesforce.com>
 * David Glick <dglick@salesforce.com>
 * James Estevez <jestevez@salesforce.com>
+* Josh Kofsky <jkofsky@salesforce.com>
```
davisagli: @jofsky heh, I forgot about this file. Could you add the rest of the team too please? (ask them to make sure it's okay to use their email addresses)

cumulusci/cli/cci.py
```diff
@@ -811,6 +811,19 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+    if (
+        org_config.organization_sobject
+        and "TrialExpirationDate" in org_config.organization_sobject.keys()
+    ):
+        if org_config.organization_sobject["TrialExpirationDate"] is None:
+            org_config.config["expires"] = "Persistent"
+        else:
+            org_config.config["expires"] = parse_api_datetime(
+                org_config.organization_sobject["TrialExpirationDate"]
+            ).strftime("%m/%d/%Y")
```
davisagli: I want to store the actual datetime object, not a string formatted from it. Then do the formatting [where it's used in output](https://github.com/SFDO-Tooling/CumulusCI/pull/1320/files#diff-34b79c3fe3ee83def757b70a69cb5770R974). That way it's not a problem if we decide we want to show it in a different format, or do some date arithmetic based on it.

Also I would use dt.date().isoformat() to format the date as YYYY-MM-DD since that's easier to understand in different locales (it's primarily just the US that uses mm/dd/YYYY)

davisagli:
@jofsky Here's what I've been trying to say we should do:
1. parse the datetime in cci_connect, and store a datetime object in the org config
2. format the datetime using .date().isoformat() in org_info

You're currently doing both the parsing and formatting in org_info, and still not formatting the date in ISO format.

cumulusci/cli/cci.py
```diff
@@ -811,6 +810,26 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+        if org_config.organization_sobject["TrialExpirationDate"] is None:
+            org_config.config["expires"] = "Persistent"
+        else:
+            print(
+                type(
+                    parse_api_datetime(
+                        org_config.organization_sobject["TrialExpirationDate"]
+                    )
+                )
+            )
```
davisagli: I assume this was for debugging -- please remove it.

cumulusci/cli/cci.py
```diff
@@ -961,6 +977,14 @@ def org_list(config, plain):
                 "username", org_config.userinfo__preferred_username
             )
             row.append(username)
+            if org_config.expires:
+                if org_config.expires != "Persistent":
+                    row.append(org_config.expires)
+                else:
+                    row.append(org_config.expires)
+            else:
+                row.append("Unknown")
```
davisagli: What's going on here? You're doing the same thing in the `if` and `else` blocks of the `if org_config.expires != "Persistent":` condition. This entire thing can be simplified down to:

    row.append(org_config.expires or "Unknown")

cumulusci/cli/tests/test_cci.py
```diff
@@ -523,8 +523,9 @@ def test_org_browser(self, browser_open):
         config.keychain.set_org.assert_called_once_with(org_config)
 
     @mock.patch("cumulusci.cli.cci.CaptureSalesforceOAuth")
+    @mock.patch("cumulusci.cli.cci.OrgConfig")
```
davisagli: The coveralls report of test coverage (https://coveralls.io/builds/27073372) shows that there is a newly uncovered line in the load_userinfo method of OrgConfig. That is a consequence of changing this to mock out the entire OrgConfig object. I'd suggest just mocking the OrgConfig._load_userinfo method instead so that OrgConfig.load_userinfo will still run.

```suggestion
    @mock.patch("cumulusci.cli.cci.OrgConfig._load_userinfo")
```

(We aim to never decrease test coverage.)

cumulusci/cli/cci.py
```diff
@@ -811,6 +810,26 @@ def org_connect(config, org_name, sandbox, login_url, default, global_org):
...
+            print(
+                type(
+                    parse_api_datetime(
+                        org_config.organization_sobject["TrialExpirationDate"]
+                    )
+                )
+            )
+            org_config.config["expires"] = parse_api_datetime(
+                org_config.organization_sobject["TrialExpirationDate"]
+            ).date()
```
davisagli: The test coverage report (https://coveralls.io/builds/27073372/) shows that lines 818-830 do not have coverage. So basically that means that there aren't any tests which confirm that the correct values are stored in the org config when connecting an org that has a TrialExpirationDate.

You'll want to add 2 new tests in test_cci.py based on the existing test_org_connect test. Add a new fake response for the call to get the organization sobject:

        responses.add(
            method="GET",
            url="https://instance/services/data/v45.0/sobjects/Organization/None",
            json={"TrialExpirationDate": None},
            status=200,
        )

In the other new test, make it return a date in the Salesforce format.

As a reminder, you can check test coverage locally by running `make coverage`

PR #1334, opened 2019-11-12
YAML Data Generator

prescod:
# Critical Changes

# Changes

Adds a new YAML format for demo data generation.

# Issues Closed


cumulusci/tasks/bulkdata/data_generation/data_gen_exceptions.py
```diff
@@ -0,0 +1,23 @@
+class DataGenError(Exception):
```
davisagli: would it make sense to extend the builtin SyntaxError?

cumulusci/tasks/bulkdata/data_generation/data_generator_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def field_funcs(self):
+        def curry(func):
+            rc = partial(func, self)
+            if hasattr(func, "lazy"):
+                rc.lazy = func.lazy
+            return rc
+
+        funcs = {name: curry(func) for name, func in template_funcs.items()}
+        return {**funcs}
```
davisagli: It looks like `field_funcs` gets called once for each SimpleValue and StructuredValue, but could be memoized so that it's only done once per context.

cumulusci/tasks/bulkdata/data_generation/data_generator_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+    """Add filename and linenumber to an exception if needed"""
+    filename, line_num = parentobj.filename, parentobj.line_num
+    if isinstance(e, DataGenError):
+        if not e.filename:
+            e.filename = filename
+        if not e.line_num:
+            e.line_num = line_num
+        raise e
+    else:
+        raise DataGenError(message, filename, line_num) from e
```
davisagli: Maybe this function should go in the exceptions module

cumulusci/tasks/bulkdata/data_generation/data_generator_runtime.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+class ObjectRow:
+    """Represents a single row
+
+    Uses __getattr__ so that the template evaluator can use dot-notation."""
+
+    def __init__(self, tablename, values=()):
+        self._tablename = tablename
+        self._values = values
```
davisagli: Would it be worth using `__slots__` to avoid the memory overhead of an instance dict?

cumulusci/tasks/bulkdata/data_generation/snowfakery.py
```diff
@@ -0,0 +1,208 @@
...
+            raise e
+        else:
+            click.echo("")
+            click.echo(
+                "An error occurred. If you would like to see a Python traceback, use the --debug-internals option."
+            )
+            raise click.ClickException(str(e)) from e
+    finally:
+        ...
+        # TODO: close output streams
```
davisagli: Maybe output streams should be context managers.

cumulusci/tasks/bulkdata/generate_and_load_data.py
```diff
@@ -81,7 +89,7 @@ def _init_options(self, kwargs):
             )
 
     def _run_task(self):
-        with temporary_dir() as tempdir:
+        with TemporaryDirectory() as tempdir:
```
davisagli: temporary_dir changes to the directory and TemporaryDirectory does not. Is that a change that matters?

cumulusci/tasks/bulkdata/tests/mapping_vanilla_sf.yml
```diff
@@ -57,11 +57,13 @@ Insert Opportunities:
...
-    ContactId: ContactId
-    AccountId: AccountId
     Amount: Amount
   lookups:
     AccountId:
       table: Account
-    ContactId:
-      table: Contact
+      key_field: AccountId
+      sf_id_table: Account
```
davisagli: Why did these need to be added?

requirements.txt
```diff
@@ -5,9 +5,11 @@ chardet==3.0.4
 click==7.0
 coloredlogs==10.0
 cryptography==2.8
+dateutil==2.8.1
```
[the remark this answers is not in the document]
prescod: I briefly had a heart attack:

https://www.zdnet.com/article/two-malicious-python-libraries-removed-from-pypi/

But no, it turns out it is a harmless typo. The malicious one was called "python3-dateutil."

Still, a good reminder of the importance of being careful with package names!

cumulusci/tasks/bulkdata/data_generation/data_generator_runtime.py
```diff
@@ -0,0 +1,413 @@
...
+
+# Look in generate_from_yaml for entry points from CCI and command line.
+# Look in parse_factory_yaml for the "generate" entry point to the data generation subsystem
+
+
+class IdManager:
+    """Keep track ofthe most recent ID per Object type"""
+
+    def __init__(self):
+        self.last_used_ids = defaultdict(lambda: 0)
```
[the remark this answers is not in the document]
prescod: I ended up not needing either (so far). Sorry @cdcarter . As the language evolves I will probably need it in the future.

requirements.txt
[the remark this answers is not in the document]
prescod: 20a15814

requirements.txt
```diff
@@ -5,9 +5,11 @@ chardet==3.0.4
...
 docutils==0.15.2
 github3.py==1.3.0
 humanfriendly==4.18
+fastnumbers==2.2.1
```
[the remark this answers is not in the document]
prescod: It actually does. The system tries to cast every single string as a number. I identified my attempt as a bottleneck. Maybe there would have been a faster technique. I was using a try/catch block (actually one for floats and one for ints) instead of a regexp-and-then-convert-if-matches. If you feel strongly I could benchmark a regexp-based solution.

This will soon become a dependency of Snowfakery rather than CCI but until we have plugins it would also be a transitive dependency of CCI.

cumulusci/core/utils.py
```diff
@@ -84,6 +84,26 @@ def process_list_arg(arg):
...
+    """Process an arg in the format "aa:bb,cc:dd" """
+    if isinstance(arg, dict):
+        return arg
+    elif isinstance(arg, str):
+        rc = {}
+        for key_value in arg.split(","):
+            subparts = key_value.split(":")
+            if len(subparts) == 2:
+                key, value = subparts
+                if rc.get(key):
```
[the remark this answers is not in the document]
prescod: e97b668e

davisagli:
Closing this since we are putting this feature in its own repository (https://github.com/SalesforceFoundation/Snowfakery)

PR #1136, opened 2019-11-12
Allow bypassing scratch org recreation prompt via option and environment variable

jstvz:
## Original Issue
```console
$ ccfr dependencies --org fev
The scratch org is expired
Attempt to recreate the scratch org? [Y/n]:
```
Setting `always_recreate: True` should bypass the prompt and always recreate the scratch org.

---
This PR allows bypassing the confirmation prompt for expired org recreation by setting the `CUMULUSCI_NO_PROMPT` environment variable, or by passing the `--no-prompt` option to the following commands:

- task_run
- flow_run
- org_info
- org_browser

Also includes some refactoring of environment variables to constants in
`cumulusci.cli.config`.

- [x] Add YAML configuration option
- [ ] Add configuration docs

---
# Critical Changes

# Changes
- We've added support for bypassing confirmation prompts for expired scratch orgs via the `--no-prompt` option on some commands, or via setting the `CUMULUSCI_NO_PROMPT` environment variable globally.

# Issues Closed

### DIFF
--- a/cumulusci/cli/options.py
+++ b/cumulusci/cli/options.py
@@ -1,40 +1,28 @@
 """Process environment variables and global configurations for some options.
 
 Returns option values in the following order of precedence:
-1. Environment variables
-2. Global configuration options
-3. Click Option
+1. Global configuration options
+2. Click Option
+3. Environment variables
+4. Click Option default value
 """
 
-import os
 
 from cumulusci.core.config import BaseGlobalConfig
 
 # CLI Environment Variables
-NO_PROMPT_ENV = "CUMULUSCI_NO_PROMPT"
-PLAIN_OUTPUT_ENV = "CUMULUSCI_PLAIN_OUTPUT"
+CCI_ENV_PREFIX = "CCI__"
+CLI_ENV_PREFIX = CCI_ENV_PREFIX + "CLI__"
+NO_PROMPT_ENV = CLI_ENV_PREFIX + "ALWAYS_RECREATE"
+PLAIN_OUTPUT_ENV = CLI_ENV_PREFIX + "PLAIN_OUTPUT"
 
 
-def no_prompt_callback(ctx, param, value):
-    """Process the "No Prompt". output global setting."""
-    print(param)
+def global_option_lookup(ctx, param, value):
+    """Process Click Option and return global config setting, if it exists."""
+    global_attr = param.envvar.replace(CCI_ENV_PREFIX, "").lower()
     global_config = BaseGlobalConfig()
-    global_value = global_config.cli__no_prompt
-    return _select_value(NO_PROMPT_ENV, global_value, value)
-
-
-def plain_output_callback(ctx, param, value):
-    """Process the Plain Table output global setting."""
-    global_config = BaseGlobalConfig()
-    global_value = global_config.cli__plain_output
-    return _select_value(PLAIN_OUTPUT_ENV, global_value, value)
-
-
-def _select_value(env_var, global_value, value):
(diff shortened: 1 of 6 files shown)


jstvz:
Sorry for the abnormal PR, thought it'd just reference the issue when creating the PR.

jstvz:
@Br4nd0R Yeah, this was an issue I created in June. I created the PR in Emacs, but with a command I hadn't used yet and it deleted the issue and created a PR with the same number, as far as I can tell.

I'll just reference the issue next time.

jstvz:
That's what I get for rebasing at the end of the day. Not ready for re-review just yet.

jstvz:
The other question I had was about where to document this. I could include an early (skeletal, really) version of the planned configuration doc, or wherever else folks think is appropriate.

davisagli:
@jstvz I think options passed explicitly on the command line should have highest precedence (seems surprising for environment vars or the user's local config to take precedence over what the user typed for that particular invocation).

I like the use of a click option callback -- I didn't know about those. I think it could be genericized somewhat: we could make a subclass of click.Option where you pass in what the config path and env var are, and it takes care of checking the config and env without needing to write a separate callback for each option. We could also have a standard mapping from env vars to config paths (e.g. CCI__cli__no_prompt in the environment corresponds to cli__no_prompt in the yaml config) so that that doesn't need to be specified for each one.

I'm still not sure about calling this no_prompt instead of something more specific like always_recreate_expired_scratch_orgs

Re docs, I think we probably do need to start the section that discusses the different places CCI configuration can go.

cumulusci/cli/options.py
```diff
@@ -0,0 +1,28 @@
...
+"""
+
+
+from cumulusci.core.config import BaseGlobalConfig
+
+# CLI Environment Variables
+CCI_ENV_PREFIX = "CCI__"
+CLI_ENV_PREFIX = CCI_ENV_PREFIX + "CLI__"
+NO_PROMPT_ENV = CLI_ENV_PREFIX + "ALWAYS_RECREATE"
+PLAIN_OUTPUT_ENV = CLI_ENV_PREFIX + "PLAIN_OUTPUT"
```
davisagli: The config setting is cli__plain_output so I think this needs to be "CCI__CLI__PLAIN_OUTPUT"

cumulusci/cli/options.py
```diff
@@ -0,0 +1,28 @@
...
+4. Click Option default value
+"""
+
+
+from cumulusci.core.config import BaseGlobalConfig
+
+# CLI Environment Variables
+CCI_ENV_PREFIX = "CCI__"
+CLI_ENV_PREFIX = CCI_ENV_PREFIX + "CLI__"
+NO_PROMPT_ENV = CLI_ENV_PREFIX + "ALWAYS_RECREATE"
```
davisagli: Can we:
1. Make this more specific, i.e. "CCI__CLI__RECREATE_EXPIRED_SCRATCH_ORGS"
2. Update the name and help for the `--no-prompt` options to match? I'm okay that this is a backwards-incompatible change; I doubt the `--no-prompt` option is used much and I consider it a bug that the option does something different than what its name and help suggest.

jstvz:
@davisagli The original reason for this PR got fixed in #1685. Is the option/envvar handling worth keeping? (This was held up waiting for Click to release 7.1 and expose `get_parameter_source`, which we needed to distinguish options passed from the command line vs environment variables.)

davisagli:
@jstvz I still like the idea of providing this way to configure some CLI options, if we find time to bring it up to date and document.

PR #1335, opened 2019-11-12
retrieve_changes improvements

davisagli:
Improvements to the retrieve_changes task:
- When you specify the include or exclude option to filter the retrieved components, the task will now retrieve only those components rather than all components listed in package.xml in the target directory.
- Changes can now be retrieved into folders in DX source format. Namespace tokenization is not supported in DX format, since there isn't currently a way to deploy DX format source including namespace tokens.
- The target directory now defaults to src if the project is using mdapi format or the default entry in packageDirectories in sfdx-project.json if the project is using sfdx format.

Along the way I also added some options to the `sfdx` util, refactored some of the namespace injection utils to support operating on a directory instead of a zip file, and gave the temporary_dir util an option to skip changing the current directory.

Caveats:
- When retrieving in DX format, the task will update sfdx's revision number if and only if all changed components were retrieved. As a result, after retrieving a filtered subset of components, `sfdx force:source:status` may show conflicts.
- Retrieving into mdapi format is accomplished by temporarily converting to dx format, running force:source:retrieve, then converting back. I've noticed a few spurious changes that can happen as a result of the conversion:
  - Reordering of components and removing blank lines inside a multi-component file (such as a .object file)
  - Changing the format of entities e.g. `&apos;` to `&#39;`

# Critical Changes

# Changes

# Issues Closed


cumulusci/utils.py
```diff
@@ -181,18 +181,51 @@ def zip_subfolder(zip_src, path):
     return zip_dest
 
 
-def zip_inject_namespace(
-    zip_src,
+def process_text_in_directory(path, process_file):
```
Br4nd0R: Docstrings on these util functions would be helpful.

cumulusci/utils.py
```diff
@@ -181,18 +181,51 @@ def zip_subfolder(zip_src, path):
...
     managed=None,
     filename_token=None,
     namespace_token=None,
     namespaced_org=None,
     logger=None,
 ):
-    """ Replaces %%%NAMESPACE%%% for all files and ___NAMESPACE___ in all
-        filenames in the zip with the either '' if no namespace is provided
-        or 'namespace__' if provided.
+    """ Replaces %%%NAMESPACE%%% in file content and ___NAMESPACE___ in filename
```
Br4nd0R:
```suggestion
    """ Replaces %%%NAMESPACE%%% in file content and ___NAMESPACE___ in name
```
Did you mean 'name' instead of 'filename'? It's not apparent to me what 'filename' refers to here.

cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -45,33 +53,19 @@ def _load_snapshot(self):
...
-
     def _run_task(self):
         changes = self._get_changes()
-        if changes["totalSize"]:
+        if changes:
             self.logger.info(
-                "Found {} changed components in the scratch org.".format(
-                    changes["totalSize"]
-                )
+                "Found {} changed components in the scratch org.".format(len(changes))
```
Br4nd0R:
```suggestion
                f"Found {len(changes)} changed components in the scratch org."
```

cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -45,33 +53,19 @@ def _load_snapshot(self):
...
         else:
             self.logger.info("Found no changes.")
 
-        filtered = self._filter_changes(changes)
-        ignored = len(changes["records"]) - len(filtered)
+        filtered, ignored = self._filter_changes(changes)
         if ignored:
             self.logger.info(
-                "Ignored {} changed components in the scratch org.".format(ignored)
+                "Ignored {} changed components in the scratch org.".format(len(ignored))
```
Br4nd0R:
```suggestion
                f"Ignored {len(ignored)} changed components in the scratch org."
```

Br4nd0R:
@davisagli - curious about this:
* When you specify the include or exclude option to filter the retrieved components, the task will now retrieve only those components rather than all components listed in package.xml in the target directory.

I definitely see how this is applicable to the `include` options, but how does this apply to the exclude option? If I specify to exclude profiles, wouldn't we want to pickup everything in the package.xml minus what we want to exclude?

Other than that, only a few minor things I can comment on. Was my first time looking through `sourcetracking.py` as well, so a lot to take in, but well worth spending some time on. I think that `RetrieveChanges._run_task()` could definitely be refactored so that it gives a good overview of all the steps involved. I'd be happy to do that as a learning exercise on my own after this is all wrapped up.

cumulusci/utils.py
davisagli: The `name` arg contains the filename that `content` came from.

davisagli:
@Br4nd0R 

> If I specify to exclude profiles, wouldn't we want to pickup everything in the package.xml minus what we want to exclude?

The filters are applied to the list of components that have changed (you can see this list without retrieving anything using `cci task run list_changes`), not the existing contents of the package in the repo. It should have the same effect as what you said though, because if a component in the package didn't change then there's no need to retrieve it.

Br4nd0R:
@davisagli - That makes complete sense, I didn't realize it's only pulling the components that have changed.

Br4nd0R:
I also realize now that I should have just read the command name... `retrieve_changes` would imply that we're retrieving... changes. Need more ☕️

davisagli:
@jstvz this is working as designed (but maybe we need a better design). include and exclude both do a substring match on the string "{metadata type}: {metadata name}" ... so `-o include Profile` includes any component with Profile as part of the name or type. You should be able to retrieve the specific component you're aiming for using `-o include "Profile: Admin"`

jstvz:
Currently, retrieve<sub>changes</sub> offers two options to control what metadata is retrieved:

    • include: Include changed components matching this string.
    • exclude: Exclude changed components matching this string.

This has led to some confusion on the part of users ([including me](https://github.com/SFDO-Tooling/CumulusCI/pull/1335#pullrequestreview-318765762)) about what the expected behavior of these options is. I think we should update these tasks as follows:

1.  Change the help text for the existing options to include an example of how its matching works, e.g. "Example: `-o include Foo` matches `CustomField: Foo__c.Baz__c` and `CustomObject: Foo__c`"
2.  Add an option to filter based on metadata type. This option would accept a comma-separated list of metadata types (e.g. `Profile`) rather than a string. While this is possible with the existing option by matching against the record prefix (`-o include Profile:`), users often fail to include a colon and retrieve unwanted metadata.

davidmreed:
Paper-cuts: 

- Running `retrieve_changes` without `--org` or a default org set results in a traceback rather than a note indicating that an org is required.
- Not including the `namespace_tokenize` parameter results in a KeyError on line 218.

In some simple testing, the changes seem to work successfully on unpackaged MDAPI directories (not over-pulling).

davisagli:
@davidmreed thanks; both of those papercuts should be fixed now.

davisagli:
@Br4nd0R @jstvz @davidmreed I think I've addressed all the feedback and have this ready for final review.

PR #1338, opened 2019-11-13
Make it possible to switch a project to a non-default devhub

davisagli:
# Critical Changes

# Changes
* Added a `devhub` service. This can be used to switch a project to a non-default sfdx dev hub using `cci service connect devhub --project`

# Issues Closed


davisagli:
Now with docs.

PR #1339, opened 2019-11-14
Implementing handling of the OrgNotFound exception

robertbwatson:
# Critical Changes

# Changes
Implementing better handling of the OrgNotFound exception thrown when specifying an org that does not exist in the keychain. When a user specifies an org that does not exist in the keychain, they'll now see a simple sentence instead of a full stack trace.

This was implemented for the following commands:
- `cci org browser`
- `cci org default`
- `cci org info` (already implemented in a previous release of CumulusCI)
- `cci org remove` (already implemented in a previous release of CumulusCI)
- `cci org scratch_delete`
- `cci org shell`
- `cci task run`
- `cci flow run`

Note: While `cci org info` already had this handling, the error message was modified to match the same language as the rest of the commands.

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -764,7 +764,10 @@ def service_info(config, service_name, plain):
 @click.argument("org_name", required=False)
 @pass_config
 def org_browser(config, org_name):
-    org_name, org_config = config.get_org(org_name)
+    try:
+        org_name, org_config = config.get_org(org_name)
+    except OrgNotFound:
+        raise click.ClickException(f"Org {org_name} does not exist in the keychain")
     org_config.refresh_oauth_token(config.keychain)
 
     webbrowser.open(org_config.start_url)
@@ -833,7 +836,10 @@ def org_default(config, org_name, unset):
         config.keychain.unset_default_org()
         click.echo(f"{org_name} is no longer the default org.  No default org set.")
     else:
-        config.keychain.set_default_org(org_name)
+        try:
+            config.keychain.set_default_org(org_name)
+        except OrgNotFound:
+            raise click.ClickException(f"Org {org_name} does not exist in the keychain")
         click.echo(f"{org_name} is now the default org")
 
 
@@ -874,8 +880,8 @@ def org_info(config, org_name, print_json):
     try:
         org_name, org_config = config.get_org(org_name)
         org_config.refresh_oauth_token(config.keychain)
-    except OrgNotFound as e:
-        raise click.ClickException(e)
+    except OrgNotFound:
+        raise click.ClickException(f"Org {org_name} does not exist in the keychain")
 
     if print_json:
         click.echo(
@@ -1059,7 +1065,11 @@ def org_scratch(config, config_name, org_name, default, devhub, days, no_passwor
 @click.argument("org_name")
 @pass_config
 def org_scratch_delete(config, org_name):
-    org_config = config.keychain.get_org(org_name)
+    try:
+        org_config = config.keychain.get_org(org_name)
+    except OrgNotFound:
+        raise click.ClickException(f"Org {org_name} does not exist in the keychain")
+
     if not org_config.scratch:
         raise click.UsageError(f"Org {org_name} is not a scratch org")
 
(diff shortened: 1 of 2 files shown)


robertbwatson:
@jstvz Can you take a look at this when you get a chance? Not sure if there's any other process I should follow - just let me know if I need to do anything else :)

Note: I did not add the improved error handling to the `cci org import` and `cci org connect` tasks... I don't know them that well and their functionality is a bit different than the rest.

cumulusci/cli/cci.py
```diff
@@ -833,7 +836,10 @@ def org_default(config, org_name, unset):
         config.keychain.unset_default_org()
         click.echo(f"{org_name} is no longer the default org.  No default org set.")
```
robertbwatson: I discovered this is a funny one - you can specify *anything* for the org name, the task just unsets whatever the default org is regardless of what's entered as the org name. 🙃

cumulusci/cli/cci.py
```diff
@@ -764,7 +764,10 @@ def service_info(config, service_name, plain):
 @click.argument("org_name", required=False)
 @pass_config
 def org_browser(config, org_name):
-    org_name, org_config = config.get_org(org_name)
+    try:
+        org_name, org_config = config.get_org(org_name)
+    except OrgNotFound:
+        raise click.ClickException(f"Org {org_name} does not exist in the keychain")
```
robertbwatson: I copied this error message from the `cci org remove` task, which already had this exception handling in place.

cumulusci/cli/cci.py
```diff
@@ -764,7 +764,10 @@ def service_info(config, service_name, plain):
 @click.argument("org_name", required=False)
 @pass_config
 def org_browser(config, org_name):
-    org_name, org_config = config.get_org(org_name)
+    try:
```
jstvz: I'll defer to @davisagli on this, but I think it makes sense to apply _Extract Method_ to this try..catch to keep it DRY (Don't Repeat Yourself).

cumulusci/cli/cci.py
jstvz: 🤔 :
```suggestion
        click.echo(f"Unsetting default org...")
        click.echo(f"No default org set.")
```

davisagli:
@robertbwatson Thanks for working on this--it's something that multiple people have mentioned.

@jstvz The problem with centralizing this logic by moving it into the keychain's get_org method is that use of `click`'s exceptions should be isolated to the `cci` module. i.e. we don't want to be raising `click` exceptions when CumulusCI is used as a library within a web app rather than as a CLI.

I'd like to take a look at whether there's a way we can centralize this by moving in the other direction; that is, moving the exception handling out to a wrapper around the `click` entry point. We already have a way of designating exceptions that shouldn't show a traceback (by making them subclass CumulusCIUsageError) but that logic is currently only applied in a few of the `click` commands.

There are a few related questions about exception handling that have come up recently so I'll try to consider this within that broader context in the next few days.

jstvz:
I meant wrapping the try-except calls that are the same into a helper function in this module. Maybe it's an unnecessary abstraction.

davisagli:
@jstvz got it. that sounds like a good idea if my experiments don't work out re handling exceptions in a wrapper of the click entrypoint.

cumulusci/cli/cci.py
robertbwatson: @jstvz @davisagli (Sorry for the delay in getting back to this...)

It's hard to tell from the conversations if I should do anything here or not. One thing that I could see being beneficial is moving `raise click.ClickException(f"Org {org_name} does not exist in the keychain")` into its own method so that the error message only exists in a single place.

cumulusci/cli/cci.py
robertbwatson: @jstvz Should I do this now or should it happen in a different PR? Happy to commit the suggestion now if that's ok.

davisagli:
Hey @robertbwatson -- sorry for the radio silence on this. It got me thinking about some refactoring that I want to do to centralize how the CLI handles exceptions. I'm working on that and hope to have something to review next week. That will probably take the place of a lot of your work here, which I'm sorry about, but I would be happy to have you review what I'm working on to make sure it accomplishes similar goals.

robertbwatson:
> Hey @robertbwatson -- sorry for the radio silence on this. It got me thinking about some refactoring that I want to do to centralize how the CLI handles exceptions. I'm working on that and hope to have something to review next week. That will probably take the place of a lot of your work here, which I'm sorry about, but I would be happy to have you review what I'm working on to make sure it accomplishes similar goals.

No worries @davisagli, thanks for the update!

davisagli:
Closing this since #1395 will solve it a different way (by centralizing our error handling and never showing the stack trace unless requested)

PR #1340, opened 2019-11-14
Feature/brandon/more f string

Br4nd0R:
* More f-string swapovers


cumulusci/core/tests/test_keychain.py
```diff
@@ -424,8 +424,9 @@ def _mk_temp_project(self):
 
     def _create_git_config(self):
         filename = os.path.join(self.tempdir_project, ".git", "config")
-        content = '[remote "origin"]\n' + "  url = git@github.com:TestOwner/{}".format(
-            self.project_name
+        content = (
+            f'[remote "origin"]\n'
+            + "  url = git@github.com:TestOwner/{self.project_name}"
```
davisagli:
```suggestion
            + f"  url = git@github.com:TestOwner/{self.project_name}"
```

PR #1341, opened 2019-11-14
Update bulk data task defaults

davisagli:
# Critical Changes
- We changed the default path for the mapping file created by the `generate_dataset_mapping` task to `datasets/mapping.yml` so that it matches the defaults for `extract_dataset` and `load_dataset`
- We changed the `extract_dataset` and `load_dataset` tasks to default to storing data in an SQL file, `datasets/sample.sql`, instead of a binary SQLite database file.

# Changes
- We added a predefined `qa` scratch org. It uses the same scratch org definition file as the `dev` org, but makes it easier to spin up a second org for QA purposes without needing to first create it using `cci org scratch`.
- The `database_url` option for the `extract_dataset` and `load_dataset` tasks is no longer required. Either `database_url` or `sql_path` must be specified. If both are specified, the `sql_path` will be ignored.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata/extract.py
+++ b/cumulusci/tasks/bulkdata/extract.py
@@ -1,4 +1,3 @@
-import os
 import tempfile
 import unicodecsv
 
@@ -29,8 +28,7 @@ from salesforce_bulk.util import IteratorBytesIO
 class ExtractData(BulkJobTaskMixin, BaseSalesforceApiTask):
     task_options = {
         "database_url": {
-            "description": "A DATABASE_URL where the query output should be written",
-            "required": True,
+            "description": "A DATABASE_URL where the query output should be written"
         },
         "mapping": {
             "description": "The path to a yaml file containing mappings of the database fields to Salesforce object fields",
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/bulkdata/extract.py
```diff
@@ -268,8 +266,6 @@ def _drop_sf_id_columns(self):
 
     def _sqlite_dump(self):
         path = self.options["sql_path"]
-        if os.path.exists(path):
-            os.remove(path)
```
davisagli: Removed this because `open` in "w" mode will overwrite the file anyway.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -53,17 +51,13 @@ def _init_options(self, kwargs):
...
         if self.options.get("sql_path"):
-            if self.options.get("database_url"):
-                raise TaskOptionsError(
-                    "The database_url option is set dynamically with the sql_path option.  Please unset the database_url option."
-                )
             self.options["sql_path"] = os_friendly_path(self.options["sql_path"])
-            if not os.path.isfile(self.options["sql_path"]):
-                raise TaskOptionsError(
-                    f"File {self.options['sql_path']} does not exist"
-                )
```
davisagli: Removed this because `open` in "r" mode will raise a very similar exception if the file doesn't exist.

jstvz:
I know this is a draft, so maybe you just haven't gotten to it yet, but gentle reminder that `bulk_data.rst` will also need an update.

davisagli:
@jstvz I don't see anything in bulk_data.rst that needs to be updated. It describes what task options are available but not what their defaults are.

jstvz:
Here's what I'm seeing on [line 317](https://github.com/sfdo-tooling/CumulusCI/blob/feature%2Fetl-task-defaults/docs/bulk_data.rst#L317):

```rst
Options
+++++++

* ``path``: Location to write the mapping file. Default: datasets/generated_mapping.yml
```

davisagli:
@jstvz thanks, I missed that

jstvz:
(It'd be nice if we could add suggested changes for any file...)

PR #1357, opened 2019-11-22
Changes to make coverage checking less strict.

prescod:
Don't include abstract methods (running them is an error in itself)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -2,3 +2,8 @@
 omit = */tests/*
     cumulusci/files/*
 source=cumulusci
+[report]
+exclude_lines =
+    pragma: no cover
+    @abstract
+    def __repr__
--- a/.coveragerc
+++ b/.coveragerc
@@ -6,4 +6,3 @@ source=cumulusci
 exclude_lines =
     pragma: no cover
     @abstract
-    def __repr__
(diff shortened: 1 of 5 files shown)


.coveragerc
```diff
@@ -2,3 +2,8 @@
 omit = */tests/*
     cumulusci/files/*
 source=cumulusci
+[report]
+exclude_lines =
+    pragma: no cover
+    @abstract
+    def __repr__
```
davisagli: Ignoring abstract methods makes sense -- but isn't the decorator `@abstractmethod` rather than `@abstract`?

I disagree about ignoring `__repr__`. It's a common place for there to be lingering typos _because_ it's not run in the normal course of operation, and the way to avoid that is to make sure it's covered by tests. Otherwise mistakes won't be discovered until you're actually debugging, which causes friction.

prescod: I think it is interpreted as a substring and thus supports both abstractmethods and abstractproperties.

.coveragerc
```diff
@@ -2,3 +2,7 @@
 omit = */tests/*
     cumulusci/files/*
 source=cumulusci
+[report]
+exclude_lines =
+    pragma: no cover
```
davisagli: We currently use "pragma: no cover" in 1 place and "pragma: nocover" in 5 places. Can you either make this exclude both of them, or update them to be consistent?

PR #1358, opened 2019-11-22
Refactor CliTable to support logger

jstvz:
This PR updates `CliTable` to support the use of loggers where needed by adding a `table` property that returns a properly formatted string.

Also, updated `PackageUpload` to use `self.logger` for reporting apex test failures.

---
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cli/ui.py
+++ b/cumulusci/cli/ui.py
@@ -89,27 +89,26 @@ class CliTable:
         Automatically falls back to AsciiTable if there's an encoding error.
         """
         if plain or os.environ.get("TERM") == "dumb":
-            self.ascii_table()
-            return None
+            table = self.ascii_table()
+        else:
+            try:
+                table = self.pretty_table()
+            except UnicodeEncodeError:
+                table = self.ascii_table()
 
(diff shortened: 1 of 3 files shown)


cumulusci/cli/ui.py
```diff
@@ -89,27 +89,29 @@ def echo(self, plain=False):
...
         if plain or os.environ.get("TERM") == "dumb":
-            self.ascii_table()
-            return None
+            table = self.ascii_table()
+        else:
+            table = self.table
+        click.echo(table)
 
+    @property
+    def table(self):
```
davisagli: What about `__str__`?

jstvz: Huh, forgot about `__str__`. Instead of this, or in addition to it?

davisagli: Instead of. I think `str(table)` is a clearer API than `table.table`

jstvz: Well, when you write it out like that... :smirk:

I'll update.

PR #1359, opened 2019-11-23
Add opt-in defaults, templates, type-checking

prescod:
A prototype of a few different Task option declarative features:

1. type checking and coercion using Click's checkers.

2. value defaulting.

3. very simple templating language based on format-strings

The basic principle is that all features are opt-in, so there SHOULD be generally no change for tasks that do not opt-in. (Brian's robot linter did actually have a latent bug in it which was exposed by my code: he had specified incorrect defaults even though there was no default-handling mechanism that I am aware of. My code turned his incorrect, ignored defaults into real defaults and tests failed.)

### DIFF
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -93,11 +93,9 @@ class BaseTask(object):
         # Handle default values
         for name, config in list(self.task_options.items()):
             if name not in self.options and "default" in config:
-                print("XXX1", name)
                 self.options[name] = config["default"]
 
             if "do_replacement" in config:
-                print("XXX2", name)
                 self.options[name] = self.options[name].format(
                     **{"project_config": self.project_config}
                 )
@@ -118,7 +116,6 @@ class BaseTask(object):
                 missing_required.append(name)
 
             elif config.get("type") and name in self.options:
-                print("XXX3", name)
                 datatype = config.get("type")
                 try:
                     self.options[name] = datatype.convert(
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -8,6 +8,8 @@ import os
 import time
 import threading
 
+import click
+
 from cumulusci.utils import cd
 from cumulusci.core.exceptions import TaskRequiresSalesforceOrg
 from cumulusci.core.exceptions import TaskOptionsError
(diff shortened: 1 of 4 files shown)


cumulusci/core/tasks.py
```diff
@@ -103,6 +115,15 @@ def _validate_options(self):
...
                 missing_required.append(name)
 
+            elif config.get("type") and name in self.options:
+                datatype = config.get("type")
+                try:
+                    self.options[name] = datatype.convert(
+                        self.options[name], name, None
+                    )
+                except click.exceptions.UsageError as e:
+                    raise TaskOptionsError(str(e)) from e
```
davisagli: Using click's types for this is an interesting idea. I need to think through whether that covers everything we'll want to do, plays nice with options both from the CLI and from YAML, and is compatible with the possibility of starting to use Python type checking. Can we start a quip doc to reach consensus on what the design goals are?

cumulusci/core/tasks.py
```diff
@@ -88,6 +90,16 @@ def _init_options(self, kwargs):
         if kwargs:
             self.options.update(kwargs)
 
+        # Handle default values
+        for name, config in list(self.task_options.items()):
```
davisagli: No need for `list()`

cumulusci/core/tasks.py
```diff
@@ -88,6 +90,16 @@ def _init_options(self, kwargs):
...
+            if name not in self.options and "default" in config:
+                self.options[name] = config["default"]
+
+            if "do_replacement" in config:
+                self.options[name] = self.options[name].format(
+                    **{"project_config": self.project_config}
+                )
```
davisagli: In the long run I don't want to have 2 different ways to do variable interpolation into option values (this one, and the one just below).

This also assumes every option name is in self.options and every value in self.options is a string, both of which are invalid.

cumulusci/core/tasks.py
```diff
@@ -88,6 +90,16 @@ def _init_options(self, kwargs):
         if kwargs:
             self.options.update(kwargs)
 
+        # Handle default values
+        for name, config in list(self.task_options.items()):
+            if name not in self.options and "default" in config:
+                self.options[name] = config["default"]
```
davisagli: Should probably do a deepcopy here. In case there are mutable defaults and multiple instances of the same task class in a flow.

cumulusci/core/tasks.py
prescod: The old way of doing templates seems pretty limiting so it should be removed someday, but that's a backwards incompatible change so it will take some time.

cumulusci/core/tasks.py
prescod: https://salesforce.quip.com/oHIAALXWwR0P

davisagli:
Closing for now. We intend to revisit this but there are some additional considerations based on our conversation today.

PR #1376, opened 2019-12-02
Fix mypy errors

prescod:
Created PR to to have a conversation context.

1. How does the pre-commit hook know to run with --ignore-missing-imports ?

2. In Quip you said that you had to declare some args as "Any" but I don't see that in the commit. What am I missing?

3. What is it that mypy doesn't like about 

        class Unserializable(object):	

4. Why didn't you make this into a PR?

5. What do the renamings have to do with mypy?

6. Were the specific annotations you added mandatory? Or did you just want to experiment with it?


davisagli:
Withdrawing this for now.

PR #1387, opened 2019-12-09
Feature/activate process builders process

jofsky:
Adding for David to look at pathing issue, unsure why task is not being identified. To the best of my knowledge I had checked in working code. 
# Critical Changes

# Changes

# Issues Closed


cumulusci/cumulusci.yml
```diff
@@ -1,7 +1,14 @@
 cumulusci:
     keychain: cumulusci.core.keychain.EncryptedFileProjectKeychain
-
+    
 tasks:
+    activate_process_builder:
+        group: Salesforce Bulk API
```
davidmreed: This is probably not the right group. Let's see if there's an existing group that fits, but I wonder if we need to start a new "Metadata Transformations" group.

cumulusci/cumulusci.yml
```diff
@@ -1,7 +1,14 @@
...
+        description: "Takes a selected list of developer names and runs activating flows against these orgs"
```
davidmreed: `
```suggestion
        description: "Activates Flows identified by a given list of Developer Names"
```

cumulusci/cumulusci.yml
```diff
@@ -1,7 +1,14 @@
...
+        class_path: cumulusci.tasks.salesforce.ActivateProcessBuilderProcesses.ActivateProcessBuilderProcesses
+        options:
+            developer-names: 
+              - ServiceDelivery
```
davidmreed: I don't think we want to ship any options here - that could expose side effects in user orgs.

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,36 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+
+class ActivateProcessBuilderProcesses(BaseSalesforceApiTask):
+    """
+    Create a Salesforce Activation Process Builder which takes in a list as options and
+    runs flows only on those listed DeveloperName's.
```
davidmreed: +1 for docstring, I'm just going to propose an edit for clarity.

"""Activate the Flows with the supplied Developer Names"""

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,36 @@
...
+                            "VersionNumber"
+                        ]
+                    }
+                }
+
+                response = self.tooling._call_salesforce("PATCH", urlpath, json=data)
+                self.logger.info(response)
+            else:
+                self.logger.info(
+                    f'Skipping DeveloperName: {listed_flow["DeveloperName"]} not in given developer-names list.'
```
davidmreed: I think it's safe to omit this branch. Some orgs may have a _lot_ of Processes and versions, so this could put a lot of chatter into the log.

cumulusci/cumulusci.yml
```diff
@@ -1,7 +1,14 @@
 cumulusci:
     keychain: cumulusci.core.keychain.EncryptedFileProjectKeychain
-
+    
 tasks:
+    activate_process_builder:
```
davidmreed: Sorry for _not_ getting this suggestion in earlier, but I'm going to propose we standardize on the term "flow" rather than "process builder". This is for two reasons:

1. The underlying metadata is called `FlowDefinition` for the entities that are called Process and Flow in the UI.
2. It seems to be more widely accepted that way.
3. It pluralizes more cleanly.

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,36 @@
...
+
+class ActivateProcessBuilderProcesses(BaseSalesforceApiTask):
+    """
+    Create a Salesforce Activation Process Builder which takes in a list as options and
+    runs flows only on those listed DeveloperName's.
+    """
+
+    def _run_task(self):
+        self.logger.info(
+            f'Allowing only rows with the following DeveloperNames {self.options.get("developer-names")}'
```
davidmreed: I'd rephrase this log to express what it's doing in Salesforce-y terms. Maybe something like "Activating the following Flows:".

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,36 @@
...
+        )
+        self.logger.info("Querying flow definitions...")
+        res = self.tooling.query(
+            "SELECT Id,ActiveVersion.VersionNumber,LatestVersion.VersionNumber,DeveloperName FROM FlowDefinition"
```
davidmreed: Nit-pick: Let's add spaces between the fields in the SELECT.

Did you experiment at all with adding a `WHERE` clause here? The Tooling API is tetchy about how you can filter on certain fields, but if it will allow a filter on `DeveloperName` (and I think it will), we could add something like

`WHERE DeveloperName IN ({0})"`.format(",".join([f"'{n}'" for n in self.options.get("developer_names")]))`

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,36 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+
+class ActivateProcessBuilderProcesses(BaseSalesforceApiTask):
+    """
+    Create a Salesforce Activation Process Builder which takes in a list as options and
+    runs flows only on those listed DeveloperName's.
+    """
+
```
davidmreed: Let's put a `task_options` in here so that we can show the options at the command line. (I'd suggest calling the option `developer_names` with an underscore for consistency).

Then, add an `_init_options()` method and you can use `process_list_arg()` from `cumulusci.core.utils` to turn the user input into a real Python list. There's an example in `testrunner.py`.

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,36 @@
...
+            f'Allowing only rows with the following DeveloperNames {self.options.get("developer-names")}'
+        )
+        self.logger.info("Querying flow definitions...")
+        res = self.tooling.query(
+            "SELECT Id,ActiveVersion.VersionNumber,LatestVersion.VersionNumber,DeveloperName FROM FlowDefinition"
+        )
+        for listed_flow in res["records"]:
+            if listed_flow["DeveloperName"] in self.options.get("developer-names"):
+                self.logger.info(f'Processing: {listed_flow["DeveloperName"]}')
+                path = "tooling/sobjects/FlowDefinition/{0}".format(listed_flow["Id"])
```
davidmreed: fstring?

cumulusci/tasks/salesforce/tests/test_ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,19 @@
+from unittest import mock
+import unittest
+
+from cumulusci.core.config import BaseGlobalConfig
+from cumulusci.core.config import BaseProjectConfig
+from cumulusci.tasks.salesforce import CreatePackage
+from .util import create_task
+
+
+class TestCreatePackage(unittest.TestCase):
```
davidmreed: This one doesn't go here. :)

cumulusci/cumulusci.yml
jofsky: Ok should I pass in as parameter?

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
jofsky: Implemented

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
jofsky: Im not following do you mean continue in the else statement?

cumulusci/cumulusci.yml
davidmreed: We just don't want the configuration to be in the global `cumulusci.yml`. We'd want to have no options declared here, and then we'd declare them in, e.g., PMDM's `cumulusci.yml` (since the options are going to be project-specific).

For testing, you can supply it on the command line, e.g., `-o developer_names ServiceDelivery`.

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
davidmreed: No, I just mean remove the `else:` branch entirely here - I just don't think we need to do this log.

cumulusci/cumulusci.yml
jofsky: ah makes sense to me

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
jofsky: I Like that!

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
jofsky: Yeah that looks right I can do this since it will probably make it faster.

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
jofsky: Am I just passing in **kwargs into process_list_arg I' ve gotten the changes to work but only when I hard code the values, I'm going to push and submit to show.

cumulusci/cumulusci.yml
```diff
@@ -1,7 +1,11 @@
 cumulusci:
     keychain: cumulusci.core.keychain.EncryptedFileProjectKeychain
-
+
```
davisagli: Please don't add random whitespace.

docs/tasks.rst
```diff
@@ -2,6 +2,14 @@
 Tasks Reference
 ==========================================
 
+activate_process_builder
+==========================================
+
+**Description:** Takes a selected list of developer names and runs activating flows against these orgs
```
davisagli: Hmm, this doesn't match what I see above. `docs/tasks.rst` should be generated from the actual code by running `cci task doc > docs/tasks.rst`, not edited by hand.

cumulusci/tasks/salesforce/ActivateFlowProcesses.py
```diff
@@ -0,0 +1,53 @@
...
+        )
+
+    def _run_task(self):
+        self.logger.info(
+            f"Activating the following Flows: {self.options['developer_names']}"
+        )
+        self.logger.info("Querying flow definitions...")
+        res = self.tooling.query(
+            "SELECT Id, ActiveVersion.VersionNumber, LatestVersion.VersionNumber, DeveloperName FROM FlowDefinition WHERE DeveloperName IN ({0})".format(
+                ",".join([f"'{n}'" for n in self.options.get("developer_names")])
```
davisagli:
```suggestion
                ",".join([f"'{n}'" for n in self.options["developer_names"]])
```

_init_options made sure it exists, so now we can assume it does. Also, `.get()` returns None if not found, which would break since it's not iterable.

cumulusci/tasks/salesforce/ActivateFlowProcesses.py
```diff
@@ -0,0 +1,53 @@
...
+            )
+        )
+
+        for listed_flow in res["records"]:
+            if listed_flow["DeveloperName"] in self.options["developer_names"]:
```
davisagli: Why would it not be, given the WHERE clause in the SOQL query?

cumulusci/tasks/salesforce/tests/test_ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,20 @@
...
+class TestActivateFlowProcesses(unittest.TestCase):
+    @mock.patch("cumulusci.tasks.salesforce.ActivateFlowProcesses")
+    def test_activate_flow_processes(self, ActivateFlowProcesses):
+        project_config = BaseProjectConfig(
+            BaseGlobalConfig(),
+            {"project": {"package": {"name": "TestPackage", "api_version": "43.0"}}},
+        )
+        task = create_task(ActivateFlowProcesses, project_config=project_config)
+        task._get_ActivateFlowProcesses()
+        ActivateFlowProcesses.assert_called()
```
davisagli: Looks like you're still working on the test?

cumulusci/tasks/salesforce/ActivateProcessBuilderProcesses.py
davisagli: @jofsky The usual pattern is that `_init_options` should start by calling `super()._init_options(kwargs)` which will do some default processing including initializing `self.options` from `kwargs`. Then your code can make adjustments to `self.options`.

cumulusci/cumulusci.yml
```diff
@@ -8,8 +8,7 @@ tasks:
         class_path: cumulusci.tasks.salesforce.ActivateFlowProcesses.ActivateFlowProcesses
         options:
             developer_names:
-                - Auto_Populate_Date_And_Name_On_Program_Engagement
-                - ape
+                -
```
davisagli: Just leave this option out of cumulusci.yml; it doesn't make sense to set it here unless there really is a default that makes sense for all projects.

cumulusci/tasks/salesforce/ActivateFlowProcesses.py
```diff
@@ -12,28 +13,34 @@ def _init_options(self, kwargs):
         self.options["developer_names"] = process_list_arg(
             self.task_config.options["developer_names"]
         )
```
davisagli: You should be setting a task_options dict in this class to define `developer_names` as a required option for the task. This is used to display help for the options when you run `cci task info activate_flow` and to build the autogenerated reference docs for tasks when you run `cci task doc`

cumulusci/cumulusci.yml
```diff
@@ -1,7 +1,14 @@
 cumulusci:
     keychain: cumulusci.core.keychain.EncryptedFileProjectKeychain
-
+    
 tasks:
+    activate_flow:
+        group: Metadata Transformations
+        description: Activates Flows identified by a given list of Developer Names
+        class_path: cumulusci.tasks.salesforce.ActivateFlowProcesses.ActivateFlowProcesses
```
davisagli: Python modules should be named with snake_case and should not have the same name as a class they contain. (There are some counterexamples to this in the existing codebase, but we want to fix them and definitely don't want to add more.)

Please rename the module to activate_flow.py and the class to ActivateFlow

cumulusci/tasks/salesforce/ActivateFlowProcesses.py
```diff
@@ -0,0 +1,46 @@
...
+    Activate the Flows with the supplied Developer Names
+    """
+
+    def _init_options(self, kwargs):
+        self.options = {}
+        self.options["developer_names"] = process_list_arg(
+            self.task_config.options["developer_names"]
+        )
+
+    api_version = "43.0"
```
davisagli: It makes sense to specify a particular API version that we know has the FlowDefinition object, but why such an old one?

cumulusci/tasks/salesforce/ActivateFlowProcesses.py
```diff
@@ -0,0 +1,46 @@
...
+                            "VersionNumber"
+                        ]
+                    }
+                }
+                response = self.tooling._call_salesforce("PATCH", urlpath, json=data)
+                self.logger.info(response)
+        else:
+            click.echo(
+                "Error you are missing developer_names definition in your task cumulusci.yml file. Please pass in developer_names for your task configuration or use -o to developer_names as a commandline arg"
+            )
```
davisagli: Tasks should never call `click.echo` -- it makes sense in the CLI context, but not when a task is running as part of a flow in one of our web apps.

Also validation of options should happen in `_init_options` so that problems are caught before the flow starts running. Check there to make sure there is at least one developer_name, and if not, raise a TaskOptionsError

cumulusci/tasks/salesforce/tests/test_ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,94 @@
+import unittest
```
davisagli: This test module should be renamed to match the name of the module it's testing.

cumulusci/tasks/salesforce/tests/test_ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,94 @@
...
+import json
+import responses
+from .util import create_task
+from cumulusci.tasks.salesforce.ActivateFlowProcesses import ActivateFlowProcesses
+
+task_options = {
+    "description": "Activates Flows identified by a given list of Developer Names",
+    "developer_names": ["Auto_Populate_Date_And_Name_On_Program_Engagement", "ape"],
+    "required": True,
+}
```
davisagli: description and required belong in the task_options dict on the task class itself which defines what options the task accepts. Not here where you are preparing specific _values_ of those options to use with an instance of the task.

cumulusci/tasks/salesforce/tests/test_ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,94 @@
...
+            status=200,
+        )
+        data = {"Metadata": {"activeVersionNumber": 1}}
+        responses.add(method=responses.PATCH, url=activate_url, status=204, json=data)
+
+        cc_task()
+        self.assertEqual(2, len(responses.calls))
+
+
+class TestActivateMultipleFlowProcesses(unittest.TestCase):
```
davisagli: You don't need a separate class for each test case.

cumulusci/tasks/salesforce/tests/test_ActivateProcessBuilderProcesses.py
```diff
@@ -0,0 +1,94 @@
...
+        self.assertEqual(3, len(responses.calls))
+
+
+class TestNoDeveloperFlowProcesses(unittest.TestCase):
+    @responses.activate
+    def test_activate_flow_processes(self):
+        result = mock.Mock()
+        final = result.create_task(ActivateFlowProcesses, task_no_developer_options)
+        self.assertEqual(0, len(responses.calls))
+        final.assert_not_called()
```
davisagli: This test doesn't make sense to me. You didn't even construct the actual test; you just called the create_task method of a mock object, which of course doesn't do anything.

cumulusci/tasks/salesforce/ActivateFlowProcesses.py
jofsky: So I think @davidmreed  had mentioned that there was a specific reason we were testing with version 43.0 but I could be making this up was a little bit ago.

cumulusci/cumulusci.yml
jofsky: @davisagli  I wanted to chat with you about the project scopings and where to put the options and when to document/how to handle if its project specific.

davisagli: the flow names for this task will be project-specific, so they should go in the project's cumulusci.yml like this:

    tasks:
        activate_flow:
            options:
                developer_names:
                    - Auto_Populate_Date_And_Name_On_Program_Engagement

This gets merged with the global config from CumulusCI's cumulusci.yml, so it'll inherit the task description and class_path

cumulusci/cumulusci.yml
jofsky: Im leaving as is since theres a space in master?

davidmreed: @jofsky David's pointing at the extra four spaces on the blank line - you can see it show up in the diff for the PR (https://github.com/SFDO-Tooling/CumulusCI/pull/1387/files) or in `git diff`.

davisagli: There isn't in master; that's why it shows up in the diff. Master has a blank line with no spaces, but you added 4 spaces.

cumulusci/tasks/salesforce/activate_flow.py
```diff
@@ -0,0 +1,53 @@
...
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super(ActivateFlow, self)._init_options(kwargs)
+        self.options["developer_names"] = process_list_arg(
+            self.options.get("developer_names")
+        )
+        self.api_version = "43.0"
+        if not self.options["developer_names"]:
```
davidmreed: This will never be `False` because it's initialized in lines 20-21 (and the option is required). Tip: you can recognize unneeded code like this because you will never be able to cover it in a unit test.

cumulusci/tasks/salesforce/activate_flow.py
```diff
@@ -0,0 +1,53 @@
...
+            urlpath = self.sf.base_url + path
+            data = {
+                "Metadata": {
+                    "activeVersionNumber": listed_flow["LatestVersion"]["VersionNumber"]
+                }
+            }
+            response = self.tooling._call_salesforce("PATCH", urlpath, json=data)
+            self.logger.info(response)
+
+        self.logger.info(
```
davidmreed: This doesn't appear to belong here.

cumulusci/tasks/salesforce/tests/test_activate_flow.py
```diff
@@ -0,0 +1,88 @@
+import unittest as test_activate_flow
```
davidmreed: What's the purpose for aliasing `unittest`? This is generally not something you need to do.

jofsky: Agreed I thought I had read David Glick had asked for it didn't really make sense to me but didn't really impact my life so i did it.

davisagli: This is definitely not what I was asking for. I was asking you to rename the file; I put my comment on this line because I needed to put it somewhere.

cumulusci/tasks/salesforce/tests/test_activate_flow.py
```diff
@@ -0,0 +1,88 @@
...
+
+        cc_task()
+        self.assertEqual(2, len(responses.calls))
+
+    @responses.activate
+    def test_both_activate_flow_processes(self):
+        cc_task = create_task(
+            ActivateFlow,
+            {
+                "description": "Activates Flows identified by a given list of Developer Names",
```
davidmreed: You don't need to provide `description` here.

cumulusci/tasks/salesforce/activate_flow.py
jofsky: I think it is needed other wise I get `File "/Users/jkofsky/Documents/python_3.7/CumulusCI/cumulusci/core/tasks.py", line 140, in __call__
    self.result = self._run_task()
  File "/Users/jkofsky/Documents/python_3.7/CumulusCI/cumulusci/tasks/salesforce/activate_flow.py", line 36, in _run_task
    ",".join([f"'{n}'" for n in self.options["developer_names"]])
TypeError: 'NoneType' object is not iterable` but if this is acceptable I'm ok with that, this way handled in a more user friendly way in my opinion.

cumulusci/tasks/salesforce/tests/test_activate_flow.py
```diff
@@ -0,0 +1,88 @@
...
+            body=json.dumps(
+                {
+                    "records": [
+                        {
+                            "Id": record_id,
+                            "DeveloperName": "Auto_Populate_Date_And_Name_On_Program_Engagement",
+                            "LatestVersion": {"VersionNumber": 1},
+                        },
+                        {
+                            "Id": record_id,
```
davidmreed: A real query result could not include two entries with the same record Id. This test works, but the setup differs from reality in a way that _might_ have impact (although I'm fairly sure it does not here).

I'd suggest adding a second record Id and URL. That will allow you to verify that exactly one activate call is made per record returned.

cumulusci/tasks/salesforce/tests/test_activate_flow.py
```diff
@@ -0,0 +1,88 @@
+import unittest as test_activate_flow
+import json
+import responses
+from .util import create_task
+from cumulusci.tasks.salesforce.activate_flow import ActivateFlow
+
+
+class TestActivateFlow(test_activate_flow.TestCase):
+    @responses.activate
+    def test_half_activate_flow_processes(self):
```
davidmreed: Does "half" here mean we're looking at the case where half (1 of 2) of the supplied Developer Names is actually found in the org?

The test looks good, but it raises an important question - should this situation be an error? That is, should we either (a) warn the user or (b) fail the build if a Flow we're asked to activate does not exist? 

At minimum, let's add a `self.logger.warn()` call in the main task. If the count of returned records isn't equal to the count of developer names, we can warn the user that we expected (X, Y, Z) and only got back (X, Y).

cumulusci/tasks/salesforce/activate_flow.py
davidmreed: Sorry, I need to fully engage my brain before reviewing code. It shouldn't be `None`, but it may be _empty_ (`[]`) after `process_list_arg()`, which you'd still want to warn the user about.

cumulusci/tasks/salesforce/tests/test_activate_flow.py
```diff
@@ -0,0 +1,91 @@
...
+    @responses.activate
+    def test_both_activate_flow_processes(self):
+        cc_task = create_task(
+            ActivateFlow,
+            {
+                "developer_names": [
+                    "Auto_Populate_Date_And_Name_On_Program_Engagement",
+                    "ape",
+                ],
+                "required": True,
```
davisagli: This doesn't belong here (the task doesn't have an option named "required")

cumulusci/tasks/salesforce/activate_flow.py
```diff
@@ -0,0 +1,59 @@
...
+    def _init_options(self, kwargs):
+        super(ActivateFlow, self)._init_options(kwargs)
+        self.options["developer_names"] = process_list_arg(
+            self.options.get("developer_names")
+        )
+        self.api_version = "43.0"
+        if not self.options["developer_names"]:
+            raise TaskOptionsError(
+                "Error you are missing developer_names definition in your task cumulusci.yml file. Please pass in developer_names for your task configuration or use -o to developer_names as a commandline argument"
+            )
```
davisagli: This line is missing test coverage. It should be pretty easy to add: just a test that tries to create the task with developer_names set to an empty list, and asserts that that raises a TaskOptionsError.

PR #1392, opened 2019-12-12
Proposal - Sample Task Doc Change

Br4nd0R:
# Changes
* Suggestion for how we may want to provide documentation for individual tasks. Largely based off of current sfdx command documentation format.
* These would eventually be grouped in logical categories based on task trying to be accomplished.
* Would prefer to use [line block notation](http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#line-blocks) for creating line breaks instead of the `raw-html` role. Couldn't get it to work as I wanted though :(


davisagli:
I like it. Looks like it should be possible to update `cci task doc` to generate this format.

Br4nd0R:
Just a proof of concept. Closing out.

PR #1394, opened 2019-12-12
bulkdata/extract.py: stabilize autogenerated PK-IDs; support mappings without "fields"

spelak-salesforce:
# Critical Changes

# Changes
`cumulusci.tasks.bulkdata.ExtractData`
- Added `ORDER BY Id` to extraction query to try to stabilize the order records are extracted
- Supports extracting bulk data if a mapping doesn't have any `fields` (e.g. only extracting autogenerated primary key).

# Issues Closed


spelak-salesforce:
If Release Engineer supports this, do any tests need to be updated? e.g.
- Test `ExtractData` with a mapping that doesn't have `fields`?

davidmreed:
@spelak-salesforce  Can you tell me about the use case for extracting an object that has no fields?

davisagli:
@spelak-salesforce I'd also like to hear more about the context for these changes before I review them.

spelak-salesforce:
> If Release Engineer supports this, do any tests need to be updated? e.g.
> 
> * Test `ExtractData` with a mapping that doesn't have `fields`?

@davidmreed @davisagli 

I wanted to support only capturing auto-generated, ordered Primary Keys (`oid_pk`).   To do this, the mapping would have no `fields`.   

I attempted to support the above in this Pull Request by supporting `map_step.get("fields")` to be not defined.   Previous an error was raised if `map_step.get("fields")` was not defined.   However, in practice, if `fields` is not defined, no records are extracted, so it doesn't work as intended.   Furthermore, I am finding out it would be bad practice to only capture `oid_pk` if it were supported since there would be no way to verify the corresponding record is what it is thought to be.

The main concern of this Pull Request is to `ORDER BY Id` when extracting.   Supporting extracting only `oid_pk` is not necessary.    Could we instead raise an error if `map_step.get("fields")` is not defined instead of getting an equivalent to a null pointer?

davidmreed:
I think the stability we get from the `ORDER BY` clause will be confined to mutation of small data sets. I'm fairly sure that (at least) record deletion and data volume > 10k records (or one Bulk API batch, whichever is smaller) will break the stability of ordering on record Ids. That's not a deal-breaker, but I also don't want us to make an explicit promise that won't be true.

I don't _think_ `ORDER BY Id` has substantive performance impact, but I also don't want to break anything for @prescod. Paul, do you have any thoughts?

spelak-salesforce:
> I think the stability we get from the `ORDER BY` clause will be confined to mutation of small data sets. I'm fairly sure that (at least) record deletion and data volume > 10k records (or one Bulk API batch, whichever is smaller) will break the stability of ordering on record Ids. That's not a deal-breaker, but I also don't want us to make an explicit promise that won't be true.
> 
> I don't _think_ `ORDER BY Id` has substantive performance impact, but I also don't want to break anything for @prescod. Paul, do you have any thoughts?

@davidmreed @davisagli 

I found a realistic use case for a mapping not to have `fields`:
- Using auto-generated ordered primary key `oid_pk` so there is no `Id: sf_id` in `fields`
- The SObject is a pure junction object, i.e. only has lookups, plus its `Name` is <kbd>Auto Number</kbd> hence not writable and cannot exist in mapping.

prescod:
I haven't done much work with extraction, so I don't have strong opinions or code to break. The use-case (of the ORDER BY) seems reasonable enough to me to justify a performance compromise: I'm not sure if anyone is actually using this task at scale or planning to. It does seem primarily to be used for small bits of sample data.

prescod:
If its true that the ORDER BY won't work past 10,000 records then we should document that.

spelak-salesforce:
Do we use the `Sforce-Enable-PKChunking` in the Bulk API query header?   The [PK Chunking Header](https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/async_api_headers_enable_pk_chunking.htm) doc ( suggest with PK Chunking, the query effectively gets an `ORDER BY Id` applied.

> The default starting ID is the first record in the table.

Also,

> Salesforce recommends that you enable PK chunking when querying tables with more than 10 million records or when a bulk query consistently times out. However, the effectiveness of PK chunking depends on the specifics of the query and the queried data.

spelak-salesforce:
The idea behind the `ORDER BY Id` clause in the extract task is to generate sql files with "minimal" diffs.  If we include `sf_id`, every time we extract/capture, the IDs change in the sql file and so appear in the diff.  However, the Id-diffs are really false positives since the data didn't really change.    This leads to devs not really paying attention to bulk data sql file diffs.

A solution could be to use `oid_pk`.  Currently, there is no `ORDER BY Id` clause and if a user extracts the same data many times in a row, the user will see changes of the order of records returned even though the records never changed between extractions.   This leads, again, to false positives (though substantially less false positives than using `sf_id`).   If we can stabilize the order of records returned in the query, it should stabilize the `oid_pk`.

davidmreed:
> If its true that the ORDER BY won't work past 10,000 records then we should document that.

`ORDER BY` will work >10k, but @spelak-salesforce's desired stability of PKs will be broken on a load/extract sequence for an object with >10k rows, unless the load is forced to use serial mode (which we did just merge support for).

davidmreed:
I am okay with this change. I'd like to do one E2E test before approving, but it looks like we don't have any negative impact on LDV, and Scott understands the limitations of what these changes buy.

davisagli:
@spelak-salesforce Sorry for losing track of this pull request -- I just found it while looking through old ones.

We just merged #1814 which should make it possible to have a mapping without `fields` specified.

I think ordering by Id to stabilize the order of records across multiple extractions (at least for datasets < 10k records) is a worthwhile goal. Could I ask you to re-submit that change in a fresh branch & pull request?

PR #1395, opened 2019-12-12
Error handling

Br4nd0R:
# Changes
* CumulusCI now stores the logs from the last five executions under `~/.cumulusci/logs`
* CumulusCI has a new top-level command: `cci gist` . This command creates a secret [gist](https://help.github.com/en/github/writing-on-github/creating-gists#about-gists) which includes: The user's current CumulusCI version, current Python version, path to python binary, sysname (e.g. Darwin), machine (e.g. x86_64), and the most recent log (`~/.cumulusci/logs/cci.log`). The command outputs a link to the created gist and opens a browser tab with the new GitHub gist. This can be helpful for sharing information regarding errors and issues encountered when working with `cci`. This feature uses a users GitHub access token for creation of gists. *If your access token does not have the 'gist (Create gists)' scope this command will result in a 404 error*.


Br4nd0R:
@jstvz - All good points, thanks for the feedback. I definitely see the benefit in using rotating logs with separate command that allows us to send back the last log file (or an option to send back the last n?). I can look into generating it in nice markdown format too.

Br4nd0R:
@jstvz - Ready for review/QA again.

Br4nd0R:
@jstvz - This is all great feedback, thanks for taking the time! 

I agree that if possible we should include a new file handler on the existing logger inside of `init_logger`. I tried that initially and was unsuccessful in capturing the output. Let me play with it again to see if I can get it to work as you mentioned above.

I can definitely implement the following:
* Better exception message for 404 errors
* Mention the `cci gist` command when we catch an error in the `cci.main()` method.

I'll have to look into the lift on outputting token scopes. I had initially opted for writing the log file to `~/.cumulusci/`. @davisagli  mentioned that is for things that should be shared if there are multiple clones of the same repository.

cumulusci/cli/cci.py
```diff
@@ -216,25 +165,68 @@ def new_func(*args, **kw):
...
+        # Configure logging
+        debug = "--debug" in args
+        if debug:
+            args.remove("--debug")
+
+        # Only create logfiles for commands
+        # that are not `cci gist`
+        is_gist_command = len(args) > 1 and args[1] == "gist"
+        if not is_gist_command:
+            logger = get_gist_logger(RUNTIME.project_config.repo_root)
```
davisagli: This breaks when running a command outside of a project directory, such as `cci project init`:

```
  File "/Users/dglick/.pyenv/versions/cci/bin/cci", line 11, in <module>
  File "/Users/dglick/Work/CumulusCI/cumulusci/cli/cci.py", line 204, in main
    logger = get_gist_logger(RUNTIME.project_config.repo_root)
AttributeError: 'NoneType' object has no attribute 'repo_root'
```

I'm actually reconsidering my earlier suggestion that logs should go inside the project directory. I don't see any particular problem with just having a central location for logs in `~/.cumulusci/logs`

cumulusci/cli/cci.py
```diff
@@ -259,42 +251,94 @@ def version():
...
+def shell(runtime):
+    # alias for backwards-compatibility
+    config = runtime
 
     code.interact(local=dict(globals(), **locals()))
 
 
+GIST_404_ERR_MSG = """A 404 error code was returned when trying to create your gist.
+Please ensure that your GitHub personal access token has the 'Create gists' scope."""
+CCI_LOG_NOT_FOUND_MSG = """No logfile to open at path: {log_path}
```
davisagli: log_path is not getting formatted

cumulusci/utils.py
```diff
@@ -564,3 +553,29 @@ def get_git_config(config_key):
...
+    def write(s):
+        output = strip_ansi_sequences(s)
+        logger.debug(output)
+        real_write(s)
+
+    sys.stdout.write = write
+    try:
+        yield
+    finally:
+        sys.stdout.write = real_write
```
davisagli: Perhaps we should also do this for sys.stderr

cumulusci/cli/cci.py
```diff
@@ -216,25 +165,68 @@ def new_func(*args, **kw):
         return decorate(func)
 
 
-# Root command
+SUGGEST_GIT_GIST_COMMAND = """\nIt looks like you may have run into error. Did you know cci has a command for sending this error to a GitHub gist?
+Just run `$ cci gist` and make sure that your GitHub access token as the 'create gist' scope.\n"""
```
```suggestion
Just run `$ cci gist` and make sure that your GitHub access token has the 'create gist' scope.\n"""
```

cumulusci/cli/cci.py
```diff
@@ -259,42 +251,94 @@ def version():
...
 
 
+GIST_404_ERR_MSG = """A 404 error code was returned when trying to create your gist.
+Please ensure that your GitHub personal access token has the 'Create gists' scope."""
+CCI_LOG_NOT_FOUND_MSG = """No logfile to open at path: {log_path}
+Please ensure you're running this command from the same directory you were experiencing an issue."""
+LAST_CMD_HEADER = "\n\n\nLast Command Run\n================================\n"
+
+
+@cli.command(name="gist", help="Create a gist from the latest logfile")
```
```suggestion
@cli.command(name="gist", help="Create a GitHub gist from the latest logfile")
```

cumulusci/cli/cci.py
```diff
@@ -259,42 +251,94 @@ def version():
...
+@pass_runtime(require_project=False)
+def gist(runtime):
+    repo_root = RUNTIME.project_config.repo_root
+    log_path = f"{repo_root}/.cci/cci.log" if repo_root else f"cci.log"
+    try:
+        last_cmd_log = open(log_path, "r")
+    except FileNotFoundError:
+        click.echo(CCI_LOG_NOT_FOUND_MSG)
+        sys.exit(1)
```
davisagli: Raise a CumulusCIException. That leaves the decisions about how to report the error and how to exit in the hands of the `main` wrapper so it's consistent with other commands. For example that way it's possible to run `cci gist` with the `--debug` flag to see the traceback.

cumulusci/core/github.py
```diff
@@ -160,3 +160,13 @@ def find_previous_release(repo, prefix=None):
             most_recent = release
         else:
             return release
+
+
+def create_gist(github, description, files):
+    """Creates a gist with the given description an files.
```
davisagli:
```suggestion
    """Creates a gist with the given description and files.
```

cumulusci/core/github.py
```diff
@@ -160,3 +160,13 @@ def find_previous_release(repo, prefix=None):
...
+
+    github - an
+    description - str
+    files - A dict of files in the form of {filename:{'content': content},...}
+    """
+    return github.create_gist(description, files, public=False)
```
davisagli: Is there a good reason for this function to exist instead of just calling github.create_gist directly?

Br4nd0R: I figured that keeping all references to `github3` in `core/github.py` would: (1) minimize effort needed to switch to a new github library if we ever wanted to, and (2) improve testability by only having to thoroughly test expected results from the github3 library in one module.

Br4nd0R:
@jstvz @davisagli - Ready for a final pass when you get a chance.

cumulusci/cli/cci.py
```diff
@@ -216,25 +167,68 @@ def new_func(*args, **kw):
...
+        try:
+            cli(standalone_mode=False)
+        except Exception as e:
+            # Display the error
+            if debug:
+                traceback.print_exc()
+                pdb.post_mortem()
+            else:
+                click.echo(click.style(f"Error: {e}", fg="red"))
+                click.echo(click.style(SUGGEST_GIT_GIST_COMMAND, fg="yellow"))
```
davisagli: @Br4nd0R we should make sure the traceback gets logged to the file even when it's not printed to stdout/stderr (i.e. even when the `--debug` flag is not set)

cumulusci/cli/cci.py
```diff
@@ -259,42 +253,93 @@ def version():
...
+Please ensure you're running this command from the same directory you were experiencing an issue."""
+LAST_CMD_HEADER = "\n\n\nLast Command Run\n================================\n"
+CCI_LOGFILE_PATH = Path.home() / ".cumulusci/logs/cci.log"
+
+
+@cli.command(name="gist", help="Create a GitHub gist from the latest logfile")
+@pass_runtime(require_project=False)
+def gist(runtime):
+    try:
+        last_cmd_log = open(CCI_LOGFILE_PATH, "r")
```
davisagli: Use open as a context manager to make sure the file gets closed when we're done with it.

cumulusci/cli/cci.py
```diff
@@ -259,42 +253,93 @@ def version():
...
+            get_github_api(gh.config["username"], gh.config["password"]),
+            "CumulusCI Error Output",
+            files,
+        )
+    except Exception as e:
+        if e.response.status_code == 404:
+            click.echo(GIST_404_ERR_MSG)
+        else:
+            click.echo(f"An error occurred attempting to create your gist:\n{e}")
+            sys.exit(1)
```
davisagli: For both lines 298 and 300 I think it would be more consistent to raise an exception and let the wrapper take care of printing it and exiting?

cumulusci/cli/cci.py
```diff
@@ -1362,3 +1382,17 @@ def flow_run(config, flow_name, org, delete_org, debug, o, skip, no_prompt):
...
+#   - make sure traceback gets logged (by BuildFlow.run, I guess)
+#   - but not for failures
+#   - make sure contextual info is there for sentry
+#     - build
+#     - task path
+#     - repo
+#     - commit
+#     - branch
+#     - plan
+#     - org
```
davisagli: We can remove this comment; I've tested what I wanted to in MetaCI.

docs/features.rst
```diff
@@ -729,4 +729,25 @@ the source should get pushed successfully from the ``force-app`` directory.
 Caveats:
 
 * It’s possible there may be some minor errors in the automatic conversion that need special attention in order to deploy successfully.
-* See https://ntotten.com/2018/05/11/convert-metadata-to-source-format-while-maintain-git-history/ for some tips on preserving git history while converting your source format.
+* See `this link <https://ntotten.com/2018/05/11/convert-metadata-to-source-format-while-maintain-git-history/>`_ for some tips on preserving git history while converting your source format.
+
+
+GitHub Gists
+=============
+Use the ``cci gist`` command to send the log of your last ``cci`` command to a github gist so you can submit it for support if needed.
```
davisagli:
```suggestion
Use the ``cci gist`` command to send the log of your last ``cci`` command to a GitHub gist so you can submit it for support if needed.
```

docs/features.rst
```diff
@@ -729,4 +729,25 @@ the source should get pushed successfully from the ``force-app`` directory.
...
+    * The path to the python executable
+    * The ``sysname`` of the host (e.g. Darwin)
+    * The machine name of the host (e.g. x86_64)
+    * The most recent logfile (cci.log) that CumulusCI has created.
+
+The URL for the gist is displayed on the terminal of the user as output, and a web browser will automatically open a tab to the gist.
+
+**Logfiles**
+
+CumulusCI creates a logfile everytime a cci command is run. The only exception to this is when the cci gist command is run. Logfiles are stored under `~/.cumulusci/logs`.
```
davisagli:
```suggestion
CumulusCI creates a logfile every time a cci command is run. The only exception to this is when the cci gist command is run. Logfiles are stored under `~/.cumulusci/logs`.
```

docs/features.rst
```diff
@@ -729,4 +729,25 @@ the source should get pushed successfully from the ``force-app`` directory.
 Caveats:
 
 * It’s possible there may be some minor errors in the automatic conversion that need special attention in order to deploy successfully.
-* See https://ntotten.com/2018/05/11/convert-metadata-to-source-format-while-maintain-git-history/ for some tips on preserving git history while converting your source format.
+* See `this link <https://ntotten.com/2018/05/11/convert-metadata-to-source-format-while-maintain-git-history/>`_ for some tips on preserving git history while converting your source format.
+
+
+GitHub Gists
```
davisagli: Maybe we should give this a title like `Reporting error logs` that's more likely to be meaningful to the reader. Gists are a means to an end.

jstvz:
@davisagli Is your error handling change in this PR ready to review as well?

davisagli:
@jstvz I believe so.

Br4nd0R:
@davisagli - Thanks for reviewing again; changes pertaining to your review have been made.

cumulusci/tasks/github/tests/util_github_api.py
```diff
@@ -466,3 +467,94 @@ def _get_expected_label(self, name=None, desc=None):
...
         return [self._get_expected_label(name=label) for label in labels]
+
+    def _get_expected_gist(self, description, files, public=False):
+        """Gist creationg returns 201 on success"""
+        gh_id = self.create_id(20)
+
+        gist_files = {}
+        for f in files:
+            for filename in f.keys():
+                gist_files[filename] = {
```
jstvz: 🤔: Why pass a list of dicts? If you pass a dict you can iterate over `files.items()` instead.

cumulusci/tests/test_utils.py
```diff
@@ -461,24 +479,29 @@ def test_get_git_config_error(self, Command):
...
+        expected_stdout_text = "This is expected stdout."
+        expected_stderr_text = "This is expected stderr."
+        with utils.tee_stdout_stderr(args, logger):
+            sys.stdout.write(expected_stdout_text)
+            sys.stderr.write(expected_stderr_text)
+
+        logger.debug.call_count == 2
+        logger.debug.call_args_list[0] == "cci test"
+        logger.debug.call_args_list[1] == expected_stdout_text
+        logger.debug.call_args_list[2] == expected_stderr_text
```
jstvz: ±: Should these be `assert`s?

cumulusci/cli/logger.py
```diff
@@ -1,9 +1,10 @@
 """ CLI logger """
 import logging
+import logging.handlers
```
jstvz: ±: Did you mean?:
```suggestion
from logging.handlers import RotatingFileHandler
```
(with a corresponding change on line 55 of this file)

Otherwise I don't think you need this line.

cumulusci/cli/tests/test_cci.py
```diff
@@ -1167,90 +1335,14 @@ def test_task_run_debug_after(self, set_trace):
...
-            None,
-            config={
-                "tasks": {
-                    "test": {"class_path": "cumulusci.cli.tests.test_cci.DummyTask"}
-                }
-            },
-        )
-        DummyTask._run_task.side_effect = TaskOptionsError
-
-        with self.assertRaises(click.UsageError):
```
jstvz: 🤔: Presumably, these were removed because we're catching these globally, but are we certain that we don't need any of them to confirm that malformed input results in an exception?

cumulusci/cli/tests/test_cci.py
```diff
@@ -157,34 +136,194 @@ def test_render_recursive(self):
...
+            run_click_command(cci.gist, runtime=runtime)
+        assert (
+            "An error occurred attempting to create your gist"
+            in context.exception.args[0]
+        )
+
+        create_gist.side_effect = ExceptionWithResponse(404)
+        with self.assertRaises(CumulusCIException) as context:
+            run_click_command(cci.gist, runtime=runtime)
+        assert cci.GIST_404_ERR_MSG in context.exception.args[0]
```
jstvz: Oh, now I see why they're module level constants. (Well, `GIST_404_ERR_MSG` at least)

cumulusci/cli/cci.py
```diff
@@ -216,25 +167,68 @@ def new_func(*args, **kw):
         return decorate(func)
 
 
-# Root command
+SUGGEST_GIT_GIST_COMMAND = """\nIt looks like you may have run into error. Did you know cci has a command for sending this error to a GitHub gist?
```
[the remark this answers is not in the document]
davisagli: Why? Then there's an extra step to open the other file to check what's in the message.

cumulusci/cli/cci.py
```diff
@@ -259,42 +256,95 @@ def version():
...
+LAST_CMD_HEADER = "\n\n\nLast Command Run\n================================\n"
+CCI_LOGFILE_PATH = Path.home() / ".cumulusci/logs/cci.log"
+
+
+@cli.command(name="gist", help="Create a GitHub gist from the latest logfile")
+@pass_runtime(require_project=False)
+def gist(runtime):
+    try:
+        with open(CCI_LOGFILE_PATH, "r") as last_cmd_log:
+            log_content = last_cmd_log.read()
```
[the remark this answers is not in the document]
davisagli: There's a theoretical race condition in your version @jstvz if the file gets deleted between the is_file check and the read_text call. But that's unlikely enough in this context that I wouldn't worry about it.

cumulusci/cli/tests/test_cci.py
davisagli: Yeah, I removed some of these tests because they were for exception handling codepaths that got replaced by the centralized handling. Testing that task classes validate their options properly belongs in the tests for the task class. I did keep test_task_run_invalid_option which tests the extra option validation that is done by task_run.

cumulusci/cli/cci.py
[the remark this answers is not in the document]
jstvz: Personal preference? The content of the message isn't relevant here; it isn't an f-string and we're not manipulating it. Including it here breaks line length and reduces readability. 

If you need to check/change this or any other message's content, `jump-to-definition` is only a keystroke away.

cumulusci/cli/cci.py
```diff
@@ -259,42 +253,95 @@ def version():
...
 
     code.interact(local=dict(globals(), **locals()))
 
 
+GIST_404_ERR_MSG = """A 404 error code was returned when trying to create your gist.
+Please ensure that your GitHub personal access token has the 'Create gists' scope."""
+CCI_LOG_NOT_FOUND_MSG = """No logfile to open at path: {}
+Please ensure you're running this command from the same directory you were experiencing an issue."""
+LAST_CMD_HEADER = "\n\n\nLast Command Run\n================================\n"
+CCI_LOGFILE_PATH = Path.home() / ".cumulusci/logs/cci.log"
```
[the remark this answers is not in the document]
Br4nd0R: `CCI_LOGFILE_PATH` is also now used by main to capture the traceback. I'll move the others into the `gist` command method.

cumulusci/cli/cci.py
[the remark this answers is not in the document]
Br4nd0R: I personally didn't like how much indentation there there was in this method. I'll opt to use James' approach.

cumulusci/tests/test_utils.py
Br4nd0R: doh

PR #1399, opened 2019-12-14
Enable type-guard.

prescod:
# Critical Changes

# Changes

# Issues Closed


prescod:
Yeah I also noticed the stack issue and I spent about 10 minutes looking at TypeGuard and it already did have some code supposedly to help improve that but it doesn't seem to work. When Advent of Code is done I may look at it on a weekend.

PR #1404, opened 2019-12-16
improve doc_task

Br4nd0R:
# Changes
* doc_task now outputs in a format that closely resembles the structure of[ sfdx documentation](https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_force_org.htm) for commands.

## Outstanding Questions
* The 'command syntax' does not conform to the typical [usage message](https://en.wikipedia.org/wiki/Usage_message) format. Do we want to stick with a more conventional usage statement syntax?


cumulusci/utils.py
```diff
@@ -382,37 +381,118 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
...
-    doc.append("**Description:** {}\n".format(task_config.description))
-    doc.append("**Class::** {}\n".format(task_config.class_path))
+    doc.append(f"**{task_name}**\n==========================================\n")
+    doc.append(f"**Description:** {task_config.description}\n")
+    doc.append(f"**Class:** {task_config.class_path}\n")
 
     task_class = import_global(task_config.class_path)
-    if "task_docs" in task_class.__dict__:
-        task_docs = textwrap.dedent(task_class.task_docs.strip("\n"))
-        doc.append(task_docs + "\n")
```
davisagli: @Br4nd0R did you intend to stop including the task_docs? I think it has useful information.

cumulusci/utils.py
```diff
@@ -382,37 +381,118 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
...
+        option_type = option.get("option_type")
+        if option_type:
+            doc.append(f"\n\t Type: {option_type}")
+
+    return doc
+
+
+def create_task_help_doc(task_config):
+    """Create documentation for the 'help' section of a task doc"""
+    return []
```
davisagli: ?

cumulusci/utils.py
```diff
@@ -382,37 +381,118 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
...
+    return usage_str
+
+
+def create_task_options_doc(task_options):
+    """Generate the 'Options' section for a given tasks documentation"""
+    doc = []
+    for option in task_options:
+        usage_str = option.get("usage")
+        if usage_str:
+            doc.append(f"\n``[{usage_str}]``")
```
davisagli: I think square brackets typically indicate optional arguments in a usage message. We should leave them out or use angle brackets for required options.

cumulusci/utils.py
```diff
@@ -382,37 +381,118 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
...
+            f"Help for task {task_name}\n------------------------------------------\n"
+        )
+        doc.extend(task_help_doc)
 
     return "\n".join(doc)
 
 
+def get_command_syntax(task_name):
+    """Return an example command syntax string in .rst format"""
+    return f"``$ cci task run {task_name}``\n\n"
```
davisagli: It would be nice to include `--org [org name]` here for the tasks that require an org_config. However, we don't currently have a way for tasks to declare whether they do or not.

cumulusci/utils.py
Br4nd0R: My intention here was to mimic what is currently in the help section of sfdx command documentation. I'm thinking that if I include task_docs I can take this part out for now.

cumulusci/utils.py
Br4nd0R: These should definitely be in there. I got a little too wrapped up with matching what is in docs for sfdx commands.

Br4nd0R:
@davisagli - Let me know if there's anything else this one needs.

cumulusci/tests/test_utils.py
```diff
@@ -337,33 +445,7 @@ def test_zip_clean_metaxml__handles_nonascii(self):
...
-extra docs
-
-Options:
-------------------------------------------
-
-* **flavor** *(required)*: What flavor
-* **color**: What color **Default: black**"""
-            == result
-        )
+        assert b"<root>\xc3\xb1</root>" in zf.read("classes/test-meta.xml")
```
davisagli: @Br4nd0R I'm curious why you had to change these to test using `in`? Were there newline differences or something?

cumulusci/utils.py
```diff
@@ -404,37 +404,122 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
...
+    if task_option_doc:
+        doc.append("Options\n------------------------------------------\n")
+        doc.extend(task_option_doc)
+
+    task_help_doc = create_task_help_doc(task_config)
+    if task_help_doc:
+        doc.append(
+            f"Help for task {task_name}\n------------------------------------------\n"
+        )
+        doc.extend(task_help_doc)
```
davisagli: These lines don't have coverage. You can probably just remove create_task_help_doc

cumulusci/tests/test_utils.py
Br4nd0R: Switched this back. Not sure why the change was made.

PR #1407, opened 2019-12-19
Allow tasks to output rather than consume mapping.yml

prescod:


prescod:
I don't think that there is anything very relevant to the release notes. It changes internal APIs that I am probably the only person using. It makes some things that were previously required optional, so it should be backwards compatible.

prescod:
It's not clear to me how to make tasks.rst . I couldn't find anything in CONTRIBUTING or Makefile that seemed appropriate. 

"make docs" didn't seem to generate it.

"grep tasks.rst" didn't turn up anything.

davisagli:
@prescod it's `cci task doc > docs/tasks.rst`

Including that in `make docs` is a good idea; I'll put in a PR to add that.

prescod:
It turns out that tasks.rst doesn't change because this task is not in cumulusci.yml yet. I did take the opportunity to clean up some naming and add a bit of inline documentation however.

davisagli:
@prescod oh yeah, that makes sense.

coveralls shows missing coverage since snowfakery isn't installed when the tests run on CircleCI. Doing that will get simpler once it's public so maybe we should just not worry about it for now.

PR #1408, opened 2019-12-20
Call Snowfakery from CCI task if both are installed

prescod:
The task is lazy-loaded so it shouldn't cause a problem for someone who doesn't need snowfakery for their flows.

The tests are eager-loaded, so I put some tricks in to skip them if snowfakery is not installed.


cumulusci/tasks/bulkdata/tests/tests/test_generate_from_snowfakery_task.py
```diff
@@ -0,0 +1,116 @@
...
+from cumulusci.tasks.bulkdata.tests.test_bulkdata import _make_task, TaskOptionsError
+from tempfile import NamedTemporaryFile
+
+import yaml
+from sqlalchemy import create_engine
+
+try:
+    import snowfakery
+except ImportError:
+    snowfakery = None
```
davisagli: I think you could use `pytest.importorskip`: http://doc.pytest.org/en/latest/skipping.html#skipping-on-a-missing-import-dependency

PR #1414, opened 2019-12-23
Public Repo Publishing

jlantz:
Added task for publishing a release from a private repo to a stripped down public repo designed to be used as an endpoint for external extensions and implementations.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -126,6 +126,16 @@ tasks:
             children_only: True
             source_branch: $project_config.repo_branch
         group: GitHub
+    github_publish_release:
+        description: Publishes a release from the project repo to a target repo with filtering of files
+        class_path: cumulusci.tasks.github.public_repo.PublishRelease
+        options:
+            version: latest
+            branch: master
+            exclude:
+                - force-app
+                - push
+                - src
     github_pull_requests:
         description: Lists open pull requests in project Github repository
         class_path: cumulusci.tasks.github.PullRequests
--- a/cumulusci/tasks/github/public_repo.py
+++ b/cumulusci/tasks/github/public_repo.py
@@ -1,0 +1,148 @@
+import os
+import shutil
+from datetime import datetime
+import github3.exceptions
+
+from cumulusci.core.exceptions import GithubException
+from cumulusci.core.github import get_github_api_for_repo
+from cumulusci.core.utils import process_bool_arg
+from cumulusci.core.utils import process_list_arg
+from cumulusci.tasks.github.base import BaseGithubTask
+from cumulusci.tasks.github.util import CommitDir
+from cumulusci.utils import cd
+from cumulusci.utils import download_extract_github_from_repo
+from cumulusci.utils import temporary_dir
+
+
+class PublishRelease(BaseGithubTask):
+    task_options = {
+        "repo_url": {"description": "The url to the public repo", "required": True,},
+        "branch": {
+            "description": "The branch to update in the target repo",
+            "required": True,
+        },
+        "version": {
+            "description": "The version number to release.  Also supports latest and latest_beta to look up the latest releases.",
+            "required": True,
+        },
+        "exclude": {"description": "A list of paths from repo root to exclude",},
+        "release_body": {
+            "description": "If True, the entire release body will be published to the public repo.  Defaults to False",
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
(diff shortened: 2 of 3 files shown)


jlantz:
Known issues:

1. I'm not sold on the naming of this task but can't come up with a better option.  Suggestions requested.

2. The CommitDir class doesn't seem to be working the way it should.  It should be handling the deletion of trees from the repo if they're not in the dir being committed.  It also seems like it shouldn't be doing a POST for each blob if they blob hasn't changed locally.

cumulusci/tasks/github/public_repo.py
```diff
@@ -0,0 +1,148 @@
...
+            self._target_repo_info["owner"], self._target_repo_info["name"],
+        )
+
+    def _run_task(self):
+        self._target_repo_info = self.project_config._split_repo_url(
+            self.options["repo_url"]
+        )
+        self.target_repo = self._get_target_repo_api()
+
+        with temporary_dir(chdir=False) as target:
```
davisagli: In Python 3 you can now just use tempfile.TemporaryDirectory as a context manager, if you don't want to cd to it.

cumulusci/tasks/github/public_repo.py
```diff
@@ -0,0 +1,148 @@
...
+        # Get current release info
+        tag_name = self.project_config.get_tag_for_version(self.options["version"])
+        repo = self.get_repo()
+        # Get the ref
+        try:
+            ref = repo.ref(f"tags/{tag_name}")
+        except github3.exceptions.NotFoundError:
+            message = f"Ref not found for tag {tag_name}"
+            self.logger.error(message)
+            raise GithubException(message)
```
davisagli: Why are you both logging at the error level and also raising an exception with the same message? Exceptions _should_ make it into the logs.

cumulusci/tasks/github/public_repo.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # Check for tag in target repo
+        target_ref = None
+        try:
+            target_ref = self.target_repo.ref(f"tags/{tag_name}")
+        except github3.exceptions.NotFoundError:
+            pass
+        if target_ref:
```
davisagli:
```suggestion
        else:
```

cumulusci/cumulusci.yml
```diff
@@ -126,6 +126,16 @@ tasks:
             children_only: True
             source_branch: $project_config.repo_branch
         group: GitHub
+    github_publish_release:
```
davisagli: What about calling it "clone release"?

cumulusci/tasks/github/public_repo.py
```diff
@@ -0,0 +1,148 @@
...
+        },
+        "exclude": {"description": "A list of paths from repo root to exclude",},
+        "release_body": {
+            "description": "If True, the entire release body will be published to the public repo.  Defaults to False",
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+        self.options["exclude"] = process_list_arg(self.options.get("exclude", []))
```
davisagli: Using an exclude list has the downside that if something new gets added to the repo which is not supposed to be public, we have to remember to add it to the list. Would it be feasible to use a list of what to include instead?

cumulusci/tasks/github/public_repo.py
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        # Create the release
+        body = release.body if self.options["release_body"] else ""
+        target_release = self.target_repo.create_release(
+            tag_name=tag_name,
+            name=self.options["version"],
+            prerelease=release.prerelease,
+        )
+
```
davisagli: Please add at least one logline at the info level to confirm that the release was created.

Also this blank line at the end of the file is causing the lint check to fail. You might need to double check that you have the precommit hooks installed (which should catch this before committing -- see CONTRIBUTING.rst) or reinstall requirements to make sure you have the latest version of black.

cumulusci/cumulusci.yml
jlantz: Actually, thinking about this more and `github_clone_release` might be confusing since clone has specific meaning in git.  Perhaps `github_copy_release`?

davisagli: @jlantz yeah ok

davisagli: @jlantz although we do have an existing github_clone_tag task

davisagli:
Replaced by #1899

PR #1429, opened 2020-01-01
Add task to manage TDTM handlers

davidmreed:
# Critical Changes

# Changes

- A task class, `SetTDTMHandlerStatus`, allows projects that use TDTM trigger handlers (via EDA or NPSP or both) to automate changing, and later restoring, the status of TDTM handlers.

# Issues Closed


davidmreed:
n.b. I've not added a named task to `cumulusci.yml` because each project (EDA, NPSP, or extensions of either or both) will need to specify which namespace's Trigger Handlers it wants to touch. I don't think there's a globally-applicable default.

prescod:
It looks good at a high level to me: but why can't you declare namespace to be a required field and then add it to cumulusci.yml? Specific projects could override with their namespace defaulted.

prescod:
I implemented something similar in the context of Abacus but I hadn't generalized it and tested it yet.

One question I had had was: if you try to disable triggers and none of them match, is that benign?

In the circumstance where the lack of match is caused by the trigger simply not existing, it's probably benign.

If you try to enable it and the trigger doesn't exist, that's a bit less clear.

And if the failure to disable is caused by a misspelled trigger then it is certainly not benign.

I had created an if_missing option to leave the decision about what to do in the hands of the caller.

        "if_missing": {
            "description": ("'error' or 'ignore' situation "
                            "if no trigger matches anything in your list. "
                            "Defaults to 'error'"),
            "required": False,
        },

I don't feel strongly about it, but I thought I would mention it in case you felt it was a useful idea to carry forward.

davidmreed:
@davisagli, good thoughts - whether this task belongs in CumulusCI might benefit from a more in-depth discussion.

> It makes me wonder if it would be better to build this as a generic "update one field on a set of objects" task

Were it only to support mutating _all_ records in a given sObject, yes, or to filter by a single field on the sObject, also yes; however, TDTM handlers are uniquely identified by the combination of _two_ fields, `Class__c` and `Object__c`. A stock NPSP org contains ~50 TDTM handlers, and I do think there's a use case for allowing them to be selectively disabled and enabled, needing some logic that is TDTM-specific.

> Do you think there's much risk of the EDA and NPSP implementations of TDTM diverging at some point in the future such that it no longer makes sense to share this task between them?

I really don't know, but it would be a pretty substantive architecture change in either product if the assumptions made by this task were to be altered.

davidmreed:
@prescod  I agree with the impetus towards loud/explicit failures. It's complicated a little bit here because the list of handlers to be modified includes wildcarding - for example, the user can specify an sObject, a TDTM class name, or both. The user could provide a set of requests that would achieve their desired end, but wouldn't involve affecting handlers in a 1:1 association with the provided options.

NPSP doesn't ship with any disabled trigger handlers; I think that the vast majority of use cases for this task is to disable, and then re-enable, specific handlers rather than to activate any that are in a disabled state at the outset.

prescod:
David Reed,

We're sort of recapitulating the discussion that knocked my own TDTM enable/disable task off track. (although you did take it farther than I did, with the ability to remember the state of them)

You are probably right that we need a specific task (as opposed to one for Updates) but I'd like to drill into it a bit more, because we had the same discussion a few weeks ago without coming to a conclusion

Our DeleteData task has a WHERE clause which allows you to query on more than one column at a time. Would that allow an UpdateWhere task to do almost the same thing as this task?

davidmreed:
> Our DeleteData task has a WHERE clause which allows you to query on more than one column at a time. Would that allow an UpdateWhere task to do almost the same thing as this task?

I think so, but I would argue that that'd be a poorer user experience. Wherever we can insulate the user from actually writing SOQL in favor of structured abstraction, I'd like to do so (although I recognize there's a lot of scope for discussion there).

cumulusci/cumulusci.yml
```diff
@@ -410,7 +410,20 @@ tasks:
         description: Remove specified XML elements from one or more metadata files
         class_path: cumulusci.tasks.metadata.modify.RemoveElementsXPath
         group: Salesforce Metadata
-
+    disable_tdtm_trigger_handlers:
+        class_path: cumulusci.tasks.salesforce.trigger_handlers.SetTDTMHandlerStatus
+        description: "Disable specified TDTM trigger handlers"
+        group: Utilities
```
davisagli: Maybe `group: Salesforce.org`?

cumulusci/tasks/salesforce/trigger_handlers.py
```diff
@@ -0,0 +1,91 @@
...
+        sobject_names = [x["name"] for x in global_describe["sobjects"]]
+
+        if self.options["restore"]:
+            with open(self.options["restore_file"], "r") as f:
+                target_status = yaml.safe_load(f)
+                self.options["handlers"] = list(target_status.keys())
+        else:
+            target_status = defaultdict(lambda: self.options["active"])
+
+        namespace = self.options.get("namespace", "npsp") + "__"
```
davisagli: I'm not totally sold on using npsp as a default here, but I don't feel strongly about it.

davidmreed:
@codexmarcel  Would you be up for doing a hands-on test to validate behavior in an NPSP org?

cumulusci/cumulusci.yml
davidmreed: What about `group: NPSP/EDA`? I don't want to give the implication that they're just for our use (extenders of the packages need this too).

cumulusci/tasks/salesforce/trigger_handlers.py
davidmreed: No, you're right - there shouldn't be a default. I'll remove it, with the result that you can use the task with no namespace in a non-namespaced org, and it'll throw an exception asking you to specify if the org's namespaced and it cannot locate the handler object.

PR #1430, opened 2020-01-02
Added ability to register locators under a custom strategy name.

boakley:
RFC for this feature: https://salesforce.quip.com/XfkLAZA7b4r8

Most of the functionality is managed through the new locator_manager.py
module. This necessitated moving the keyword 'Initialize Location
Strategies' from Salesforce.robot to Salesforce.py so it could call some
internal functions.

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/tests/test_locator_manager.py
```diff
@@ -0,0 +1,151 @@
+import unittest
```
davisagli: This module should be located in cumulusci/robotframework/tests. Our usual pattern is that if the actual code is in x/y.py, the tests for it should be in x/tests/test_y.py

cumulusci/robotframework/locator_manager.py
```diff
@@ -0,0 +1,149 @@
...
+0. Locators are stored in a global LOCATORS dictionary. The keys
+   are the locator prefixes, the values are dictionaries containing
+   the locators
+
+1. Libraries can register a dictionary of locators with a prefix
+   (eg: register_locators("npsp", {...})). These get added to
+   LOCATORS
+
+2. Open Test Brower calls selenium's add_location_strategy for each
+   registered set of locators. (Note: This cannot be called before a
```
davisagli:
```suggestion
   registered set of locators. (Note: Location strategies cannot be added before a
```

to remove the ambiguous "this" which made me think this was saying you can't call Open Test Browser until a browser is open, until I figured out what you meant.

cumulusci/core/tests/test_locator_manager.py
```diff
@@ -0,0 +1,151 @@
...
+            locators={
+                "foo": {
+                    "message": "{} {}",
+                    "bar": {
+                        "baz": "//div[@class='baz']",
+                        "hello": "//span[text()='Hello, {}']",
+                    },
+                }
+            },
+        )
```
davisagli: These test classes that register locators in setUpClass should probably also have a tearDownClass that clears LOCATORS, lest we try to add new tests that use the global locators in the future and need to track down where the pollution is coming from.

Personally I would consider using a class-scoped pytest fixture instead: https://computableverse.com/blog/pytest-sharing-class-fixtures

cumulusci/robotframework/Salesforce.py
```diff
@@ -81,6 +82,21 @@ def builtin(self):
     def cumulusci(self):
         return self.builtin.get_library_instance("cumulusci.robotframework.CumulusCI")
 
+    def initialize_location_strategies(self):
+        """Initialize the Salesforce location strategies 'text' and 'title'
+        plus any strategies registered by other keyword libraries
+
+        Note: This keyword is called automatically from Open Test Browser
+        """
+        locator_manager.register_locators("cci", lex_locators)
```
davisagli: Possibly we should use `sf` as the prefix for the core set of Salesforce locators -- i.e. name them after the functional domain that they're related to, rather than the fact that they happen to be defined by CumulusCI.

cumulusci/robotframework/locator_manager.py
```diff
@@ -0,0 +1,149 @@
...
+    # SeleniumLibrary's filter mechanism is a private function. In
+    # practice it probably won't matter <shrug>.
+    selenium = BuiltIn().get_library_instance("SeleniumLibrary")
+    loc = translate_locator(prefix, locator)
+    BuiltIn().log(f"locate_element: '{prefix}:{locator}' => {loc}", "DEBUG")
+    try:
+        element = selenium.get_webelement(loc)
+    except Exception:
+        raise Exception(
+            f"Element with locator '{prefix}:{locator}' not found\ntranslated: '{loc}'"
```
davisagli: Including the translated locator here is a nice touch.

cumulusci/robotframework/locator_manager.py
```diff
@@ -0,0 +1,149 @@
...
+    # exists, so we use a flag to make sure this code is called
+    # only once.
+    selenium = BuiltIn().get_library_instance("SeleniumLibrary")
+    for (prefix, strategy) in LOCATORS.items():
+        try:
+            BuiltIn().log(f"adding location strategy for '{prefix}'", "DEBUG")
+            if isinstance(strategy, dict):
+
+                def strategy(parent, locator, tag, constraints):
+                    locate_element(prefix, parent, locator, tag, constraints)
```
davisagli: This strategy function is no longer used since you switched to functools.partial (thanks coverage)

cumulusci/robotframework/locator_manager.py
boakley: This helped me find a bug that Selenium was sometimes unregistering the locators when I wasn't expecting it. Unfortunately, if the location strategy is unregistered, robot/selenium will try to use the pattern as a plain text string rather than saying something like "unknown strategy x". This is both helpful when debugging but also acts as a canary in the coal mine for when something goes wrong with the locators.

cumulusci/robotframework/Salesforce.py
boakley: After letting Jason weigh in, I'll rename it to `sf`. Thanks for the suggestion!

cumulusci/robotframework/locator_manager.py
boakley: Hah! Good catch. I saw the coverage report and thought it was wrong, because I "knew" the code was working. I had forgotten I switched to `functools.partial` instead of using the embedded function. Now I'm up to 100% coverage of the new code. w00t!

cumulusci/core/tests/test_locator_manager.py
boakley: I added a tearDownClass method in the interest of time, since I'm unfamiliar with pytest fixtures. Thanks for the link, though. I may switch to using a fixture once I've played around with them a bit.

cumulusci/core/tests/test_locator_manager.py
boakley: I've moved it. Thanks for catching it.

cumulusci/robotframework/locator_manager.py
```diff
@@ -0,0 +1,145 @@
...
+
+0. Locators are stored in a global LOCATORS dictionary. The keys
+   are the locator prefixes, the values are dictionaries containing
+   the locators
+
+1. Libraries can register a dictionary of locators with a prefix
+   (eg: register_locators("npsp", {...})). These get added to
+   LOCATORS
+
+2. Open Test Brower calls selenium's add_location_strategy for each
```
davidmreed:
```suggestion
2. Open Test Browser calls selenium's add_location_strategy for each
```

cumulusci/robotframework/locator_manager.py
```diff
@@ -0,0 +1,145 @@
...
+   registered set of locators. (Note: Location strategies cannot be added before a
+   browser is open.
```
davidmreed:
```suggestion
   browser is open).
```

PR #1431, opened 2020-01-02
Allow batches of records to be generated and loaded with Snowfakery

prescod:
The generate_and_load_data task was created last year to allow a "generate records" and "publish" records loop to upload records in an arbitrary number of batches. Any Data Generator task can be plugged into this loop.

Snowfakery has a new concept of "continuation files" which allow it to generate a series of batches which might have backwards foreign key references.

This code mostly glues those two features together to allow Snowfakery to be started and stopped.

Per our other discussions, this required a change to LoadData to allow it to persist IDs.

Over the long term, if Snowfakery obsoletes other Data Generator tasks, we may want to collapse some of these layers which provide generality which might end up going unused.


cumulusci/tasks/bulkdata/generate_and_load_data.py
```diff
@@ -29,7 +33,17 @@ class GenerateAndLoadData(BaseSalesforceApiTask):
     Use 'batch_size' to specify how many records to generate and upload in every batch.
 
     By default it creates the data in a temporary file and then cleans it up later. Specify database_url if you
-    need more control than that. The use of both database_url and batch_size together is not currently supported.
+    need more control than that. Existing data tables will be emptied before being refilled.
+    Your database will be completely deleted!
+
+    If you use database_url and batch_size together, latter batches will overwrite
+
+    If you use database_url and batch_size together, latter batches will overwrite
```
Br4nd0R: Duplicate sentence

cumulusci/tasks/bulkdata/generate_and_load_data.py
```diff
@@ -80,22 +104,38 @@ def _init_options(self, kwargs):
...
+                    "but `replace_database` was not specified"
+                )
 
     def _run_task(self):
         with TemporaryDirectory() as tempdir:
             for current_batch_size, index in generate_batches(
                 self.num_records, self.batch_size
             ):
+                self.logger.info(
+                    f"Generating a bunch, current_bunch_size={current_batch_size} "
```
davidmreed: The terminology here switches between `bunch` and `batch` - do we want `batch` across the board?

cumulusci/tasks/bulkdata/generate_and_load_data.py
```diff
@@ -121,15 +161,22 @@ def _dataload(self, subtask_options):
...
+
         subtask_options = {
             **self.options,
             "mapping": mapping_file,
+            "reset_oids": False,
             "database_url": database_url,
             "num_records": batch_size,
             "current_batch_number": index,
+            "working_directory": tempdir,
+            "reset_oids": False,
```
davidmreed: This argument is specified twice.

cumulusci/tasks/bulkdata/generate_and_load_data.py
```diff
@@ -139,3 +186,22 @@ def _generate_batch(self, database_url, tempdir, mapping_file, batch_size, index
             if not subtask_options.get("mapping"):
                 subtask_options["mapping"] = generated_mapping_file.name
             self._dataload(subtask_options)
+
+    def _setup_engine(self, database_url):
+        """Find all tables that do not relate to id->OID mapping"""
```
davidmreed: I don't think I understand this docstring for this method.

cumulusci/tasks/bulkdata/generate_from_yaml.py
```diff
@@ -32,7 +36,24 @@ class GenerateDataFromYaml(BaseGenerateDataTask):
             "description": "A path to put a mapping file inferred from the generator_yaml",
             "required": False,
         },
+        "num_records_tablename": {
+            "description": "A string representing which table to count records in."
+        },
+        "continuation_file": {
```
davidmreed: The descriptions for these two options might not provide enough guidance on how to use them.

cumulusci/tasks/bulkdata/generate_from_yaml.py
```diff
@@ -55,19 +87,79 @@ def _generate_data(self, db_url, mapping_file_path, num_records, current_batch_n
...
+                open_yaml_file=open_yaml_file,
+                user_options=self.vars,
+                output_stream=output_stream,
+                stopping_criteria=self.stopping_criteria,
+                continuation_file=old_continuation_file,
+                generate_continuation_file=new_continuation_file,
             )
             output_stream.close()
+
+            if (
```
davidmreed: I don't think I follow the full lifecycle of the continuation file (where each is created and how it's moved through the load process). Let's talk that through today.

cumulusci/tasks/bulkdata/generate_and_load_data.py
prescod: The Generate and load infrastructure has a mechanism for "batching" which is different than the batching used in the LoadData task. I've added a picture here:

https://salesforce.quip.com/Yh4fA1OA1hAj

One of the things to discuss is whether it is helpful to have two different words (e.g. bunch and batch or set and batch) or whether using batch at two different levels is okay.

cumulusci/tasks/bulkdata/generate_and_load_data.py
prescod: Cut and paste error.

cumulusci/tasks/bulkdata/generate_and_load_data.py
prescod: 7b52b9d4

cumulusci/tasks/bulkdata/generate_and_load_data.py
prescod: 974ba34cff42

cumulusci/tasks/bulkdata/generate_from_yaml.py
prescod: Maybe this will help:

7fd12dac

cumulusci/tasks/bulkdata/generate_and_load_data.py
prescod: Fixed in c5b77fd6

cumulusci/tasks/bulkdata/generate_from_yaml.py
prescod: num_records_tablename and continuation_file or

continuation_file and generate_continuation_file?

davidmreed: The latter two.

prescod: Is it clearer in 15e5f3ad ?

davidmreed: Very much so - thank you!

cumulusci/tasks/bulkdata/generate_and_load_data.py
davisagli: @prescod I thought you decided to use batch for both, but this still says bunch.

davisagli: @prescod oh never mind, that was from https://github.com/SFDO-Tooling/CumulusCI/pull/1435 which came after this.

prescod:
Coverage is high if snowfakery is installed and lower if it is missing. I think that's the same as before the PR. 

I do see one exception-throwing line that I missed which does not depend on Snowfakery, so I can fix that one.

The same thing is true of the documentation: I had mentioned that it makes sense to document this stuff when Snowfakery is available. This is the same reason we're not making a CCI.yml task for it yet.

PR #1441, opened 2020-01-07
Bump version: 3.3.0 → 3.3.1dev0

prescod:
Bump version to allow deployment to MetaCI.


prescod:
Can I do that with bump2version or I have to do it "by hand"?

davisagli:
@prescod I always do version updates by hand, so I'm not sure if there's a better way.

PR #1446, opened 2020-01-08
Failed deletes (and loads) throw an exception

prescod:
If a batch job has numberRecordsFailed then _wait_for_job will throw an exception. (this part is controversial)

Unless it is caught, the task that called it will throw an exception.

This means that the keywords Run Task and Run Task Class will also error.

All of this is tested below.

As far as we know, only DeleteData and LoadData uses these functions, but we don't know for sure about customers in the market.

It is unclear whether LoadData behaviour changes significantly, because it already has some error checking and exception raising.

====

This MAY NOT be an entirely backwards compatible change. Who knows how many errors are currently being swallowed in DeleteData (and perhaps also LoadData)? It is possible that some flows which currently seem to succeed will be revealed to actually have failing subtasks.
 
Arguably, we might need "exception behaviour" as a task option on effected tasks to give a way to get the old behaviour back.

Or arguably the old behaviour might be the default (during a transition period?)

### DIFF
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -101,7 +101,7 @@ class LoadData(BulkJobTaskMixin, BaseSalesforceApiTask):
             self._load_record_types([mapping["sf_object"]], conn)
         mapping["oid_as_pk"] = bool(mapping.get("fields", {}).get("Id"))
         job_id, local_ids_for_batch = self._create_job(mapping)
-        result = self._wait_for_job(job_id)
+        result = self._wait_for_job(job_id, error_behaviour="return")
 
         self._process_job_results(mapping, job_id, local_ids_for_batch)
 
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -102,6 +102,8 @@ class LoadData(BulkJobTaskMixin, BaseSalesforceApiTask):
         mapping["oid_as_pk"] = bool(mapping.get("fields", {}).get("Id"))
         job_id, local_ids_for_batch = self._create_job(mapping)
         result = self._wait_for_job(job_id, error_behaviour="return")
+        if result == "Failed":
+            errors = self.error_messages  # noQA   # Todo: what to do with these errors?
 
         self._process_job_results(mapping, job_id, local_ids_for_batch)
 
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -71,10 +71,7 @@ class LoadData(BulkJobTaskMixin, BaseSalesforceApiTask):
         self.bulk_mode = (
             self.options.get("bulk_mode") and self.options.get("bulk_mode").title()
         )
-        if self.bulk_mode and self.bulk_mode not in [
-            "Serial",
-            "Parallel",
-        ]:
+        if self.bulk_mode and self.bulk_mode not in ["Serial", "Parallel"]:
             raise TaskOptionsError("bulk_mode must be either Serial or Parallel")
 
     def _run_task(self):
@@ -99,7 +96,7 @@ class LoadData(BulkJobTaskMixin, BaseSalesforceApiTask):
                 for after_name, after_step in self.after_steps[name].items():
                     self.logger.info(f"Running post-load step: {after_name}")
                     result = self._load_mapping(after_step)
-                    if not result.startswith("Completed"):
+                    if result not in ("Completed", "CompletedWithFailures"):
                         raise BulkDataException(
                             f"Job {after_name} did not complete successfully"
                         )
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -93,7 +93,7 @@ class LoadData(BulkJobTaskMixin, BaseSalesforceApiTask):
 
             self.logger.info(f"Running Job: {name}")
             result = self._load_mapping(mapping)
-            if not result.startswith("Completed"):
+            if result not in ("Completed", "CompletedWithFailures"):
                 raise BulkDataException(f"Job {name} did not complete successfully")
             if name in self.after_steps:
                 for after_name, after_step in self.after_steps[name].items():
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/bulkdata/utils.py
```diff
@@ -85,6 +84,12 @@ def _parse_job_state(self, xml):
         elif "Failed" in statuses:
             return "Failed", state_messages
 
+        failures = tree.find(f".//{{{ns}}}numberRecordsFailed")
+        if failures is not None:
+            num_failures = int(failures.text)
+            if num_failures:
+                return "Failed", [f"Failures detected: {num_failures}"]
```
davidmreed: This appears to break existing functionality in `LoadData` that allows row-level errors to be ignored at user option via setting `ignore_row_errors` to `True`. I haven't done an E2E test though.

cumulusci/tasks/bulkdata/utils.py
```diff
@@ -101,6 +106,7 @@ def _wait_for_job(self, job_id):
         if result == "Failed":
             for state_message in messages:
                 self.logger.error(f"Batch failure message: {state_message}")
+            raise BulkDataException("Job Error", messages)
```
davidmreed: I still feel that the decision to respond to the job status logically belongs to the Task, not here in the mixin. When I call `_wait_for_job()`, I'd expect an exception to be thrown if an HTTP error ensues, but otherwise to be returned the status, whatever it is.

That is, at this layer, exception => "cannot determine state of job at the API level". At the Task layer, exception => "the operation did not succeed".

prescod: I've implemented a compromise in the code.

If the client is written naively (as DeleteData was) then it will throw an exception. If the caller wants to manage job state themselves, they can do that easily.

cumulusci/tasks/bulkdata/utils.py
prescod: I think I've addressed this now. `LoadData` still doesn't use the error messages returned from the lower level, but you would know better than I do whether the same errors will be discovered through another channel.

davisagli: @prescod you've managed to make it so that LoadData will actually process the results, and raise or not for record-level errors based on the ignore_row_errors setting. But you've changed what state _parse_job_state returns in the case where all batches completed but some records failed. Since it now returns "Failed" in that case instead of "Completed", this line will now raise an exception and prevent processing of other mappings: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/tasks/bulkdata/load.py#L86

The point of the ignore_row_errors=True option is to allow further mappings to be processed in the case where a small number of rows failed but most succeeded. I think I added this because I was doing a load where one object was prone to row lock errors (it was the load of GAU allocations where one of the GAUs had thousands of related allocations) and it was preferable to complete the data load with some records missing than to abort, because the whole thing took many hours to run.

I'd be willing to add a test for the correct behavior with regard to ignore_row_errors, but may not be able to get to it for a few days.

prescod: Half of the goal of the PR was to provoke conversation, because this behaviour is literally not backwards compatible in any context. For all I know, it will cause many of my Performance Tests which previously claimed to succeed to now claim to fail. It may be a messy process to determine which of the failures are acceptable and which are not.

I guess we should have states for Completed, Failed and CompletedWithErrors. Then the caller can decide how serious a CompletedWithErrors situation is.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -82,13 +82,13 @@ def _run_task(self):
 
             self.logger.info(f"Running Job: {name}")
             result = self._load_mapping(mapping)
-            if result != "Completed":
+            if not result.startswith("Completed"):
```
davisagli: Here and below, I think it'd be easier to read the intent of the code if we checked for specific values (e.g. `if result not in ("Completed", "CompletedWithFailures"):`) instead of checking for a substring

cumulusci/tasks/bulkdata/load.py
```diff
@@ -101,7 +101,9 @@ def _load_mapping(self, mapping):
             self._load_record_types([mapping["sf_object"]], conn)
         mapping["oid_as_pk"] = bool(mapping.get("fields", {}).get("Id"))
         job_id, local_ids_for_batch = self._create_job(mapping)
-        result = self._wait_for_job(job_id)
+        result = self._wait_for_job(job_id, error_behaviour="return")
+        if result == "CompletedWithFailures":
+            errors = self.error_messages  # noQA   # Todo: what to do with these errors?
```
davisagli: This is just the `f"Failures detected: {num_failures}"` message. It was already logged in _wait_for_job and I don't think there's a need to do anything more with it. The individual record errors will be found and reported while processing the job results.

davidmreed: The actual error handling (a bit counterintuitively) is in `_generate_results_id_map()`, which first consumes the downloaded results file.

prescod:
I guess the questions that are open in my mind are:

a) do we have to consider the impact on any classes other than ones in CCI itself?

b) do we want DeleteData to be permissive by default (backwards compatible) or crashing-by-default?

c) do we want any changes to the behaviour of LoadData at all?

davisagli:
@prescod 

> a) do we have to consider the impact on any classes other than ones in CCI itself?

I think it's unlikely that it'll matter (and we can mention the potential impact in release notes). Searching github, the only use of BulkJobTaskMixin outside of CumulusCI is in this task that @davidmreed worked on at the sprint, and it doesn't look like it'll be affected. https://github.com/SFDO-Community-Sprints/DataGenerationTool/blob/master/tasks/GenerateRecordPermutations.py

> b) do we want DeleteData to be permissive by default (backwards compatible) or crashing-by-default?

Reporting failures sounds better to me even if it's not backwards compatible. If you're using DeleteData you probably are aiming for the data to be gone and would want to know if it isn't.

> c) do we want any changes to the behaviour of LoadData at all?

I don't think so, since it already provides a way to control whether row-level failures should abort or not, via the ignore_row_errors option.

cumulusci/tasks/bulkdata/load.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1446/commits/c723246f05e15c48f39a03d1c0ea221716ada8d6#diff-a9551a5c355a86c23df29e1d9686489cL96

prescod:
@davisagli Once the merge tests complete I think that this is done.

PR #1450, opened 2020-01-10
more helpful?

Br4nd0R:
# Changes
* We now suggest how to show configured services when catching errors related to establishing a connection with GitHub.

### DIFF
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -51,6 +51,7 @@ def get_github_api_for_repo(keychain, owner, repo):
                 raise GithubException(
                     f"Could not access {owner}/{repo} using GitHub app. "
                     "Does the app need to be installed for this repository?"
+                    "\nUse the `cci service list` command to show what services are currently configured."
                 )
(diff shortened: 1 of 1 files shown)


Br4nd0R:
I'm could have read this scenario wrong. Let me know if I'm off base at all.

cumulusci/core/github.py
```diff
@@ -51,6 +51,7 @@ def get_github_api_for_repo(keychain, owner, repo):
                 raise GithubException(
                     f"Could not access {owner}/{repo} using GitHub app. "
                     "Does the app need to be installed for this repository?"
+                    "\nUse the `cci service list` command to show what services are currently configured."
```
davisagli: @Br4nd0R in this code path (starting line 44), CumulusCI is using a GitHub app configured using the GITHUB_APP_ID and GITHUB_APP_KEY environment variables, not an individual user's username and password configured in the github service.

Br4nd0R: Ahh, ya. Not applicable here. Will close this out.

PR #1451, opened 2020-01-10
Add check_only and test_level parameters to deploy task

### LINKED ISSUE #1066: Support "checkOnly" option for Deploy task
I see it is currently hard-coded to `false` in the soap envelope in `cumulusci.salesforce_api.metadata.ApiDeploy`. Could it be exposed as a parameter instead, and eventually made as an option for the `deploy` task?

davisagli:
This is based on @mrcmac's work and replaces #1066 with a branch within the SFDO-Tooling github organization so that the builds will run.

I also made some changes:
- Avoid changing the order of parameters for ApiDeploy.__init__ in case someone is passing the named arguments positionally.
- Provide the `test_level` option as a string instead of `run_tests` as a boolean, in order to better parallel testLevel in the actual API and support passing all of its options.
- Expect that check_only must be passed to ApiDeploy as a boolean, not a string.

@mrcmac please let me know if you have concerns about any of these changes.

# Critical Changes

# Changes

# Issues Closed
- Fixes #1066

### DIFF
--- a/docs/tasks.rst
+++ b/docs/tasks.rst
@@ -144,6 +144,8 @@ Options:
 * **namespace_inject**: If set, the namespace tokens in files and filenames are replaced with the namespace's prefix
 * **namespace_strip**: If set, all namespace prefixes for the namespace specified are stripped from files and filenames
 * **namespace_tokenize**: If set, all namespace prefixes for the namespace specified are replaced with tokens for use with namespace_inject
+* **check_only**: If True, performs a test deployment (validation) of components without saving the components in the target org
+* **run_tests**: If True, all tests from the package are run during the deploy (along with local tests in salesforce org)
 * **static_resource_path**: The path where decompressed static resources are stored.  Any subdirectories found will be zipped and added to the staticresources directory of the build.
 * **namespaced_org**: If True, the tokens %%%NAMESPACED_ORG%%% and ___NAMESPACED_ORG___ will get replaced with the namespace.  The default is false causing those tokens to get stripped and replaced with an empty string.  Set this if deploying to a namespaced scratch org or packaging org.
 * **clean_meta_xml**: Defaults to True which strips the <packageVersions/> element from all meta.xml files.  The packageVersion element gets added automatically by the target org and is set to whatever version is installed in the org.  To disable this, set this option to False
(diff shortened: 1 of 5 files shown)


docs/tasks.rst
```diff
@@ -144,6 +144,8 @@ Options:
 * **namespace_inject**: If set, the namespace tokens in files and filenames are replaced with the namespace's prefix
 * **namespace_strip**: If set, all namespace prefixes for the namespace specified are stripped from files and filenames
 * **namespace_tokenize**: If set, all namespace prefixes for the namespace specified are replaced with tokens for use with namespace_inject
+* **check_only**: If True, performs a test deployment (validation) of components without saving the components in the target org
+* **test_level**: Specifies which tests are run as part of a deployment.
```
mrcmac: I would consider adding supported values (as test level is an enum). The changed line would look like belo:\
`* **test_level**: Specifies which tests are run as part of a deployment (valid values: NoTestRun, RunLocalTests, RunAllTestsInOrg).`

I skipped `RunSpecifiedTests` intentionally because that option requires list of the tests to run.

mrcmac: Applies to all below.

mrcmac:
One note: this PR replaces #1447

davidmreed:
Looks great to me. I'm running an E2E test locally.

Any reason not to add support for `RunSpecifiedTests` while we're mucking around inside `deploy`? I'm happy to tackle that facet.

davisagli:
@davidmreed I'm fine with that.

mrcmac:
> Looks great to me. I'm running an E2E test locally.
> 
> Any reason not to add support for `RunSpecifiedTests` while we're mucking around inside `deploy`? I'm happy to tackle that facet.

To support `RunSpecifiedTests` user should have an option to pass the list of test classes which is currently not included in the deploy envelope - check `RunTests` deploy option https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_deploy.htm

Initially I wanted to limit the scope of changes but of course that can be added.

davidmreed:
Adding RunSpecifiedTests support in child branch, #1453

PR #1455, opened 2020-01-11
Support Serial mode for bulk data loads

davidmreed:
# Critical Changes

# Changes
- Bulk data mappings now allow `bulk_mode: Serial` to select serial mode.

# Issues Closed


prescod:
Would you be open to an amendment (coded by you or me) that allows it to be set at a task level? I'd propose that if the task says serial then it overrides the step. If the step says serial, then it overrides the task. If they both are silent, it means parallel.

In my use-cases, I generate mapping files and it gets messy to try to find an object-by-object place to put this configuration.

Also: my understanding is that the locking errors only become a problem when you want to upload more than 10,000 of some row. So the serial/parallel mode may be more a property of the specific data than of the data shape. I imagine that my test cases would turn on serial mode for large uploads and turn it off for small ones.

davidmreed:
> Would you be open to an amendment (coded by you or me) that allows it to be set at a task level? I'd propose that if the task says serial then it overrides the step. If the step says serial, then it overrides the task. If they both are silent, it means parallel.

Yes, I'm fine with having it available at the task level. I would suggest that the lowest-level setting takes precedence for that step (if explicitly set).

> Also: my understanding is that the locking errors only become a problem when you want to upload more than 10,000 of some row. So the serial/parallel mode may be more a property of the specific data than of the data shape. I imagine that my test cases would turn on serial mode for large uploads and turn it off for small ones.

Strictly speaking, the behavior changes when you hit the batch size, which is pinned to 10,000 (the maximum). If we made that configurable in the future it would change.

Below BATCHSIZE rows, you can set either mode and it'll behave the same.

davidmreed:
@prescod Added Serial mode at the task level, with the lowest-level setting (step > task > default) taking precedence. How's that look?

PR #1456, opened 2020-01-11
Make UpdateAdminProfile deploy profiles only

jstvz:
Prerelease [build failure](https://sfdo-metaci.herokuapp.com/builds/165142) when running `update_admin_profile`:

```
MetadataParseError: Could not process MDAPI response: Update of CustomObject Contact: Error on line 7891, col 20: No more than 10 columns may be specified in lookupPhoneDialogsAdditionalFields
```

This commit changes `UpdateAdminProfile` so that it only deploys the modified Admin profile. While it is necessary to retrieve profiles along their associated metadata objects, we don't need to do that for deployments.

----

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/UpdateAdminProfile.py
+++ b/cumulusci/tasks/salesforce/UpdateAdminProfile.py
@@ -1,5 +1,4 @@
 import os
-import shutil
 import tempfile
 
 from pathlib import Path
@@ -56,11 +55,19 @@ class UpdateAdminProfile(Deploy):
         }
 
     def _run_task(self):
-        self.tempdir = tempfile.mkdtemp()
-        self._retrieve_unpackaged()
-        self._process_metadata()
-        self._deploy_metadata()
-        shutil.rmtree(self.tempdir)
+        self.retrieve_dir = None
+        self.deploy_dir = None
+        with tempfile.TemporaryDirectory() as tempdir:
+            self._create_directories(tempdir)
+            self._retrieve_unpackaged()
+            self._process_metadata()
+            self._deploy_metadata()
+
+    def _create_directories(self, tempdir):
+        self.retrieve_dir = Path(tempdir, "retrieve")
+        self.deploy_dir = Path(tempdir, "deploy")
+        self.retrieve_dir.mkdir()
+        self.deploy_dir.mkdir()
 
     def _retrieve_unpackaged(self):
         path = self.options.get("package_xml") or os.path.join(
@@ -83,11 +90,11 @@ class UpdateAdminProfile(Deploy):
             self.project_config.project__package__api_version,
         )
         unpackaged = api_retrieve()
-        unpackaged.extractall(self.tempdir)
+        unpackaged.extractall(self.retrieve_dir)
 
     def _process_metadata(self):
-        self.logger.info("Processing retrieved metadata in {}".format(self.tempdir))
-        path = os.path.join(self.tempdir, "profiles", "Admin.profile")
+        self.logger.info(f"Processing retrieved metadata in {self.retrieve_dir}")
+        path = self.retrieve_dir / "profiles" / "Admin.profile"
         self.tree = elementtree_parse_file(path)
 
         self._set_apps_visible()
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/UpdateAdminProfile.py
```diff
@@ -175,6 +176,19 @@ def _set_tabs_visibility(self):
             elem.find("sf:visibility", self.namespaces).text = "DefaultOn"
 
     def _deploy_metadata(self):
-        self.logger.info("Deploying updated Admin.profile from {}".format(self.tempdir))
-        api = self._get_api(path=self.tempdir)
+        deploy_dir = tempfile.mkdtemp()
```
davidmreed: Don't we need to clean up `deploy_dir` after deployment?

davisagli: Actually we already are working in self.tempdir; it doesn't look to me like there's a need to create another temporary directory.

We could handle the self.tempdir cleanup better in _run_task though, by using tempfile.TemporaryDirectory as a context manager.

cumulusci/tasks/salesforce/tests/test_UpdateAdminProfile.py
```diff
@@ -83,67 +85,68 @@
...
-        def _check_result():
-            with open(
-                os.path.join(task.tempdir, "profiles", "Admin.profile"), "r"
-            ) as f:
-                result = f.read()
-            self.assertMultiLineEqual(ADMIN_PROFILE_EXPECTED, result)
-
-        task._retrieve_unpackaged = _retrieve_unpackaged
-        task._deploy_metadata = _check_result
+def test_run_task():
```
davisagli: @jstvz is there any particular benefit to getting rid of the test class? I'm all for using pytest assertions instead of unittest, but I still find it kind of useful organizationally to keep the tests for methods of the same class grouped within one test class. Probably mostly a matter of personal preference though.

jstvz: I take your point, but I'm of the opinion the class isn't providing any additional organization since there's a 1:1 correspondence between the class under test and the module. 

In this case, the main reason for getting rid of `unittest.TestCase` is that it allows using pytest fixtures (`tmpdir`) and parameterization.

cumulusci/tasks/salesforce/UpdateAdminProfile.py
jstvz: Looks like the consensus is that my general approach makes sense. I can take a second pass at making it a bit smarter about how we're handling the temporary directory and re-request review.

cumulusci/tasks/salesforce/UpdateAdminProfile.py
```diff
@@ -176,19 +183,18 @@ def _set_tabs_visibility(self):
...
         <types><members>Admin</members><name>Profile</name></types><version>39.0</version></Package>
         """
         )
 
-        source_profile_dir = Path(self.tempdir, "profiles")
-        target_profile_dir = Path(deploy_dir, "profiles")
-        source_profile_dir.replace(target_profile_dir)
+        retreived_profile_dir = Path(self.retrieve_dir, "profiles")
+        target_profile_dir = Path(self.deploy_dir, "profiles")
+        retreived_profile_dir.replace(target_profile_dir)
```
davisagli: retrieved

I don't really understand the reason for using separate directories for retrieval and deployment. We can have a single temporary working directory where we retrieve, then overwrite package.xml and then deploy just the profile, no?

jstvz: > retrieved

Thanks. Right up there with "receipt" among my most misspelled words. 

I misread your comment above, I think. My rationale is that we shouldn't send extraneous data over the network if we can help it. I haven't done any testing of deployment times when sending extra data, so I can't say whether it's any slower. I think it's worth the extra ~20 LoC, but I'm receptive to counterarguments.

davisagli: @jstvz oh right, I hadn't considered that our deploy task is naive and will include everything in the folder even if it's not in package.xml. okay, you convinced me.

jstvz: Okay. Went back and tested this for my own edification:
```
Update of CustomField Contact.Phone: Error: Not in package.xml
```

PR #1464, opened 2020-01-14
Feature/robot spring 20

boakley:
This is a collection of small fixes to address issues when running robot tests against Spring '20. The 
tests seem to mostly work for me locally, though I am still getting more `StaleElementReferenceException` failures in Spring '20 than Winter '20. I currently am thinking of a way to fix that, but that fix may be in a later changeset since QEs are blocked by some things this PR addresses.

Roughly speaking, this PR addresses the following broad issues:

1) some of the DOM has changed significantly, requiring improved locators
2) more of the DOM lives in shadow elements which are hard to click on, so some of our keywords now use javascript to click on elements
3) The app menu button now has a different behavior - it shows a menu rather than immediately opening the app modal
4) more processing seems to be happening when pages are loaded, causing our tests to jump the gun, so to speak. I've added a couple sleeps to minimize the problem, but I'll want to revisit this later to see if I can find a better solution. 
5) a couple new tests have been added, but our coverage of keywords still needs improvement.

I'm seeing more flakiness on circleci, so I've added a little extra debugging information in a few spots.

For background on the switch to using javascript for some of the clicks, see the knowledge article [Fixing Automated UI Tests in Spring '20](https://help.salesforce.com/articleView?id=000352057&language=en_US&mode=1&type=1)

There is still at least one known issue yet to be fixed, and that is with the `Populate lookup field` keyword. I've got it so that it's able to click the field and wait for the popup, but it's still not able to click an item on the menu. 
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -783,6 +783,15 @@ class Salesforce(object):
             self.builtin.log("waiting for a refresh...", "DEBUG")
             self.selenium.capture_page_screenshot()
             time.sleep(interval)
+            location = self.selenium.get_location()
+            if "//test.salesforce.com" in location:
+                # did we land on https://test.salesforce.com? If so, maybe
+                # the authentication succeeded so let's try the url without
+                # the frontdoor servelet
+                login_url = self.cumulusci.org.config["instance_url"]
+                self.builtin.log(
+                    f"setting login_url temporarily to {login_url}", "DEBUG"
+                )
             self.selenium.go_to(login_url)
 
     def breakpoint(self):
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -12,7 +12,7 @@ from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.common.action_chains import ActionChains
 
 from simple_salesforce import SalesforceResourceNotFound
-from cumulusci.robotframework.utils import selenium_retry
+from cumulusci.robotframework.utils import selenium_retry, capture_screenshot_on_error
 from SeleniumLibrary.errors import ElementNotFound, NoOpenBrowser
 from urllib3.exceptions import ProtocolError
 
@@ -178,14 +178,19 @@ class Salesforce(object):
         self._jsclick(locator)
         self.wait_until_modal_is_open()
 
+    @capture_screenshot_on_error
     def click_related_item_link(self, heading, title):
         """Clicks a link in the related list with the specified heading.
 
         This keyword will automatically call `Wait until loading is complete`
         """
+        self.builtin.set_log_level("DEBUG")
+        self.builtin.log("loading related list...", "DEBUG")
         self.load_related_list(heading)
         locator = lex_locators["record"]["related"]["link"].format(heading, title)
+        self.builtin.log("clicking...", "DEBUG")
         self._jsclick(locator)
+        self.builtin.log("waiting...", "DEBUG")
         self.wait_until_loading_is_complete()
 
     def click_related_item_popup_link(self, heading, title, link):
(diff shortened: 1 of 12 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -97,6 +97,22 @@ def initialize_location_strategies(self):
...
+
+        self.selenium.wait_until_page_contains_element(locator)
+        element = self.selenium.get_webelement(locator)
+        # Setting the focus first seems to be required as of Spring'20
+        # (read: without it, tests started failing in that release). I
+        # suspect it's because there is a focusOut handler on form
+        # fields which need to be triggered for data to be accepted.
+        self.selenium.driver.execute_script(
+            "arguments[0].focus(); arguments[0].click()", element
+        )
```
davisagli: Did you consider monkeypatching selenium library's click command to do this? Or are there still cases where we prefer to do a standard click?

cumulusci/robotframework/Salesforce.py
```diff
@@ -695,6 +724,12 @@ def wait_until_loading_is_complete(self, locator=None):
         try:
             self.selenium.wait_until_page_contains_element(locator)
             self.wait_for_aura()
+            # this knowledge article recommends waiting a second. I don't
+            # like it, but it seems to help. We should do a wait instead,
+            # but I can't figure out what to wait on.
```
davisagli: wait_for_aura should already be waiting for Aura ajax calls to complete (unless they changed the interface for that). I suppose there's still potential for a delay between when they are done and when things have rendered. I wish they exposed a way to get a callback when the event loop is idle.

cumulusci/robotframework/Salesforce.py
```diff
@@ -757,6 +792,15 @@ def wait_until_salesforce_is_ready(self, locator=None, timeout=None, interval=5)
             self.builtin.log("waiting for a refresh...", "DEBUG")
             self.selenium.capture_page_screenshot()
             time.sleep(interval)
+            location = self.selenium.get_location()
+            if "//test.salesforce.com" in location:
+                # did we land on https://test.salesforce.com? If so, maybe
```
davisagli: Note that it would probably be login.salesforce.com if running tests against an org that's not on a sandbox instance.

cumulusci/robotframework/Salesforce.py
boakley: I considered it, but it seemed a bit risky to make the change everywhere. I'm open to the idea, but I'm not a fan of monkey-patching selenium.

cumulusci/robotframework/Salesforce.py
boakley: From experimentation, it seems like something is still tweaking the DOM after wait_for_aura returns.

cumulusci/robotframework/Salesforce.py
boakley: Thanks for the heads up.  This was initially a fix for the fact my machine was _always_ hitting the redirect Monday afternoon. I'll add a check for either 'login.salesforce.com' or 'test.salesforce.com'

cumulusci/robotframework/Salesforce.py
davisagli: @boakley yeah, that's fair, especially since this does more than just clicking. It seems like ultimately the ability to interact with shadow DOM elements without resorting to custom javascript is something that belongs in webdriver/selenium, though -- but maybe LWC's efforts to isolate components get in the way.

cumulusci/robotframework/Salesforce.py
davisagli: I wonder if https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback would be helpful

boakley: Hmmmm.... interesting idea.  Perhaps we could set a flag in an idle callback, and then our keywords could wait for that flag to be set. That might work, unless the work that the LWCs are doing is also scheduled with that function. 

I don't think I'll try to do that in this PR in the interest of time, but it's definitely an idea worth experimenting on.

boakley:
I'm marking this as ready because there are some fixes for some blockers that some QEs are having. There is still one more keyword I know I need to fix, but we really need to get some of these changes in their hands.

PR #1465, opened 2020-01-14
Make the options case-insensitive.

prescod:
Make the task_mode ("Serial" or "Parallel") case-insensitive. Also I fixed a small bug in one of your tests.

### DIFF
--- a/cumulusci/tasks/bulkdata/tests/test_bulkdata.py
+++ b/cumulusci/tasks/bulkdata/tests/test_bulkdata.py
@@ -285,7 +285,7 @@ class TestLoadDataWithSFIds(unittest.TestCase):
         task._create_job(mapping)
 
         task.bulk.create_update_job.assert_called_once_with(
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/bulkdata/tests/test_bulkdata.py
```diff
@@ -285,7 +285,7 @@ def test_create_job__serial_task_level(self):
         task._create_job(mapping)
 
         task.bulk.create_update_job.assert_called_once_with(
-            "Account", contentType="CSV", concurrency="Serial"
+            "Account", contentType="CSV", concurrency="serial"
```
davisagli: @prescod did you actually test it with the lowercase value? The documentation mentions the concurrencyMode options with initial caps, and I don't know if they are case insensitive in the API or not. https://developer.salesforce.com/docs/atlas.en-us.api_asynch.meta/api_asynch/asynch_api_reference_jobinfo.htm

PR #1476, opened 2020-01-16
Gist fixes

### LINKED ISSUE #1474: Don't suggest `cci gist` command if `cci gist` fails.

### LINKED ISSUE #1475: Remove reference to `os.uname` in `cci gist` command
It's not available on Windows.

Br4nd0R:
# Issues Closed
* Only suggest `cci gist` command we that command wasn't the last one run. (Fixes #1474) 
* Use `platform.uname` instead of `os.uname` (Fixes #1475)

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -6,6 +6,7 @@ import functools
 import json
 import re
 import os
+import platform
 import pdb
 import shutil
 import sys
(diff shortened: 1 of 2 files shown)


Br4nd0R:
Thinking through this more last night, we should have a test to assert that `cci.SUGGEST_GIT_GIST_COMMAND` is displayed for a non-gist cci command, and isn't displayed for `cci gist`.

PR #1479, opened 2020-01-18
Add ignore_types option to uninstall_packaged_incremental

jstvz:
This commit exposes the `self.ignore_types` option as a task option. As shown by the included test change, this is primarily intended to allow entire metadata types to be ignored by this task. 

The next logical step would be to allow wildcard (`*`) matches against component names to allow different translations of the same object to be ignored without listing each component individually.

# Critical Changes

# Changes
- We've added the option `ignore_types` to the `uninstall_packaged_incremental` task to allow all components of the specified metadata type to be ignored without having to explicitly list each one.

# Issues Closed


cumulusci/tasks/salesforce/UninstallPackagedIncremental.py
```diff
@@ -36,7 +40,10 @@ def _init_options(self, kwargs):
         self.options["purge_on_delete"] = process_bool_arg(
             self.options.get("purge_on_delete", True)
         )
-        self.options["ignore"] = self.options.get("ignore") or {}
+        self.options["ignore"] = self.options.get("ignore", {})
```
davisagli: I prefer `self.options.get("ignore") or {}` because it handles the case where the key was explicitly set to None.

jstvz: 👨‍🎓 👍

PR #1489, opened 2020-01-21
Refactor API access for Bulk Data tasks

davidmreed:
# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +41,46 @@ def _init_options(self, kwargs):
...
+            query = f"SELECT Id FROM {obj}"
+            if self.options["where"]:
+                query += f" WHERE {self.options['where']}"
+
+            if not self.options["where"]:
+                # FIXME: Perform a count() query to determine whether we have work to do.
+                pass
+
+            qs = BulkApiQueryStep(obj, {}, self, query)
+            qs.query()
```
prescod: In my quick testing, qs.status is not set when qs.query runs.

Also, it is unclear to me how the actual query results get fed into the delete step below.

davidmreed: I'll check on the status update issue.

The records form query are fed in at line 75. The query's result generator becomes the delete's input generator (through a `map()` to change its structure).

prescod: Okay I see it now. I was imagining the stream being submitted in the step constructor.

Is it important that it be exposed to the caller as a 4-step process? construct, start, feed, end? 

In the two places its used, the steps seem identical and therefore could perhaps be hidden in a 1-step or 2-step process?

cumulusci/tasks/bulkdata/step.py
```diff
@@ -0,0 +1,256 @@
...
+
+
+Result = namedtuple("Result", ["id", "success", "error"])
+
+
+@contextmanager
+def download_file(uri, bulk_api):
+    """Download the bulk API result file for a single batch"""
+    (handle, path) = tempfile.mkstemp(text=False)
+    resp = requests.get(uri, headers=bulk_api.headers(), stream=True)
```
prescod: Please add

`resp.raise_for_status()`

cumulusci/tasks/bulkdata/step.py
```diff
@@ -0,0 +1,256 @@
...
+    with open(path, "r") as f:
+        yield f
+
+    pathlib.Path(path).unlink()
+
+
+class BulkJobTaskMixin:
+    def _job_state_from_batches(self, job_id):
+        uri = f"{self.bulk.endpoint}/job/{job_id}/batch"
+        response = requests.get(uri, headers=self.bulk.headers())
```
prescod: Please add

`response.raise_for_status()`

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +41,46 @@ def _init_options(self, kwargs):
...
-            query += f" WHERE {where}"
+            query = f"SELECT Id FROM {obj}"
+            if self.options["where"]:
+                query += f" WHERE {self.options['where']}"
+
+            if not self.options["where"]:
+                # FIXME: Perform a count() query to determine whether we have work to do.
+                pass
+
+            qs = BulkApiQueryStep(obj, {}, self, query)
```
prescod: Would it be virtuous and helpful to have these be Tasks? We should be conservative about adding new abstractions (but of course should do so when it is helpful).

davidmreed: My design goal was to isolate the API access layer to act as sort of Bulk API primitives that are easily composable. It's expected that individual tasks will use multiple `Step` instances in many use cases - for example, `DeleteData` uses both a `QueryStep` and a `DmlStep`. for that reason and because the steps do not have a concept of a storage layer, I think they belong as a separate abstraction, but I'm open to pushback and discussion on that.

cumulusci/tasks/bulkdata/delete.py
davidmreed: I fixed the issue with `status` not being set. It turns out we have all along been silently ignoring failures at the query batch level, which I've fixed.

As to the 4-step process: I'm open to condensing this. Mostly, I think that would look like moving the logic of `start()` and `end()` into `load_records()`. But I liked keeping the option open for `load_records()` to be called multiple times to help ensure it's backend-storage-agnostic.

cumulusci/tasks/bulkdata/delete.py
prescod: I like the isolation and code reuse. I'm in favour of that. I'm suggesting that the coupling be even looser than it is, rather than tighter.

I'm suggesting roughly that instead of

            qs = BulkApiQueryStep(obj, {}, self, query)

It would delegate to a subtask more like this:

            BulkApiQueryTask({output_file=my_csv_tempfile_path, query=query})

...

           BulkApiDmltask({operation = Operation.HARD_DELETE, id_csv=my_csv_tempfile_path})

Although the actual syntax might not be so clean unless we make a special helper for spawning sub-tasks. Which we probably could and should. I could have used it here:

https://github.com/SFDO-Tooling/CumulusCI/blob/4cf572a6703b7c3244bfd7514ad2efc24b1e5162/cumulusci/tasks/bulkdata/generate_and_load_data.py#L144

That would mean, among other things, that I could generate my own QueryResults file on the command line (using whatever complicated Python code or spreadsheet hackery I want) and then I could also pass it to a DmlStep to do the actual deleting (or Loading, or Updating). This would give CCI a user-accessible mechanism for deleting a list of IDs (wherever they came from) and *on top of that* a mechanism for deleting from a query, if the list of objects I want to delete can be expressed easily that way. (just this week I ran into a case where I had to write custom Python code to delete a list of objects that was not easily expressible as a simple query result).

Another thing it would mean is that the end user could try and redo a DmlTask if it fails (as long as they have a way of generating the list of records to delete, insert or update).

cumulusci/tasks/bulkdata/delete.py
prescod: Yes we have a lot of silently ignored failures to find and replace!

I didn't understand your rationale: "I liked keeping the option open for load_records() to be called multiple times to help ensure it's backend-storage-agnostic."

What is the use-case for calling it multiple times, what is the behaviour if you call it multiple times and how does that behaviour make it more storage-agnostic?

cumulusci/tasks/bulkdata/delete.py
davidmreed: I would prefer to build that functionality as a Task layer on top of these classes. As implemented here, they know nothing about input and output - they just consume a generator and produce another generator of results. I'd like to separate the logic that relates to storage and persistence.

prescod: Okay that works too.

cumulusci/tasks/bulkdata/delete.py
davidmreed: Well, my thought was that there could occur situations where we wanted to pull out multiple subsets of data from a data store, or source data in batches from the underlying data store, to upload in a single job. 

If we separate out the initiation of a job, loading of records into a job, and closure of a job, we aren't bound to the assumption that we have to have all of the records present at the point that the first call to `load_records()` occurs. 

In the current implementation, if you call it multiple times, you just get multiple batches in the same job (each of will will be further chunked based on the maximum batch size).

cumulusci/tasks/bulkdata/step.py
```diff
@@ -0,0 +1,246 @@
...
+        result = self._wait_for_job(self.job_id)
+        if result == "Completed":
+            self.status = Status.SUCCESS
+        else:
+            self.status = Status.FAILURE
+
+    def load_records(self, records):
+        self.batch_ids = []
+
+        for count, batch_file in enumerate(self._batch(records)):
```
davisagli: It's not really a batch_file now; it's a generator of CSV lines, right?

cumulusci/tasks/bulkdata/step.py
```diff
@@ -0,0 +1,246 @@
...
+
+        content.seek(0)
+        yield content.read().encode("utf-8")
+        for rec in records:
+            content = io.StringIO()
+            writer = csv.writer(content)
+            writer.writerow(rec)
+            content.seek(0)
+
+            yield content.read().encode("utf-8")
```
davisagli: It seems like there should be a way to do this without constructing a new csv writer for each row.

cumulusci/tasks/bulkdata/tests/test_step.py
```diff
@@ -0,0 +1,462 @@
...
+    <stateMessage>{first_message}</stateMessage>
+</batch>
+<batch>
+    <state>{second_state}</state>
+    <stateMessage>{second_message}</stateMessage>
+</batch>
+</root>"""
+
+
+class test_download_file(unittest.TestCase):
```
davisagli: Can you stick with camel case for class names, please?

cumulusci/tasks/bulkdata/extract.py
```diff
@@ -91,90 +89,81 @@ def _init_mapping(self):
...
+            if column:
+                columns.append(column)
+
         if not columns:
             return
         record_type = mapping.get("record_type")
         if record_type:
             columns.append("record_type")
 
-        processor = log_progress(
```
davisagli: Looks like we've lost the logging of progress as records are extracted.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -86,19 +79,21 @@ def _run_task(self):
...
                 continue
+
             started = True
 
-            self.logger.info(f"Running Job: {name}")
+            self.logger.info(f"Running step: {name}")
             result = self._load_mapping(mapping)
-            if result not in ("Completed", "CompletedWithFailures"):
-                raise BulkDataException(f"Job {name} did not complete successfully")
+            if result is Status.FAILURE:
```
davisagli: Are you intentionally switching to always considering row-level errors as a failure of the entire task regardless of the ignore_row_errors option?

cumulusci/tasks/bulkdata/extract.py
```diff
@@ -91,90 +89,81 @@ def _init_mapping(self):
...
+                    self._sql_bulk_insert_from_records(
+                        conn, mapping["table"], columns, csv.reader(data_file_values)
                     )
-                    self._sql_bulk_insert_from_csv(
-                        conn, mapping["sf_id_table"], ["sf_id"], data_file_ids
+                    self._sql_bulk_insert_from_records(
+                        conn,
+                        mapping["sf_id_table"],
+                        ["sf_id"],
+                        csv.reader(data_file_ids),
```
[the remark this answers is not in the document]
davidmreed: I have not benchmarked the change.

More generally, I'm very eager to decouple as much as possible from the details of underlying database storage. The complexity of the logic around database interaction makes it hard to significantly extend the Bulk Data toolchain in the directions we want to go.

In specific, I asked @prescod if he'd ever used that functionality and he said no, but that's perhaps not enough to justify its removal. More broadly, with Paul's work on Snowfakery in place, do we have a use case for storing a static LDV dataset in Postgres? I believe only NPSP's original LDV data set is stored that way at present, and I don't believe the Postgres database has been used since I started at SFDO.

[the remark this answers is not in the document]
davisagli: I agree that decoupling logic from the underlying data storage is a good goal -- I just want to make sure we're considering the implications. It sounds like removing the special postgres handling won't affect how we are doing things these days, but let me fill in a bit more historic context.

In late 2018 when I started to use Jason's original bulkdata tasks to load LDV test data for testing DSO, performance was an issue. This was true for both phases of the process: generating data in the local db, and loading it from there into Salesforce. When I profiled the latter in particular, I found that uploading the batches was quick, processing the batches was out of our control (aside from switching to parallel batches) and downloading the results was quick, but loading the result ids into the local db was slow. One change I made as a result was to put the ids in a separate table -- this allowed us to do an insert instead of update, which helped a lot. I also did quite a bit of experimentation to figure out what the fastest way was to get the ids inserted, and that's what led to using postgres rather than sqlite and using postgres' COPY FROM feature rather than sqlalchemy to load the ids.

What I'm trying to say is that I'll sign off on this refactoring even if we're losing the special postgres support, but:
1. I want to make sure Paul's aware that there's probably better performance available for loading result ids with an approach like this.
2. I would like to see a benchmark comparison for loading a significant-sized dataset before and after these changes, to make sure we're not introducing significant regressions that are unrelated to the special postgres support.

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +41,42 @@ def _init_options(self, kwargs):
...
+                obj,
+                Operation.HARD_DELETE
+                if self.options["hardDelete"]
+                else Operation.DELETE,
+                {},
+                self,
+                ["Id"],
+            )
+            ds.start()
+            ds.load_records(map(lambda result: [result[0]], qs.get_results()))
```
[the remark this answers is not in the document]
davidmreed: I think this was held over from an earlier implementation and is no longer needed. Simplifying to `ds.load_records(qs.get_results())`

cumulusci/tasks/bulkdata/extract.py
```diff
@@ -1,5 +1,6 @@
+import csv
 import tempfile
-import unicodecsv
```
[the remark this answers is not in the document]
davidmreed: Looks like it is only used now in `pushfails.py`, which is one of the EREs' collective next targets for reworking.

cumulusci/tasks/bulkdata/load.py
davidmreed: Nope. I missed this in doing a merge. Having different error semantics here and in the mixin is confusing; I will rework it to use an enum throughout and consistent logic.

cumulusci/tasks/bulkdata/step.py
davidmreed: Correct. Renaming for clarity.

cumulusci/tasks/bulkdata/utils.py
```diff
@@ -66,82 +47,13 @@ def setup_epoch(inspector, table, column_info):
...
-            table = self.metadata.tables[table]
-            rows = list(reader)
-            if rows:
-                conn.execute(table.insert().values(rows))
+class SqlAlchemyMixin:
+    def _sql_bulk_insert_from_records(self, conn, table, columns, records):
+        table = self.metadata.tables[table]
+
+        for batch in BatchIterator(records, n=100):
+            conn.execute(table.insert(), [dict(zip(columns, row)) for row in batch])
```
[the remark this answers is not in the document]
davidmreed: This was Jason's solution to get EDAEASY to work at all (I had a similar problem with NPSP's new data set). 

It seems reasonable to make it conditional on the underlying database type, but I don't know the field of options well at all. Are you aware of any other backends that require the same workaround?

cumulusci/tasks/bulkdata/delete.py
davidmreed: I've added context manager support for both query and DML operations.

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +41,42 @@ def _init_options(self, kwargs):
...
-        # Query for rows to delete
-        delete_rows = self._query_salesforce_for_records_to_delete(obj, where)
-        if not delete_rows:
-            self.logger.info(f"  No {obj} objects found, skipping delete")
-            return
+            self.logger.info(f"Querying for {obj} objects")
+            qs = BulkApiQueryStep(obj, {}, self, query)
+            qs.query()
+            if qs.status is not Status.SUCCESS:
+                raise BulkDataException(f"Unable to query records for {obj}")
```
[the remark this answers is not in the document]
davidmreed: I'm going to aim at addressing both of these thoughts by reworking how we track job-level results in general.

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +43,57 @@ def _init_options(self, kwargs):
...
-            query += f" WHERE {where}"
+            query = f"SELECT Id FROM {obj}"
+            if self.options["where"]:
+                query += f" WHERE {self.options['where']}"
+
+            self.logger.info(f"Querying for {obj} objects")
+            qs = BulkApiQueryOperation(obj, {}, self, query)
+            qs.query()
+            if qs.job_result.status is not DataOperationStatus.SUCCESS:
+                raise BulkDataException(
```
davidmreed: @prescod Particularly want your thoughts on these ~20 lines - I altered Delete Data to make it throw an exception if the query fails, do nothing if the query returns no results, and throw an exception if the delete job fails or if it has a row-level error and `ignore_row_errors` is unset.

cumulusci/tasks/bulkdata/extract.py
davidmreed: Restoring it, although I think when we add REST API support progress logging needs to be handled differently.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -107,56 +102,39 @@ def _load_mapping(self, mapping):
...
         statics = self._get_statics(mapping)
         query = self._query_db(mapping)
 
         total_rows = 0
-        batch_num = 1
-
-        batch_file, writer, batch_ids = self._start_batch(columns)
-        for row in query.yield_per(batch_size):
-            total_rows += 1
+        self.local_ids = []
```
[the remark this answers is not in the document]
davidmreed: Done!

cumulusci/tasks/bulkdata/step.py
davidmreed: I did some experiments and sure enough, it's considerably faster to reuse the `csv.writer` and `StringIO` instances (truncating the `StringIO` each row) than it is to recreate them. Fixing!

cumulusci/tasks/bulkdata/step.py
```diff
@@ -0,0 +1,246 @@
...
+import tempfile
+import time
+import xml.etree.ElementTree as ET
+
+from collections import namedtuple
+from contextlib import contextmanager
+from enum import Enum
+from cumulusci.tasks.bulkdata.utils import BatchIterator
+from cumulusci.core.exceptions import BulkDataException
+from cumulusci.core.utils import process_bool_arg
```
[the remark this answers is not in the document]
jstvz: I'd like to put in a plug for [`isort`](https://github.com/timothycrosley/isort), which follows the [same pattern](https://github.com/timothycrosley/isort#how-does-isort-work).

I would suggest adding a pre-commit check for it, but it sometimes conflicts with `black` and I haven't had the time to get them to play together nicely.

cumulusci/tasks/bulkdata/utils.py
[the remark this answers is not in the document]
davisagli: Not sure about other databases. For SQLite I think it actually depends on a compile-time setting :-/ https://stackoverflow.com/a/15313495

cumulusci/tasks/bulkdata/step.py
```diff
@@ -287,16 +287,16 @@ def _csv_generator(self, records):
         content = io.StringIO()
         writer = csv.writer(content)
         writer.writerow(self.fields)
-
         content.seek(0)
         yield content.read().encode("utf-8")
+        content.truncate(0)
```
davisagli: @davidmreed TIL about truncate :)

cumulusci/tasks/bulkdata/extract.py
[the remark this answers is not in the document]
prescod: I do have a long-range goal of storing some large data sets in cloud-hosted Postgres some day: partially because of performance and partially because of repeatability. I don't have first-hand experience with the performance implications of the optimizations and its hard for me to have a strong opinion until one of us has a chance to do before and after tests. My most recent experiences would tend to colour my opinion towards "Salesforce is 99% of the problem and SQL is 1%" but in a situation with more reasonable data skew or some other workaround, that might not be true anymore. Salesforce is multi-threaded and this code is single-threaded, so Salesforce import might actually be more reasonable if I could use it as it was designed to be used.

I can imagine an abstraction that would allow the caller to detect that the recipient prefers CSV and it would serve CSV (not unlike content-negotiation on the Web). You'd need a Postgres sub-class of the SQL Alchemy mixin, I guess (or some smarts in the mixin analogous to the old code).

Maybe it's not much more crazy than

if self._storage_prefersCSV():
   self._import_from_CSV()
else:
   self._import_from_records()

But on the other hand, I don't totally understand what (present or future) refactorings are made difficult by having CSV as the standardized interchange format so perhaps I'm blocking the same refactorings by making it an option.

But if David Reed sees it as too messy then I can treat it as a "future optimization opportunity" and deal with it when the time comes.

cumulusci/tasks/bulkdata/utils.py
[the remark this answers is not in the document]
prescod: Why is it 100 in particular? The links all say 999 is the max?

In general there is something deeply non-intuitive about all of this string parsing and generating in a single process, but that's mostly independent of David's refactoring. It's just literally the way that SQL Alchemy and SQLite work. It's almost as weird in Snowfakery. Generating strings for talking to an in-process library is kind of weird. Being limited to 999 inserts at the same time is even weirder.

cumulusci/tasks/bulkdata/step.py
davidmreed: Stack Overflow scared me into believing it was not performant. Lesson learned - Stack Overflow lies; do the benchmarking yourself.

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +41,42 @@ def _init_options(self, kwargs):
...
+                query += f" WHERE {self.options['where']}"
 
-    def _create_job(self, obj, where=None):
-        # Query for rows to delete
-        delete_rows = self._query_salesforce_for_records_to_delete(obj, where)
-        if not delete_rows:
-            self.logger.info(f"  No {obj} objects found, skipping delete")
-            return
+            self.logger.info(f"Querying for {obj} objects")
+            qs = BulkApiQueryStep(obj, {}, self, query)
```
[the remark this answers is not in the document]
davidmreed: I've made all of the operation classes (renamed from Step!) use mandatory keyword arguments.

cumulusci/tasks/bulkdata/delete.py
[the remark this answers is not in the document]
davidmreed: I've tried to solve this by adding a `DataOperationJobResult` object to wrap the status, associated errors, and total row-level completed/failed counts. This facilitates surfacing the errors, but also makes it easier for consumers to react to the record counts by (e.g.) not running a Delete operation if the Query didn't return anything.

cumulusci/tasks/bulkdata/extract.py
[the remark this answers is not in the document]
davidmreed: Memorializing our conversation this morning: I'm going to add the original Postgres optimization back in (dumping the record generator to a CSV file first), and we'll see if we can benchmark that.

cumulusci/tasks/bulkdata/utils.py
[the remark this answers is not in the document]
davidmreed: I think this should be refined, but I'm not 100% sure how to trigger the error in the first place, or how to reason about it in the code. We set the limit at 100 _records_, arbitrarily, because that's what Jason found by experiment worked on EDAEASY. The SQLite limit I believe is on variables in the INSERT statement.

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -38,77 +43,61 @@ def _init_options(self, kwargs):
...
+            if ds.job_result.status not in [
+                DataOperationStatus.SUCCESS,
+                DataOperationStatus.ROW_FAILURE,
+            ]:
+                raise BulkDataException(
+                    f"Unable to delete records for {obj}: {','.join(qs.job_result.job_errors)}"
+                )
+
+            for result in ds.get_results():
+                if not result.success and not self.options["ignore_row_errors"]:
```
prescod: Can we include the text for the failure reason in this (at least one row)?

Arguably we could also save the results to a file and let them peruse them themselves, but we'd get 90% of the benefit if we could see the first row.

davidmreed: Yep. Changed this to parallel how it already works in `load.py`.

cumulusci/tasks/bulkdata/extract.py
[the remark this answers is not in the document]
davidmreed: I've re-added the Postgres optimization and added some unit tests. @prescod Do you have any active Postgres data sets now that could be used to test it end-to-end?

cumulusci/tasks/bulkdata/utils.py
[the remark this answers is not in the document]
davidmreed: Because I could not reproduce the original error with or without it, I removed this workaround for the moment. Once I can document how to reproduce it, I'd like to do a new revision with a chunking strategy that's less ad-hoc.

cumulusci/tasks/bulkdata/extract.py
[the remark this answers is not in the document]
prescod: @davisagli : I think you have a large Postgres dataset around somewhere?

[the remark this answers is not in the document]
davisagli: @prescod alas, not any more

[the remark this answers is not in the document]
davidmreed: Hmm. If we don't have a way to test it, what about segregating this optimization off into a branch for the time down the line when @prescod starts using Postgres infrastructure? I don't want to ship a code path we can't test, and doing a lot of work to set up Postgres to test an optimization we are not using feels misplaced.

cumulusci/tasks/bulkdata/tests/test_load.py
```diff
@@ -0,0 +1,918 @@
...
+                    "mapping": "mapping.yml",
+                }
+            },
+        )
+
+        step = mock.Mock()
+        step.get_results.return_value = iter(
+            [
+                DataOperationResult("001000000000000", True, None),
+                DataOperationResult(None, False, None),
```
prescod: @davidmreed  I know this is just a test case but for work I'm doing I need to know whether a  result row like this could ever really exist? i.e. Error==None and Success=False .

cumulusci/tasks/bulkdata/load.py
```diff
@@ -107,56 +108,43 @@ def _load_mapping(self, mapping):
...
-                mapping["sf_object"], contentType="CSV", concurrency=mode
-            )
+        step = BulkApiDmlOperation(
+            sobject=mapping["sf_object"],
+            operation=(
+                DataOperationType.INSERT
+                if mapping.get("action") == "insert"
+                else DataOperationType.UPDATE
+            ),
+            api_options={"bulk_mode": self.bulk_mode or "Parallel"},
```
prescod: bulk_mode was previously set from the mapping. That code seems to have been deleted. Was that deliberate?

davidmreed:
End to end testing completed successfully. Test charter: https://salesforce.quip.com/agLwAj0DwuQ4

cumulusci/tasks/bulkdata/tests/test_load.py
davidmreed: @prescod  I'm not entirely sure.

davisagli:
@prescod any further concerns before approving?

davidmreed:
@davisagli  I'm tracking one possible regression that is hard to pin down. Hoping to finish that today.

davidmreed:
@davisagli  @prescod  Bug fixed (added a `commit()` call, which stopped SQLite leaving the database in a locked state when extracting Record Types) and regression passed.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -113,6 +113,7 @@ def _load_mapping(self, mapping):
         if mapping.get("fields", {}).get("RecordTypeId"):
             conn = self.session.connection()
             self._load_record_types([mapping["sf_object"]], conn)
+            self.session.commit()
```
prescod: Should this be in a try/finally block?

davisagli: Why would we want to commit if there was an exception?

PR #1490, opened 2020-01-21
Support chained batches.

prescod:
Chained batches are batches that kick off a successor before the previous batch completes.

I tested this with Abacus (and a unit test below) and it seems to work.

It also seems to work with normal batches that don't have successors.


davidmreed:
I think there's still a potential race condition here, but check my logic.

Suppose we enqueue Batch A. Batch A is `current_batch`, and it runs really fast.  We poll, it is in progress. Batch A completes, enqueues Batch B. Batch B completes very quickly and enqueues Batch C. We poll again. Batch C is now `current_batch`, but we _never saw_ Batch B. 

If Batch B has individual batch errors, we won't get them.

What would you think about polling for all batches and filtering by `CreatedDate`, instead of doing `LIMIT 1`? Poll all batches which were created at or after the creation date of the first batch we observe?

prescod:
I could go either way on this. 

It complexifies the code a bit because we have to treat the first query differently (because we have no start-date to drive a WHERE). My understanding -- correctly me if I'm wrong -- is that this chaining is used for extremely large jobs. Abacus doesn't switch to "LDV Mode" (as they call it) until 1,000,000 records. And then it defaults to "chunk" sizes of 1,000,000. So each chunk job runs for hours. 

https://github.com/SalesforceFoundation/Abacus/blob/b805b628ca0c954867eecabb173a29c2eacd790f/src/classes/UTIL_AbstractChunkingLDV_BATCH.cls#L68

Like I said, I'm open to changing it but just wanted to discuss whether it was really worth it.

davidmreed:
Batch chaining can be used at any data volume; Abacus just doesn't use it except at LDV. I would suggest that we do make this task fully general since applying it in projects that make different assumptions about batch chaining could theoretically produce some deeply inscrutable behavior.

cumulusci/tasks/apex/batch.py
```diff
@@ -26,78 +25,105 @@ class BatchApexWait(BaseSalesforceApiTask):
...
+        failed_batches = self.failed_batches()
+
+        if failed_batches:
             self.logger.info("There have been some batch failures.")
             self.logger.info("Error values:")
-            self.logger.info(repr(vals))
-            raise SalesforceException("There were import errors: %s" % repr(vals))
+            self.logger.info(repr(failed_batches))
+            raise SalesforceException(
+                f"There were import errors: {repr(failed_batches)}"
```
Br4nd0R: Are these alway guaranteed to be import errors, or should wording be more general?

cumulusci/tasks/apex/batch.py
```diff
@@ -32,83 +34,116 @@ def _run_task(self):
...
+        if failed_batches:
             self.logger.info("There have been some batch failures.")
             self.logger.info("Error values:")
-            self.logger.info(repr(vals))
-            raise SalesforceException("There were import errors: %s" % repr(vals))
-        elif not self.done_for_sure:
+            self.logger.info(repr(failed_batches))
+            raise SalesforceException(
+                f"There were batch errors: {repr(failed_batches)}"
+            )
```
davisagli: It seems redundant to show the failed_batches both in the log and in the exception message. (I see that's how it was before, but this seems like the best time to fix it.)

cumulusci/tasks/apex/batch.py
```diff
@@ -32,83 +34,116 @@ def _run_task(self):
...
     def _poll_action(self):
         # get batch status
-        query_results = self.tooling.query(self._batch_query)
 
-        self.batch = query_results["records"][0]
+        if not self.original_created_date:
+            query_results = self.tooling.query(self._batch_query(date_limit=None))
+            self.original_created_date = parse_api_datetime(
+                query_results["records"][0]["CreatedDate"]
+            )
```
davisagli: We should handle the case where no job was found and show something more helpful than an IndexError.

cumulusci/tasks/apex/batch.py
```diff
@@ -32,83 +34,116 @@ def _run_task(self):
...
-            self.batch["TotalJobItems"],
-            self.batch["NumberOfErrors"],
+            f"{self.options['class_name']}: "
+            f"Job{'s' if len(summary['Jobs'])>1 else ''}: {summary['Jobs']}"
+            f"{summary['JobItemsProcessed']} of {summary['TotalJobItems']} "
+            f"({summary['NumberOfErrors']} failures)"
         )
 
-        self.poll_complete = not self._poll_again()
+        self.poll_complete = current_batch["Status"] in COMPLETED_STATUSES
```
davisagli: What if the most recent batch is complete but a different one that started since original_created_date is not?

prescod: This whole branch is supporting a "coding convention" which is not directly supported by the platform, so we have to make some assumptions about how it works. My understanding of the coding convention is that you trigger Job N+1 after Job N is finished. That's what makes them chained. There is no parallel processing at all. That's certainly the case for Abacus.

I guess a parallel job run could be started by a 3rd party (e.g. a schedule) .In that case the safest thing would be to wait for both jobs to finish which would, I guess entail that this line might need to deal with that inconsistent situation.

davisagli: I was imagining a fan pattern where one job starts multiple other jobs rather than everything being sequential. I don't know if that is realistic -- @davidmreed?

cumulusci/tasks/apex/batch.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1490/files#diff-defe95b2261349328fbef8dd3bceeb93R44

cumulusci/tasks/apex/batch.py
prescod: I feel like sometimes logs go one place and exceptions another. Especially in Robot. So I'd rather leave it as-is. Better too much information about errors than too little. I'll keep an eye out in practice as I use it to see if it does consistently duplicate in the relevant contexts.

davisagli: The intention is that exceptions should always be included in logs. I think we do need to do some work to make sure that Robot Framework doesn't hide output from the MetaCI build log.

davisagli: But okay, we can revisit this later.

cumulusci/tasks/apex/batch.py
prescod: e09332e6d0185b0c2ab675c37c5ea0b232103204

cumulusci/tasks/apex/batch.py
prescod: In theory the code should support it now but testing parallel stuff is always hard and especially so if I don't have a real-world example yet.

https://github.com/SFDO-Tooling/CumulusCI/pull/1490/files#diff-defe95b2261349328fbef8dd3bceeb93R124

davidmreed: Fanout is not supported. Batch and queueable chaining is always N->N+1->N+2. However, @prescod's right that a second parallel chain could be invoked by a 3rd party.

PR #1492, opened 2020-01-22
Enhance EnsureRecordTypes to work properly with Case, Lead, and Solution

davidmreed:
# Critical Changes

# Changes

- The `EnsureRecordTypes` class is now exposed as `ensure_record_types` and correctly supports the Case, Lead, and Solution sObjects (in addition to other standard objects).

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -98,6 +98,13 @@ tasks:
         options:
             command: 'force:source:push'
         group: Salesforce DX
+    ensure_record_types:
+        description: Ensure that a default Record Type is extant on the given standard sObject (custom objects are not supported). If Record Types are already present, do nothing.
+        class_path: cumulusci.tasks.salesforce.EnsureRecordTypes
+        options:
+            record_type_label: Default
+            record_type_developer_name: Default
+        group: 'Salesforce Metadata'
     execute_anon:
         description: Execute anonymous apex via the tooling api.
         class_path: cumulusci.tasks.apex.anon.AnonymousApexTask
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/salesforce/EnsureRecordTypes.py
```diff
@@ -116,11 +120,19 @@ def _build_package(self):
                         self.options["sobject"]
                     )
                 )
+                # Lead, Case, and Solution require a default value in their Business Process
+                # Opportunity prohibits it.
+                if self.options["sobject"] == "Opportunity":
+                    default = "false"
+                else:
+                    default = "true"
```
jstvz: 💭 :
```python
       default = self.options["sobject"] != "Opportunity"
```

Then again, the existing version is probably more readable.

cumulusci/cumulusci.yml
```diff
@@ -98,6 +98,13 @@ tasks:
         options:
             command: 'force:source:push'
         group: Salesforce DX
+    ensure_record_types:
```
jstvz: 🤔 : Why did I think this was already a task?

davidmreed: It was, but it wasn't exposed in a defined task - just per-project.

cumulusci/tasks/salesforce/EnsureRecordTypes.py
davidmreed: Yeah, I was expecting more branches here and then discovered Case, Lead, and Solution all play by the same rules, but Opportunity doesn't. Adopting your proposal.

davidmreed: (Not before I make a couple of bone-headed mistakes, of course)

PR #1495, opened 2020-01-22
New features to run the Python code from CCI command line.

prescod:
# Critical Changes

# Changes

Add new features for running Python code (in a file or string) without bringing up an interactive shell. `--python` and `--script` arguments to `cci shell` and `cci org shell`

    $ cci org shell --python "print(sf.query('select Count(Id) from Opportunity'))"
    $ cci org shell --script foo.py

    $ cci  shell --python "print(config.get_org('dev'))"
    $ cci shell --script foo.py

I'm not sure if I need to document this somewhere else or whether the `--help` is the documentation.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -16,6 +16,7 @@ from datetime import datetime
 import webbrowser
 import contextlib
 from pathlib import Path
+import runpy
 
 import click
 import github3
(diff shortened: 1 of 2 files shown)


davisagli:
> I'm not sure if I need to document this somewhere else or whether the --help is the documentation.

`--help` doesn't count as documentation. But the right place for this to be documented is in a CumulusCI CLI command reference document which doesn't exist yet.

Did you consider implementing this as a Python task rather than new options for the shell commands? Then it could be incorporated into flows.

prescod:
@davisagli  I didn't really consider that. I thought of it more as an extension to the pre-existing Python shell command for interactive use. There is probably some virtue in encouraging people to think of Python extensions to CCI as reusable tasks instead of "snippets". I wouldn't be opposed to it as a task but I also like it the way it is. Meh.

PR #1496, opened 2020-01-23
Robot spring20 part2

boakley:
This fixes keyword issues that have been reported for Spring '20. The biggest change is in
the populate keywords, where I rewrote the basic algorithm for finding the input elements. There are also fixes for clicking on related item links. 

Unfortunately, an input element and its label can be marked up in at least two different ways, and the ways are so different as to make it almost impossible to craft a single xpath that works in both cases. With this change, we find the label and then use the 'for' attribute on that element to get the id of
the related input element. All of our tests pass with this change, and I've also spot-checked it with several NPSP tests that were brought to my attention.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -417,19 +417,24 @@ class Salesforce(object):
         self.wait_until_modal_is_open()
 
     def populate_field(self, name, value):
-        """Enters a value into a text field.
+        """Enters a value into an input or textarea field.
+
+        'name' represents the label on the page (eg: "First Name"),
+        and 'value' is the new value.
 
         Any existing value will be replaced.
         """
-        locator = lex_locators["object"]["field"].format(name)
+        locator = self._get_input_field_locator(name)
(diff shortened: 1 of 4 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -444,7 +452,38 @@ def populate_lookup_field(self, name, value):
...
+        'for' attribute is then extracted from the label and used
+        to create a new locator with that id.
+
+        For example, the locator 'abc123' will be returned
+        for the following html:
+
+        <label for='abc123'>First Name</label>
+        -or-
+        <label for='abc123'><span>First Name</span>
+        """
```
Br4nd0R: This doc string is great!

cumulusci/robotframework/tests/salesforce/ui.robot
```diff
@@ -1,9 +1,10 @@
-*** Settings ***
```
davisagli: @boakley ???

PR #1500, opened 2020-01-23
Remove section until we have up-to-date docs

Br4nd0R:
# Critical Changes

# Changes
* Temporarily removing CI section of cookbook as it is not up-to-date. We don't want users to waste time following examples that are no longer valid.


Br4nd0R:
I'm not sure whether I should mention the issue on here to close it out, or if you would like to reword the open issue, or if you want to make a work item for the kanban board (or even gus 🤷‍♂).

davisagli:
I'll try to work on updated docs in the next day or two so we can replace this instead of removing it.

davisagli:
Closing this; it was replaced in #1719

PR #1507, opened 2020-01-28
Allow for FindReplace to take a list of file_patterns

bethbrains:
# Critical Changes

# Changes
-- The base cci task FindReplace can now accept a list of strings for the file_pattern option

# Issues Closed


bethbrains:
FYI we ended up *not* using the solution that this would've made better, so there is zero urgency here. Still nice to have for someone in the future, though, so I'll leave the PR open.

cumulusci/tasks/util.py
```diff
@@ -149,14 +150,21 @@ def _run_task(self):
...
+
+        for fp in filePatterns:
+            findReplace(
+                find=self.options["find"],
+                replace=self.options["replace"],
+                directory=self.options["path"],
+                filePattern=fp,
+                logger=self.logger,
+                **kwargs
+            )
```
davisagli: It would be ideal to make findReplace handle a list of patterns itself so that we don't have to walk through the directories and write files multiple times. But I'm not worried about doing it this way.

bethbrains:
Thanks, @davisagli !!! Great feedback. Happy to have another small contribution! :)

davisagli:
@bethbrains I also moved the option processing into _init_options (that way if the task is used as part of a flow, the problem is caught before the flow starts running) and made it optional to specify the file_pattern option.

PR #1510, opened 2020-01-29
Metadata ETL Framework and Initial Tasks

davidmreed:
# Critical Changes

# Changes

- Added framework for building tasks that extract, transform, and reload metadata from a Salesforce org.
- Added task `add_standard_value_set_entries` to add entries to a `StandardValueSet`.
- Added task `add_page_layout_related_lists` to add Related Lists to a Page Layout.
- Added task `add_permission_set_perms` to add field permissions and Apex class accesses to a Permission Set.
- Added task `set_organization_wide_defaults` to set the Organization-Wide Defaults for one or more objects and wait for the operation to complete. 

# Issues Closed


cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+    @property
+    def _namespace_injector(self):
+        return lambda text: inject_namespace(
+            "",
+            text,
+            self.options.get("namespace_inject"),
+            not self.options["unmanaged"],
+        )[1]
+
+    def _get_package_xml_content(self, deploy):
```
Br4nd0R: I think I'm missing something here; is the `deploy` param being utilized implicitly?

davidmreed: @Br4nd0R Its purpose is to allow subclasses to generate a _different_ `package.xml` for deployment than what they requested to pull.

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+        for entity, api_names in self._get_entities().items():
+            members = "\n".join(
+                f"        <members>{api_name}</members>" for api_name in api_names
+            )
+            types += base.format(members=members, name=entity)
+
+        return types
+
+    def _get_package_xml_content(self, deploy):
+        return f"""<?xml version="1.0" encoding="UTF-8"?>
```
Br4nd0R: Same question from above: how is the `deploy` param being utilized?

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+                        "path": self.deploy_dir,
+                        "namespace_inject": self.options.get("namespace_inject"),
+                        "unmanaged": self.options.get("unmanaged"),
+                    }
+                }
+            ),
+            self.org_config,
+        )
+        result = api()
+        self._post_deploy(result)
```
davisagli: Should this be called from _run_task so that ETL tasks which override _deploy don't have to remember to call it?

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+        with tempfile.TemporaryDirectory() as tempdir:
+            self._create_directories(tempdir)
+            if self.retrieve:
+                self._retrieve()
+            self._transform()
+            if self.deploy:
+                self._deploy()
+
+
+class BaseMetadataSynthesisTask(BaseMetadataETLTask):
```
davisagli: metaclass=ABCMeta

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def _transform(self):
+        self._synthesize()
+
+    def _synthesize(self):
+        # Create metadata in self.deploy_dir
+        pass
+
+
+class BaseMetadataTransformTask(BaseMetadataETLTask):
```
davisagli: metaclass=ABCMeta

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+    </types>
+"""
+        types = ""
+        for entity, api_names in self._get_entities().items():
+            members = "\n".join(
+                f"        <members>{api_name}</members>" for api_name in api_names
+            )
+            types += base.format(members=members, name=entity)
+
+        return types
```
davisagli: I'd sort of like to factor this out into a separate utility for generating package.xml from Python (and/or give PackageXmlGenerator a better interface for specifying types from Python rather than from components in a directory) ...but that doesn't have to happen right now.

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+        return metadata
+
+    def _transform(self):
+        # call _transform_entity once per retrieved entity
+        # if the entity is an XML file, provide a parsed version
+        # and write the returned metadata into the deploy directory
+
+        parser = PackageXmlGenerator(
+            None, self.api_version
+        )  # We'll use it for its metadata_map
```
davisagli: Sounds like we should refactor things so the metadata map can be used without constructing a PackageXmlGenerator

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+            filter(
+                lambda entry: any(
+                    [
+                        subentry["type"] == self.entity
+                        for subentry in parser.metadata_map[entry]
+                    ]
+                ),
+                parser.metadata_map,
+            )
+        )
```
davisagli: entity_configurations = [entry for entry in parser.metadata_map if any( ... )]

But maybe the metadata_map should be wrapped by a class that provides a way to look up by type, instead of implementing that here.

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+        extension = configuration["extension"]
+        directory = entity_configurations[0]
+
+        for api_name in self.api_names:
+            path = self.retrieve_dir / directory / f"{api_name}.{extension}"
+            if not path.exists():
+                raise CumulusCIException(f"Cannot find metadata file {path}")
+
+            transformed_xml = self._transform_entity(
+                elementtree_parse_file(path), api_name
```
davisagli: Could we use lxml instead? @prescod has a PR #1391 which is trying to move us toward using it more consistently.

That said, I think we might want to figure out a way to do operations on the element tree without needing to specify the XML namespace everywhere.

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+                    "utf-8",
+                    xml_declaration=True,
+                    default_namespace=self.namespaces["sf"],
+                )
+
+
+def get_new_tag_index(tree, tag, namespaces):
+    # All top-level tags must be grouped together in XML file
+    tags = tree.findall(f".//sf:{tag}", namespaces)
+    if 0 < len(tags):
```
davisagli:
```suggestion
    if len(tags):
```

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+        # Insert new tag after the last existing tag of the same type
+        return list(tree.getroot()).index(tags[-1]) + 1
+    else:
+        # There are no existing tags of this type; insert new tag at the top.
+        return 0
```
davisagli: Bottom might be a less surprising default?

cumulusci/tasks/salesforce/metadata_etl/permissions.py
```diff
@@ -0,0 +1,132 @@
...
+
+from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.tasks.salesforce.metadata_etl import (
+    MetadataSingleEntityTransformTask,
+    get_new_tag_index,
+)
+
+
+class AddPermissions(MetadataSingleEntityTransformTask):
+    entity = "PermissionSet"
```
davisagli: How hard would it be to make this usable for Profiles too?

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+        if self.retrieve:
+            self.retrieve_dir = Path(tempdir, "retrieve")
+            self.retrieve_dir.mkdir()
+        if self.deploy:
+            self.deploy_dir = Path(tempdir, "deploy")
+            self.deploy_dir.mkdir()
+
+    def _retrieve(self):
+        api_retrieve = ApiRetrieveUnpackaged(
+            self, self._generate_package_xml(False), self.api_version
```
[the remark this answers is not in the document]
davidmreed: What would you think about using a typed enum to define the operation rather than a Boolean?

cumulusci/tasks/salesforce/metadata_etl/base.py
davidmreed: Yes. Fixing that.

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+from cumulusci.core.exceptions import CumulusCIException
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask, Deploy
+from cumulusci.salesforce_api.metadata import ApiRetrieveUnpackaged
+from cumulusci.tasks.metadata.package import PackageXmlGenerator
+from cumulusci.core.utils import process_bool_arg, process_list_arg
+from cumulusci.utils import elementtree_parse_file, inject_namespace
+from cumulusci.core.config import TaskConfig
+
+
+class BaseMetadataETLTask(BaseSalesforceApiTask):
```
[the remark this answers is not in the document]
davidmreed: I didn't know about that module. Thank you! Correcting this throughout and removing all pseudo-unit tests covering abstract methods.

cumulusci/tasks/salesforce/metadata_etl/base.py
davidmreed: Would you feel comfortable deferring that (this PR is already too big), or do you see this pattern as too poor to move forward?

cumulusci/tasks/salesforce/metadata_etl/permissions.py
davidmreed: I think it is doable. Profiles and Permission Sets are largely the same at the metadata level; Profiles are just far more frustrating to pull via MDAPI.

This is the logic CaseMan require(d) for their installer and which they built. I'd like to extend it to cover Profiles, Permission Sets, and all of the components they permission.

cumulusci/tasks/salesforce/metadata_etl/base.py
davidmreed: The comment is slightly confusing. If there are extent tags of this type (e.g `<relatedList>`), we should append after the last tag of that type. If there are no such tags, we should insert at the head of the file.

This is because MDAPI wants all like-typed tags grouped together.

cumulusci/tasks/salesforce/metadata_etl/sharing.py
```diff
@@ -0,0 +1,116 @@
...
+    def _poll_action(self):
+        elapsed = datetime.now() - self.time_start
+        if elapsed.total_seconds() > self.options["timeout"]:
+            raise CumulusCIException(
+                f'Sharing enablement not completed after {self.options["timeout"]} seconds'
+            )
+
+        for sobject in self.owds:
+            result = self.sf.query(
+                f"SELECT ExternalSharingModel, InternalSharingModel "
```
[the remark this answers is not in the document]
davidmreed: Got it from Chris Peterson: https://www.ca-peterson.com/apex/2020/01/21/apex-describe-calls-what-you-need-to-know.html

cumulusci/tasks/salesforce/metadata_etl/base.py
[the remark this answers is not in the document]
davidmreed: I replaced it with an enum in an attempt to respond to David's comment.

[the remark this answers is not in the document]
jstvz: Okay, I think that makes sense. 

I do believe docstrings would be helpful here. While you've got comments throughout, I don't think they work as docstrings unless they're [PEP 257-style](https://www.python.org/dev/peps/pep-0257/) triple quoted strings.

I don't think we need them _everywhere_, but I think they're important for describing how the base class works. Take `_transform()`, for example: What does it transform? Should it transform and return something or does it transform some instance variable?

cumulusci/tasks/salesforce/metadata_etl/base.py
davisagli: Yes, it can wait.

cumulusci/tasks/salesforce/metadata_etl/base.py
davisagli: Yes, that's what I understood from the code. If there aren't any existing tags of this type then it doesn't really matter whether the new tag gets added at the top or bottom in order to satisfy MDAPI's grouping requirement, but I'm saying adding them at the bottom would better match what a human would probably do if adding them to the file manually.

cumulusci/tasks/salesforce/metadata_etl/permissions.py
davisagli: What I'm wondering is whether that could be done in the same task class or would require a different class for adding to profiles. If this one will remain for permission sets only then we might want a more specific name like AddPermissionsToPermissionSet

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -11,10 +13,17 @@
...
+class MetadataOperation(enum.Enum):
+    DEPLOY = "deploy"
+    RETRIEVE = "retrieve"
+
+
 class BaseMetadataETLTask(BaseSalesforceApiTask):
     deploy = False
     retrieve = False
 
+    __metaclass__ = abc.ABCMeta
```
davisagli: @davidmreed This is the Python 2 way of specifying metaclasses and I don't think it has special meaning in Python 3. The Python 3 way puts them alongside the base classes: https://docs.python.org/3/reference/datamodel.html#metaclasses

cumulusci/tasks/salesforce/metadata_etl/base.py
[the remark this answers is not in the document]
davisagli: It seems like this is optimizing the API for use cases where deploy and retrieve operate on the same set of components, which I'm not sure is going to be the case. But I don't feel strongly about it and the enum is an improvement over the boolean.

cumulusci/tasks/salesforce/metadata_etl/sharing.py
```diff
@@ -0,0 +1,116 @@
...
+                record = result["records"][0]
+                if (
+                    self.owds[sobject]["internal_sharing_model"]
+                    and record["InternalSharingModel"]
+                    != self.owds[sobject]["internal_sharing_model"]
+                ) or (
+                    self.owds[sobject]["external_sharing_model"]
+                    and record["ExternalSharingModel"]
+                    != self.owds[sobject]["external_sharing_model"]
+                ):
```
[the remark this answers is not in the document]
davidmreed: Yeah. I just found this route slightly clearer. Open to argument.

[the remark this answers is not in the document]
jstvz: I thought about it a bit more and I think your approach is the right one. Otherwise you'd have to have some logic to toggle whether a field should be included in the filter.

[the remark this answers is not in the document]
jstvz: Oh, another observation would be that a NamedTuple in place of a dict is something to consider the next time around. 

Define the NamedTuple as:
```python
SharingModel = namedtuple("SharingModel", ["internal", "external"])
self.owds[self._namespace_injector(elem["api_name"])] = SharingModel(
    internal=elem.get("internal_sharing_model"),
    external=elem.get("external_sharing_model"),
)
```

This section would become:
```python
models = self.owds[sobject]
if result["totalSize"] == 1:
    record = result["records"][0]
    if (
        record["InternalSharingModel"] != models.internal
        or record["ExternalSharingModel"] != models.external
    ):
        return
```

cumulusci/tasks/salesforce/metadata_etl/base.py
davidmreed: Proposal subsequent to discussion this morning: suppose instead of using `elementtree_parse_file()` this module explicitly uses `xml.etree` to guarantee to subclasses what they will receive, while preserving the insulation of the concrete tasks from disk access.

Then, when we move forward with switching to `lxml` across more domains (or wrapping it in an abstraction layer), I'll commit to refactoring the entire MD-ETL framework to use that.

davisagli: @davidmreed I'd rather proactively switch this over to lxml's etree, since that's very likely what we'll end up using as our standard internal representation of parsed xml. I realized just after our call that as long as we do that now, there's no particular urgency to come up with a convenience API that adapts the lxml etree.

I'll try to help with switching these tasks over to lxml in the next couple days.

davisagli: Done in https://github.com/SFDO-Tooling/CumulusCI/pull/1544

cumulusci/tasks/salesforce/metadata_etl/tests/test_base.py
```diff
@@ -0,0 +1,300 @@
...
+from cumulusci.tasks.salesforce.metadata_etl import (
+    BaseMetadataETLTask,
+    BaseMetadataSynthesisTask,
+    BaseMetadataTransformTask,
+    MetadataSingleEntityTransformTask,
+    get_new_tag_index,
+)
+
+
+class test_BaseMetadataETLTask(unittest.TestCase):
```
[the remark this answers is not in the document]
davisagli: I removed the subclassing of unittest.TestCase in https://github.com/SFDO-Tooling/CumulusCI/pull/1544

cumulusci/tasks/salesforce/metadata_etl/base.py
davisagli: Fixed in https://github.com/SFDO-Tooling/CumulusCI/pull/1544

cumulusci/tasks/salesforce/metadata_etl/base.py
davidmreed: Thank you, @davisagli!

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+
+
+class BaseMetadataETLTask(BaseSalesforceApiTask):
+    deploy = False
+    retrieve = False
+
+    namespaces = {"sf": "http://soap.sforce.com/2006/04/metadata"}
+
+    task_options = {
+        "unmanaged": {
```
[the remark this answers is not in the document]
davidmreed: I switched to `managed` throughout.

cumulusci/tasks/salesforce/metadata_etl/base.py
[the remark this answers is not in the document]
davidmreed: Good push, James. I've added docstrings throughout.

cumulusci/tasks/salesforce/metadata_etl/permissions.py
davidmreed: As implemented, it will require at least a different facade class since it's one MDAPI entity per `MetadataSingleEntityTransformTask`. I'll change the name.

cumulusci/tasks/salesforce/metadata_etl/base.py
davidmreed: Ah, pardon my misunderstanding. Yes, I see your point, and I've changed this behavior.

cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def _get_entities(self):
+        return {}
+
+    def _get_types_package_xml(self):
+        base = """    <types>
+{members}
+        <name>{name}</name>
+    </types>
+"""
```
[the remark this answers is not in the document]
davidmreed: Yeah, this is a little cockeyed; I did it this way to get the indentation in the merged output how I wanted it, and so I could more cleanly assert against the generated XML. Open to doing it a better way.

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -0,0 +1,293 @@
...
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+
+        self.api_names = list(
+            map(
+                self._inject_namespace,
+                process_list_arg(self.options.get("api_names", [])),
+            )
+        )
```
davisagli: I just noticed this one, so I won't block approval on it, but this could be:

    self.api_names = [self._inject_namespace(name) for name in process_list_arg(self.options.get("api_names", []))]

PR #1511, opened 2020-01-29
handle connection error

### LINKED ISSUE #1460: CCI Connect Error Reporting Problem
Somehow I created an org with the REST API turned off.

I authenticate successfully but get an exception trying to save my org:

```
$ cci org connect pprescod-abacus-perf-playground-two

Launching web browser for URL https://login.salesforce.com/services/oauth2/authorize?response_type=code&client_id=3MVG9i1HRpGLXp.or6OVlWVWynz5l.LVm9GuZp5eDSX7s0YDDWfCkUMC0LrHzIQa070sI3nGyOUloCFOsocsh&redirect_uri=http://localhost:8080/callback&scope=web%20full%20refresh_token&prompt=login
Spawning HTTP server at http://localhost:8080/callback with timeout of 300 seconds.
If you are unable to log in to Salesforce you can press ctrl+c to kill the server and return to the command line.
127.0.0.1 - - [12/Jan/2020 16:18:54] "GET /callback?code=aPrxsIcPf9nU.yMj6tJ3NRgRDp9hBVMZjlFRcilOZQy3PJpFAW_Q5v3nPgwAjO2EfHo76DL6Mg%3D%3D HTTP/1.1" 200 -
  File "/Users/pprescod/.venvs/abacus-py37/bin/cci", line 10, in <module>
    sys.exit(main())
  File "/Users/pprescod/.venvs/abacus-py37/lib/python3.7/site-packages/click/core.py", line 764, in __call__
  File "/Users/pprescod/.venvs/abacus-py37/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/pprescod/.venvs/abacus-py37/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/Users/pprescod/.venvs/abacus-py37/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/Users/pprescod/.venvs/abacus-py37/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "/Users/pprescod/.venvs/abacus-py37/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "/Users/pprescod/code/CumulusCI/cumulusci/cli/cci.py", line 209, in new_func
    func(config, *args, **kw)
  File "/Users/pprescod/code/CumulusCI/cumulusci/cli/cci.py", line 807, in org_connect
    org_config._load_orginfo()
  File "/Users/pprescod/code/CumulusCI/cumulusci/core/config/OrgConfig.py", line 135, in _load_orginfo
    "org_type": self._org_sobject["OrganizationType"],
TypeError: list indices must be integers or slices, not str
```

self._org_sobject is 

```[{'message': 'The REST API is not enabled for this Organization.', 'errorCode': 'API_DISABLED_FOR_ORG'}]```

### PR DESCRIPTION
**Pull request by Br4nd0R**
# Changes
* We now give a more helpful message when network issues are detected.
* Closes #1460

### CODE REVIEW

File: cumulusci/cli/cci.py [hunk 1]
```diff
@@ -217,26 +218,41 @@ def main(args=None):
...
-            # TODO: errorsdb
-            # Return a non-zero exit code to indicate a problem
+            show_debug_info() if debug else handle_generic_error(e, is_gist_command)
             sys.exit(1)
 
 
+def handle_connection_error(error):
+    click.echo(
+        click.style(
+            f"We encountered an error with your internet connection. Please check your connection and try the last cci command again.\nError: {error}",
```
Br4nd0R: This message could probably be improved. Let me know if anyone has suggestions.

cumulusci/cli/cci.py
```diff
@@ -217,26 +218,44 @@ def main(args=None):
...
+            fg="red",
+        )
+    )
+
+
+def handle_generic_error(error, is_gist_cmd):
+    """Displays the error back to the user, prompts user investigat further
```
```suggestion
    """Displays the error back to the user, prompts user to investigate further
```

cumulusci/cli/cci.py
```diff
@@ -217,26 +218,44 @@ def main(args=None):
         try:
             cli(standalone_mode=False)
         except click.Abort:  # Keyboard interrupt
-            click.echo("\nAborted!")
+            show_debug_info() if debug else click.echo("\nAborted!")
```
davisagli: This is a change in behavior to enter pdb if the user hits Cmd+C while running a command that they ran with the `--debug` flag. It could be annoying or useful; I'm in favor of trying it.

cumulusci/cli/cci.py
davisagli: Hmm. The connection error could be due to a local network issue, but it could be a problem on the server side that the user doesn't have control over too. In that case it would be nice to have the traceback logged so we can get a clue as to what cci was trying to connect to.

davisagli: @Br4nd0R Sorry, I wasn't very clear about what I meant. I didn't mean that we should output the stack trace to stdout (we just did a bunch of refactoring to avoid doing that except for when the user asks for it). Instead what I was suggesting is that connection errors should not have their own `except` block in `main()`, because we don't want to bypass the normal error handling logic (with its support for the debug flag, logging the traceback to the log file, etc). The only difference for connection errors is that we want to print a custom message instead of the message from the exception instance.

davisagli:
Also, tests are failing.

cumulusci/cli/cci.py
Br4nd0R: Got it.

PR #1525, opened 2020-02-03
Week 6 Dependency Updates

Br4nd0R:


requirements.txt
```diff
@@ -21,7 +21,7 @@ PyYAML==5.3
 raven==6.10.0
 requests[security]==2.22.0
 robotframework==3.1.2
-robotframework-seleniumlibrary==4.1.0
+robotframework-seleniumlibrary==4.2.0
```
davisagli: @boakley could you review the changes in robotframework-seleniumlibrary 4.2.0 and let us know if we should wait to do this update?

boakley: The tests all passed for me locally after upgrading, and I didn't see any red flags in the release notes. I think it's safe to do the upgrade.

PR #1526, opened 2020-02-03
Feature/robot selenium speed

boakley:
This adds a new variable `${SELENIUM_SPEED}` which is used to initialize selenium when the `Open Test Browser` keyword is called. I also added a section to docs/robotframework.rst which mentions the variables that we define in Salesforce.robot. 

# Critical Changes

# Changes

# Issues Closed


cumulusci/robotframework/tests/salesforce/TestListener.py
```diff
@@ -1,32 +1,84 @@
...
         """Called whenever a message is added to the log"""
         if self.capture_log:
             self.log_messages.append(message)
 
+    def _start_test(self, name, attrs):
+        self.reset_test_listener_keyword_cache()
+
+    def _end_keyword(self, name, attrs):
+        attrs_subset = {name: attrs[name] for name in ("status", "args")}
+        self.keyword_cache.append((name, attrs_subset))
```
davisagli: Maybe this should just be a 3-tuple of `(name, attrs["status"], attrs["args"])`?

cumulusci/robotframework/tests/salesforce/TestListener.py
```diff
@@ -1,32 +1,84 @@
...
+                if attrs["status"] != expected_status:
+                    message = (
+                        f"Status of keyword {keyword_name} with args {args} "
+                        f"expected to be {expected_status} but was {attrs['status']}"
+                    )
+                    raise Exception(message)
+        if not keyword_was_found:
+            raise Exception(
+                f"No keyword with name '{keyword_name}' with args '{args}' was found"
+            )
```
davisagli: Should these exceptions be AssertionErrors?

cumulusci/robotframework/tests/salesforce/TestListener.py
```diff
@@ -1,32 +1,84 @@
-"""
-    This hybrid library/listener can be used to verify
-    messages that have been logged.
+"""This hybrid library/listener can be used to verify messages that
+    have been logged and keywords have been called.
+
+    This works by listening for log messages and keywords via the
+    listener interface, and saving them in a cache. Keywords are
```
davisagli: It seems like more of a log than a cache to me (cache sounds like things would get accessed randomly, and invalidated individually), but I don't feel strongly about what it's called.

docs/robotframework.rst
```diff
@@ -160,6 +160,76 @@ This simple test file can then be run via the **robot** task in CumulusCI:
...
+       Default: ``1280x1024``
+
+   * - ``${IMPLICIT_WAIT}``
+     - This is automatically passed to the `Set Selenium Implicit Wait
+       <http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Set%20Selenium%20Implicit%20Wait>`_ keyword.
+       Default: ``7 seconds``
+
+   * - ``${SELENIUM_SPEED}``
+     - This defines a delay added after every selenium command. It is
+       automaticaly passed to the `Set Selenium Speed
```
```suggestion
       automatically passed to the `Set Selenium Speed
```

cumulusci/robotframework/tests/salesforce/TestListener.py
boakley: I'm going to leave as-is. I may want to extend the listener to include more of the attributes, and keeping them in a dictionary makes that a bit easier. Thanks for the comment though.

cumulusci/robotframework/tests/salesforce/TestListener.py
boakley: Yes, they should. Thanks for catching that. I've made that fix.

cumulusci/robotframework/tests/salesforce/TestListener.py
boakley: You're right, I was being lazy. I've refactored a good bit of the code to introduce some consistency.

PR #1528, opened 2020-02-04
Feature/retry forever option

prescod:
This could use a bit of work on tests but I thought I'd get some feedback first.

### DIFF
--- a/cumulusci/tasks/bulkdata/delete.py
+++ b/cumulusci/tasks/bulkdata/delete.py
@@ -22,6 +22,9 @@ class DeleteData(BaseSalesforceApiTask, BulkJobTaskMixin):
         "hardDelete": {
             "description": "If True, perform a hard delete, bypassing the recycle bin. Default: False"
         },
+        "on_connection_failure": {
+            "description": "What to do after several connection retries fail: 'keep_trying' forever or 'quit' (the default)"
+        },
     }
 
     def _init_options(self, kwargs):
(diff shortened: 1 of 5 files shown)


davidmreed:
@prescod  Are you OK with holding for the API refactor (it is almost ready)? Right now every change to direct calls to the Bulk API will be a merge conflict because all of that has been refactored.

I'd like to hear more about the use case for a "retry forever" mode. When are we seeing that N retries fail but N + M retries don't?

prescod:
Yes it can wait.

With respect to the use-case:

I start a job that takes "around" 18 hours at 5 PM. Around 2AM the computer loses network access, perhaps for 15 minutes, if the problem is transient, perhaps until 8AM when I wake up and fix something (e.g. bounce the wifi connection, or a router, or whatever). Either way, I've lost around 9 hours "work" for no reason. If the crash happened in _wait_for_job (which is the most common case), then I may have lost more than 9 hours "work" because the Org might still be processing 100% correctly and it might be at 99% or 100% complete when I realize I have to delete all of the data and start again.

davisagli:
I don't like adding these retry wrappers throughout the LoadData code...what if we miss adding it somewhere? It seems like something that should be encapsulated in the bulk API client. Unfortunately salesforce-bulk doesn't expose access to a requests `session` where we could do this by configuring an urllib3 retry policy.

prescod:
I don't like it either, but the alternative is even worse than you've said: we don't just use Salesforce Bulk. We also use direct requests in some places.

https://github.com/SFDO-Tooling/CumulusCI/pull/1528/files#diff-0fbd63c7646d7de2efb1c16f1aeffb19R75

Also: the retry infrastructure don't just do HTTP retries: it also re-authenticates if necessary. So urllib is not the best place to do it either.

The only clean alternative I can think of is to wrap all of SF Bulk in a new wrapper class which also adds the additional features we bypass it to get access to (assuming that there is a good reason we bypass it). Or we could try a whole new library, although it would probably have the same issue.

Encapsulating SF Bulk would not be the worst thing in the world but that abstraction is more code for us to manage.

Perhaps the same abstraction would smooth out the rough corners of simple_salesforce but that's even more code and effort.

davidmreed:
simple_salesforce's Bulk API implementation is broken, and SalesforcePy only supports Bulk API 2.0.

#1489 factors all direct Bulk API access into a separate API layer, which would make this kind of logic easier to add.

prescod:
A more architectural, but also more reliable, way of handling the same issue is to make every Task and Step retry-able/continue-able as described here:

https://salesforce.quip.com/AFJ0AohovoHb

That could allow load processes to continue even in the face of machines being rebooted or dynos being cycled.

PR #1530, opened 2020-02-04
Fix windows sfdx shell quoting

davisagli:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/core/sfdx.py
+++ b/cumulusci/core/sfdx.py
@@ -1,5 +1,6 @@
 import io
 import logging
+import platform
 import sarge
 import sys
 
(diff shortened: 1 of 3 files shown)


davisagli:
Hmm, are we not reporting coverage from the windows build, or is it not running the right code?

cumulusci/core/sfdx.py
```diff
@@ -47,3 +48,18 @@ def sfdx(
...
         raise Exception(f"Command exited with return code {p.returncode}")
     return p
+
+
+def shell_quote(s):
+    if platform.system() == "Windows":
+        assert isinstance(s, str)
+        if not s:
+            result = '""'
+        elif '"' not in s:
```
Br4nd0R: I'm not following the logic here. When we looked at this yesterday, the `access_token` option worked when having no surrounding quotations _or_ if it had double quotations, but not when using single quotations. If we pass in `'access_token'` this code would return the same string. So how is this working?

davisagli: @Br4nd0R The input passed to this function is the actual value that we want to pass to the subprocess, without quotes around it. For Windows, we need to escape double quotes so that they won't be interpreted as special characters by cmd.exe, but single quotes don't have special meaning to cmd.exe so we don't have to escape them.

Br4nd0R: Thanks for the explanation, and the quick turnaround on this fix!

Issue #1529, opened 2020-02-04
flow info throws AttributeError on cross-project flows


jstvz:
Steps to reproduce (EDA):

``` shell
cci flow info upgraded_org
```

results in:

``` python-traceback
Traceback (most recent call last):
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/cumulusci/cli/cci.py", line 218, in main
    cli(standalone_mode=False)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/cumulusci/cli/cci.py", line 165, in new_func
    func(RUNTIME, *args, **kw)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/cumulusci/cli/cci.py", line 1330, in flow_info
    coordinator = runtime.get_flow(flow_name)
  File "/Users/jestevez/.pyenv/versions/3.7.1/envs/cci-prod/lib/python3.7/site-packages/cumulusci/core/runtime.py", line 97, in get_flow
    callbacks=callbacks,
[...]

PR #1532, opened 2020-02-06
3.6.0 Release

Br4nd0R:


HISTORY.rst
```diff
@@ -2,6 +2,30 @@
 History
 =======
 
+3.6.0 (2020-02-06)
+------------------
+Changes:
+
+* `cci task info` now has Command Syntax section and improved formatting of option information.
```
davisagli: @Br4nd0R can you run `make docs`, confirm the new generated docs get formatted successfully as html, and check in the resulting docs/tasks.rst?

HISTORY.rst
```diff
@@ -2,6 +2,30 @@
...
+
+* The `FindReplace` task now accepts a list of strings for the `file_pattern` option. 
+
+* If the `DeleteData` task fails to delete some rows, this is now reported as an error.
+
+* Robot Framework: Added a new variable `${SELENIUM_SPEED}` that is used to control the speed at which selenium runs when the `Open Test Browser` keyword is called. 
+
+Issues Closed:
+
+* Fixed an issue with how sfdx command options were being quoted in Windows.
```
davisagli: In the release notes we should aim to describe the issue by what the symptoms were rather than how we fixed it. Maybe "Fixed an issue where existing scratch orgs could sometimes not be used in Windows."

HISTORY.rst
```diff
@@ -2,6 +2,30 @@
...
+
+* Fixed a regression where the keychain was not being loaded for `flow info` and `task info` commands in cross-project flows.
```
davisagli: How about:

Fixed a regression where the `flow info` and `task info` commands could show an error `AttributeError: 'NoneType' object has no attribute 'get_service'` when trying to load tasks or flows from a cross-project source. (#1529)

HISTORY.rst
Br4nd0R: Yup, Task Reference section looks great!

davisagli: @Br4nd0R so are you going to check it in so that it will render on readthedocs?

PR #1533, opened 2020-02-06
Allow UpdateAdminProfile to run against alternate profiles

davidmreed:
# Critical Changes

# Changes

- `UpdateAdminProfile` now accepts a `profile_name` argument, which defaults to `Admin`. This allows the task to be executed against other profiles.

# Issues Closed

### DIFF
--- a/cumulusci/files/admin_profile.xml
+++ b/cumulusci/files/admin_profile.xml
@@ -12,7 +12,7 @@
     </types>
     <types>
         <members>Admin</members>
-        <name>Profile</name>
+        <name>{profile_name}</name>
     </types>
     <version>39.0</version>
 </Package>
--- a/cumulusci/files/admin_profile.xml
+++ b/cumulusci/files/admin_profile.xml
@@ -11,8 +11,8 @@
         <name>CustomObject</name>
     </types>
     <types>
-        <members>Admin</members>
-        <name>{profile_name}</name>
+        <members>{profile_name}</members>
+        <name>Profile</name>
     </types>
     <version>39.0</version>
 </Package>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,9 +1,7 @@
 version: 2.1
 
-
 orbs:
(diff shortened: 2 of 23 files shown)


force2b:
This is awesome!!

davidmreed:
Whoops, fixing the issue with the release test.

cumulusci/files/admin_profile.xml
```diff
@@ -12,7 +12,7 @@
     </types>
     <types>
         <members>Admin</members>
-        <name>Profile</name>
+        <name>{profile_name}</name>
```
jstvz: Should you be replacing `<members>` instead?

davidmreed:
> I'd suggest renaming the class from UpdateAdminProfile to UpdateProfile (but also add `UpdateAdminProfile = UpdateProfile` to the module as an alias for backwards compatibility)

There's at least one project with a custom task that imports `UpdateAdminProfile` via the "wrong" path (`from cumulusci.tasks.salesforce.UpdateAdminProfile import UpdateAdminProfile` instead of `from cumulusci.tasks.salesforce ...`), such that if I rename the class and its containing file it will be a breaking change. Is there a preferred pattern for doing that in a backwards-compatible way? Just don't change the filename?

I will fix that project, but it may not be the only one doing such a thing.

davisagli:
@davidmreed oh hmm -- really I'd like to rename all the camel case modules to snake case, but I didn't realize that it was possible to do the import that way.

It should be possible to set up module aliases something like this (within `cumulusci/tasks/salesforce/__init__.py`):

    from cumulusci.tasks.salesforce import update_profile
    UpdateProfile = UpdateAdminProfile = update_profile.UpdateProfile
    sys.modules["cumulusci.tasks.salesforce.UpdateAdminProfile"] = update_profile

davidmreed:
Had no idea I could synthesize the module like that. Thanks, @davisagli. I took the opportunity to snake-case the module as well.

PR #1541, opened 2020-02-12
Fix env var for passing persistent org access token to sfdx

davisagli:
# Critical Changes

# Changes

# Issues Closed
* Fixed a bug where using SFDXOrgTask to run an sfdx command on a non-scratch org would break with "Must pass a username and/or OAuth options when creating an AuthInfo instance."


davisagli:
@prescod this should not have been approved and merged before the builds were done running.

prescod:
Okay, I'm open to that feedback but on the other hand it really didn't seem plausible that a Robot failure would indicate a bug in your environment variables. So based on my past experience I thought that the most likely result was that if the Robot failed we'd merge it anyhow on the basis that a Robot failure would likely be just something intermittent and unrelated.

PR #1544, opened 2020-02-13
Various adjustments to metadata ETL

davisagli:
- Use lxml etree for manipulating XML (turns out f-strings make it nicer than it used to be!)
- Use camel case class names for test classes, and don't extend unittest.TestCase because we don't need to
- Declare ABCMeta metaclass the correct way for Python 3
- Add a log line before the metadata retrieve and deploy to clarify what we're showing progress of
- Fix `required` parameter of task options to be a boolean

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/metadata_etl/base.py
+++ b/cumulusci/tasks/salesforce/metadata_etl/base.py
@@ -2,6 +2,7 @@ import abc
 import enum
 import tempfile
 
+from lxml import etree
 from pathlib import Path
 
 from cumulusci.core.exceptions import CumulusCIException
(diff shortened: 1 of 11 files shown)


cumulusci/tasks/salesforce/metadata_etl/base.py
```diff
@@ -108,7 +109,6 @@ def _deploy(self):
 
         return result
 
-    @abc.abstractmethod
```
davisagli: Removed this one since it _requires_ the method to be overridden by subclasses.

PR #1545, opened 2020-02-13
Insert Record task

prescod:
# Critical Changes

New task to insert any Salesforce Record: insert_record

### DIFF
--- a/cumulusci/tasks/salesforce/insert_record.py
+++ b/cumulusci/tasks/salesforce/insert_record.py
@@ -10,7 +10,7 @@ class InsertRecord(BaseSalesforceApiTask):
     """
 
     task_options = {
-        "object": {"description": "An SObject type to insert", "required": True},
+        "object": {"description": "An sObject type to insert", "required": True},
         "values": {
             "description": "Field names and values in the format 'aa:bb,cc:dd'",
             "required": True,
--- a/cumulusci/tasks/salesforce/insert_record.py
+++ b/cumulusci/tasks/salesforce/insert_record.py
@@ -1,0 +1,24 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.core.utils import process_list_of_pairs_dict_arg
+
+
+class InsertRecord(BaseSalesforceApiTask):
+    task_docs = """
+        For example:
+
+        cci task run insert_record --org dev -o object PermissionSet -o values Name:HardDelete,PermissionsBulkApiHardDelete:true
+    """
+
+    task_options = {
+        "object": {"description": "An SObject type to insert", "required": True},
+        "values": {
+            "description": "Field names and values in the format 'aa:bb,cc:dd'",
+            "required": True,
+        },
+    }
+
+    def _run_task(self):
+        object_handler = getattr(self.sf, self.options["object"])
+        values = process_list_of_pairs_dict_arg(self.options["values"])
+        rc = object_handler.create(values)
+        self.logger.info(f"Record inserted {rc['id']}")
--- a/cumulusci/tasks/salesforce/insert_record.py
+++ b/cumulusci/tasks/salesforce/insert_record.py
@@ -1,5 +1,6 @@
 from cumulusci.tasks.salesforce import BaseSalesforceApiTask
 from cumulusci.core.utils import process_list_of_pairs_dict_arg
+from cumulusci.core.exceptions import SalesforceException
 
 
 class InsertRecord(BaseSalesforceApiTask):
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/salesforce/insert_record.py
```diff
@@ -0,0 +1,24 @@
...
+        "object": {"description": "An SObject type to insert", "required": True},
+        "values": {
+            "description": "Field names and values in the format 'aa:bb,cc:dd'",
+            "required": True,
+        },
+    }
+
+    def _run_task(self):
+        object_handler = getattr(self.sf, self.options["object"])
+        values = process_list_of_pairs_dict_arg(self.options["values"])
```
davisagli: This should be done in _init_options so that errors parsing the arg are caught before the flow starts running.

cumulusci/tasks/salesforce/insert_record.py
```diff
@@ -0,0 +1,24 @@
...
+        rc = object_handler.create(values)
+        self.logger.info(f"Record inserted {rc['id']}")
```
davisagli:
```suggestion
        self.logger.info(f"{self.options['object']} record inserted: {rc['id']}")
```

cumulusci/tasks/salesforce/insert_record.py
```diff
@@ -0,0 +1,24 @@
...
+        "values": {
+            "description": "Field names and values in the format 'aa:bb,cc:dd'",
+            "required": True,
+        },
+    }
+
+    def _run_task(self):
+        object_handler = getattr(self.sf, self.options["object"])
+        values = process_list_of_pairs_dict_arg(self.options["values"])
+        rc = object_handler.create(values)
```
davisagli: We should check the `success` key and report the `errors` key if it's false.

cumulusci/tasks/salesforce/tests/test_insert_record.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_syntax_errors(self):
+        task = create_task(
+            InsertRecord,
+            {
+                "object": "PermissionSet",
+                "values": "Name:Hard:Delete,PermissionsB:ulkApiHardDelete:true",
+            },
+        )
+        task._init_task()
+        with self.assertRaises(TaskOptionsError):
```
davisagli: If you use `pytest.raises` here then the test class doesn't need to extend unittest.TestCase.

davisagli:
Also please run `make docs` and check in the resulting updates to docs/tasks.rst

cumulusci/tasks/salesforce/tests/test_insert_record.py
prescod: True but then I'd have to import pytest.

Also: 14 of our files use pytest.raises and 42 use self.assertRaises.

davisagli: Yes, we've been switching them over to pytest as we work on them.

cumulusci/tasks/salesforce/insert_record.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1545/commits/4e4a1a37380e77f0e87f04afbcc7760372093f78#diff-dfd20195f14b74746c715a851a4d98b8R21

cumulusci/tasks/salesforce/insert_record.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1545/commits/4e4a1a37380e77f0e87f04afbcc7760372093f78#diff-dfd20195f14b74746c715a851a4d98b8R30

cumulusci/tasks/salesforce/insert_record.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1545/commits/4e4a1a37380e77f0e87f04afbcc7760372093f78#diff-dfd20195f14b74746c715a851a4d98b8R32

cumulusci/tasks/salesforce/tests/test_insert_record.py
prescod: Okay but why? It kind of rubs me the wrong way to use identical functionality from a 3rd party package instead of the standard library unless there is a good reason.

davisagli: That's fair -- I think they are pretty identical functionality-wise, and this is partly why I've felt no particular need to switch everything over to the pytest version. However there are some pytest features that can't be used in unittest.TestCase subclasses (https://docs.pytest.org/en/latest/unittest.html#pytest-features-in-unittest-testcase-subclasses), and then it's handy to have the alternative.

PR #1546, opened 2020-02-14
Avoid leaving an incomplete codebase in the cache if zipball extraction fails

davisagli:
# Critical Changes

# Changes

# Issues Closed
- Fixed an issue where an error while extracting the codebase of a cross-project source could leave behind an incomplete copy of the codebase which would then be used by future commands.


davidmreed:
Approving based on code; I didn't do an E2E test on Windows.

PR #1547, opened 2020-02-15
Fix issues with wildcards and dropped entities in Metadata ETL

davidmreed:
This PR corrects a couple of small issues with Metadata ETL:

- Failure to correctly omit deployment of entities that the transformer task declined to act on.
- Failure to identify entities pulled by a wildcard and transform them.
- Added code to handle URL-encoded file names for entities like Page Layouts (such as `Contact-Contact %28Marketing%29 Layout.layout`)

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/metadata_etl/base.py
+++ b/cumulusci/tasks/metadata_etl/base.py
@@ -213,6 +213,9 @@ class MetadataSingleEntityTransformTask(BaseMetadataTransformTask, metaclass=ABC
             self._inject_namespace(arg)
             for arg in process_list_arg(self.options.get("api_names", ["*"]))
         }
+        self.api_names = {
+            quote(arg, safe=" ") if arg != "*" else arg for arg in self.api_names
+        }
 
     def _get_entities(self):
         return {self.entity: self.api_names}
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/metadata_etl/base.py
```diff
@@ -258,9 +258,26 @@ def _transform(self):
...
+            )
+
+        removed_api_names = set()
 
         for api_name in self.api_names:
-            path = self.retrieve_dir / directory / f"{api_name}.{extension}"
+            # Page Layout names can contain spaces, but parentheses are quoted.
+            # Those files _cannot_ be deployed back into Salesforce under their
+            # quoted names - we have to rename them.
+            quoted_api_name = quote(api_name, safe=" ")
```
davidmreed: I _think_ this is the right way to handle metadata names. For most components, the API name is required to be [A-Za-z0-9_]; Page Layouts can include parens and other characters which are URL-encoded, but spaces are not.

(@jstvz have you seen any other variants to be aware of?)

davidmreed:
After more testing, I found some edge cases where the current quoting workaround does not work (although the principle is sound). Doing a slight adjustment.

PR #1560, opened 2020-02-19
Metadata XML Tree Model 

prescod:
# Changes

A new library for making it easy to manipulate Metadata XML.

Now has 100% test coverage and replaces XML parsing in many places in CCI. (anywhere that Metadata XML is manipulated on the basis of its semantics rather than merely "as XML as in the DeleteElements task")


cumulusci/util/xml/test/test_salesforce_encoding.py
```diff
@@ -0,0 +1,68 @@
...
+
+        assert files
+
+        for file in files:
+            orig = open(file).read()
+            tree = etree.parse(file)
+            out = serialize_xml_for_salesforce(tree)
+            print(f"[{orig}]")
+            print(f"[{out}]")
+            assert orig == out
```
davisagli: Please add an assertion error with the name of the file (or this might be an opportunity to use pytest's test parametrization feature, but that could be more trouble than it's worth)

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+        """Remove an element from its parent (self)"""
+        self._element.remove(metadata_element._element)
+
+    def find(self, tag, text=None):
+        """Find a single direct child-elements with name `tag`"""
+        return next(self._findall(tag, text), None)
+
+    def findall(self, tag, text=None):
+        """Find all direct child-elements with name `tag`"""
+        return list(self._findall(tag, text))
```
davisagli: It's a pity to lose the ability to do more complex XPath queries -- but I don't know how we would add the namespaces without parsing xpath ourselves

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -250,19 +249,15 @@ def _process_static_resources(self, zip_src):
...
-        tree = ET.fromstring(package_xml)
-        section = tree.find(".//sf:types[sf:name='StaticResource']", self.namespaces)
-        if section is None:
-            section = ET.Element("{{{}}}types".format(self.namespaces["sf"]))
-            name = ET.Element("{{{}}}name".format(self.namespaces["sf"]))
-            section.append(name)
-            name.text = "StaticResource"
-            tree.find(".//sf:types[last()]", self.namespaces).addnext(section)
+        Package = metadata_tree.parse(package_xml)
+        sections = [typ for typ in Package.types if typ.name.text == "StaticResource"]
```
davisagli: We're doing this in a number of places. Would be nice if we could do something like `Package.find_by("types", name="StaticResource")`

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -287,8 +284,8 @@ def _transform(self):
                 raise CumulusCIException(f"Cannot find metadata file {path}")
 
             try:
-                tree = etree.parse(str(path))
-            except etree.ParseError as err:
+                tree = metadata_tree.parse(str(path))
```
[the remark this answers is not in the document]
prescod: It would be nice to have a cleaner interface and less code duplication. I'll see what David R. says.

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
+from typing import Union
```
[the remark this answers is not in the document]
prescod: I find packages that are also modules kind of ugly and confusing myself. os.path.X versus os.X for example.

I can't think of good examples from the stdlib. I assume urlopen moved into urllib.request, for that reason.

Either way it's going to be slightly confusing during the transition and simpler once all of the utils are moved.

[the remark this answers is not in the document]
davisagli: I'd like to stick with utils. It's what we use in most other locations in the codebase.

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+
+def parse(source):
+    """Parse a file by path or file object into a Metadata Tree
+
+    The parse() function supports any of the following sources:
+
+        * an open file object (make sure to open it in binary mode)
+        * a file-like object that has a .read(byte_count) method returning a byte string on each call
+        * a filename string
+        * an HTTP or FTP URL string
```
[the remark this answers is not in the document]
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1560/commits/965954ca39cf32490dc0ab504a6db842b0dc4e28

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+    >>> Package["types"]["name"].text
+    'StandardValueSet'
+
+    That might be convenient if you had the key names in a variable or if you had to refer to an element name
+    which would clash with a Python keyword (like `import`) or an instance method (like `text`).
+
+    You can refer to members of lists by index:
+
+    >>> Package["types"]["members"][1]
+    <members>CaseStatus</members> element
```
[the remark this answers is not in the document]
prescod: I agree but my quick experiment suggests that it is a small project in its own right. You get several errors if you run

pytest --doctest-modules

And they don't all seem to have the same solution.

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+
+    __slots__ = ["_element", "_parent", "_ns", "tag"]
+
+    def __init__(
+        self, element: etree._Element, parent: etree._Element = None,
+    ):
+        assert isinstance(element, etree._Element)
+        assert len(element.nsmap) == 1, "Only one namespace allowed"
+        self._element = element
+        self._parent = parent
```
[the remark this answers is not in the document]
prescod: No, I think that it will break logic if the parent goes away before the child. In order to implement

foo.bar[3] we do foo.bar.parent.find['foo'][3]

Which I think answers another one of your questions.

We do it that way because we want foo.bar to behave as a regular leaf UNLESS we index it and then it becomes more like a list. So faking that means we need all of the information available to treat it as either a node or a list.

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+        self._ns = next(iter(element.nsmap.values()))
+        self.tag = element.tag.split("}")[1]
+
+    @property
+    def text(self):
+        if len(self._element):
+            text_node = self._element.find(self._add_namespace("text"))
+            if len(text_node) >= 0:
```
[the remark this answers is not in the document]
prescod: The previous code did actually work but I changed it to be simpler and more robust.

The weird coding was mostly caused by the fact that some versions of etree have weird behaviour with respect to elements with no sub-elements. It treats them as false.

But anyhow, I already have code to deal with it elsewhere (including nicer error messaging) so I just called it:

https://github.com/SFDO-Tooling/CumulusCI/pull/1560/commits/4b577affd68e64fcea37ae11183114912940d7cd

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+
+    @text.setter
+    def text(self, text):
+        self._element.text = text
+
+    def _wrap_element(self, child: etree._Element):
+        return MetadataElement(child, self._element)
+
+    def _add_namespace(self, tag):
+        return "{%s}%s" % (self._ns, tag)
```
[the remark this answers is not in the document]
prescod: I used old-fashioned string interpolation because curly-braces already have a special meaning in namespaces. I think that your suggestion would leave out the curly braces that the namespace wants. Of course I can escape the curly braces but this seemed like the lesser evil by a smidgen.

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _get_child(self, childname):
+        child_element = self._element.find(self._add_namespace(childname))
```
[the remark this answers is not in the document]
prescod: This is an internal function. 

The indexing occurs here:

https://github.com/SFDO-Tooling/CumulusCI/pull/1560/files#diff-0f00573b2fbf4af992b4caf5820f88a2R116

And as I mentioned elsewhere, it does this because

x.y

Needs to be ready to behave as the base stem of either:

x.y.z

Or

x.y[3]

In other words it behaves as either a list or an element depending on what happens 
"next". The base-case is element so I need an element that can behave as a list.

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+        elif isinstance(item, str):
+            return self._get_child(item)
+
+    def append(self, tag: str, text: str = None):
+        '''Append a new element at the appropriate place.
+
+        If the parent element (self) already one or more children that match,
+        the new element follows the last one.
+
+        Otherwise, the new element goes to the bottom of the parent (self).
```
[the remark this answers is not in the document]
prescod: I thought about it but decided that we should wait to see if we need it in practice. So far the use cases seem to always be simple mutations...

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+        index = self._element.index(oldElement._element)
+        return self.insert(index, tag, text)
+
+    def insertAfter(self, oldElement: "MetadataElement", tag: str, text: str = None):
+        """Insert after some other element
+
+        Tag and text can be supplied. Return value is the new element."""
+
+        index = self._element.index(oldElement._element)
+        return self.insert(index + 1, tag, text)
```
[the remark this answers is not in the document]
prescod: Seems a bit like overkill to me. But I could add documentation that they should use the other methods in preference to this one unless they have a very specific use-case.

5f1a9e0a

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __getattr__(self, childname):
+        return self._get_child(childname)
+
+    def __getitem__(self, item: Union[int, str]):
+        if isinstance(item, int):
+            children = self._parent.findall(self._element.tag)
+            return self._wrap_element(children[item])
+        elif isinstance(item, str):
+            return self._get_child(item)
```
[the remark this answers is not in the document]
prescod: 8b6321f2c662

cumulusci/util/xml/salesforce_encoding.py
```diff
@@ -0,0 +1,45 @@
...
+
+from xml.sax.saxutils import escape
+
+xml_encoding = '<?xml version="1.0" encoding="UTF-8"?>\n'
+METADATA_NAMESPACE = "http://soap.sforce.com/2006/04/metadata"
+
+
+def serialize_xml_for_salesforce(xdoc, xml_declaration=True):
+    r = xml_encoding if xml_declaration else ""
+    if METADATA_NAMESPACE in xdoc.getroot().tag:
```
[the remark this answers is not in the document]
prescod: I think this is another case where the curly braces are going to mess us up.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
prescod: 1bddbc1eb08

cumulusci/util/xml/metadata_tree.py
prescod: Now that these things are objects it crossed my mind to use something like ObjectPath or JSONPath but so far list expressions are working reasonably well.

cumulusci/util/xml/salesforce_encoding.py
```diff
@@ -0,0 +1,45 @@
...
+        xdoc.getroot().attrib["xmlns"] = METADATA_NAMESPACE
+    for action, elem in etree.iterwalk(
+        xdoc, events=("start", "end", "start-ns", "end-ns", "comment")
+    ):
+        if action == "start-ns":
+            pass  # handle this nicely if SF starts using multiple namespaces
```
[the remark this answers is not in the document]
prescod: 95ea21b1

cumulusci/util/xml/test/test_salesforce_encoding.py
prescod: 3d4b7068

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+        etree.indent(doc, space="    ")
+        return serialize_xml_for_salesforce(doc, xml_declaration=xml_declaration)
+
+    def __eq__(self, other: "MetadataElement"):
+        return self._element == other._element
+
+    def __repr__(self):
+        return (
+            f"<{self.tag}>{self.text.strip() if self.text else ''}</{self.tag}> element"
+        )
```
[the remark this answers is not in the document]
prescod: I think there's a problem choosing which interpretation is most intuitive due to the strange way Salesforce manages lists.

```
for foo in Recipe.Ingredients:
   print(foo.tag)   # what does this print?
```
Is this the same as:

```
for foo in Recipe.find_all("Ingredients"):
```

Or:

```
for foo in Recipe.Ingredients.children:
```

cumulusci/util/xml/metadata_tree.py
```diff
@@ -0,0 +1,222 @@
...
+        doc = etree.ElementTree(self._element)
+        etree.indent(doc, space="    ")
+        return serialize_xml_for_salesforce(doc, xml_declaration=xml_declaration)
+
+    def __eq__(self, other: "MetadataElement"):
+        return self._element == other._element
+
+    def __repr__(self):
+        return (
+            f"<{self.tag}>{self.text.strip() if self.text else ''}</{self.tag}> element"
```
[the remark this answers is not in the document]
prescod: e2d59f16c2

cumulusci/tasks/salesforce/Deploy.py
prescod: Sounds like you want an ORM. ;)

cd3dde5e

cumulusci/util/xml/metadata_tree.py
prescod: I think with the new matching feature we cover 90% of the use-cases. Metadata files are not recursive and are not super-deeply nested either.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
prescod: By the way, this was the central trick that made any of it possible. In my first prototype I was parsing the schema to figure out what were "just" elements and what were lists. Then I realized I could use this trick to "do what the user expects" in either case.

cumulusci/tasks/metadata_etl/base.py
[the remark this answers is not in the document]
davisagli: @prescod sounds like David was on board with making this change. Let's make sure we call it out in the "critical changes" section of the release notes.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
davisagli: Fair enough, I'll add a user story to work on that separately.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
davisagli: That helps. We can add a check later if we're running into it.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
davisagli: Ah yes, okay, this is the key trick that I wasn't quite spotting.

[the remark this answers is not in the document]
davisagli: It would really be worth a comment or docstring in `__getitem__`.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
davisagli: Oops, yeah you're right.

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
davisagli: The `for foo in Recipe.find_all("Ingredients"):` interpretation would be less surprising to me, but it's a fair point that we can avoid the ambiguity by not implementing either one.

cumulusci/tasks/salesforce/Deploy.py
davisagli: I like the generalization to kwargs in findall.

davisagli:
There are 10 lines missing coverage: https://coveralls.io/builds/28961956

cumulusci/utils/xml/metadata_tree.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class MetadataElement:
+    '''A class for representing Metadata in a Pythonic tree.
+
+    After you parse into a MetadataElement tree, you can refer
+    to child elements like this:
+
+    >>> from cumulusci.utils.xml import metadata_tree
+    >>> Package.types.name.text
```
davidmreed: These two lines are duplicated from below.

cumulusci/utils/xml/metadata_tree.py
```diff
@@ -0,0 +1,251 @@
...
+
+        If all else fails then you can use this one to precisely insert right
+        where you want it but you're responsible for adhering to Salesforce's
+        grouping rules.
+        """
+        newchild = self._create_child(tag, text)
+        self._element.insert(index, newchild._element)
+        return newchild
+
+    def insertBefore(self, oldElement: "MetadataElement", tag: str, text: str = None):
```
davidmreed: Is there a reason for lowerCamel rather than snake_case?

cumulusci/utils/xml/test/test_metadata_tree.py
```diff
@@ -0,0 +1,165 @@
...
+        assert Data.tostring() == f'<Data xmlns="{METADATA_NAMESPACE}"/>'
+
+    def test_error_handling(self):
+        Data = fromstring(f"<Data xmlns='{METADATA_NAMESPACE}'><Foo/></Data>")
+        Data.Foo
+        with pytest.raises(AttributeError) as e:
+            assert Data.Bar
+        assert "not found in" in str(e.value)
+
+    def test_getxxxx___(self):
```
davidmreed: Could this method name be
```suggestion
    def test_getattr_getitem(self):
```

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -60,20 +56,15 @@ def _create_new_related_list(self, metadata, api_name, related_list, index):
...
         ]
 
-        elem = etree.Element(f"{MD}relatedLists")
-        metadata.getroot().insert(index, elem)
+        elem = metadata.append("relatedLists")
 
         for f in fields:
-            elem_field = etree.SubElement(elem, f"{MD}fields")
-            elem_field.text = f
+            elem.append(tag="fields", text=f)
```
davidmreed:
```suggestion
            elem.append("fields", text=f)
```

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -60,20 +56,15 @@ def _create_new_related_list(self, metadata, api_name, related_list, index):
...
+            elem.append("excludeButtons", button)
 
         for button in custom_buttons:
-            elem_button = etree.SubElement(elem, f"{MD}customButtons")
-            elem_button.text = button
+            elem.append("customButtons", button)
 
-        elem_related_list = etree.SubElement(elem, f"{MD}relatedList")
-        elem_related_list.text = related_list
+        elem.append(tag="relatedList", text=related_list)
```
davidmreed:
```suggestion
        elem.append("relatedList", text=related_list)
```

cumulusci/tasks/metadata_etl/permissions.py
```diff
@@ -48,24 +43,22 @@ def _upsert_class_accesses(self, metadata, api_name):
...
-                f".//{MD}classAccesses[{MD}apexClass='{class_access['apexClass']}']"
+                "classAccesses", apexClass=class_access["apexClass"]
             )
-            if 0 < len(existing_permissions):
+
+            if len(existing_permissions):
                 # Permission exists: update
                 for elem in existing_permissions:
-                    elem.find(f"{MD}enabled").text = str(
+                    elem.find(f"enabled").text = str(
```
davidmreed:
```suggestion
                    elem.find("enabled").text = str(
```

cumulusci/tasks/metadata_etl/permissions.py
```diff
@@ -48,24 +43,22 @@ def _upsert_class_accesses(self, metadata, api_name):
...
-                elem = etree.Element(f"{MD}classAccesses")
-                metadata.getroot().insert(new_permission_index, elem)
-
-                elem_apexClass = etree.SubElement(elem, f"{MD}apexClass")
-                elem_apexClass.text = class_access.get("apexClass")
-
-                elem_enabled = etree.SubElement(elem, f"{MD}enabled")
-                elem_enabled.text = str(class_access.get("enabled", True)).lower()
+                elem = metadata.append("classAccesses")
+                elem.append(tag="apexClass", text=class_access.get("apexClass"))
```
davidmreed:
```suggestion
                elem.append("apexClass", text=class_access.get("apexClass"))
```

cumulusci/tasks/metadata_etl/permissions.py
```diff
@@ -48,24 +43,22 @@ def _upsert_class_accesses(self, metadata, api_name):
...
+                elem.append(
+                    tag="enabled", text=str(class_access.get("enabled", True)).lower()
```
davidmreed:
```suggestion
                    "enabled", text=str(class_access.get("enabled", True)).lower()
```

cumulusci/tasks/metadata_etl/permissions.py
```diff
@@ -88,27 +79,25 @@ def _upsert_field_permissions(self, metadata, api_name):
...
-                f".//{MD}fieldPermissions[{MD}field='{field_permission['field']}']"
+                "fieldPermissions", field=field_permission["field"]
             )
-            if 0 < len(existing_permissions):
+
+            if len(existing_permissions):
                 # Permission exists: update
                 for elem in existing_permissions:
-                    elem.find(f"{MD}readable").text = str(
+                    elem.find(f"readable").text = str(
```
davidmreed:
```suggestion
                    elem.find("readable").text = str(
```

cumulusci/tasks/metadata_etl/permissions.py
```diff
@@ -88,27 +79,25 @@ def _upsert_field_permissions(self, metadata, api_name):
...
                         field_permission.get("readable", True)
                     ).lower()
-                    elem.find(f"{MD}editable").text = str(
+                    elem.find(f"editable").text = str(
```
davidmreed:
```suggestion
                    elem.find("editable").text = str(
```

cumulusci/tasks/metadata_etl/value_sets.py
```diff
@@ -49,39 +44,27 @@ def _transform_entity(self, metadata, api_name):
                         "CaseStatus standard value set entries require the key 'closed'"
                     )
 
-            new_entry_index = get_new_tag_index(metadata, "standardValue")
+            metadata.append(tag="standardValue")
```
davidmreed: I don't think this is right. We don't want to insert an empty `<standardValue>` tag yet.

cumulusci/tasks/metadata_etl/value_sets.py
```diff
@@ -49,39 +44,27 @@ def _transform_entity(self, metadata, api_name):
...
+
             if not existing_entry:
                 # Entry doesn't exist. Insert it.
-                elem = etree.Element(f"{MD}standardValue")
-                metadata.getroot().insert(new_entry_index, elem)
-
-                elem_fullName = etree.SubElement(elem, f"{MD}fullName")
-                elem_fullName.text = entry["fullName"]
+                elem = metadata.append(tag="standardValue")
+                elem.append(tag="fullName", text=entry["fullName"])
```
davidmreed:
```suggestion
                elem.append("fullName", text=entry["fullName"])
```

cumulusci/tasks/metadata_etl/value_sets.py
```diff
@@ -49,39 +44,27 @@ def _transform_entity(self, metadata, api_name):
...
 
-                elem_label = etree.SubElement(elem, f"{MD}label")
-                elem_label.text = entry["label"]
+                elem.append(tag="label", text=entry["label"])
```
davidmreed:
```suggestion
                elem.append("label", text=entry["label"])
```

cumulusci/tasks/metadata_etl/value_sets.py
```diff
@@ -49,39 +44,27 @@ def _transform_entity(self, metadata, api_name):
...
 
-                elem_default = etree.SubElement(elem, f"{MD}default")
-                elem_default.text = "false"
+                elem.append(tag="default", text="false")
```
davidmreed:
```suggestion
                elem.append("default", text="false")
```

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -250,19 +249,15 @@ def _process_static_resources(self, zip_src):
...
-            section = ET.Element("{{{}}}types".format(self.namespaces["sf"]))
-            name = ET.Element("{{{}}}name".format(self.namespaces["sf"]))
-            section.append(name)
-            name.text = "StaticResource"
-            tree.find(".//sf:types[last()]", self.namespaces).addnext(section)
+        Package = metadata_tree.parse(package_xml)
+        sections = Package.findall("types", name="StaticResource")
+        section = sections[0] if sections else None
+        if not section:
+            section = Package.append(tag="types")
```
davidmreed:
```suggestion
            section = Package.append("types")
```

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -250,19 +249,15 @@ def _process_static_resources(self, zip_src):
...
+            section.append(tag="name", text="StaticResource")
```
davidmreed:
```suggestion
            section.append("name", text="StaticResource")
```

cumulusci/util/xml/metadata_tree.py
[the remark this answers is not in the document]
prescod: e9deea920e2b8e2c

prescod:
> This doesn't look like an approach that is likely to work on Windows. Can we just check isinstance(source, pathlib.PurePath), and if so use `str(source)`?

Actually th

prescod:
> This doesn't look like an approach that is likely to work on Windows. Can we just check isinstance(source, pathlib.PurePath), and if so use `str(source)`?

Actually the example code seems to be on Windows:

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

I don't feel strongly about it, I guess just an experiment in duck typing. If you feel strongly I can still change it.

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -298,23 +295,12 @@ def _transform(self):
                     parent_dir.mkdir()
                 destination_path = parent_dir / f"{api_name}.{extension}"
 
-                with destination_path.open(mode="wb") as f:
-                    transformed_xml.write(f, encoding="utf-8", xml_declaration=True)
+                with destination_path.open(mode="w", encoding="utf-8") as f:
+                    f.write(
+                        transformed_xml.tostring(xml_declaration=True)
+                    )  # todo: figure out encoding declaration and general salesforce-y encoding
```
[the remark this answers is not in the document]
prescod: 851ec51f

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -28,23 +25,22 @@ class AddRelatedLists(MetadataSingleEntityTransformTask):
         **MetadataSingleEntityTransformTask.task_options,
     }
 
-    def _transform_entity(self, metadata, api_name):
-        new_related_list_index = get_new_tag_index(metadata, "relatedLists")
+    def _transform_entity(
+        self, metadata: MetadataElement, api_name
```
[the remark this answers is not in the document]
prescod: 29d78288

davisagli:
@prescod that example is using `as_posix` to convert from a Windows path to a POSIX path. I think we need to do the opposite (convert from a path of whatever sort was passed to the correct type of path for the OS we're running on, before handing that path to lxml). At the least we should create a copy of this branch with `windows` in the branch name so that we can confirm the tests pass on Windows (the windows build is only triggered for branches with windows in the name, and master, because CircleCI limits our windows builds)

cumulusci/utils/xml/metadata_tree.py
prescod: I guess it was a subconscious reference to this:

https://www.w3schools.com/jsref/met_node_insertbefore.asp

I don't feel strongly about it though. I could go either way.

davisagli: Let's use camel case. This is Python, not Javascript.

cumulusci/utils/xml/metadata_tree.py
prescod: 17a5e7cac4

cumulusci/utils/xml/metadata_tree.py
prescod: 9eae9fee

Although I will note that insertBefore shows up twice in the stdlib.

cumulusci/tasks/metadata_etl/value_sets.py
prescod: Good point. Fixed in af6c60eb

prescod:
I thought of a better way to handle pathlib objects: 2337cac36660d

cumulusci/utils/xml/metadata_tree.py
```diff
@@ -0,0 +1,265 @@
...
+        """
+        if isinstance(item, int):
+            siblings = self._parent.findall(self._element.tag)
+            return MetadataElement(siblings[item], self._parent)
+        elif isinstance(item, str):
+            return self._get_child(item)
+        else:
+            raise TypeError(
+                f"Indices must be integers or strings, not {type(item)}"
+            )  # # pragma: no cover
```
davisagli: Ugh, so we have code that can be reached at runtime but is untestable due to typeguard :(

cumulusci/utils/xml/metadata_tree.py
```diff
@@ -0,0 +1,265 @@
...
+    The parse() function supports any of the following sources:
+
+        * an open file object (make sure to open it in binary mode)
+        * a file-like object that has a .read(byte_count) method returning a byte string on each call
+        * a filename string or pathlib Path
+        * an HTTP or FTP URL string
+
+    Note that passing a filename or URL is usually faster than passing an open file or file-like object. However, the HTTP/FTP client in libxml2 is rather simple, so things like HTTP authentication require a dedicated URL request library, e.g. urllib2 or requests. These libraries usually provide a file-like object for the result that you can parse from while the response is streaming in.
+    """
+    if hasattr(source, "open"):  # for pathlib.Path objects
```
davisagli: While I generally like duck typing, this feels a bit too open (pun intended). For example if someone tries to pass in a ZipFile, this will try to call its open method and will raise a TypeError due to missing a required argument.

davidmreed:
Changes look good to me. I would like to do a sanity test on the MDETL tasks as re-implemented before release.

PR #1563, opened 2020-02-20
Specify excluded and custom buttons to new Related Lists

davidmreed:
# Critical Changes

# Changes

- `add_page_layout_related_lists` now supports suppressing standard buttons and adding custom buttons on related lists.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/metadata_etl/layouts.py
+++ b/cumulusci/tasks/metadata_etl/layouts.py
@@ -19,6 +19,12 @@ class AddRelatedLists(MetadataSingleEntityTransformTask):
             "description": "Array of field API names to include in the related list",
             "required": False,
         },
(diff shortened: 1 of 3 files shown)


cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -19,6 +19,12 @@ class AddRelatedLists(MetadataSingleEntityTransformTask):
             "description": "Array of field API names to include in the related list",
             "required": False,
         },
+        "exclude_buttons": {
+            "description": "Array of button names to suppress from the related list"
+        },
+        "custom_buttons": {
+            "description": "Array of button names to add to the related list"
+        },
```
davisagli: We should be talking about lists, not arrays

PR #1577, opened 2020-02-25
metadeploy_publish task: add support for extracting and publishing translations

davisagli:
Note: Merging needs to be coordinated with https://github.com/SFDO-Tooling/MetaDeploy/pull/1121

# Critical Changes

# Changes
* The `metadeploy_publish` task now has a `labels_path` option which specifies a folder to store translations. After publishing a plan, `labels_en.json` will be updated with the untranslated labels from the plan. Before publishing a plan, labels from other languages will be published to MetaDeploy.

# Issues Closed


cumulusci/tasks/metadeploy.py
```diff
@@ -219,6 +282,8 @@ def _find_or_create_version(self, product):
         return version
 
     def _find_or_create_plan_template(self, product, plan_name, plan_config):
+        # TODO: Add preflight_message, post_install_message, and error_message here
+        # once the MetaDeploy API supports them.
```
codexmarcel: Is this in the backlog?

cumulusci/tasks/tests/test_metadeploy.py
```diff
@@ -189,10 +207,48 @@ def test_run_task(self):
                         "checks": [],
                     },
                 },
+                {
+                    "name": "util_sleep",
```
codexmarcel: Why is the sleep required in this case?

cumulusci/tasks/metadeploy.py
davisagli: @codexmarcel they are added in the MetaDeploy PR you looked at: https://github.com/SFDO-Tooling/MetaDeploy/pull/977/files#diff-7b9b524e363875f277fe63e6a912f5c2R73 (This is a Django Rest Framework serializer used by the view that serves the "admin API" that CumulusCI interacts with. These fields were not included by the default serializer because we have a special field type to handle Markdown and it's not fully integrated with DRF.)

I should be able to make this adjustment here once I deploy the MetaDeploy changes.

cumulusci/tasks/tests/test_metadeploy.py
davisagli: Sleep was just a handy task without side effects to use for the test. It didn't matter what task but I needed to add a step with a step-level preflight check (as opposed to plan-level) to make sure that this line gets test coverage: https://github.com/SFDO-Tooling/CumulusCI/pull/1577/files#diff-0b667355598f51590b2539b31b3b3ca3R169 (and it's a good thing I did, because it revealed that I was looking for the checks in the wrong place)

PR #1580, opened 2020-02-27
Fix test runner for Spring '20

davidmreed:
# Critical Changes

# Changes

- CumulusCI now will not attempt to retry class-level concurrency failures when running Apex unit tests in a managed package due to changes made to the Tooling API in Spring '20. Such failures will be logged but will not cause a build failure.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/apex/testrunner.py
+++ b/cumulusci/tasks/apex/testrunner.py
@@ -331,7 +331,7 @@ class RunApexTests(BaseSalesforceApiTask):
             )
 
             # In Spring '20, we cannot get symbol tables for managed classes.
-            if "__" in class_name:
+            if self.options["managed"]:
                 self.logger.error(
                     f"Cannot access symbol table for managed class {class_name}. Failure will not be retried."
                 )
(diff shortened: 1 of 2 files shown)


davidmreed:
Validated on a parallelized NPSP test run.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -330,6 +330,13 @@ def _get_test_results(self, allow_retries=True):
                 f"Class {class_name} failed to run some tests with the message {error}. Applying error to unit test results."
             )
 
+            # In Spring '20, we cannot get symbol tables for managed classes.
```
davisagli: @davidmreed didn't it appear to be an issue with getting the symbol table for any class with a namespace prefix, whether managed or not?

davidmreed: Yes, but the response from Core was that that appears to be a bug rather than the intended behavior.

The test runner does not discriminate between namespaced and managed, so this should hit both. The only downside is that we'll in after Core fixes the bug treat namespaced runs like managed, but a larger refactor would be needed to fix that.

davisagli: Ah yes, I see how we are using the `managed` option for unmanaged namespaced feature builds as well.

PR #1581, opened 2020-02-27
Truncate warnings

prescod:
I have had problems in LDV orgs when the logs got gargantuan due to per-row errors. This avoids that.

### DIFF
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -8,7 +8,11 @@ import yaml
 
 from cumulusci.core.exceptions import BulkDataException, TaskOptionsError
 from cumulusci.core.utils import process_bool_arg
-from cumulusci.tasks.bulkdata.utils import get_lookup_key_field, SqlAlchemyMixin
+from cumulusci.tasks.bulkdata.utils import (
+    get_lookup_key_field,
+    SqlAlchemyMixin,
+    check_for_row_error,
+)
 from cumulusci.tasks.bulkdata.step import (
     BulkApiDmlOperation,
     DataOperationStatus,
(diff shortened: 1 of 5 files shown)


prescod:
> Are you coordinating this with @davidmreed? I would guess it'll take some care to merge with his refactor.

This is a sub-branch of his.

davisagli:
@prescod oops, I didn't see that. I'll let him do the main review then, but I have no objection.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -320,14 +325,13 @@ def _generate_results_id_map(self, step, local_ids):
...
-                if self.options["ignore_row_errors"]:
-                    self.logger.warning(
-                        f"Error on record with id {local_id}: {result.error}"
-                    )
-                else:
-                    raise BulkDataException(
-                        f"Error on record with id {local_id}: {result.error}"
-                    )
+                check_for_row_error(
+                    self,
```
davidmreed: I find the way this function stashes state inside the task unintuitive. Lacking that, the function doesn't need a reference to the task at all. Could we just use a local counter?

davidmreed: (Otherwise, I'm fine with this change).

prescod: Okay I got rid of the weirdness but it did push some extra complexity into the Task class. I wasn't totally happy the way it was but not the way it is either. I probably should have just left it as two files that shared 10 lines of identical code. Which was also not ideal.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -315,19 +320,16 @@ def _process_job_results(self, mapping, step, local_ids):
     def _generate_results_id_map(self, step, local_ids):
```
prescod: @davidmreed : I'm inclined to change this to take just the generator instead of the whole step. Do you agree?

prescod: My first argument is that it obscures the control flow somewhat to have get_results happen in a sub-method instead of the main method.

My second argument is that it makes clearer that the step itself is not modified or even read by the method. Just the results are used.

davidmreed: @prescod  I'm fine with that change.

PR #1582, opened 2020-02-27
Fix docs

davisagli:
Okay, I got a bit carried away but with this, you can run `make docs` and then look at docs/_build/html/api/cumulusci.utils.xml.html#module-cumulusci.utils.xml.metadata_tree for the generated metadata_tree docs. I would suggest adding a module docstring to provide an introduction.

The other things in the PR are:
- fixing some reST errors that were showing warnings while building the docs
- switched to the "alabaster" sphinx theme which is the current default and which I find nicer
- fixed the main table of contents to avoid listing all tasks and release versions
- tweaked sphinx-apidoc arguments for better output and to exclude test modules

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/apex/testrunner.py
+++ b/cumulusci/tasks/apex/testrunner.py
@@ -90,20 +90,20 @@ class RunApexTests(BaseSalesforceApiTask):
     may automatically be retried. Note that retries are supported whether or not
     the org has parallel Apex testing enabled.
 
-    The `retry_always` option modifies this behavior: if a test run fails and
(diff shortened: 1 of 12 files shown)


prescod:
Could this be a PR against master instead?

davisagli:
> Could this be a PR against master instead?

@prescod I'm fine with that as long as the __init__.py gets added in cumulusci/utils/xml

prescod:
Okay if you rebase it I'll approve it.

davisagli:
@prescod rebased

PR #1585, opened 2020-02-28
Rebuild UpdateAdminProfile as a Metadata ETL task

davidmreed:
This PR converts UpdateAdminProfile/UpdateProfile to a Metadata ETL task called ProfileGrantAllAccess.

- [x] Unit testing
- [x] End to end testing
- [ ] MetaDeploy testing
- [x] Deal with any breaking changes in projects that subclasses UpdateAdminProfile
  - [x] NPSP
  - [x] EDA
  - [x] K12
  - [x] CaseMan
  - [x] FGM-Portal
  - [x] FC-Reviewer
- [x] Validate fix for CCI Trailhead

# Critical Changes

# Changes

- The `UpdateAdminProfile`/`UpdateProfile` task has been revised to use the new Metadata ETL framework as `ProfileGrantAllAccess`. This is a breaking change for local tasks which subclassed `UpdateAdminProfile`. The new task supports updating arbitrary Profiles in addition to System Administrator. 
- **Note**: At time of release, please update `update_profile.py` line 89 to actual version number.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
+++ b/cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
@@ -356,3 +356,11 @@ def test_generate_package_xml__retrieve():
         task._generate_package_xml(MetadataOperation.RETRIEVE)
         task._expand_profile_members.assert_called_once()
         task._expand_package_xml.assert_not_called()
+
+
+def test_init_options_raises_combined_options():
+    with pytest.raises(TaskOptionsError):
+        create_task(
+            ProfileGrantAllAccess,
+            {"package_xml": "admin_profile.xml", "profile_name": "test"},
+        )
--- a/cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
+++ b/cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
@@ -5,7 +5,7 @@ from unittest import mock
 
 import pytest
 
-from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.core.exceptions import TaskOptionsError, CumulusCIException
 from cumulusci.tasks.metadata_etl import MetadataOperation
 from cumulusci.tasks.salesforce import ProfileGrantAllAccess
 from cumulusci.utils import CUMULUSCI_PATH
@@ -150,9 +150,7 @@ def test_run_task():
 
         dest_path = task.deploy_dir / "profiles" / "Admin.profile"
         assert dest_path.exists()
-        assert dest_path.read_bytes().decode("utf-8") == ADMIN_PROFILE_EXPECTED.decode(
-            "utf-8"
-        )
+        assert dest_path.read_bytes() == ADMIN_PROFILE_EXPECTED
 
 
 def test_transforms_profile():
@@ -191,7 +189,7 @@ def test_throws_exception_record_type_not_found():
 def test_expand_package_xml():
     task = create_task(ProfileGrantAllAccess, {"api_names": ["Admin"]})
     task.tooling = mock.Mock()
-    task.tooling.query.return_value = {
+    task.tooling.query_all.return_value = {
         "totalSize": 2,
         "records": [
             {"DeveloperName": "Test", "NamespacePrefix": "ns"},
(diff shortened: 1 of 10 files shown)


davidmreed:
Looks like I have a couple of extra lines to cover; more unit tests forthcoming.

davidmreed:
I've tested this revision extensively (see https://salesforce.quip.com/jDZhAxJAUjbM for test charter) and found that it works, corrects the issue we need to address by mid-March, and also fixes a small bug in the existing `update_admin_profile` task, which does not properly update Apex class accesses.

For the two projects that have in-use subclasses of UpdateAdminProfile (EDA and K12), I'm suggesting that those projects establish a local copy of the previous version of that task for subclassing, until figuring out how to insert this new task into their automation.

cumulusci/tasks/salesforce/update_profile.py
```diff
@@ -47,168 +65,182 @@ def _init_options(self, kwargs):
...
+    def _expand_profile_members(self, package_xml):
+        profile_names = package_xml.find("types", name="Profile")
+        for profile in self.api_names:
+            profile_names.append("members", text=profile)
+
+    def _expand_package_xml(self, package_xml):
+        # Query the target org for all namespaced objects
+        # Add these entities to the package.xml
+
+        results = self.tooling.query(
```
davisagli: I think we need to use query_all in case there is more than one page of results

cumulusci/tasks/salesforce/update_profile.py
```diff
@@ -47,168 +65,182 @@ def _init_options(self, kwargs):
...
+            "SELECT DeveloperName, NamespacePrefix FROM CustomObject WHERE ManageableState != 'unmanaged'"
         )
 
-    def _set_apps_visible(self):
-        xpath = ".//sf:applicationVisibilities[sf:visible='false']"
-        for elem in self.tree.findall(xpath, self.namespaces):
-            elem.find("sf:visible", self.namespaces).text = "true"
+        custom_objects = package_xml.find("types", name="CustomObject")
```
davisagli: If it's not there we should add it or throw a helpful exception.

cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
```diff
@@ -0,0 +1,366 @@
...
+        task.deploy_dir = pathlib.Path(tempdir, "deploy")
+        task.deploy_dir.mkdir()
+
+        task()
+
+        dest_path = task.deploy_dir / "profiles" / "Admin.profile"
+        assert dest_path.exists()
+        assert dest_path.read_bytes().decode("utf-8") == ADMIN_PROFILE_EXPECTED.decode(
+            "utf-8"
+        )
```
davisagli: Why not just compare the bytes, or convert ADMIN_PROFILE_EXPECTED to text and use dest_path.read_text()?

cumulusci/tasks/salesforce/update_profile.py
davidmreed: Yeah, you're right. I wish `simple_salesforce` did not use `query_all` to mean something other than "include deleted records" (which is what it means at the API level).

davisagli: yep. I wonder if we can goad @prescod into writing a better Salesforce REST API client as his next weekend project. :-p

davidmreed: We should look at https://github.com/forcedotcom/SalesforcePy

Not least because we could probably get contributor rights.

cumulusci/tasks/salesforce/update_profile.py
davidmreed: Opting for throwing an exception in both of these cases because they indicate a package.xml that is not properly configured for this task; taking action could produce unexpected results.

cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
davidmreed: Tied myself in a knot with an earlier version of this unit test. Corrected to just compare the bytes.

davidmreed:
> Have you thought through possible implications for uses of update_admin_profile in MetaDeploy plans that were already published? There could be big problems if there's an existing plan using a custom subclass of this task because we don't have the ability to update the subclass. At first glance I think it is only used in a couple MD plans though (Abacus limited release and NPSP trailhead).

I can't locate any MetaDeploy plans, at least the current versions of MetaDeploy plans, that utilize a subclass of `UpdateAdminProfile`. CaseMan's was used elsewhere and is superseded; NPSP didn't actually have a subclass; and the subclasses present in two foundationConnect projects were spurious and unused.

Abacus appears to be using the released version (no subclassing).

I do want to test this on Staging to catch any issues around freezing steps.

> It sounds like updates to the subclasses in EDA and K12 would need to happen before merging this?

Yes. It would break EDA and K12 for any build running on CCI `master`. I'll work with James and Josh on that.

davidmreed:
Prerequisite PRs:

EDA: https://github.com/SalesforceFoundation/EDA/pull/1113
K12: https://github.com/SalesforceFoundation/k12-architecture-kit/pull/176

cumulusci/tasks/salesforce/update_profile.py
```diff
@@ -47,168 +65,190 @@ def _init_options(self, kwargs):
...
-            package_xml_content = f.read()
+        # We enable new functionality to extend the package.xml to packaged objects
+        # by default only if we meet specific requirements: the project has to require
+        # CumulusCI >= 3.8.1 (i.e., creation date or opt-in after release), and it must
+        # not be using a custom `package.xml`
+        min_cci_version = self.project_config.minimum_cumulusci_version
+        if min_cci_version and "package_xml" not in self.options:
+            parsed_version = pkg_resources.parse_version(min_cci_version)
+            default_packages_arg = parsed_version >= pkg_resources.parse_version(
+                "3.8.1"
```
davidmreed: @davisagli  Would the pre-trail release go out as 3.8.1 or as 3.9.0? (I suppose this will work either way, but for clarity's sake).

davisagli: Depends on whether the release includes new features. Probably 3.9.0. You can leave this is, put a note in the PR description to remind me, and I'll adjust when I prepare the release notes.

PR #1593, opened 2020-03-05
Add support for GitHub Enterprise

jlantz:
- Use a real parser for repo url instead of a dumb split.  Convert all repo urls to https format since that's what's required for calling the GitHub API
- Changes cumulusci.core.github.get_github_api_for_repo to accept args for the project_config and the repo_url
- Use a mapping of service names to github site base url in project.git.sites in cumulusci.yml to look up service names for different repo base urls.  The custom service still has to be defined under services for the mapping to work

# Critical Changes

# Changes

# Issues Closed


jstvz:

PR #1596, opened 2020-03-07
Convert NBSP to space in cci.yml

prescod:
Only converts in leading space blocks.

Outputs a warning.


cumulusci/core/config/project_config.py
```diff
@@ -86,15 +88,15 @@ def _load_config(self):
...
+            project_config = cci_safe_load(f_config)
 
         if project_config:
             self.config_project.update(project_config)
 
         # Load the local project yaml config file if it exists
         if self.config_project_local_path:
             with open(self.config_project_local_path, "r") as f_local_config:
-                local_config = yaml.safe_load(f_local_config)
+                local_config = cci_safe_load(f_local_config)
```
davisagli: There are two other calls to yaml.safe_load in the project_config module that should probably also be adjusted.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -0,0 +1,34 @@
...
+            "Future versions of CumulusCI may disallow these characters.\n"
+        )
+    return rc
+
+
+def _replace_nbsp(data):
+    return pattern.sub(_replacer_func, data)
+
+
+def cci_safe_load(f_config: IO[Text]):
```
davisagli: docstring please

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -0,0 +1,34 @@
...
+from logging import getLogger
+from io import StringIO
+
+import yaml
+
+NBSP = "\u00A0"
+
+pattern = compile(r"^\s*[\u00A0]+\s*", MULTILINE)
+
+logger = getLogger("CumulusCI.yml")
```
davisagli:
```suggestion
logger = getLogger(__name__)
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def _replacer_func(matchobj):
+    string = matchobj.group(0)
+    rc = string.replace(NBSP, " ")
+    if rc != string:  # this is fast for identical strings
+        logger.warn(
+            "Note: Non-breaking space character was detected in Cumulusci.yml.\n"
```
davisagli: We merge together the project config from cumulusci.yml files in several different sources. It would be ideal to identify the path to the file (which would require passing it in to cci_safe_load)

cumulusci/utils/yaml/tests/test_cumulusci_yml.py
```diff
@@ -0,0 +1,40 @@
+from io import StringIO
+
+from cumulusci.utils.yaml.cumulusci_yml import cci_safe_load, _replace_nbsp
+
+
+# fill this out.
```
davisagli: ?

cumulusci/utils/yaml/tests/test_cumulusci_yml.py
```diff
@@ -0,0 +1,40 @@
...
+    def test_convert_nbsp(self, caplog):
+        yaml = """xyz:
+           \u00A0 y: abc"""
+        cciyml = cci_safe_load(StringIO(yaml))
+        assert "space character" in caplog.text
+
+        assert isinstance(cciyml, dict)  # should parse despite funny character
+        assert cciyml["xyz"]["y"] == "abc", cciyml
+
+    def test_converter(self, caplog):
```
davisagli: caplog not used

cumulusci/utils/yaml/tests/test_cumulusci_yml.py
```diff
@@ -0,0 +1,40 @@
...
+        inp = """xyz:
+           \u00A0 y: abc"""
+        outp = """xyz:
+             y: abc"""
+
+        rc = _replace_nbsp(inp)
+        assert rc == outp
+
+    def test_converter_is_selective(self, caplog):
```
davisagli: caplog not used

cumulusci/core/config/project_config.py
prescod: I had done that but then undid it on the premise that I should make this kind-of-rushed change as low-impact as possible, especially given that there is no reason to think that those obscure CCI.yml files would ever end up with an NBSP in them.

Once the real CCI.yml parser is integrated, all CCI.yml files will be treated the same way, I agree.

If you feel strongly it should be fairly low-risk to change them all, though.

cumulusci/utils/yaml/cumulusci_yml.py
prescod: That's another feature I removed in the interest of simplicity for this small short-term patch. I can put it back in if you feel strongly.

prescod: Most of it will be replaced when I merge the YAML validator thing, which does take the filename and a logger argument etc..

cumulusci/utils/yaml/cumulusci_yml.py
prescod: Fixed.

https://github.com/SFDO-Tooling/CumulusCI/pull/1596/files#diff-9c0b3c5a896ba84e51be4be79b9a8414R12

cumulusci/utils/yaml/cumulusci_yml.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1596/files#diff-9c0b3c5a896ba84e51be4be79b9a8414R38

cumulusci/core/config/project_config.py
davisagli: I disagree that there's no risk of them ending up with an NBSP, but I'm fine holding off since we anticipate introducing a more robust and consistent validation soon.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -0,0 +1,41 @@
+from typing import IO, Text
+from re import compile, MULTILINE
+from logging import getLogger
+from io import StringIO
+
+import yaml
+
+NBSP = "\u00A0"
+
+pattern = compile(r"^\s*[\u00A0]+\s*", MULTILINE)
```
davisagli: This is a nitpick but I would prefer importing the `re` module and then using `re.compile` here so that it doesn't look like the `compile()` builtin.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -0,0 +1,41 @@
...
+        string = matchobj.group(0)
+        rc = string.replace(NBSP, " ")
+        return rc
+
+    data = pattern.sub(_replacer_func, origdata)
+
+    if counter:
+        plural = "s were" if counter > 1 else " was"
+        logger.warn(
+            f"Note: {counter} non-breaking space character{plural} detected in cumulusci.yml.\n"
```
davisagli: The code is counting the number of lines that matched; not the number of characters.

prescod: I changed the message to clarify. If this message frequently gets invoked, in a future version, could try to help the user pinpoint to a line number.

prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1596/commits/1b455898665ceedec330dea68e53ece4719b2898#diff-9c0b3c5a896ba84e51be4be79b9a8414R30

cumulusci/utils/yaml/cumulusci_yml.py
prescod: Fixed.

PR #1606, opened 2020-03-13
Integration testing task.

prescod:
Draft pull request of a feature for integration testing CumulusCI. Using pytest CLI syntax instead of CCI task syntax.

Once this is merged we could later extend it to:

 a) run automatically in CI (perhaps after merge to master, or daily, or something like that)

 b) expose these capabilities to other repositories with custom tasks.

(the latter might just be minimal conftest.py files which import and re-export names from this one, or it might be a new CCI task which does that "automatically")


tests/README.rst
```diff
@@ -0,0 +1,15 @@
+This directory contains the Pytest tests which depend on a Salesforce org
+or other remote services.
```
davisagli, line 2: Does this file belong in /tests/integration instead of /tests?

tests/README.rst
```diff
@@ -0,0 +1,15 @@
...
+
+Pytest integration tests are hidden from ordinary `pytest` because they are in
+a different directory.
+
+You can invoke these test with the CCI task:
```
davisagli, line 7:
```suggestion
You can invoke these tests with the command:
```

PR #1609, opened 2020-03-13
Add support for CHROME_LOG variable in robot tests

boakley:
Setting this variable to a filename will cause chrome logs to be
written to that file.

I'm creating this PR even though I'm having a dickens of a time getting all of the tests to pass on circleci. The new test passes, but I am seeing timeout failures on the app menu in one of the other tests. <scratches head>

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.robot
+++ b/cumulusci/robotframework/Salesforce.robot
@@ -35,6 +35,7 @@ ${INITIAL_TIMEOUT}  180.0
 ${TIMEOUT}          30.0
 ${LOCATION STRATEGIES INITIALIZED}  ${False}
 ${DEFAULT BROWSER SIZE}  1280x1024
+${CHROME LOG}       ${EMPTY}
 
 *** Keywords ***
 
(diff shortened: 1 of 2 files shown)


davidmreed:
Bryan, is the use case here local runs? Do we have any route to access this log file if a task were run with CHROME_LOG active on MetaCI?

davisagli:
https://chromedriver.chromium.org/logging says "By default ChromeDriver logs only warnings/errors to stderr." I assume the condition we're investigating counts as an error. So maybe we only need to add log-path and not verbose, for a smaller log.

boakley:
The use case was to do this on MetaCI. However, over the weekend I was
looking at that, and MetaCI will require a change to persist the data. I
may end up having to press the pause button on this until I can update
MetaCI, and that may take some work. I don't want to have to do a quick
patch of MetaCI, I'd rather do a proper fix that makes it easy to persist
any data, not just this one file.

davisagli:
@boakley MetaCI Already has a BuildFlowAsset model for storing a file in S3 that's associated with a particular BuildFlow. So I think we would just need to:
- decide how to configure which files should get stored as artifacts
- expose a way to access artifacts in the UI

davisagli:
Closing; not opposed to the goal here but in my testing it was causing chromedriver to log rather than chrome itself.

PR #1621, opened 2020-03-18
Improve handling when repo or release are not found

davisagli:
# Critical Changes

# Changes

# Issues Closed

* Show a more helpful error message if CumulusCI can't find a project's repository or release on GitHub.

### DIFF
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -4,6 +4,7 @@ import re
 
 API_VERSION_RE = re.compile(r"^\d\d+\.0$")
 
+import github3
 import yaml
 
 from cumulusci.core.utils import merge_config
@@ -11,6 +12,7 @@ from cumulusci.core.config import BaseTaskFlowConfig
 from cumulusci.core.exceptions import (
     ConfigError,
     DependencyResolutionError,
+    GithubException,
     KeychainNotFound,
     NamespaceNotFoundError,
     NotInProject,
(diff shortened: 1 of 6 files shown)


cumulusci/core/config/project_config.py
```diff
@@ -385,6 +398,9 @@ def _get_latest_tag_for_prefix(self, repo, prefix):
             if not release.tag_name.startswith(prefix):
```
prescod: I wrote this alternative but it may not be better enough to be worth including. Up to you:

```
    def _get_latest_tag_for_prefix(self, repo, prefix):
        relevant_releases = (
            release
            for release in repo.releases()
            if release.tag_name.startswith(prefix)
        )
        first_release = next(relevant_releases, None)
        if first_release:
            return first_release.tag_name
        else:
            raise GithubException(
                f"No release found for {self.repo_url} with tag prefix {prefix}"
            )
```

davisagli: To be honest I think it's easier to follow with a simple loop instead of a generator expression.

prescod: Just for the record, my thinking is that you can follow the logic just by reading the variable names:

````
    relevant_releases = (whatever)
    first_release = next(relevant_releases, None)
    if first_release:
            return first_release.tag_name
 ```

prescod: But its more lines of code so, shrug.

davisagli: And you have to know about generators to understand that it's lazy.

PR #1622, opened 2020-03-18
Proposed change to resolve import cycle

davisagli:
This is my attempt to fix the import cycle (W-7346064) by:
1. moving BaseSalesforceTask into cumulusci.core.tasks so that cumulusci.tasks.metadata_etl.base can import it without trying to import other tasks that use the metadata ETL framework
2. moving cumulusci.tasks.metadata_etl.base's import of the Deploy task to happen at runtime

I also switched BaseMetadataETLTask to extend BaseSalesforceTask rather than BaseSalesforceApiTask since metadata ETL tasks do not necessarily use the REST API. Let me know if you have concerns about this @davidmreed -- and could you help confirm that I didn't break the set_organization_wide_defaults task with this change?

# Critical Changes

# Changes

# Issues Closed
* Fixed a bug that caused errors when trying to import metadata ETL tasks.

### DIFF
--- a/cumulusci/core/tasks.py
+++ b/cumulusci/core/tasks.py
@@ -8,7 +8,9 @@ import os
 import time
 import threading
 
+from cumulusci import __version__
 from cumulusci.utils import cd
+from cumulusci.core.exceptions import ServiceNotValid, ServiceNotConfigured
 from cumulusci.core.exceptions import TaskRequiresSalesforceOrg
 from cumulusci.core.exceptions import TaskOptionsError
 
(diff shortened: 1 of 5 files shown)


davidmreed:
This makes sense on a read-through, @davisagli, and I don't mind at all have MDETL tasks not subclass something (`BaseSalesforceApiTask`) that they may not need. I will do a hands-on test of the org wide defaults task tomorrow. (That's also the task that is most badly broken in namespaced scratch orgs, which I need to fix separately).

davidmreed:
Regression on MDETL tasks passes with this change, @davisagli.

davidmreed:
This is a potential breaking change for any custom Metadata ETL tasks. Because the framework is so new, I'm hoping there are few to none of those. I will address the one that I'm aware of, on fC to patch namespaced org issues pending a global fix.

PR #1623, opened 2020-03-18
Add support for an `include_stopped` parameter to BatchApexWait

force2b:
# Critical Changes

# Changes

- Modify the batch_apex_wait task to include Aborted and Failed jobs by default

# Issues Closed


prescod:
I don't think the failing test is your fault. Don't worry about that.

But I wanted to discuss the include_stopped parameter. What are the use-cases where you would want it to be false? If its false the CCI process will presumably wait forever even though it does not have to. Also, shouldn't we also handle the Failed state?

https://help.salesforce.com/articleView?id=code_apex_job.htm&type=5

force2b:
@prescod Good point. I had though that Completed might include a failed job, and though it would if only one batch iteration failed out of more than one. However, if there is a failure of 1 out of 1 batch iteration, or the job fails with an initial query timeout then it would be listed as "Failed". 

I think your other question is why not always include Failed and Aborted in the initial check. Honestly I'm not really sure. I was mainly avoiding modifying the default behavior of the task, however it's feasible to say that maybe this should always be the case. And if so, then maybe the options would be there to _disable_ including those particular statuses in cases there really is a flow that needs to check explicitly for "Completed".

- `include_stopped` (or `include_aborted`): Check for Status="Aborted" (default to true)
- `include_failed': Check for Status="Failed" (default to true)

prescod:
I meant to do this as a pull request to your pull request but I pushed before I branched. Didn't seem worth undoing.

In the ProdRelEng discussion we decided to try to apply these philosophies:

 * noisy failure are better than quiet ones
 * only add options when someone really describes why they need them ("YAGNI")

The backwards compatible behaviour is to wait forever for stopped jobs, which is basically a bug. So I presume that nobody has really run into this issue in the past. Still, it's better to fix the bug and not hang forever. YAGNI is not an excuse for bugs.

But since the backwards compatible behaviour is a bug we should think through what we would have done if we had implemented it completely 2.5 years ago. And I think that throwing exceptions is how we would have done it.

If the Flow-writer wants the flow to continue even in the face of the aborted or failed job, they can do so with the feature called `ignore_failure` on the flow steps.

Also, it may not have been clear from the code but a failed batch will not generate a failed job and that's part of why we've probably(?) never run into this before. Failed batches are common but failed jobs seem quite rare.

force2b:
@prescod Thanks for the cleanup on this. Much appreciated. Hopefully the next PR on a field value wait state will be in better shape.

PR #1624, opened 2020-03-19
CumulusCI.yml linter based on Pydantic

prescod:
Validates CumulusCI.yml files now. If they do not match the pattern, it will generate a warning.

Incorporates the NBSP cleanup tool from the other branch.


cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -27,15 +237,22 @@ def _replacer_func(matchobj):
...
+    loc: Sequence[Union[str, int]]
+    msg: str
+    type: str
+
+
+def _log_yaml_error(logger, error: ErrorDict):
+    "Format and log a Pydantic-style error dictionary"
+    logger.warning("CumulusCI Parsing Error:")
+    loc = " -> ".join((repr(x) for x in error["loc"]))
+    logger.warning("%s : %s", loc, error["msg"])
```
Br4nd0R, line 258: can we not use an fstring here?

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,18 +1,228 @@
...
+    logger = logger or default_logger
+
+    with load_from_source(source) as (filename, data_stream):
+        # this is inelegant but the _replace_nbsp code is a lot easier
+        # to write with regexps and Python regexps don't work with streams
+        cleaned_up_data = _replace_nbsp(data_stream.read(), data_stream, logger)
+        data = safe_load(StringIO(cleaned_up_data))
+        context = context or filename
+
+        on_error = on_error or (lambda error: _log_yaml_error(logger, error))
```
Br4nd0R, line 200: potentially move this closer to 191

Br4nd0R (dismissed):
This looks good to me. Thanks for letting me walk myself through the code with you @prescod!

davisagli (changes requested):
I'd like to refresh my memory on this one and do some checking to make sure existing cumulusci.yml files pass validation before approving.

Also will need to have master merged in.

davisagli:
@prescod A big part of why I let this one stall out is because I'm nervous about breaking things because there is something used in the wild that we missed including in the validator. Could we start with this in a mode that is either opt-in or logs warnings rather than aborting, so that we have a way to try it out and find the gaps?

prescod:
@davisagli 

If my time machine is properly functioning then that's how it already works. :) 

It returns the data along with printing warnings about problems.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    install_class: str = None
+    uninstall_class: str = None
+    api_version: str = None
+
+
+class Test(CCIDictModel):
+    name_match: str
+
+
+class Parser(CCIDictModel):
```
davisagli, line 71:
```suggestion
class ReleaseNotesParser(CCIDictModel):
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    class_path: PythonClassPath
+    title: str
+
+
+class ReleaseNotes(CCIDictModel):
+    parsers: Dict[int, Parser]  # should probably be a list
```
davisagli, line 77:
```suggestion
    parsers: Dict[int, Parser]
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    prefix_feature: str = None
+    prefix_beta: str = None
+    prefix_release: str = None
+    push_prefix_sandbox: str = None
+    push_prefix_production: str = None
+    release_notes: ReleaseNotes = None
+    two_gp_context: str = Field(None, alias="2gp_context")
+
+
+class ApexDoc(CCIDictModel):
```
davisagli, line 92: Remove this one, it's no longer used anywhere.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    homepage: str = None  # can't find these in the CCI Python code
+    banner: str = None
+    branch: str = None
+    repo_dir: Path = None
+
+
+class Check(CCIDictModel):
```
davisagli, line 99:
```suggestion
class PreflightCheck(CCIDictModel):
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    when: str = None
+    action: str = None
+    message: str = None
+
+
+class Plan(CCIDictModel):  # MetaDeploy plans
+    title: str
+    description: str = None
+    tier: str = None
```
davisagli, line 108: enum choice between primary, secondary, additional

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    slug: str = None
```
davisagli, line 109: required

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    is_listed: bool = None
```
davisagli, line 110:
```suggestion
    is_listed: bool = True
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    steps: Dict[str, Union[FlowReference, TaskReference]]
+    checks: List[Check] = None
```
davisagli, line 112:
```suggestion
    checks: List[Check] = []
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    group: str = None
+    error_message: str = None
+    error_message: str = None
```
davisagli, line 115:
```suggestion
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    post_install_message: str = None
+    preflight_message: str = None
+
+
+class Project(CCIDictModel):
+    name: Optional[str]
```
davisagli, line 121: required

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    package: Package = None
+    test: Test = None
+    git: Git = None
```
davisagli, line 124: We don't need to make these optional if they have values in the universal config, right?

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    dependencies: List[Dict[str, URL]] = None  # TODO
```
davisagli, line 125: The values of the dict are not all URLs. We should follow up to make this use the models that David Reed has been defining in his dependencies refactor.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    apexdoc: ApexDoc = None
```
davisagli, line 126:
```suggestion
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    source_format: Literal["sfdx", "mdapi"] = "mdapi"
```
davisagli, line 127: use enum?

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+
+
+class ScratchOrg(CCIDictModel):
+    config_file: Path
+    days: int = None
+    namespaced: str = None
```
davisagli, line 133: Add setup_flow: str = None

(It's not used by CumulusCI yet but it can be used by Metecho)

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+
+
+class Orgs(CCIDictModel):
+    scratch: Dict[str, ScratchOrg]
+
+
+class Attribute(CCIDictModel):
```
davisagli, line 140:
```suggestion
class ServiceAttribute(CCIDictModel):
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    validator: PythonClassPath = None
+
+
+class CumulusCIConfig(CCIDictModel):
+    keychain: PythonClassPath
+
+
+class Source(CCIDictModel):
+    github: URL = None
+    release: str = None
```
davisagli, line 157: There are other keys supported here; see https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/source/github.py#L43

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+
+
+class CumulusCLIConfig(CCIDictModel):
+    show_stacktraces: bool = None
```
davisagli, line 161:
```suggestion
    show_stacktraces: bool = False
```

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+
+def _log_yaml_error(logger, error: ErrorDict):
+    "Format and log a Pydantic-style error dictionary"
+    logger.warning("CumulusCI Parsing Error:")
+    loc = " -> ".join(repr(x) for x in error["loc"] if x != "__root__")
+    logger.warning("%s : %s", loc, error["msg"])
+    logger.error(
+        "NOTE: These errors will cause major problems in future versions of CumulusCI."
+    )
+    logger.error("Please correct them before May 1, 2021.")
```
davisagli, line 217: Let's wait and see how many errors there are in the wild before committing to a timeline.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    on_error = on_error or (lambda error: _log_yaml_error(logger, error))
+
+    logger = logger or default_logger
+
+    with load_from_source(source) as (data_stream, filename):
+        data = load_yaml_data(data_stream, filename)
+        context = context or filename
+
+        try:
+            validate_data(data, context=context, on_error=on_error)
```
davisagli, line 241: I think we want to validate the final merged project config, not individual config files as they are loaded. Otherwise we can't show errors about things that are required but missing.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+                        "loc": (context,),
+                        "msg": f"Error validating cumulusci.yml {e}",
+                        "type": "exception",
+                    }
+                )
+            pass
+        return data
+
+
+# validate YML files as a CLI for smoke testing
```
davisagli, line 257: Why not make this a full-fledged cci command instead of a separate entry point?

cumulusci/utils/yaml/safer_loader.py
```diff
@@ -16,15 +16,11 @@
...
+default_logger = getLogger(__name__)
 
 
-def _replace_nbsp(origdata):
-    """Replace non-breaking spaces with spaces.txt
-
-    They sometimes arise in copy and paste from Trailhead or
-    elsewhere and are hard to debug."""
-
+def _replace_nbsp(origdata, filename, logger=default_logger):
```
davisagli, line 22: I think we probably can always use the default logger. Generally when we want to redirect log output we do that by changing the handlers attached to the "cumulusci" logger (which also processes log entries from other loggers starting with `cumulusci.`)

cumulusci/utils/yaml/safer_loader.py
```diff
@@ -16,15 +16,11 @@
...
+    "Replace nbsp characters in leading whitespace in a YAML file."
```
davisagli, line 23:
```suggestion
    """Replace nbsp characters in leading whitespace in a YAML file."""
```

davisagli (changes requested):

cumulusci/utils/yaml/cumulusci_yml.py
prescod: If I use enums, consuming code will need to make more pervasive changes when it starts consuming the models instead of the YAML parser output.

cumulusci/utils/yaml/cumulusci_yml.py
prescod: I think it is more helpful to the user if we validate each file by itself. How can we tell them where their error is if the "missing value" could be anywhere in four files.

cumulusci/utils/yaml/cumulusci_yml.py
prescod: Should my PR depend on his? Or should I use ANY or object for the time being?

cumulusci/utils/yaml/cumulusci_yml.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1624#discussion_r606113698

cumulusci/utils/yaml/cumulusci_yml.py
prescod: I only intended it for the CCI developer team to use.

cumulusci/utils/yaml/cumulusci_yml.py
prescod: I think we'll need to add additional methods to each model to do an extra "semantic" validation after merger. That would be a separate WI.

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: Isn't that the goal -- to be able to make the more pervasive changes enabled by being able to assume the input is already validated?

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,276 @@
...
+    task: str
+    options: Dict[str, Any] = {}
+    ui_options: Dict[str, Any] = {}
+    ignore_failure: bool = None
+    when: str = None  # is this a documented feature?
+    # classpath, description allowed?
+
+
+class Task(CCIDictModel):
+    name: str = None  # is this real or a typo
```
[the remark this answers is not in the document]
davisagli: As we discussed, this is actually needed.

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: > How can we tell them where their error is if the "missing value" could be anywhere in four files.

We fundamentally can't; we can just tell them what is missing and let them decide what scope is most appropriate to add it (likely the project config).

But I suppose you're thinking more of errors that are wrong types rather than missing values, and that it's better if we can tell the user which file and line the problem is at. That's true but I'm not sure it's worth creating models that incorrectly declare fields as optional. I wonder if we need the dictmerge to construct a sort of sourcemap so that we can identify where each value came from (does the yaml parser even give us a way to preserve info about lines?)

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: I didn't intend to make this depend on his work, but URL is clearly wrong here. List[Dict[str, str]] should be accurate except for when someone entered a version number in a form that yaml interprets as a float :-[

This is going to end up with us writing our own yaml parser, isn't it

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: I know, but it seems like it could be useful for users to have a way to validate their configuration without trying to run any command in particular

cumulusci/utils/yaml/cumulusci_yml.py
prescod: My goal is to give better error messages by catching errors earlier in the project lifespan (of an individual CCI project) and earlier in a process execution process. Rewriting "string code" to "enum code" doesn't do a lot to advance those. I can do it if you insist but it just seems like a different project to figure out how and when we consume these models.

davisagli: The reason I suggested an enum was because I figured then the error can be more helpful by indicating what the valid options are -- which I think is in service of the overall goal here. That can probably be done either using an enum or by a custom validator that still returns a string. And I see this as an opportunity but not a blocker for this PR.

cumulusci/utils/yaml/cumulusci_yml.py
prescod: It isn't incorrect in saying that those values are optional. They are optional in the "CumulusCI.yml file format" because the can be specified elsewhere.

The Python "file format parser" allows you to call len() with zero arguments, but the runtime will complain that you've left something out. The parser isn't wrong: it's just responsible for a particular set of concerns.

Sourcemap seems very complicated and providing very little benefit over doing it the way I suggest. You can't catch every error at parse time and I don't see that as a problem even a little bit. You can do a semantic review later with dramatically less effort than trying to build a sourcemap for Python primitive objects.

"Task foo does not have a class_path" doesn't tie back to a particular source file anyhow.

I did the sourcemap thing in Snowfakery because the benefit there was large but as you probably remember and have noticed, line number handling had to suffuse the whole code base. It was a lot of work and I don't see the cost/benefit the same way in CumulusCI. We'd need to identify WHICH errors can only be found after merge AND validate that they actually even relate to a single file as opposed to the project as a whole.

As an aside: there's a part of me that thinks that dictmerge may be a clever hack that has outlived its usefulness. As you know I have another way I prefer for projects to inject steps and if we enumerate every dict-merge use-case maybe we'd find that dedicated features work better for all of them. Just a guess and an aside.

davisagli: In that sense *everything* is optional in cumulusci.yml. But I don't want to have to maintain two different models of our configuration, one which pretends everything is optional and one which actually indicates what is required at runtime. I think we want one model that indicates what's required at runtime, and then figure out how to ignore missing required items while parsing a specific file if we need to do it that way.

I agree a sourcemap is complicated and don't want to go that way unless we have to. Just throwing ideas out.

I also don't like dictmerge much, but I care a lot about not forcing users to rewrite their existing configuration files, so changing it feels daunting.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -104,19 +92,19 @@ class PreflightCheck(CCIDictModel):
...
-    checks: List[Check] = []
+    steps: Dict[str, Step]
+    checks: List[PreflightCheck] = []
     group: str = None
     error_message: str = None
     post_install_message: str = None
     preflight_message: str = None
 
 
 class Project(CCIDictModel):
```
davidmreed, line 106: I have Pydantic models in `feature/dependency-refactor` for the dependency list.

cumulusci/utils/yaml/cumulusci_yml.py
prescod: "But I don't want to have to maintain two different models of our configuration, one which pretends everything is optional and one which actually indicates what is required at runtime."

I don't think we need two models. We need two algorithms. We can have one algorithm driven by Pydantic that looks primarily for types and extra keys and another driven by Python code that validates various semantic aspects, including the presence of required-at-runtime keys.

"I think we want one model that indicates what's required at runtime, and then figure out how to ignore missing required items while parsing a specific file if we need to do it that way."

It seems far easier to me to layer on an additional check than to "turn off" a required item
I know how to write that code straightforwardly.

In fact, I wrote that kind of code at your request:

https://github.com/SFDO-Tooling/CumulusCI/pull/1624/files#diff-66f8c248cf465233668398ce87f73e4cf911807618fa87b39f5224a39d67fe8fR29

In that case, I had accurately expressed true "requirements" at the syntactic Iayer and I switched it to be a post-validation at your request because I don't see why it is important that the schema be "accurate" as long as it generates the right error messages for the user.

I literally don't have a good idea about how to tell Pydantic: "this string is required except when it isn't." I can't imagine that they would be open to such a feature request.

I am sure that with a day or two of programming there is SOME way, but it seems like a lot of complication for not much benefit to me.

If I understand you, you want something like this:

```
class Task(CCIDictModel):
    class_path: OptionalAtParseTimeRequiredAtRuntime[str] = None
```

And I want something like this:

```
class Task(CCIDictModel):
    class_path: str = None
    
    def runtime)check(self):
        if not self.class_path:
            raise ValidationError("No class_path")
```

cumulusci/utils/yaml/cumulusci_yml.py
prescod: Okay yeah, I'll do this instead:

```
    tier: Literal["primary", "secondary", "additional"] = None
```

Which generates:

```
'foo.yml' -> 'plans' -> 'plan1' -> 'tier' : unexpected value; permitted: 'primary', 'secondary', 'additional'
```

cumulusci/utils/yaml/cumulusci_yml.py
prescod: Actually, in that specific case, Pydantic will coerce.

davisagli: I'd rather have an error than coercion for version numbers, because I think 1.10 gets de-yamled into 1.1 (float) which then gets coerced to "1.1", which is not what the user intended.

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: No, I want this:

```
class Task(CCIDictModel):
    class_path: str
```

because that's the type that accurately describes the fact that we want to write code which can assume the presence of a class_path once it has a validated task configuration. I'm anticipating that we'll eventually want to use these models in type annotations in the codebase, not just for validation, and I don't want my type checker to tell me I need to handle None everywhere.

And then I want 2 different kinds of validation while loading the configuration:
1. A _partial_ validation of each file being loaded (i.e. validate the keys that are present to make sure they match the types from the model, but don't yet check that all required fields are there) so we can indicate what file has the wrong type.
2. A full validation of the merged configuration, so that we can catch missing required fields -- without needing to write custom validators.

https://github.com/samuelcolvin/pydantic/issues/1673 has some discussion of doing partial validation, with a suggestion from Samuel to use `validate_model` directly.

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: @prescod Fixing how yaml handles version numbers is clearly out of scope here, but I still think this needs to change to `List[Dict[str, str]]` so we don't complain about values that aren't URLs (i.e. all keys other than `github`)

cumulusci/utils/yaml/cumulusci_yml.py
davisagli: I'm fine with calling a command like that out of scope for this PR, but my actual concern here is that code in a `if __name__ == "__main__"` block is hard to test, so is more likely to break.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -1,4 +1,264 @@
...
+    logger.warning("CumulusCI Parsing Error:")
+    loc = " -> ".join(repr(x) for x in error["loc"] if x != "__root__")
+    logger.warning("%s : %s", loc, error["msg"])
+    logger.error(
+        "NOTE: These errors will cause major problems in future versions of CumulusCI."
+    )
+    logger.error(
+        "If you think your YAML has no error, please report the bug to the CumulusCI team."
+    )
+    logger.error("https://github.com/SFDO-Tooling/CumulusCI/issues/")
```
davisagli, line 209: The output is a little hard to read:

Screen Shot 2021-04-07 at 11 40 39 AM

Suggestions:
- Only show "CumulusCI Parsing Error", the note, and the link to the issue tracker once
- Put a newline and indentation between the location and error message
- Log an empty line after all the warnings

There are a couple other issues here too:
- "Steps must have a task or flow" shouldn't be shown here; the step in question has a task or flow inherited from the universal config
- `checks` is allowed in a flow step

prescod: Overriding the options without specifying what task you are overriding seems quite fragile to me. Do we want to encourage it?

davisagli: I don't think we've encouraged it, but it's possible and clearly done.

davisagli (approved):

davisagli (approved):

PR #1625, opened 2020-03-19
A feature for allowing classes to move with a deprecation warning.

prescod:
Guide to this PR (updated):

cumulusci/core/utils.py has a new method which allows one to label deprecated classes but only gets assertive if Pytest is running and only outputs warning if you run with python -W. So it won't be noisy except for those who should pay attention to it.

https://github.com/SFDO-Tooling/CumulusCI/pull/1625/files#diff-9c057c3ee7455a5f3399d2b1170f0c82

cumulusci/core/tests/test_utils.py tests this function.

https://github.com/SFDO-Tooling/CumulusCI/pull/1625/files#diff-0cc3155d74646dc569a53602bb6a9789

Deploy.py is renamed to deploy_metadata.py so it can have a new lower-case name without clashing with itself.

A script renames everything else from WordWord.py to word_word.py. History is preserved in this rename.

WordWord.py becomes a redirect to word_word.py like this:

https://github.com/SFDO-Tooling/CumulusCI/pull/1625/files#diff-2c9659eba7aa06c30e7934e0ca533e08

That file also has a history of how it changed to end up like that.

`__ini__.py` gets a bunch of redirects

https://github.com/SFDO-Tooling/CumulusCI/pull/1625/files#diff-726f7a88e0e118566b6ab6b249ba65cb

cumulusci.yml and all other references to these files are updated.


davisagli:
Zope has a feature like this, and used to make careful use of it. But they found that people tended to just ignore the warnings until the old location was actually removed. So the main accomplishment of the deprecation warnings was showing a bunch of warnings that got in the way of noticing more critical messages. So Zope stopped using the warnings and now just tries to keep aliases from old locations as long as possible.

So feel free to make the case for it, but that's why my initial reaction is dubious.

prescod:
If I understand our current situation, importing *any* of the 30-ish tasks in cumulusci.tasks.salesforce imports ALL of them. That's slow and buggy as we've just seen. So what is the better solution than deprecate and then delete?

davisagli:
@prescod I agree it's non-ideal, but I think there are higher priority things for us to be working on for now. And if/when it is at the top of the list, I think I would be more inclined to do a hard removal of the `__init__` aliases in a new major version, document it well, and not bother with deprecation warnings.

prescod:
I updated the description at the top of the PR to represent the current variant of this PR.

cumulusci/core/utils.py
```diff
@@ -26,6 +32,26 @@ def import_global(path):
...
+    cls = import_global(fully_qualified)
+    clsname = cls.__name__
+    message = f"Instead of importing from {__name__}.{clsname}, please import from {fully_qualified}"
+
+    def __init__(self, *args, **kwargs):
+        conditional_deprecation_warning(message)
+        super(cls, self).__init__(*args, **kwargs)
+
+    init_dict = {"__init__": __init__}
+    return type(clsname, (cls,), init_dict)
```
davisagli: This looks a bit too magical. If there's an isinstance check using the class from the old location, it will fail to return true for instances of the class from the new location.

I guess the point is to defer warnings until pytest has initialized sufficiently to determine whether we're supposed to raise or warn? How about this as an alternative:
1. Warn unconditionally at import time
2. Make deprecated_import add the class to a global set, and use a pytest hook to raise assertion errors between test discovery and starting to run the tests.

cumulusci/core/utils.py
```diff
@@ -26,6 +32,22 @@ def import_global(path):
...
+    clsname = cls.__name__
+    message = f"Instead of importing from {__name__}.{clsname}, please import from {fully_qualified}"
+    warnings.warn(message, DeprecationWarning)
+
+    if "PYTEST_CURRENT_TEST" in os.environ:
+
+        def __init__(self, *args, **kwargs):
+            assert False, message
+
+        cls.__init__ = {"__init__": __init__}
```
davisagli:
```suggestion
        def __init__(self, *args, **kwargs):
            raise AssertionError(message)

        cls.__init__ = __init__
```

cumulusci/core/utils.py
prescod: If the name bound in `__init__.py` is bound to exactly the same object as foo_bar.py then there is no way to know which one is being imported by the client file.

deprecated_import happens unconditionally as soon as `__init__.py` is imported for any reason, whether legitimate or not.

I don't know that there is a way to preserve bidirectional isinstance and also issue the deprecation warnings only when they are really merited. Bidirectional isinstance is probably (?) the same as "is" 

if we deprecate every single class in `__init__.py` maybe we could deprecate importing the whole module but I'm not sure if its imported automatically under some circumstances. I haven't tested that.

So I'm a bit stuck.

davisagli:
I think we'll want to reimplement this in terms of the lazy imports I added in #1857 -- but also, that lazy import system makes it much less pressing to do this reorganization since it avoids loading all the salesforce tasks at once.

PR #1627, opened 2020-03-20
Add CCI task to wait for custom setting value

force2b:
# Critical Changes

# Changes

- Add tasks.apex.custom_settings.CustomSettingValueWait task to wait for custom setting value

# Issues Closed


cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,96 @@
...
+        self.logger.info(
+            "{0}: Looking for {1} and found {2}".format(
+                self.field_name, self.check_value, self.field_value
+            )
+        )
+        return self.field_value == self.check_value
+
+    @property
+    def _object_query(self):
+        return "SELECT {0} FROM {1} LIMIT 1".format(self.field_name, self.object_name)
```
prescod: The versions of Python we use these days allow this:

```suggestion
        return f"SELECT {self.field_name} FROM {self.object_name} LIMIT 1"
```
Even if it does look less like bind variables. ;)

cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,96 @@
...
+
+    @property
+    def success(self):
+        self.field_value = self.record[self.field_name]
+
+        if isinstance(self.field_value, (bool)):
+            self.check_value = process_bool_arg(self.check_value)
+            self.field_value = process_bool_arg(self.field_value)
+        elif isinstance(self.field_value, (int, float)):
+            self.check_value = float(self.check_value)
```
prescod: Just out of curiosity, do you know of any use-cases for floats? Comparing floats always makes me slightly nervous so I'd like to know what range they would be in.

force2b: In testing, any numeric field in Salesforce is retrieved with a number to right decimal point, even if that's zero and even if the Number field in Salesforce is configured with a scale of 0. This is consistent with how all Number fields are handled in Salesforce as well.

cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,94 @@
...
+        self._poll()  # will block until poll_complete
+
+        self.logger.info("Value Matched.")
+
+        return True
+
+    def _poll_action(self):
+        query_results = self.sf.query(self._object_query)
+
+        self.record = query_results["records"][0]
```
davisagli: Hmm, so we're just assuming that the first record returned is the one we care about? That doesn't sound like a safe assumption when there is more than one instance of the custom settings object.

force2b: All of the Hierarchical custom settings objects in NPSP will always have only a single row unless someone manually creates their own individual record. That's extremely rare. The best I can think of to get the org default record is to find the record where the SetupOwnerId starts with "00D". That would have to be done in a separate loop of the returned records. It's feasible to do that if you'd like. Doing this would also limit this to only work on Hierarchical Custom Settings records - preventing the User from trying to use it against any other SObject or even a List-type Custom Settings object.

cumulusci/tasks/apex/custom_settings.py
prescod: Comparing floats always sets off a clarion in my head, but I don't know that it will cause a problem in this situation so I think we should go ahead with it.

cumulusci/tasks/apex/custom_settings.py
force2b: I added logic to verify that the object is a Hierarchical Custom Settings object and to get only the Org Default record if there is more than one record in the object

cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+            raise
+
+        self.record = None
+        for row in query_results["records"]:
+            setupOwnerId = str(row["SetupOwnerId"])
+            if setupOwnerId.startswith("00D"):
+                self.record = row
+
+        # print(self.record)
```
prescod:
```suggestion
```

force2b:
Thanks @prescod. Change made.

cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def _run_task(self):
+        self.poll_interval_s = int(self.options.get("poll_interval", 10))
+
+        self.object_name = self.options["object"]
+        self.field_name = self.options["field"]
+        self.check_value = self.options["value"]
+
+        # Process namespace tokens
+        namespace = self.project_config.project__package__namespace
```
force2b: @prescod and @davisagli I had though (clearly incorrectly) that this would return a namespace if it was needed (kind of like what would happen in Apex). Is there a better way to do this so that the Namespace is applied when it's _needed_, or is the best approach to use the same `managed` and `namespace` option logic that the apex anonymous task implements?

force2b: I ended up copying the logic from ApexAnonymous so this now uses the `managed` and `namespaced` options. Works much better now.

force2b:
@davisagli and @prescod Are there other changes I should make to this or do you think it's good as is?

cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,126 @@
...
+        return True
+
+    def _poll_action(self):
+        try:
+            query_results = self.sf.query(self._object_query)
+        except:
+            print(
+                "Only Hierarchical Custom Settings objects can be used with this task"
+            )
+            raise
```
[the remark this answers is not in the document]
davisagli: Oh, and the import: `from cumulusci.core.exceptions import TaskOptionsError`

cumulusci/tasks/apex/custom_settings.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if isinstance(self.field_value, (bool)):
+            self.check_value = process_bool_arg(self.check_value)
+            self.field_value = process_bool_arg(self.field_value)
+        elif isinstance(self.field_value, (int, float)):
+            self.check_value = float(self.check_value)
+            self.field_value = float(self.field_value)
+        elif isinstance(field_value, (str)):
+            self.check_value = str(self.check_value).lower()
+            self.field_value = str(self.field_value).lower()
```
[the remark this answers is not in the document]
force2b: In apex string comparisons are case insensitive by default so I was kind of keeping it consistent. Doesn't need to be though.

cumulusci/tasks/apex/custom_settings.py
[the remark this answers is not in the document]
force2b: Thanks. Is there a way to get more about the error since it could be a bad object or field name too? At one point I tried to parse the error response, but didn't have any luck with it.

cumulusci/tasks/apex/custom_settings.py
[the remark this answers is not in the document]
davisagli: Oh wow, I knew Apex did that for variables and object/field names but didn't know it did for string comparisons too.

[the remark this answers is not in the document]
force2b: Yeah. `==` is case-insensitive while using `.equals()` is case-sensitive.

cumulusci/tasks/apex/custom_settings.py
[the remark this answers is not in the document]
force2b: Thanks for the tips on this. I added some better error parsing to provide a more accurate error.

force2b:
Thanks @davisagli. Requested changes made.

davisagli:
@force2b I just added a commit with a few tweaks:
- Moved the module from cumulusci/tasks/apex/custom_settings.py to cumulusci/tasks/salesforce/custom_settings_wait.py since there's nothing particularly Apex-y about it.
- Only catch simple_salesforce.exceptions.SalesforceError from the query. Other Python exceptions do not necessarily have a `content` attribute, and it's ok if we let them fall through to the default error reporting.
- Fixed a reference to `field_value` that needed to be `self.field_value`. This one was caught by the flake8 utility which I would recommend setting up to run automatically for Python files in your IDE.

force2b:
Thanks @davisagli!

PR #1629, opened 2020-03-20
Proposed solution for namespaced orgs

davidmreed:
# Critical Changes

# Changes

- Correct a bug in how the `set_organization_wide_defaults` task operated in namespaced scratch orgs.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/metadata_etl/sharing.py
+++ b/cumulusci/tasks/metadata_etl/sharing.py
@@ -105,10 +105,21 @@ class SetOrgWideDefaults(MetadataSingleEntityTransformTask, BaseSalesforceApiTas
             )
 
         for sobject in self.owds:
+            # The Tooling API requires that we use fully-qualified sObject names in namespaced scratch orgs.
+            # However, the Metadata API requires no namespaces in that context.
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/metadata_etl/sharing.py
```diff
@@ -105,10 +105,21 @@ def _poll_action(self):
             )
 
         for sobject in self.owds:
+            # The Tooling API requires that we use fully-qualified sObject names in namespaced scratch orgs.
+            # However, the Metadata API requires no namespaces in that context.
+            # Dynamically inject the namespace if required.
+            real_api_name = (
+                f"{self.project_config.project__package__namespace}__{sobject}"
+                if hasattr(self.org_config, "namespaced")
+                and self.org_config.namespaced
```
davisagli:
```suggestion
                if self.org_config.namespaced
```

It's a BaseConfig subclass, so it has a `__getattr__` which returns None if the key isn't in the config.

davidmreed:
Marking this PR "in progress" because I've identified a potential issue in ProfileGrantAllAccess I need to explore further.

PR #1632, opened 2020-03-22
1. bulk_mode. 2. _execute_step. 3. Type checking.

prescod:
1. I forgot bulk_mode in the schema because there were no example mapping files with it. Fixed now.

2. Renamed _load_mapping to _execute_step because _load_mapping always implied to me that it was loading the mapping file. This is subjective so tell me if you disagree.

3. I added stronger type checking which is how I found problem 1. Integration tests don't use bulk_mode but unit tests do

4. I deleted the test that checks what happens if a bulk_mode is set to a wrong value in mapping because the mapping parser should handle that now and it is hard to trigger with both pytest and typeguard working at the same time. In any case, the Salesforce API would give you an error pretty quickly if you found a way around all of this type checking...

### DIFF
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -1,5 +1,8 @@
 from collections import defaultdict
 import datetime
+from unittest.mock import MagicMock
+from typing import Union
+
 from sqlalchemy import Column, MetaData, Table, Unicode, create_engine, text
 from sqlalchemy.orm import aliased, Session
 from sqlalchemy.ext.automap import automap_base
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/bulkdata/load.py
```diff
@@ -101,13 +105,15 @@ def _run_task(self):
...
+                    result = self._execute_step(after_step)
                     if result.status is DataOperationStatus.JOB_FAILURE:
                         raise BulkDataException(
                             f"Step {after_name} did not complete successfully: {','.join(result.job_errors)}"
                         )
 
-    def _load_mapping(self, mapping):
+    def _execute_step(
+        self, mapping: MappingStep
+    ) -> Union[DataOperationJobResult, MagicMock]:
```
davisagli: I don't like having to include mocks in the type signature. Is there no way to tell pydantic to always allow them?

davisagli:
@prescod the tests are failing because this imports from typing_extensions which is not in requirements

cumulusci/tasks/bulkdata/load.py
prescod: I think you mean typeguard.

And yes, there is a plan to fix it at the Typeguard level but not merged yet.

https://github.com/agronholm/typeguard/pull/105

I'll bug the maintainer again, but if he doesn't respond we might have to figure out whether to fork, abandon or continue to work around small bugs in typeguard. I found another one on the weekend. Easy to fix, but only if he accepts my PRs.

PR #1637, opened 2020-03-23
Robot task now prints correct message based off of error code

boakley:
Previously, the robot task assumed the error code represented the number of failed tests. That is only correct up to the value 250. After that, different values mean different things.

For more information see [return codes](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#return-codes
) in the robot user guide 

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -89,7 +89,7 @@ class Robot(BaseSalesforceTask):
         # http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#return-codes
         if 0 < num_failed < 250:
             raise RobotTestFailure(
-                "{} test{} failed.".format(num_failed, "" if num_failed == 1 else "s")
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -84,8 +84,23 @@ def _run_task(self):
...
 
         num_failed = robot_run(self.options["suites"], **options)
-        if num_failed:
-            raise RobotTestFailure("{} tests failed".format(num_failed))
+
+        # These numbers are from the robot framework user guide:
+        # http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#return-codes
+        if 0 < num_failed < 250:
+            raise RobotTestFailure(
+                "{} test{} failed.".format(num_failed, "" if num_failed == 1 else "s")
```
davisagli:
```suggestion
                f"{num_failed} test{'' if num_failed == 1 else 's'} failed."
```

PR #1640, opened 2020-03-25
Prepare 3.9.1

davisagli:
# Critical Changes

# Changes

# Issues Closed


.github/workflows/release.yml
```diff
@@ -28,8 +28,9 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_TOKEN }}
       - name: Create tag
         run: |
-          git tag -a -m "version $(python setup.py --version)" v$(python setup.py --version)
-          git push --follow-tags
+          curl -s -X POST https://api.github.com/repos/$GITHUB_REPOSITORY/git/refs -H "Authorization: token $GITHUB_TOKEN" -d @- << EOF
```
davidmreed: I think this needs an `env` entry to acquire `GITHUB_TOKEN` from `secrets`, doesn't it?

davisagli: @davidmreed good catch that it's a secret, not automatically in the env. I think we can refer to it with this syntax: `${{ secrets.GITHUB_TOKEN }}`

davidmreed: I don't think Actions allows that kind of interpolation inside a shell script, based on the examples (https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions) of expression syntax, but I can't find an explicit statement to that effect.

davisagli: @davidmreed I see it in an example here: https://help.github.com/en/actions/configuring-and-managing-workflows/authenticating-with-the-github_token#example-calling-the-rest-api

davidmreed: I know it does work in `env` because I tried it (https://github.com/davidmreed/amaxa/blob/master/.github/workflows/publish.yml)

davidmreed: Cool! That streamlines the markup nicely.

PR #1641, opened 2020-03-26
Feature/metecho subprocess

davisagli:
These are a couple changes needed by Metecho in order to run flows in a subprocess:
1. Make it possible to avoid trying to refresh the org access token (by setting the CUMULUSCI_DISABLE_REFRESH env var)
2. Make it possible to log in to the github API using an oauth token from the GITHUB_TOKEN env var instead of using GitHub app credentials or a username + password or personal access token

I also updated the release github action to abort releasing if version.txt contains a dev version.

# Critical Changes

# Changes

# Issues Closed


.github/workflows/release.yml
```diff
@@ -13,6 +13,8 @@ jobs:
     runs-on: ubuntu-18.04
     steps:
       - uses: actions/checkout@master
+      - name: Abort for dev version
```
codexmarcel: Cool!

davisagli: It didn't actually work to prevent the release. But fortunately .dev* versions count as prerelease and our code to check whether there's an upgrade available only includes final releases...so no real harm done.

PR #1643, opened 2020-03-28
Namespaces support and other small fixes for salesforce_encoding.

prescod:
Add support for arbitrary namespaces.

Add newlines after empty elements.

Add tests for both.


cumulusci/utils/xml/salesforce_encoding.py
```diff
@@ -4,46 +4,85 @@
...
+    return "".join(
+        f" xmlns{format_ns(prefix, url)}" for prefix, url in declarations.items()
+    )
+
+
+def _render_attr_name(name, namespaces):
+    if name[0] == "{":
+        return _render_name(name, namespaces)
+    else:
+        return name
```
davisagli: This line is missing coverage

PR #1644, opened 2020-03-28
Better feedback for GetInstalledPackages task

prescod:
Michael Beaty asked me if we could make a task for listing installed packages.

I said: "don't we already have that?"

So he and I tested it. It had no visible output.

Thus this PR.


davisagli:
@prescod I revised the test to at least assert that the right API helper was called, and to avoid using a pseudo-private function from the bulkdata test suite.

PR #1649, opened 2020-04-02
prepare 3.10.0

davisagli:
- Prepare 3.10.0 release
- Add a test file that was missed from an earlier commit
- Try adding a github action to build the homebrew formula when releases are created.

# Critical Changes

# Changes

# Issues Closed


.github/workflows/build_homebrew.yml
```diff
@@ -0,0 +1,36 @@
...
+          python-version: 3.7
+      - name: Build formula
+        run: |
+          utility/build-homebrew.sh cumulusci.rb
+      - name: Test formula
+        run: |
+          brew audit cumulusci.rb
+          brew install cumulusci.rb
+          brew test cumulusci.rb
+      - name: Create branch in homebrew-sfdo repository
```
jstvz, line 26: Neat. If this works we'll toss `utility/push-homebrew.py`?

davisagli: @jstvz I think I'll need to iterate on this, but yeah.

PR #1654, opened 2020-04-06
Move CI to github actions

davisagli:
This moves all of CumulusCI's own CI builds from CircleCI to GitHub Actions.

I also used this opportunity to split the robot tests into 3 separate builds that run in parallel.

It is expected that the CircleCI builds fail on this PR since I've removed the CircleCI configuration. Once the PR is approved but before merging I will remove the CircleCI commit statuses from branch protection and add the new builds.

# Critical Changes

# Changes

# Issues Closed


davisagli:
@Br4nd0R I suspect we may be able to use YAML anchors for that (https://confluence.atlassian.com/bitbucket/yaml-anchors-960154027.html) but I didn't try yet.

PR #1656, opened 2020-04-06
create_community improvements

prescod:
- Remove call to servlet endpoint that is no longer needed
- Add the ability to try the community creation multiple times


davisagli:
I am working on testing this with the `create_advisee_community` task in SAL. And I'm wondering whether the necessity of the workaround depends on which community template is selected.

davidmreed:
I don't know about template dependency, but it is definitely dependent on whether or not there's already a Community in the org. The original workaround was required only on the _first_ Community creation event.

cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -26,33 +28,43 @@ class CreateCommunity(BaseSalesforceApiTask):
...
-        # Let's simulate that without actually using a browser.
-        self.logger.info("Preparing org for Communities")
-        s = requests.Session()
-        s.get(self.org_config.start_url).raise_for_status()
-        r = s.get(
-            "{}/sites/servlet.SitePrerequisiteServlet".format(
-                self.org_config.instance_url
+        community = self._get_community()
+        if community is not None:
+            raise Exception(
```
davidmreed, line 47: Should we use something more specific like a `TaskOptionsError` (sort of fits) or a `CumulusCIException` here?

cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -26,33 +28,43 @@ class CreateCommunity(BaseSalesforceApiTask):
             "description": "URL prefix for the community.",
             "required": False,
         },
+        "retries": {
```
davidmreed, line 31: Need to regenerate the task doc for this change, right?

PR #1669, opened 2020-04-14
Use pathlib

Br4nd0R:
# Critical Changes

# Changes
* Minor cleanup of `os.path` use to `pathlib.Path`

# Issues Closed


cumulusci/core/config/project_config.py
```diff
@@ -421,9 +433,9 @@ def get_previous_version(self):
     def config_project_path(self):
         if not self.repo_root:
             return
-        path = os.path.join(self.repo_root, self.config_filename)
-        if os.path.isfile(path):
-            return path
+        path = Path(self.repo_root) / self.config_filename
+        if path.is_file():
+            return str(path)
```
davisagli, line 438: What's your logic for when to convert a Path to a str and when to leave it as a Path? You're doing the former here but the latter in config_project_local_path.

Br4nd0R: good catch, let me make both of them return a string. I think it's a little safer than returning a `Path` object.

cumulusci/core/config/project_config.py
```diff
@@ -240,21 +241,33 @@ def _split_repo_url(self, url):
         git_info = {"url": url, "owner": owner, "name": name}
         return git_info
 
+    def git_path(self, tail=None):
+        """Returns a Path to self.repo_root with
```
davisagli, line 245: you mean to the .git folder in repo_root...

cumulusci/core/config/project_config.py
```diff
@@ -240,21 +241,33 @@ def _split_repo_url(self, url):
...
         if path:
             return path
 
-        path = os.path.splitdrive(os.getcwd())[1]
+        path = Path(os.path.splitdrive(Path.cwd())[1])
         while True:
-            if os.path.isdir(os.path.join(path, ".git")):
+            if (path / ".git").is_dir():
+                # if os.path.isdir(os.path.join(path, ".git")):
                 return path
```
davisagli, line 265: This is another one we may want to convert back to a string

cumulusci/core/config/project_config.py
```diff
@@ -266,7 +279,7 @@ def repo_name(self):
             return
 
         in_remote_origin = False
-        with open(os.path.join(self.repo_root, ".git", "config"), "r") as f:
+        with open(self.git_path("config"), "r") as f:
```
davisagli, line 282: The logic below for reading the file to find the "origin" url is duplicated between several of these functions. I'm thinking we should move all of it (except for the final step of parsing the url) into a helper function instead of just `git_path`.

cumulusci/core/config/project_config.py
Br4nd0R: 🤦

cumulusci/core/config/project_config.py
```diff
@@ -234,27 +236,51 @@ def _log_detected_overrides_as_warning(self, info):
     def _split_repo_url(self, url):
         url_parts = url.rstrip("/").split("/")
         name = url_parts[-1]
-        owner = url_parts[-2]
+        owner = url_parts[-2].split(":")[-1]
```
Br4nd0R, line 239: `_apply_repo_env_var_overrides()` was the only other method that was using `_split_repo_url()`. I don't *think* that this change impacts it by changing the return value of 'owner', but you might want to double check as well.

cumulusci/core/config/project_config.py
```diff
@@ -304,16 +315,8 @@ def repo_owner(self):
...
-        in_remote_origin = False
-        with open(os.path.join(self.repo_root, ".git", "config"), "r") as f:
-            for line in f:
-                line = line.strip()
-                if line == '[remote "origin"]':
-                    in_remote_origin = True
-                    continue
-                if in_remote_origin and line.find("url =") != -1:
-                    line_parts = line.split("/")
-                    return line_parts[-2].split(":")[-1]
```
Br4nd0R, line 316: this is what prompted me to change the `_split_repo_url()` method to return the actual 'owner' segment of the url

cumulusci/core/config/project_config.py
davisagli: @Br4nd0R let's just make sure we have tests for this that cover both https and ssh git URLs

cumulusci/core/config/project_config.py
```diff
@@ -234,27 +236,51 @@ def _log_detected_overrides_as_warning(self, info):
...
+        is not set."""
+        path = None
+        if self.repo_root:
+            path = Path(self.repo_root) / ".git"
+            if tail is not None:
+                path = path / str(tail)
+        return path
+
+    def git_config_remote_origin_line(self, start_string):
+        """Rerturns the first line under the [remote origin]
```
davisagli, line 257:
```suggestion
        """Returns the first line under the [remote origin]
```

cumulusci/core/config/project_config.py
```diff
@@ -234,27 +236,51 @@ def _log_detected_overrides_as_warning(self, info):
...
+        section of the .git/config file that that starts
+        with start_string. Returns None if .git/config file
+        not present or no matching line is found."""
+        with open(self.git_path("config"), "r") as f:
+            for line in dropwhile(self.line_not_remote_origin, f):
+                if start_string in line.strip():
+                    return line.strip()
```
davisagli, line 264: Hmm, how about

```suggestion
            while next(f).strip() != '[remote "origin"]':
                continue
            for line in f:
                if start_string in line.strip():
                    return line.strip()
```

cumulusci/core/config/project_config.py
Br4nd0R: sounds good.

cumulusci/core/config/project_config.py
Br4nd0R: I don't mind your suggested version here, I thought I was being slick with the use of `dropwhile()`. Are you thinking that readability is better this way, or don't like the use of dropwhile() here? Just curious.

davisagli: I think mostly I didn't like having to look elsewhere to find self.line_not_remote_origin when it turned out to be a one-liner.

cumulusci/core/config/project_config.py
```diff
@@ -234,27 +236,53 @@ def _log_detected_overrides_as_warning(self, info):
...
+        """Returns the first line under the [remote origin]
+        section of the .git/config file that that starts
+        with start_string. Returns None if .git/config file
+        not present or no matching line is found."""
+        with open(self.git_path("config"), "r") as f:
+            while next(f).strip() != '[remote "origin"]':
+                continue
+            for line in f:
+                if start_string in line.strip():
+                    return line.strip()
```
jstvz, line 266: There's also `configparser`:
```python
from configparser import ConfigParser
config = ConfigParser(strict=False)
config.read('.git/config')
return config['remote "origin"']['url']
```

Br4nd0R: oh, heck ya!

cumulusci/core/tests/test_config.py
```diff
@@ -341,7 +341,7 @@ def test_repo_url_no_repo_root(self):
 
     def test_repo_url_from_git(self):
         config = BaseProjectConfig(BaseGlobalConfig())
-        self.assertIn("/CumulusCI", config.repo_url)
+        self.assertEquals("git@github.com:SFDO-Tooling/CumulusCI.git", config.repo_url)
```
davisagli, line 344: unittest.TestCase.assertEquals is deprecated; use assertEqual (singular) or a pytest assertion instead.

cumulusci/core/config/project_config.py
```diff
@@ -233,28 +235,59 @@ def _log_detected_overrides_as_warning(self, info):
...
+        config = ConfigParser(strict=False)
+        try:
+            config.read(self.git_path("config"))
+            url = config['remote "origin"']["url"]
+        except (KeyError, TypeError):
+            url = None
+        return url
+
+    def line_not_remote_origin(self, line):
+        return True if line.strip() != '[remote "origin"]' else False
```
davisagli, line 274: Remove this now that we aren't using it.

PR #1671, opened 2020-04-14
Fix CommitDir

davisagli:
Fix CommitDir util and improve tests.

Without this change, existing files were getting detected as new files and duplicated in the list of blobs to commit but potentially with the wrong file mode.

# Critical Changes

# Changes

# Issues Closed


Br4nd0R:
Is `CommitDir` only used for testing? I'm currently only seeing it referenced in `cumulusci.tasks.github.tests`. Just trying to understand what it's providing from a functionality standpoint.

davisagli:
@Br4nd0R It was used by a task that no longer exists to commit updates to the generated Apex docs. But it is also used by Metecho.

davisagli:
And also by Jason's draft task to publish a release from a private repo to a public repo https://github.com/SFDO-Tooling/CumulusCI/pull/1414

davisagli:
@jstvz could you re-approve please? I snuck in a fix for an unstable test that was introduced when I merged from master.

cumulusci/tasks/github/util.py
```diff
@@ -53,15 +53,15 @@ def __call__(
         self.local_dir, self.repo_dir = self._validate_dirs(local_dir, repo_dir)
         self._set_git_data(branch)
 
-        new_tree_list = [self._create_new_tree_item(item) for item in self.tree]
-        self._add_new_files_to_tree(new_tree_list)
+        self.new_tree_list = [self._create_new_tree_item(item) for item in self.tree]
```
Br4nd0R, line 56: Is there a benefit to having this as a member variable rather than a standalone variable? My initial though would be that we may want to reference this elsewhere in a different task, but I'm not sure on the context.

davisagli: @Br4nd0R this change was for the test, so that it can easily verify that the correct new tree was produced.

PR #1673, opened 2020-04-15
Small improvements for data dictionary

prescod:
1. Allow Lookups to declare their lookup table in the same column
that picklists declare their valid values.

2. MultiselectPicklists are supported the same way that regular Picklists were.


davidmreed:
@prescod  What's the impetus for the Lookup Field change? Can we vet it by CCE, who are among the primary consumers of this tool for building NPSP's data dictionary?

prescod:
Judi and Cori agreed that it was useful to have the referenceTo information for Lookups.

PR #1680, opened 2020-04-16
Feature/brandon/10 mil byte bulk limit

Br4nd0R:
# Critical Changes

# Changes
* We now truncate batches that would exceed the ten million character limit during `BulkApiDmlOperation._batch(records)`.
 
# Issues Closed


cumulusci/tasks/bulkdata/utils.py
```diff
@@ -13,15 +12,66 @@
 from cumulusci.utils import convert_to_snake_case
 
 
-def batch_iterator(iterator, n=10000):
-    while True:
-        batch = list(itertools.islice(iterator, n))
-        if not batch:
-            return
+def batch_iterator(iterator, fields, n=10000, char_limit=10000000):
```
Br4nd0R, line 15: I don't love that we have to pull in fields here but didn't see a way around it.

cumulusci/tasks/bulkdata/utils.py
```diff
@@ -13,15 +12,66 @@
...
+            batch = []
+            current_chars = csv_field_chars
+
+    # give back anything leftover
+    if batch:
         yield batch
 
 
+def csv_const_chars_per_record(num_fields):
+    """Given number of fields (or values) on a record, return the
```
Br4nd0R, line 50: I know this is a lot of text for a one-liner. Please feel free to say it's too verbose; however, I thought that just saying `'''Returns constance csv chars per character''' return 3 * num_fields` wasn't quite enough.

cumulusci/tasks/bulkdata/utils.py
```diff
@@ -13,15 +12,66 @@
...
+
+    The corresponding line in a csv file would be:
+        'Id','FirstName','LastName'\n
+
+    For each field there are:
+        quotations = num_fields * 2
+        commas = num_fields - 1
+        newlines = 1
+    Some algebra gives us: 3 * num_fields
+    #TODO: Do we include spaces in between values?
```
Br4nd0R, line 65: Anyone know if we include spaces in the final csv data we send out?

davidmreed: That behavior's owned by `csv`. It doesn't include spaces.

By default, a newline is `\r\n`, though, for 2 chars.

cumulusci/tasks/bulkdata/utils.py
```diff
@@ -13,15 +12,66 @@
...
+    """
+    return 3 * num_fields
+
+
+def sum_chars(list):
+    """Given a list of strings return the sum of string lengths"""
+    return len("".join(list))
```
davidmreed, line 72: Wondering, given that this method may be called tens of millions of times in performance data loads, if there's a significant difference using `join()`, which presumably is doing a new string alloc, or summing the `len()` of each element.

cumulusci/tasks/bulkdata/utils.py
davidmreed: Actually, there's something else nasty lurking here. `csv` will _quote_ specific field values, and will also escape embedded quotes in the field data. That will result in expansion of the field values we are sending. For some `csv.writer` `w`, 

    w.writerow(['test stuff', 'fo"o'])

yields

    "test\nstuff","fo""o"\r\n

I don't immediately see a way around this other than either walking across all the record values and applying the expected CSV transformations, which is a bit nasty, or serializing the entire CSV file to disk before uploading it, which is quite nasty and a major change to the architecture. Do you have any better thoughts?

davidmreed: We use the default CSV dialect (https://docs.python.org/3/library/csv.html#csv-fmt-params).

Br4nd0R: @davidmreed - Since the input/output of `batch_iterator()` has remained constant, this would be a bug that already exists in the code and isn't something being introduced correct? If you can confirm I can log a new bug record in GUS to address this.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -285,7 +285,9 @@ def load_records(self, records):
 
     def _batch(self, records):
         """Return a generator of generators, where each child generator is batched."""
-        for batch in batch_iterator(records, self.api_options.get("batch_size", 10000)):
+        for batch in batch_iterator(
+            records, self.fields, self.api_options.get("batch_size", 10000)
+        ):
             yield self._csv_generator(batch)
```
davisagli, line 291: I think that we're making things harder for ourselves by trying to produce a batch of the correct size first, then passing it along to `_csv_generator`. It means that batch_iterator needs to do a bunch of work to anticipate how big the records will be in CSV format and know details about the CSV dialect that will be used.

Instead I think we could replace `_batch` and `_csv_generator` with a function that does something like this:

1. Iterate over `records`
2. For each record, use a csv writer to serialize it as CSV, and read the serialized value into a variable `row`.
3. Check the length of `row`. If it will not put the current batch over the limit, append the row to a list for the current batch. If it will put the current batch over the limit, yield an iterator over the current batch, then start a new list for the next batch.

The downside is that we have to store a whole batch in memory. It's probably possible to avoid this by being more clever with the iterators but I'm having trouble figuring it out at the moment.

Br4nd0R: I like it, this would simplify things a bit. Let me play around with reworking and see what I can come up with.

Br4nd0R:
@davisagli - I keep seeing the 'No browser' Robot tests fail on this branch. Do I have to change something in the robot tests related to datagen? The logs don't make it quite clear which exact tests are failing.

davisagli:
@Br4nd0R try running those robot tests locally with `-o debug True` so it drops you into a pdb when it hits the error.

cumulusci/tasks/bulkdata/utils.py
Br4nd0R: eliminated the need for this method :)

Br4nd0R:
@davidmreed - This is ready for another once over when you have a chance.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -281,12 +291,48 @@ def load_records(self, records):
...
+        # give back anything leftover
+        if len(batch) > 1:
+            yield batch
+
+    def _serialize_csv_record(self, record):
+        """Given a list of strings (record) return
+        the corresponding record serialized in .csv format"""
+        self.csv_writer.writerow(record)
+        self.csv_buff.seek(0)
+        serialized = self.csv_buff.read().encode("utf-8")
```
davisagli, line 330: I think you can just do this:
```suggestion
        serialized = self.csv_buff.getvalue().encode("utf-8")
```

cumulusci/tasks/bulkdata/step.py
```diff
@@ -264,6 +263,17 @@ def get_results(self):
...
 
+    def __init__(self, *, sobject, operation, api_options, context, fields):
+        super().__init__(
+            sobject=sobject,
+            operation=operation,
+            api_options=api_options,
+            context=context,
+            fields=fields,
+        )
+        self.csv_buff = io.StringIO()
```
davisagli, line 274: I think we should pass `newline=""` here so that the csv writer can manage the newlines correctly if there are records with text that includes newlines.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -281,12 +291,48 @@ def load_records(self, records):
...
+        self.csv_writer.writerow(record)
+        self.csv_buff.seek(0)
+        serialized = self.csv_buff.read().encode("utf-8")
+        # flush buffer
+        self.csv_buff.truncate(0)
+        self.csv_buff.seek(0)
+
+        return serialized
 
     def _csv_generator(self, records):
```
davisagli, line 337: Looks like this is now unused.

cumulusci/tasks/bulkdata/step.py
davisagli: (But make sure that we're counting a newline at the end of each record toward the limit.)

Br4nd0R: @davisagli - We serialize the records to `.csv` format (write with the csv writer and then read the record back from the `IO.StringIO(newline='')` stream) and only then do we count the number of bytes. It looks like specifying newline as `''` is indicating that the stream should not do any translations with newline chars. Is this also adding an additional newline somewhere in the mix?

davisagli: @Br4nd0R I'm saying we should verify that serializing a list like ["col 1", "multiline\ncol 2"] results in `col 1,"multiline\ncol 2"\n` and counts as 24 characters.

PR #1682, opened 2020-04-17
Feature/robot app launcher stability

boakley:
This changes the app launcher keywords to work around the fact that sometimes the app launcher
hangs up while rendering. The `Open App Launcher` keyword tries to detect that, and will refresh
the modal. 

While this code still doesn't work 100% of the time, it went from failing every 7-8 test runs to once out of 256 last night. 

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -496,7 +496,7 @@ class Salesforce(object):
         will need to create a custom keyword.
 
         If the retry parameter is true, the keyword will
-        close and then re-open the app launcher it times out
+        close and then re-open the app launcher if it times out
         while waiting for the dialog to open.
         """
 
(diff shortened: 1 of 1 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -485,32 +485,47 @@ def log_browser_capabilities(self, loglevel="INFO"):
...
+                self.builtin.log(
+                    f"caught exception {e} waiting for app launcher; retrying", "DEBUG"
+                )
+                self.selenium.press_keys("sf:modal.is_open", "ESCAPE")
+                self.wait_until_modal_is_closed()
+                self.open_app_launcher(retry=False)
+            else:
+                self.builtin.log(
+                    f"caught exception waiting for app launcher; not retrying", "DEBUG"
+                )
```
davisagli, line 528: Shouldn't we re-raise the exception here to prevent continuation of the test case, and to trigger the capture_screenshot_on_error mechanism?

boakley:
Good catch.  I've pushed that change.

PR #1685, opened 2020-04-17
Don't prompt for scratch org recreation

### LINKED ISSUE #1136: Allow bypassing scratch org recreation prompt via config
```console
$ ccfr dependencies --org fev
The scratch org is expired
Attempt to recreate the scratch org? [Y/n]:
```
Setting `always_recreate: True` should bypass the prompt and always recreate the scratch org.

jstvz:
This commit suppreses the user prompt upon scratch org expiration.

----

# Critical Changes

# Changes
- We've removed the prompt users see when a scratch org expires, and now automatically recreate the scratch org.

# Issues Closed
- Fixes #1136

### DIFF
--- a/cumulusci/cli/runtime.py
+++ b/cumulusci/cli/runtime.py
@@ -98,23 +98,20 @@ class CliRuntime(BaseCumulusCI):
     def check_org_expired(self, org_name, org_config):
         if org_config.scratch and org_config.date_created and org_config.expired:
             click.echo(click.style("The scratch org is expired", fg="yellow"))
-            if click.confirm("Attempt to recreate the scratch org?", default=True):
-                self.keychain.create_scratch_org(
-                    org_name,
-                    org_config.config_name,
-                    days=org_config.days,
-                    set_password=org_config.set_password,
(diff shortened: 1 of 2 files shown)


cumulusci/cli/runtime.py
```diff
@@ -98,23 +98,20 @@ def get_org(self, org_name=None, fail_if_missing=True):
...
-                    "to remove the org and then recreate the config manually"
+            self.keychain.create_scratch_org(
+                org_name,
+                org_config.config_name,
+                days=org_config.days,
+                set_password=org_config.set_password,
+            )
+            click.echo(
+                click.style(
+                    "Org config was refreshed, attempting to recreate scratch org",
```
davisagli, line 109: Let's move this above the call to create_scratch_org so it's shown even if scratch org creation fails. And let's change it to "Attempting to recreate expired scratch org." I don't think we have reason to conclude that the org config was refreshed.

davisagli: never mind, I forgot create_scratch_org just overwrites the keychain entry and it doesn't actually get created until we call create_org

PR #1691, opened 2020-04-20
Week 17 dependency updates

Br4nd0R:


Br4nd0R:
@davisagli - Looks like the snowfakery dependency in `cci` will now cause issues in dependency updates if certain modules are running different versions. I'm hoping this is only a one-way collision; in that, we can update jinja2 in snowfakery without issue and then circle back to update it here.

PR #1694, opened 2020-04-21
Death to OrderedDicts

prescod:
OrderDicts are not convenient from the point of view of repr, YAML, pprint and other libraries that prefer standard dicts. They are also less efficient and offer no extra benefit in modern versions of Python.

This PR replaces their use in simple_salesforce with standard dicts.

### DIFF
--- a/cumulusci/__init__.py
+++ b/cumulusci/__init__.py
@@ -1,5 +1,6 @@
 import os
 import sys
+from simple_salesforce import api, bulk
 
 __import__("pkg_resources").declare_namespace("cumulusci")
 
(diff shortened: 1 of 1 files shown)


prescod:
I'd like to see simple_salesforce reimplemented as well, but we've been saying that for a year or more and this two-line change improves our lives until someone gets around to it. If someone submits a patch to replace simple_salesforce with something better then I'd suggest we accept that instead. In the meantime I think this helps at low risk and effort.

prescod:
I updated the PR description to be good for the README.

PR #1696, opened 2020-04-22
Feature/robot classic detection

boakley:
This contains an attempt to automatically click the "switch to lightning" link in a robot test when
we recognize that the scratch org comes up in classic.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -10,7 +10,10 @@ from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
 from robot.utils import timestr_to_secs
 from selenium.webdriver.common.keys import Keys
 from selenium.webdriver.common.action_chains import ActionChains
-from selenium.common.exceptions import StaleElementReferenceException
+from selenium.common.exceptions import (
+    StaleElementReferenceException,
+    NoSuchElementException,
+)
 import faker
 
 from simple_salesforce import SalesforceResourceNotFound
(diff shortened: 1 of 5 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -1100,27 +1105,15 @@ def wait_until_salesforce_is_ready(self, locator=None, timeout=None, interval=5)
                     self.selenium.capture_page_screenshot()
                     raise Exception("Timed out waiting for a lightning page")
```
davisagli, line 1106: Will this result in two screenshots due to the added use of the `capture_screenshot_on_error` decorator?

cumulusci/robotframework/Salesforce.py
```diff
@@ -1131,3 +1124,58 @@ def breakpoint(self):
...
+        auto-click the "switch to lightning" link
+
+        """
+        try:
+            # we don't actually want to wait here, but if we don't
+            # explicitly wait, we'll implicitly wait longer than
+            # necessary.  This needs to be a quick-ish check.
+            self.selenium.wait_until_element_is_visible(
+                "class:switch-to-lightning", timeout=2
+            )
```
davisagli, line 1142: 2 seconds is still not quick if it has to happen at the start of every test even when things are working properly. Can we invert the check so that we're looking for something that is expected to be on every lightning page (maybe the #auraLoadingBox) and only look for the switch-to-lightning link if that fails?

boakley: We're already doing that check. We only get to this section if we didn't find a specific lightning component. By the time this code runs, we've already waited for `document.readyState`, waited for aura, and waited for a lightning locator.

cumulusci/robotframework/Salesforce.py
boakley: Yes it will. Good catch.

PR #1697, opened 2020-04-22
Create record type source table automatically

prescod:
# Changes

Fix record type data loads.

The test fails but I'd rather @davidmreed validate the fix and determine whether its the test that's wrong or the "fixed" code? I don't see any other place in this file where that table is created and I think it needs to be.


davidmreed:
Can you describe the situation for which this is a fix?

If we are attempting to use the `_target_rt_mapping` table, the other table should already exist from the original extraction.

davidmreed:
Discussed a different approach with Paul; this change is not required.

PR #1698, opened 2020-04-22
Fix encoding issues when reading files in windows

### LINKED ISSUE #1619: Windows: Not able to create a gist with a successful flow output
Per @davisagli - looks like we're not handling special characters correctly on Windows.

steps:
1. Run a flow in a repo and it completes successfully 
2. Try to share the flow so can see tasks run with someone by creating a gist
3. Run `cci error gist`
Get an error:
Error: 'charmap' codec can't decode byte 0x90 in position 117552: character maps to <undefined>

See full output here (had to manually create this gist)
https://gist.github.com/annejackson/aef0088833f7f59e5362cecd7bd03e3d

Br4nd0R:
# Critical Changes

# Changes

# Issues Closed
* Fixed issue where logfile was not being read in 'utf-8' format on Windows. (closes #1619)

### DIFF
--- a/cumulusci/utils/logging.py
+++ b/cumulusci/utils/logging.py
@@ -41,7 +41,7 @@ def tee_stdout_stderr(args, logger, tempfile):
         logger.handlers[0].close()
 
         # log contents of tempfile to rotating log files
-        with open(tempfile, "r", encoding="utf-8", errors="replace") as f:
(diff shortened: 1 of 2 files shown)


davisagli:
I'm thinking we should adjust the test_tee_stdout_stderr test to include one of the characters that causes the problem. (e.g. ┌ )

PR #1700, opened 2020-04-24
Add Duplicate Rule Metadata ETL task

davidmreed:
# Critical Changes

# Changes

- A new Metadata ETL task, `set_duplicate_rule_status`, allows selective activation and deactivation of Duplicate Rules.

# Issues Closed


cumulusci/tasks/metadata_etl/duplicate_rules.py
```diff
@@ -0,0 +1,24 @@
...
+        **MetadataSingleEntityTransformTask.task_options,
+    }
+
+    def _transform_entity(
+        self, metadata: MetadataElement, api_name: str
+    ) -> Optional[MetadataElement]:
+        status = "true" if process_bool_arg(self.options["active"]) else "false"
+        metadata.find("isActive").text = status
+
+        return metadata
```
davisagli, line 24: Would it make sense for this to be a more generic UpdateMetadataProperty class, with `entity` and `name` options that could be specified in the yaml task config for `set_duplicate_rule_status`? (And the `active` option renamed to `value`)

On the other hand then we would lose the specific description for the `active` option.

davidmreed: My goal is ultimately to move all of the logic from inside these task classes to a parallel hierarchy of transform classes to make them composable inside a putative "enablement script", with a single retrieve and single atomic deploy. 

Would you be comfortable keeping the current structure pending that refactor? It's a V2MOM objective for me.

davisagli: Yes. And that sounds quite cool.

PR #1718, opened 2020-04-30
Do not include default step bulk_mode: get it from CCI config

prescod:
During the switch over to Pydantic I added a default bulk_mode but this prevents the task configuration from overriding the per-step bulk_mode.


cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -32,7 +32,9 @@ class MappingStep(CCIDictModel):
     action: str = "insert"
     oid_as_pk: bool = False  # this one should be discussed and probably deprecated
     record_type: str = None  # should be discussed and probably deprecated
-    bulk_mode: Literal["Serial", "Parallel"] = "Parallel"
+    bulk_mode: Literal[
+        "Serial", "Parallel"
+    ] = None  # default should come from task options
```
davisagli, line 37: Does this need to be an Optional literal now?

prescod:
> I think you meant to put Optional on the type, not the value, throughout.

That's what I get for rushing to finish before a meeting!

PR #1723, opened 2020-05-02
Better error handling for Metadata SOAP

prescod:
I don't know for sure if this is a backwards compatible change but it makes more sense architecturally and has a better UX in the situation tested in the test-case.


davisagli:
@prescod So this is specifically to handle the case where the response has `<done>true</done>` but also has an `errorMessage`, right?

PR #1732, opened 2020-05-05
Data dictionary: Add column with field description

jofsky:
# Critical Changes

# Changes

# Issues Closed


davisagli:
What's the context for this? My first reaction is that it might make more sense to always output both the help text and description in separate columns (it's easy enough to delete a column after the data dictionary is generated if it's not desired) rather than controlling it using a task option.

cumulusci/cumulusci.yml
```diff
@@ -178,6 +178,10 @@ tasks:
         description: Lists open pull requests in project Github repository
         class_path: cumulusci.tasks.github.PullRequests
         group: GitHub
+    github_protect_branch:
+        description: Protects master branch in project Github repository
+        class_path: cumulusci.tasks.github.protect_branch.ProtectBranch
+        group: GitHub
```
davisagli, line 184: @jofsky accidental commit?

cumulusci/tasks/datadictionary.py
```diff
@@ -4,7 +4,7 @@
 
 from distutils.version import LooseVersion
 
-
+from cumulusci.core.utils import process_bool_arg
```
davisagli, line 7: Please remove this import which you're no longer using.

And please make sure your editor runs flake8 on save, which will warn you about things like this.

cumulusci/tasks/tests/test_datadictionary.py
```diff
@@ -628,6 +638,17 @@ def test_init_options__defaults(self):
...
+    def test_init_options__defaults(self):
+        project_config = create_project_config()
+        project_config.project__name = "Project"
+
+        task = create_task(
+            GenerateDataDictionary, {"release_prefix": "rel/"}, project_config
+        )
+
+        assert task.options["object_path"] == "Project sObject Data Dictionary.csv"
+        assert task.options["field_path"] == "Project Field Data Dictionary.csv"
```
davisagli, line 650: You added another copy of test_init_options__defaults, which is simply replacing the prior definition. This is another example of a mistake that flake8 would tell you about.

cumulusci/cumulusci.yml
jofsky: indeed good catch.

PR #1740, opened 2020-05-08
Create Add Picklist Values Metadata ETL task

davidmreed:
# Critical Changes

# Changes

- A new Metadata ETL task, `add_picklist_entries`, safely adds picklist values to an existing custom field.

# Issues Closed


davidmreed:
Tested with K12 and appears to work - needs hands-on review from that team and @jofsky to make sure it hits objectives.

cumulusci/tasks/metadata_etl/__init__.py
```diff
@@ -18,6 +19,7 @@
     MetadataSingleEntityTransformTask,
     AddRelatedLists,
     AddPermissionSetPermissions,
+    AddPicklistEntries,
```
davisagli, line 22: Can we skip adding this to `__init__.py` and use the full module path for the class_path? We're trying to move away from the pattern of `__init__.py` convenience imports because a) we have to import all the tasks in order to use one of them, and b) it can lead to messy import cycles if one of the task imports another one.

cumulusci/tasks/metadata_etl/picklists.py
```diff
@@ -0,0 +1,185 @@
...
+            "ignored, and * is a wildcard. Default behavior is to do nothing."
+        },
+        "sorted": {
+            "description": "Set whether or not to sort the picklist values alphabetically. Default is no change."
+        },
+        **MetadataSingleEntityTransformTask.task_options,
+    }
+
+    def _init_options(self, kwargs):
+        self.task_config.options["api_names"] = "dummy"
```
davisagli, line 37: ?

cumulusci/tasks/metadata_etl/picklists.py
```diff
@@ -0,0 +1,185 @@
...
+                )
+            if not field.endswith("__c"):
+                raise TaskOptionsError(
+                    "This task only supports custom fields. To modify "
+                    "Standard Value Sets, use the add_standard_value_set_entries task."
+                )
+
+            self.picklists.setdefault(self._inject_namespace(obj), []).append(
+                self._inject_namespace(field)
+            )
```
davisagli, line 62: You could initialize self.picklists as `defaultdict(lambda: defaultdict(list))` and then this would just be:

    self.picklists[self._inject_namespace(obj)].append(self._inject_namespace(field))

cumulusci/tasks/metadata_etl/picklists.py
davidmreed: Quirk of the Metadata ETL framework. To be refactored when I pull out the service tasks to build Enablement Scripts.

davidmreed:
Functionality has been confirmed to meet K12 needs with Matt Blanski.

PR #1741, opened 2020-05-08
CCI Shell Enhancements:

prescod:
cci org shell enhancements:

1. Better banner message.
2. More help available
3. add query() function
4. add describe() function


cumulusci/cli/cci.py
```diff
@@ -1152,21 +1171,28 @@ def org_shell(runtime, org_name, script=None, python=None):
...
     sf = get_simple_salesforce_connection(runtime.project_config, org_config)
+    printer = type(copyright)
+
+    helpers = repl_helpers(sf)
+
     globals = {
         "sf": sf,
         "org_config": org_config,
         "project_config": runtime.project_config,
+        "cci": printer("cci", org_shell_cci_help_message),
```
davisagli, line 1182: I think people will get confused about the difference between `cci` at a shell prompt and `cci` inside the cci prompt. What about using `help` instead (i.e. make it be an object whose `__repr__` prints this message, but passes through to the help builtin when called)?

cumulusci/cli/ui.py
```diff
@@ -133,3 +133,137 @@ def _dim_value(self, val):
...
+        )
+
+    query.__doc__ = pretty_soql_query.__doc__.replace("pretty_soql_query", "query")
+
+    def describe(obj_name, detailed=False, format="pprint"):
+        return pretty_describe(sf, obj_name, detailed=detailed, format=format)
+
+    describe.__doc__ = pretty_describe.__doc__.replace("pretty_describe", "describe")
+
+    return locals()
```
davisagli, line 269: This seems like a strange and unnecessarily clever way of defining these helpers and connecting them to a simple-salesforce instance. Is there a reason not to do something like this?

    class SalesforceAPIAdapter:
        def __init__(self, sf):
            self.sf = sf

        def pretty_soql_query(self, ...):
            ...

        def pretty_describe(self, ...):
            ...

and then use it like this:

    sf_adapter = SalesforceAPIAdapter(sf)
    globals = {
        ...
        "query": sf_adapter.pretty_soql_query,
        "describe": sf_adapter.pretty_describe,
    }

cumulusci/cli/ui.py
```diff
@@ -133,3 +133,137 @@ def _dim_value(self, val):
...
+           - "obj" -- ordinary Python objects
+           - "pprint" -- string in easily readable Python dict shape
+           - "json" -- JSON
+        * max_rows -- maximum rows to output, defaults to 100
+
+    For example:
+        pretty_soql_query("select Name, Id from Account", format="pprint")
+        contact_ids = pretty_soql_query("select count(Id) from Contact", format="obj")
+    """
+    results = sf.query(query, include_deleted=include_deleted)["records"]
```
davisagli, line 157: Careful -- `query` only returns the first batch of results (200 I think). In order to get the rest you have to do additional requests using `query_more`.

cumulusci/cli/cci.py
prescod: bc673a8e

cumulusci/cli/ui.py
prescod: Okay yeah, it just evolved from a partial. I've implemented a class similar to what you described: 51ab2f4d

cumulusci/cli/ui.py
```diff
@@ -133,3 +133,128 @@ def _dim_value(self, val):
...
+                - "obj" -- ordinary Python objects
+                - "pprint" -- string in easily readable Python dict shape
+                - "json" -- JSON
+            * max_rows -- maximum rows to output, defaults to 100
+
+        For example:
+            query("select Name, Id from Account", format="pprint")
+            contact_ids = query("select count(Id) from Contact", format="obj")
+        """
+        results = self._sf.query_all(query, include_deleted=include_deleted)["records"]
```
davisagli, line 193: Now we're getting everything even if we're going to throw away the results beyond max_rows. The more careful approach would be to go one batch at a time using query_more and return once max_rows is reached. But I'm okay with this as a first pass.

prescod: Yeah, it's just for quick interactive uses, not to replace simple_salesforce itself. The max_rows feature overlaps with the LIMIT feature in any case and is primarily just so you don't accidentally dump a million rows on your terminal if you forgot which org you were talking to.

PR #1746, opened 2020-05-12
REPL namespace included everything in scope.

prescod:
Cleans up REPL and command line exec namespaces.

Previously, the namespaces included everything in scope.

Now it is opt-in instead of everything-by-default.


cumulusci/cli/cci.py
```diff
@@ -301,15 +301,16 @@ def version():
 def shell(runtime, script=None, python=None):
     # alias for backwards-compatibility
     config = runtime  # noQA
+    variables = {"config": config, "runtime": runtime}
```
davisagli, line 304:
```suggestion
    variables = {"config": runtime, "runtime": runtime}
```

davisagli: Also it would be handy if we add project_config and sf here IF the user is in a project directory.

prescod: project_config will either be a useful project_config or nothing.

sf isn't available because we don't know what org. We'd need to have a wrapper for instantiating with an org name but I don't really see the urgency or the use-cases. cci org shell works for that.

PR #1747, opened 2020-05-12
Fix allow_failure

bethbrains:
# Critical Changes

# Changes
- `allow_failure` was broken. It now works. Handy for`dx` tasks that call `sfdx` commands that are likely to fail when repeated, like PermissionSetAssignments.

# Issues Closed


bethbrains:
This works with `ignore` instead of `allow`. Oops.

PR #1748, opened 2020-05-14
Installed package access in OrgConfig + Apex test code coverage enforcement

davidmreed:
# Critical Changes

# Changes

- The `OrgConfig` object exposes an `installed_packages` property, containing a `dict` mapping package namespaces to `StrictVersion` objects with the installed version number. This property is accessible in `when` clauses and is updated automatically when CumulusCI performs a Metadata API deployment.
- `when` clauses may create `StrictVersion` objects, which can be compared with installed package versions.
- The `run_tests` task supports a new option, `required_org_code_coverage_percent`. If set, the task will fail if aggregate code coverage in the org is less than the configured value. Code coverage verification is available only in unmanaged builds.

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -563,6 +563,15 @@ def project_init(runtime):
         test_name_match = None
     context["test_name_match"] = test_name_match
 
+    context["code_coverage"] = None
+    if click.confirm(
+        click.style("Do you want to enforce Apex code coverage?", bold=True),
+        default=True,
+    ):
+        context["code_coverage"] = click.prompt(
+            click.style("Minimum code coverage percentage", bold=True), default=75
+        )
+
     # Render templates
     for name in (".gitignore", "README.md", "cumulusci.yml"):
         template = env.get_template(name)
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -565,7 +565,9 @@ def project_init(runtime):
 
     context["code_coverage"] = None
     if click.confirm(
-        click.style("Do you want to enforce Apex code coverage?", bold=True),
+        click.style(
+            "Do you want to check Apex code coverage when tests are run?", bold=True
+        ),
         default=True,
     ):
         context["code_coverage"] = click.prompt(
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -188,13 +188,13 @@ class OrgConfig(BaseConfig):
         Use a package Id to handle this circumstance."""
         installed_version = self.installed_packages.get(package_identifier)
 
-        if len(installed_version) > 1:
+        if not installed_version:
+            return False
+        elif len(installed_version) > 1:
             raise CumulusCIException(
                 f"Cannot check installed version of {package_identifier}, because multiple "
                 f"packages are installed that match this identifier."
             )
-        elif not installed_version:
-            return False
 
         return installed_version[0] >= version_identifier
 
@@ -210,7 +210,7 @@ class OrgConfig(BaseConfig):
         Beta version of a package are represented as "1.2.3b5", where 5 is the build number."""
         if not self._installed_packages:
             response = self.salesforce_client.restful(
-                "tooling/query/?q=SELECT SubscriberPackage.NamespacePrefix, SubscriSubscriberPackageVersion.MajorVersion, "
+                "tooling/query/?q=SELECT SubscriberPackage.Id, SubscriberPackage.NamespacePrefix, SubscriberPackageVersion.MajorVersion, "
                 "SubscriberPackageVersion.MinorVersion, SubscriberPackageVersion.PatchVersion,  "
                 "SubscriberPackageVersion.BuildNumber, SubscriberPackageVersion.IsBeta "
                 "FROM InstalledSubscriberPackage"
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -1,12 +1,12 @@
+from collections import defaultdict
+from distutils.version import StrictVersion
 import os
 
-from distutils.version import StrictVersion
 import requests
-
 from simple_salesforce import Salesforce
 
 from cumulusci.core.config import BaseConfig
-from cumulusci.core.exceptions import SalesforceCredentialsException
+from cumulusci.core.exceptions import SalesforceCredentialsException, CumulusCIException
 from cumulusci.oauth.salesforce import SalesforceOAuth2
 from cumulusci.oauth.salesforce import jwt_session
 
@@ -79,17 +79,20 @@ class OrgConfig(BaseConfig):
             self._client = Salesforce(
                 instance=self.instance_url.replace("https://", ""),
                 session_id=self.access_token,
-                version="45.0",
+                version=self.latest_api_version,
             )
+
         return self._client
 
     @property
     def latest_api_version(self):
         if not self._latest_api_version:
-            response = self.salesforce_client._call_salesforce(
-                "GET", f"https://{self.salesforce_client.sf_instance}/services/data"
+            headers = {"Authorization": "Bearer " + self.access_token}
+            response = requests.get(
+                self.instance_url + "/services/data", headers=headers
             )
             self._latest_api_version = str(response.json()[-1]["version"])
+
         return self._latest_api_version
 
     @property
@@ -131,12 +134,8 @@ class OrgConfig(BaseConfig):
         return False
 
     def _load_orginfo(self):
-        headers = {"Authorization": "Bearer " + self.access_token}
-        self._org_sobject = requests.get(
-            self.instance_url
-            + f"/services/data/v45.0/sobjects/Organization/{self.org_id}",
-            headers=headers,
-        ).json()
+        self._org_sobject = self.salesforce_client.Organization.get(self.org_id)
+
         result = {
             "org_type": self._org_sobject["OrganizationType"],
             "is_sandbox": self._org_sobject["IsSandbox"],
@@ -149,11 +148,7 @@ class OrgConfig(BaseConfig):
 
     def _fetch_community_info(self):
         """Use the API to re-fetch information about communities"""
-        headers = {"Authorization": "Bearer " + self.access_token}
-        response = requests.get(
-            self.instance_url + "/services/data/v45.0/connect/communities",
-            headers=headers,
-        ).json()
+        response = self.salesforce_client.restful("connect/communities")
 
         # Since community names must be unique, we'll return a dictionary
         # with the community names as keys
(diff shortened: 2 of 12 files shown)


cumulusci/tasks/apex/testrunner.py
```diff
@@ -512,6 +527,28 @@ def _run_task(self):
...
+        else:
+            self.logger.info(
+                "No code coverage level specified; not checking code coverage."
+            )
+
+    def _check_code_coverage(self):
+        result = self.tooling.query("SELECT PercentCovered FROM ApexOrgWideCoverage")
+        coverage = result["records"][0]["PercentCovered"]
+        if coverage < self.code_coverage_level:
+            raise ApexTestException(
```
davidmreed, line 544: Does using `ApexTestException` muddy the water too much? Perhaps this should be a new exception class.

cumulusci/core/config/OrgConfig.py
```diff
@@ -178,3 +180,27 @@ def get_community_info(self, community_name, force_refresh=False):
             )
 
         return self._community_info_cache[community_name]
+
+    @property
+    def installed_packages(self):
+        if not self._installed_packages:
+            response = self.salesforce_client.restful(
```
davisagli, line 187: While we're in here, can we update salesforce_client to use the org's API version rather than being hardcoded to 45.0?

cumulusci/core/config/OrgConfig.py
```diff
@@ -178,3 +180,27 @@ def get_community_info(self, community_name, force_refresh=False):
...
+                "tooling/query/?q=SELECT SubscriberPackage.NamespacePrefix, SubscriberPackageVersion.MajorVersion, "
+                "SubscriberPackageVersion.MinorVersion, SubscriberPackageVersion.PatchVersion "
+                "FROM InstalledSubscriberPackage"
```
davisagli, line 190: What about beta package versions? (IsBeta = true)

cumulusci/core/config/OrgConfig.py
```diff
@@ -178,3 +180,27 @@ def get_community_info(self, community_name, force_refresh=False):
...
+
+            self._installed_packages = {}
+            for package in response["records"]:
+                sp = package["SubscriberPackage"]
+                spv = package["SubscriberPackageVersion"]
+                # PatchVersion is a 0 on a non-patch version.
+                version = (
+                    f"{spv['MajorVersion']}.{spv['MinorVersion']}.{spv['PatchVersion']}"
+                )
+                self._installed_packages[sp["NamespacePrefix"]] = StrictVersion(version)
```
davisagli, line 201: Hmm -- In the 2GP world there could be multiple SPVs sharing the same prefix (or no prefix)

cumulusci/core/flowrunner.py
```diff
@@ -359,6 +359,7 @@ def run(self, org_config):
                     jinja2_context = {
                         "project_config": step.project_config,
                         "org_config": self.org_config,
+                        "StrictVersion": StrictVersion,
```
davisagli, line 362: fwiw it looks like StrictVersion implements comparison in a way that will cast compared strings to StrictVersion. So we could probably do this without having StrictVersion in the expression context:

    "npsp" in org_config.installed_packages and org_config.installed_packages["npsp"] <= "3.175"

but that's still a bit of a mouthful; maybe OrgConfig should have a has_sufficient_package_version(namespace, minimum_version) method?

cumulusci/tasks/apex/testrunner.py
```diff
@@ -210,6 +213,18 @@ def _init_options(self, kwargs):
...
+            try:
+                self.code_coverage_level = int(
+                    self.options["required_org_code_coverage_percent"]
+                )
+            except ValueError:
+                raise TaskOptionsError(
+                    f"Invalid code coverage level {self.options['required_org_code_coverage_percent']}"
+                )
+        else:
+            self.code_coverage_level = None
```
davisagli, line 226: I really need to find time to review Paul's PR adding declarative option validation using pydantic...

cumulusci/tasks/apex/testrunner.py
```diff
@@ -512,6 +527,28 @@ def _run_task(self):
                 )
             )
 
+        if self.code_coverage_level:
+            if self.options.get("namespace") not in self.org_config.installed_packages:
+                self._check_code_coverage()
+            else:
+                self.logger.info("This is a managed org; not checking code coverage.")
```
davisagli, line 534:
```suggestion
                self.logger.info("This is a managed package; not checking code coverage.")
```

cumulusci/tasks/apex/testrunner.py
davisagli: Using ApexTestException has the advantage of being in the list of exceptions that MetaCI will count as a failure rather than an error. Creating a new subclass of ApexTestException would also work.

cumulusci/core/config/OrgConfig.py
```diff
@@ -178,3 +180,27 @@ def get_community_info(self, community_name, force_refresh=False):
             )
 
         return self._community_info_cache[community_name]
+
+    @property
+    def installed_packages(self):
```
davisagli, line 185: docstring please

cumulusci/core/config/OrgConfig.py
davidmreed: Betas are treated as if they are the upcoming major version, i.e., 1.1.0 Beta 1 is treated as 1.1.0. I think that this will produce the correct behavior where a comparison is being performed. Automation should not be attempting to test for equality to a specific beta.

cumulusci/core/config/OrgConfig.py
davidmreed: That's a really good point and I'm not sure off the cuff what the right strategy is to handle it.

We could use `AllPackageId` alongside the namespace: populate the `AllPackageId` for every package, with or without namespace, and log a warning and do not populate a version for any package where multiple packages are sharing a namespace.

cumulusci/core/config/OrgConfig.py
davisagli: That makes sense for the preflight check use case, but it seems like there may be other use cases where it's important to be able to tell if it's a beta or not. If we synthesized a version like the ones used on platform (i.e. 1.1.0 Beta 1 -- does that correspond to the build number?) and used LooseVersion, I think it would still work okay for the comparisons:

    >>> LooseVersion("3.175 Beta 1") >= "3.175"
    True

cumulusci/core/config/OrgConfig.py
davisagli: I was thinking it might make sense to provide a namespace mapping-based API that intentionally only includes 1GP package versions, and then a separate one that requires using package ids. But I'm not sure we have a way to filter the SPV query to get only 1gp results. Possibly using the Package2ContainerOptions field?

davidmreed: I like that idea and it works nicely with your other suggestion to provide a function on `org_config` to abstract away the comparison. I'll investigate to confirm if we can tell 1GPs and 2GPs apart.

davidmreed:
@jstvz  The reason I wasn't setting a default for the coverage level is that 

 - It's not relevant to non-managed package projects
 - Some projects stub out functionality before they write tests, and putting a default coverage level in would cause all their Feature builds to fail.

What about a different tack - add a Code Coverage question to the new project interview? That'd make it more visible, and the default _there_ could be to opt in to 75% coverage checks.

jstvz:
@davidmreed Perfectly reasonable. I tend to lean towards being more opinionated, and I'd personally set a minimum coverage for any project, managed or not.

But I'm fine tabling this for now and discussing in more detail further down the road.

cumulusci/core/config/OrgConfig.py
davidmreed: I don't think that outcome is quite right, though. 3.175 Beta 1 should be <= 3.175, while 3.176 Beta 1 > 3.175 and 3.175 Beta N.

`StrictVersion` supports a `bN` suffix on the version number that seems to exhibit the right sorting behavior, e.g.,

`StrictVersion("1.10b1") < StrictVersion("1.10")` => `True`
`StrictVersion("1.10b1") < StrictVersion("1.10b2")` => `True`

The `BuildNumber` field _mostly_ corresponds to the Beta number; it's the count of builds against that major release. So 1.8 Beta 9 has BuildNumber 9, and 1.8 has BuildNumber 10. That should suffice to generate real, comparable version numbers for betas.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -210,6 +213,18 @@ def _init_options(self, kwargs):
 
         self.counts = {}
 
+        if "required_org_code_coverage_percent" in self.options:
+            try:
+                self.code_coverage_level = int(
+                    self.options["required_org_code_coverage_percent"]
```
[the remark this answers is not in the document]
davidmreed: Compassionate release engineering, that. Applying.

cumulusci/core/config/OrgConfig.py
davidmreed: I adopted the `b` model for betas. See what you think in the latest version.

cumulusci/cli/cci.py
```diff
@@ -563,6 +563,15 @@ def project_init(runtime):
...
     context["test_name_match"] = test_name_match
 
+    context["code_coverage"] = None
+    if click.confirm(
+        click.style("Do you want to enforce Apex code coverage?", bold=True),
+        default=True,
+    ):
+        context["code_coverage"] = click.prompt(
+            click.style("Minimum code coverage percentage", bold=True), default=75
+        )
```
davisagli, line 573: @davidmreed @jstvz Is this just going to mislead people into thinking they can change what coverage the _platform_ requires for their managed package?

How about: Remove the prompt. Set this to 75 if the user answered yes to the "managed package" question, otherwise don't.

jstvz: I’m not sure I entirely agree. The platform enforces 75% code coverage for any production deployment. I suggest warning the user when coverage <75% instead.

cumulusci/core/config/OrgConfig.py
```diff
@@ -178,3 +175,63 @@ def get_community_info(self, community_name, force_refresh=False):
...
+        Beta version of a package are represented as "1.2.3b5", where 5 is the build number."""
+        if not self._installed_packages:
+            response = self.salesforce_client.restful(
+                "tooling/query/?q=SELECT SubscriberPackage.NamespacePrefix, SubscriSubscriberPackageVersion.MajorVersion, "
+                "SubscriberPackageVersion.MinorVersion, SubscriberPackageVersion.PatchVersion,  "
+                "SubscriberPackageVersion.BuildNumber, SubscriberPackageVersion.IsBeta "
+                "FROM InstalledSubscriberPackage"
+            )
+
+            self._installed_packages = defaultdict(lambda: [])
```
davisagli, line 219:
```suggestion
            self._installed_packages = defaultdict(list)
```

cumulusci/core/config/OrgConfig.py
```diff
@@ -178,3 +175,63 @@ def get_community_info(self, community_name, force_refresh=False):
...
+                # PatchVersion is a 0 on a non-patch version.
+                version = (
+                    f"{spv['MajorVersion']}.{spv['MinorVersion']}.{spv['PatchVersion']}"
+                )
+                if spv["IsBeta"]:
+                    version += f"b{spv['BuildNumber']}"
+                self._installed_packages[sp["NamespacePrefix"]].append(
+                    StrictVersion(version)
+                )
+                self._installed_packages[sp["Id"]].append(StrictVersion(version))
```
davisagli, line 232: Fortunately namespaces have to start with a letter so we don't have to worry about someone who tried to set their namespace to the 033 id!

cumulusci/cli/cci.py
davisagli: Isn't the point to break feature builds that have insufficient coverage though? I think a warning would get lost.

I guess I'm fine with including this prompt here, but I think it would also be okay to just omit it, let the option default to 75, and users can override it if they aren't developing for a production environment.

jstvz: The warning would be here, in the context of the wizard, not builds. 

Omitting it works too.

davidmreed: I changed the initial prompt to try to clarify what we're actually doing here. 

> Do you want to check Apex code coverage when tests are run?

Does that help cover your concerns, @davisagli? I do want to be clear to users that what they're opting in to is a CCI behavior (a build failure), not anything at the platform level.

PR #1749, opened 2020-05-14
Feature/org prune

bethbrains:
# Critical Changes

# Changes
- Adds `cci org prune` to automatically remove all expired scratch orgs from the keychain (`cci org list`)

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1053,6 +1053,41 @@ def org_list(runtime, plain):
     persistent_table.echo(plain)
 
 
+@org.command(name="prune", help="Removes all expired orgs from the current project")
+@pass_runtime(require_project=False, require_keychain=True)
+def org_prune(runtime):
+    org_configs = {
+        org: runtime.keychain.get_org(org) for org in runtime.keychain.list_orgs()
+    }
+    scratch_configs = getattr(runtime.project_config, "orgs__scratch", {})
+
+    expired_orgs_removed = []
+    org_shapes_skipped = []
+    active_orgs_skipped = []
+    for org, org_config in org_configs.items():
+        if scratch_configs.get(org):
+            org_shapes_skipped.append(org)
+        elif org_config.active:
+            active_orgs_skipped.append(org)
+        elif isinstance(org_config, ScratchOrgConfig) and not org_config.active:
+            try:
+                runtime.keychain.remove_org(org)
+                expired_orgs_removed.append(org)
+            except Exception as e:
+                click.echo(e)
+    # see test_org_remove_delete_error for how to mock click.echo and assert the messages
+    if expired_orgs_removed:
+        click.echo(
+            f"Successfully removed {len(expired_orgs_removed)} expired scratch orgs: {', '.join(expired_orgs_removed)}"
+        )
+    else:
+        click.echo("No expired scratch orgs to delete. ✨")
+    if org_shapes_skipped:
+        click.echo(f"Skipped org shapes: {', '.join(org_shapes_skipped)}")
+    if active_orgs_skipped:
+        click.echo(f"Skipped active orgs: {', '.join(active_orgs_skipped)}")
+
+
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cci.py
```diff
@@ -1053,6 +1053,41 @@ def org_list(runtime, plain):
     persistent_table.echo(plain)
 
 
+@org.command(name="prune", help="Removes all expired orgs from the current project")
+@pass_runtime(require_project=False, require_keychain=True)
```
davisagli, line 1057:
```suggestion
@pass_runtime(require_project=True, require_keychain=True)
```

to make sure CCI raises an error proactively if you try using this command outside a cci project folder, rather than a more obscure error later when something tries to use project_config but it's None

cumulusci/cli/cci.py
```diff
@@ -1053,6 +1053,41 @@ def org_list(runtime, plain):
...
+def org_prune(runtime):
+    org_configs = {
+        org: runtime.keychain.get_org(org) for org in runtime.keychain.list_orgs()
+    }
+    scratch_configs = getattr(runtime.project_config, "orgs__scratch", {})
+
+    expired_orgs_removed = []
+    org_shapes_skipped = []
+    active_orgs_skipped = []
+    for org, org_config in org_configs.items():
```
davisagli, line 1067: I think you could do:

```suggestion
    for org, org_config in runtime.keychain.orgs.items():
```

and skip setting org_configs above.

Also if it were me I'd call it `org_name` instead of `org`

cumulusci/cli/cci.py
```diff
@@ -1053,6 +1053,41 @@ def org_list(runtime, plain):
...
+        if scratch_configs.get(org):
+            org_shapes_skipped.append(org)
```
davisagli, line 1069: So this is to avoid removing the scratch orgs that are predefined in cumulusci.yml? It might actually be fine to remove them; they should get recreated the next time the keychain is initialized.

If we _do_ need this check, I would probably call it `predefined_scratch_configs` instead of `scratch_configs`, and write the conditional as `if org in scratch_configs:` (you can use `in` to check if a dict contains a key)

cumulusci/cli/cci.py
```diff
@@ -1053,6 +1053,41 @@ def org_list(runtime, plain):
...
+        elif org_config.active:
+            active_orgs_skipped.append(org)
+        elif isinstance(org_config, ScratchOrgConfig) and not org_config.active:
+            try:
+                runtime.keychain.remove_org(org)
+                expired_orgs_removed.append(org)
+            except Exception as e:
+                click.echo(e)
```
davisagli, line 1077: I think these 2 lines are unnecessary: if there's an exception it will fall through to the generic exception handler which also prints the exception (as well as making sure the traceback gets logged to where `cci error info` can find it).

cumulusci/cli/cci.py
```diff
@@ -1053,6 +1053,41 @@ def org_list(runtime, plain):
...
+    # see test_org_remove_delete_error for how to mock click.echo and assert the messages
```
davisagli, line 1078: This looks like a comment that doesn't need to remain if you have a working test?

cumulusci/cli/cci.py
```diff
@@ -1053,6 +1053,63 @@ def org_list(runtime, plain):
...
+        elif org_config.active:
+            if include_active:
+                runtime.keychain.remove_org(org_name)
+                active_orgs_removed.append(org_name)
+            else:
+                active_orgs_skipped.append(org_name)
+
+        elif isinstance(org_config, ScratchOrgConfig):
+            runtime.keychain.remove_org(org_name)
+            expired_orgs_removed.append(org_name)
```
davisagli, line 1090: Haha, I tried to simplify this logic by reordering it and ended up with exactly the same number of lines.

davisagli:
@bethbrains also be sure to run black and merge changes from master.

Julian88Tex:
@bethbrains thanks for this!

PR #1757, opened 2020-05-19
Get rid of snake_cased mapping files entirely.

prescod:
A few code paths were still ending up with snake-cased column names.


davisagli:
Replaced by #1772

PR #1758, opened 2020-05-20
Feature/generate and load data cleanups

prescod:
Allow the user to specify rather than generate a mapping file (different than last week's PR)
    
Fix init_db so it is more similar to the (unfortunately not line-for-line identical) version in load.py when it comes to tables that appear in multiple mappings.
    
Change generate_from_yaml to reuse Snowfakery's database handling code instead of duplicating it.
    
Add a try:finally to close the outputstream "just in case".
    
Re-indent a bunch of code.
    
Never save a column named RecordType to Salesforce. It's presumably just there to represent the...actual record type. This prevents me from having to use one mapping file to generate the database tables and a different one to do the actual load.

### DIFF
--- a/cumulusci/tasks/bulkdata/base_generate_data_task.py
+++ b/cumulusci/tasks/bulkdata/base_generate_data_task.py
@@ -82,8 +82,9 @@ class BaseGenerateDataTask(BaseTask, metaclass=ABCMeta):
         metadata = MetaData()
         metadata.bind = engine
         if mappings:
-            for mapping in mappings.values():
-                create_table(mapping, metadata)
+            for name, mapping in mappings.items():
+                if "table" in mapping and mapping["table"] not in metadata.tables:
+                    create_table(mapping, metadata)
         metadata.create_all()
         base = automap_base(bind=engine, metadata=metadata)
         base.prepare(engine, reflect=True)
--- a/cumulusci/tasks/bulkdata/generate_from_yaml.py
+++ b/cumulusci/tasks/bulkdata/generate_from_yaml.py
@@ -87,11 +87,8 @@ class GenerateDataFromYaml(BaseGenerateDataTask):
                 self.mappings = yaml.safe_load(f)
         else:
             self.mappings = {}
-        session, engine, base = self.init_db(db_url, self.mappings)
         self.logger.info(f"Generating batch {current_batch_num} with {num_records}")
-        self.generate_data(session, engine, base, num_records, current_batch_num)
-        session.commit()
-        session.close()
+        self.generate_data(db_url, num_records, current_batch_num)
 
     def default_continuation_file_path(self):
         return Path(self.working_directory) / "continuation.yml"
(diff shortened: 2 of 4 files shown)


cumulusci/tasks/bulkdata/generate_from_yaml.py
```diff
@@ -137,36 +134,38 @@ def open_new_continuation_file(self) -> Optional[TextIO]:
...
+                summary = generate(
+                    open_yaml_file=open_yaml_file,
+                    user_options=self.vars,
+                    output_stream=output_stream,
+                    stopping_criteria=self.stopping_criteria,
+                    continuation_file=old_continuation_file,
+                    generate_continuation_file=new_continuation_file,
+                )
+        finally:
             output_stream.close()
```
davisagli, line 156: Maybe output streams should be context managers.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -191,6 +191,8 @@ def _get_columns(self, mapping):
         # If we're using Record Type mapping, `RecordTypeId` goes at the end.
         if "RecordTypeId" in columns:
             columns.remove("RecordTypeId")
+        if "RecordType" in columns:
+            columns.remove("RecordType")
```
davisagli, line 195: I didn't quite follow this:

> Never save a column named RecordType to Salesforce. It's presumably just there to represent the...actual record type. This prevents me from having to use one mapping file to generate the database tables and a different one to do the actual load.

I remember that Snowfakery supports using RecordType to refer to a record type by developer name rather than Id, but I forget how that actually gets loaded.

cumulusci/tasks/bulkdata/base_generate_data_task.py
```diff
@@ -82,8 +82,9 @@ def init_db(db_url, mappings):
         metadata = MetaData()
         metadata.bind = engine
         if mappings:
-            for mapping in mappings.values():
-                create_table(mapping, metadata)
+            for name, mapping in mappings.items():
+                if "table" in mapping and mapping["table"] not in metadata.tables:
+                    create_table(mapping, metadata)
```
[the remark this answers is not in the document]
prescod: Yes. It would be nicer to check but we don't anywhere else in our mapping file infrastructure. (e.g. Load.py doesn't check that the mapping file and the database correspond) Baby steps...

cumulusci/tasks/bulkdata/load.py
prescod: If Snowfakery generates the mapping.yml then the generated mapping.yml will filter out the RecordType column, but it will use it in a WHERE clause like this:

https://github.com/SalesforceFoundation/NPSP/compare/feature/pprescod/generate_and_load_test_data?expand=1#diff-1641945226b0d2a6f8cfecdfd177bb00R34

In one of my use cases, the mapping file is being fed in as input for reasons that I probably don't need to go into in this PR. This means that mapping file has a RecordType column declared which initializes the correct database column. CCI subsequently tries to generate a field with that name.

cumulusci/tasks/bulkdata/generate_from_yaml.py
prescod: I can do that in Snowfakery but I'd rather not upgrade them both in lockstep this time. I can do it in Snowfakery and then clean this up next time I'm in this code.

davisagli: Yeah, I meant it that way rather than a blocker to merge this.

cumulusci/tasks/bulkdata/load.py
davisagli: So it's because the same mapping file is used both by Snowfakery to initialize the db (where we need to populate RecordType) and by CCI to load it (where we don't want to include RecordType as a mapped field).

It feels like a bit of a leaky abstraction for CCI to hardcode the fact that something else will be using this specific standard-fieldname-that-is-not-actually-a-standard-field. But I'm not thinking of an alternate proposal at the moment. Can you please at least make sure there's a comment to remind us what use case this is for?

davisagli: This line is also missing coverage. You could cheat and put this removal inside a loop over the list of names to remove.

prescod: It's a little bit deeper than just the leaky abstraction. We do have a plan to have a first-class way of specifying RecordTypes by name rather than ID.

https://salesforce.quip.com/alywAQ9UBe7G

So this is a first step towards that special treatment as opposed to just a hack relating to Snowfakery. In fact, when I finish adding it to CCI, I'll probably be able to rip record type smarts out of Snowfakery and stop using a mapping file in this case.

I did notice, by the way, that there is one (and only one) sObject that actually does use a field named RecordType, but the sObject is read-only:

https://developer.salesforce.com/docs/atlas.en-us.sfFieldRef.meta/sfFieldRef/salesforce_field_reference_CombinedAttachment.htm

PR #1765, opened 2020-05-27
expose InstalledPackage securityType

davisagli:
# Critical Changes

# Changes
* The install_managed and update_dependencies tasks now accept a security_type option to specify which users the package should be installed for.

# Issues Closed

### DIFF
--- a/cumulusci/salesforce_api/package_zip.py
+++ b/cumulusci/salesforce_api/package_zip.py
@@ -28,6 +28,7 @@ INSTALLED_PACKAGE = u"""<?xml version="1.0" encoding="UTF-8"?>
 <InstalledPackage xmlns="http://soap.sforce.com/2006/04/metadata">
   <versionNumber>{}</versionNumber>
   <activateRSS>{}</activateRSS>
+  <securityType>{}</securityType>
   {}
 </InstalledPackage>"""
 
@@ -86,7 +87,9 @@ class CreatePackageZipBuilder(BasePackageZipBuilder):
 class InstallPackageZipBuilder(BasePackageZipBuilder):
     api_version = "43.0"
 
-    def __init__(self, namespace, version, activateRSS=False, password=None):
+    def __init__(
+        self, namespace, version, activateRSS=False, password=None, securityType="FULL"
+    ):
         if not namespace:
             raise ValueError("You must provide a namespace to install a package")
         if not version:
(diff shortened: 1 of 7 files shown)


cumulusci/tasks/salesforce/InstallPackageVersion.py
```diff
@@ -31,6 +31,9 @@ class InstallPackageVersion(Deploy):
         "retry_interval_add": {
             "description": "Number of seconds to add before each retry (default=30),"
         },
+        "security_type": {
+            "description": "Which users to install package for (FULL = all users, NONE = admins only)"
```
davidmreed, line 35: Let's expose `PUSH` too so we can replicate the user push upgrade experience in QA.

cumulusci/salesforce_api/package_zip.py
```diff
@@ -95,6 +98,7 @@ def __init__(self, namespace, version, activateRSS=False, password=None):
         self.version = version
         self.activateRSS = activateRSS
         self.password = password
+        self.securityType = securityType
```
davidmreed, line 101: Worth validating and throwing a `TaskOptionsError` just to not start an operation we know will fail?

cumulusci/tasks/salesforce/UpdateDependencies.py
```diff
@@ -305,7 +305,9 @@ def _install_dependency(self, dependency):
                     )
                 )
                 package_zip = InstallPackageZipBuilder(
-                    dependency["namespace"], dependency["version"]
+                    dependency["namespace"],
+                    dependency["version"],
+                    securityType=dependency.get("security_type", "FULL"),
```
davidmreed, line 310: What do you think about making this a task option rather than a per-dependency option?

The reason I suggest that is that I'm imagining the automation in a project wanting to be able to run `update_dependencies` in different contexts with different behaviors, but without repeating the dependency list so that it can set this option. It seems more amenable to future refactorability.

I don't see much of a use case for projects wanting to specify a different security model for each package it installs on a _package_ basis, but rather on a task or flow basis. I _think_ it's more likely to be "install everything with X model" or, for regression, "install everything with X model and then run install_beta with Y model".

cumulusci/salesforce_api/package_zip.py
davisagli: Good point re validation, but task options should really be validated in _init_options so that errors are thrown before the flow starts, not in code called by _run_task.

cumulusci/tasks/salesforce/UpdateDependencies.py
davisagli: @davidmreed good call.

PR #1766, opened 2020-05-28
Enhance Data Dictionary Generation

davidmreed:
# Critical Changes

# Changes

- The `generate_data_dictionary` task now has more features for complex projects. By default, the task will walk through all project dependencies and include them in the generated data dictionaries. Other non-dependency projects can be included with the `additional_dependencies` option. The output format has been extensively improved.

# Issues Closed

### DIFF
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -492,13 +492,11 @@ class BaseProjectConfig(BaseTaskFlowConfig):
             )
 
     def get_repo_from_url(self, url):
-        repo_owner, repo_name = url.split("/")[3:5]
-        if repo_name.endswith(".git"):
-            repo_name = repo_name[:-4]
-        gh = self.get_github_api(repo_owner, repo_name)
-        repo = gh.repository(repo_owner, repo_name)
+        splits = self._split_repo_url(url)
+        gh = self.get_github_api(splits["owner"], splits["name"])
+        repo = gh.repository(splits["owner"], splits["name"])
 
-        return (repo, repo_owner, repo_name)
+        return repo
 
     def get_ref_for_dependency(self, repo, dependency, include_beta=None):
         release = None
@@ -592,7 +590,9 @@ class BaseProjectConfig(BaseTaskFlowConfig):
             skip = [skip]
 
         # Initialize github3.py API against repo
-        repo, repo_owner, repo_name = self.get_repo_from_url(dependency["github"])
+        repo = self.get_repo_from_url(dependency["github"])
+        repo_owner = repo.owner
+        repo_name = repo.name
         if repo is None:
             raise DependencyResolutionError(
                 f"{indent}Github repository {dependency['github']} not found or not authorized."
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -591,12 +591,13 @@ class BaseProjectConfig(BaseTaskFlowConfig):
 
         # Initialize github3.py API against repo
         repo = self.get_repo_from_url(dependency["github"])
-        repo_owner = repo.owner
-        repo_name = repo.name
         if repo is None:
             raise DependencyResolutionError(
                 f"{indent}Github repository {dependency['github']} not found or not authorized."
             )
+
+        repo_owner = repo.owner
+        repo_name = repo.name
 
         # Determine the commit
         release, ref = self.get_ref_for_dependency(repo, dependency, include_beta)
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -593,7 +593,7 @@ class BaseProjectConfig(BaseTaskFlowConfig):
             )
 
         # Determine the commit
-        release, ref = self.get_ref_for_dependency(repo, dependency)
+        release, ref = self.get_ref_for_dependency(repo, dependency, include_beta)
 
         # Get the cumulusci.yml file
         contents = repo.file_contents("cumulusci.yml", ref=ref)
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -485,6 +485,42 @@ class BaseProjectConfig(BaseTaskFlowConfig):
                 + "config.set_keychain(keychain) before accessing orgs"
             )
 
+    def get_repo_from_url(self, url):
+        repo_owner, repo_name = url.split("/")[3:5]
+        if repo_name.endswith(".git"):
+            repo_name = repo_name[:-4]
+        gh = self.get_github_api(repo_owner, repo_name)
+        repo = gh.repository(repo_owner, repo_name)
+
+        return (repo, repo_owner, repo_name)
+
+    def get_ref_for_dependency(self, repo, dependency, include_beta=None):
+        release = None
+        if "ref" in dependency:
+            ref = dependency["ref"]
+        else:
+            if "tag" in dependency:
+                try:
+                    # Find the github release corresponding to this tag.
+                    release = repo.release_from_tag(dependency["tag"])
+                except NotFoundError:
+                    raise DependencyResolutionError(
+                        f"No release found for tag {dependency['tag']}"
+                    )
+            else:
+                release = find_latest_release(repo, include_beta)
+            if release:
+                ref = repo.tag(
+                    repo.ref("tags/" + release.tag_name).object.sha
+                ).object.sha
+            else:
+                self.logger.info(
+                    f"No release found; using the latest commit from the {repo.default_branch} branch."
+                )
+                ref = repo.branch(repo.default_branch).commit.sha
+
+        return (release, ref)
+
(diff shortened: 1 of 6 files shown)


cumulusci/core/config/project_config.py
```diff
@@ -485,6 +491,42 @@ def _check_keychain(self):
                 + "config.set_keychain(keychain) before accessing orgs"
             )
 
+    def get_repo_from_url(self, url):
+        repo_owner, repo_name = url.split("/")[3:5]
+        if repo_name.endswith(".git"):
+            repo_name = repo_name[:-4]
```
davisagli, line 497: Use self._split_repo_url(url) for this part.

cumulusci/core/config/project_config.py
```diff
@@ -485,6 +491,42 @@ def _check_keychain(self):
...
+                ref = repo.tag(
+                    repo.ref("tags/" + release.tag_name).object.sha
+                ).object.sha
+            else:
+                self.logger.info(
+                    f"No release found; using the latest commit from the {repo.default_branch} branch."
+                )
+                ref = repo.branch(repo.default_branch).commit.sha
+
+        return (release, ref)
```
davisagli, line 528: This doesn't have to happen right now, but I think we should be moving in the direction of making BaseProjectConfig resolve dependencies by delegating to Source objects (like we already do for resolving cross-project flows) so that we can have different Source implementations for different version control systems.

cumulusci/tasks/datadictionary.py
```diff
@@ -28,44 +54,139 @@ class GenerateDataDictionary(BaseGithubTask):
         "field_path": {
             "description": "Path to a CSV file to contain an field-level data dictionary."
         },
-        "release_prefix": {
-            "description": "The tag prefix used for releases.",
-            "required": True,
+        "include_dependencies": {
+            "description": "Process all of the GitHub dependencies of this project and include their schema in the data dictionary.",
```
davisagli, line 58: Really tempted to say schemata here but I guess it won't help comprehension...

cumulusci/tasks/datadictionary.py
```diff
@@ -28,44 +54,139 @@ class GenerateDataDictionary(BaseGithubTask):
...
         },
     }
 
+    def _init_task(self):
+        # Walking GitHub releases seems to be especially prone to API timeouts.
+        # Increase the defaults substantially to try to cope.
+        self.project_config.github_session = GitHubSession(
+            default_read_timeout=30, default_connect_timeout=30
+        )
+        super()._init_task()
```
davisagli, line 73: Why not always use this customized session? Would simplify things a bit.

cumulusci/core/config/project_config.py
```diff
@@ -485,6 +491,42 @@ def _check_keychain(self):
...
             )
 
+    def get_repo_from_url(self, url):
+        repo_owner, repo_name = url.split("/")[3:5]
+        if repo_name.endswith(".git"):
+            repo_name = repo_name[:-4]
+        gh = self.get_github_api(repo_owner, repo_name)
+        repo = gh.repository(repo_owner, repo_name)
+
+        return (repo, repo_owner, repo_name)
```
davisagli, line 501: Probably should get the owner and name from the repo object in case we were redirected from an old URL. Also can we just return repo and let code get the name and owner from it?

cumulusci/tasks/datadictionary.py
```diff
@@ -28,44 +54,139 @@ class GenerateDataDictionary(BaseGithubTask):
...
+                namespace,
+                self.project_config.project__git__prefix_release,
+            )
+        ]
+        if "additional_dependencies" in self.options:
+            repos += self._get_repo_dependencies(
+                [
+                    {"github": url}
+                    for url in process_list_arg(self.options["additional_dependencies"])
+                ],
```
davisagli, line 169: I think additional_dependencies should be provided in the same form as dependencies usually are (i.e. a list of dicts where `github` can be a key, rather than a list of github urls). It'll make it easier to support other VCS in the future.

cumulusci/tasks/datadictionary.py
```diff
@@ -28,44 +54,139 @@ class GenerateDataDictionary(BaseGithubTask):
...
         self._write_results()
 
     def _init_schema(self):
         """Initialize the structure used for schema storage."""
-        self.schema = defaultdict(
-            lambda: {"version": None, "fields": defaultdict(lambda: {"version": None})}
-        )
+        self.sobjects = defaultdict(lambda: [])
+        self.fields = defaultdict(lambda: [])
+        self.package_versions = defaultdict(lambda: [])
```
davisagli, line 184:
```suggestion
        self.sobjects = defaultdict(list)
        self.fields = defaultdict(list)
        self.package_versions = defaultdict(list)
```

cumulusci/tasks/datadictionary.py
```diff
@@ -189,102 +386,160 @@ def _process_field_element(self, sobject_name, field, version):
...
+            ]
+        )
 
-    def _write_results(self):
-        """Write the stored schema details to our destination CSVs"""
-        with open(self.options["object_path"], "w") as object_file:
-            writer = csv.writer(object_file)
+        for sobject_name, versions in self.sobjects.items():
+            versions = sorted(
+                versions, key=lambda ver: ver.version.version, reverse=True
```
davisagli, line 431: I got stuck here for a bit deciphering the different kinds of versions. Also, why reverse?

cumulusci/tasks/datadictionary.py
davidmreed: I love it. I think it's okay stylistically to use schema as an uncounted noun here but I will never pass up a chance at a third-decension plural.

cumulusci/tasks/datadictionary.py
davidmreed: That makes sense and removes some rather ugly code. Will make that change.

cumulusci/tasks/datadictionary.py
davidmreed: One day I will internalize the "you don't need a lambda here" voice, but not just yet. Fixed.

cumulusci/tasks/datadictionary.py
davidmreed: An older iteration of the algorithm for finding the right versions for each column required walking the version list in reverse-chronological order. That's no longer needed, so I'll remove it.

I'll add comments to clarify the nature of the data model here.

cumulusci/tasks/datadictionary.py
davidmreed: Sorry, I'm not sure I follow the alternative.

cumulusci/tasks/datadictionary.py
davidmreed: Scratch that, it's actually still required!

davidmreed:
I'm trying to figure out why a few lines are not obtaining coverage. A couple of them most certainly should be.

davidmreed:
Checking on one possible issue for NPSP, likely just data that is lost in the annals rather than a bug.

cumulusci/tasks/datadictionary.py
davisagli: @davidmreed I'm just saying, these customized timeouts feel like something that we could apply to all github3 clients constructed by cci, rather than just for this task.

davidmreed:
No bug, just NPSP's GitHub history doesn't reach before version 3.24

cumulusci/core/config/project_config.py
```diff
@@ -550,40 +590,17 @@ def process_github_dependency(  # noqa: C901
...
-            repo_name = repo_name[:-4]
-        gh = self.get_github_api(repo_owner, repo_name)
-        repo = gh.repository(repo_owner, repo_name)
+        repo = self.get_repo_from_url(dependency["github"])
         if repo is None:
             raise DependencyResolutionError(
                 f"{indent}Github repository {dependency['github']} not found or not authorized."
             )
 
+        repo_owner = repo.owner
```
davisagli, line 599: @davidmreed can you double check that repo.owner is a string? I'm having a little trouble telling from the github3 docs whether it's a string or an object that returns the username when stringified.

cumulusci/tasks/datadictionary.py
davidmreed: Oh, got it. I'm fine with that - I was aiming just to reduce the scope of intervention and testing but I suppose increasing the timeout is pretty unlikely to produce any regression behavior.

cumulusci/core/config/project_config.py
davidmreed: It is not a string; it's a `github3.users.ShortUser`. Fixed to explicitly convert.

cumulusci/tasks/datadictionary.py
davidmreed: Revised to use the longer timeout everywhere unless explicitly overridden.

PR #1768, opened 2020-05-29
Feature/robot dropdown keyword

boakley:
New modal page object keyword 'Select Dropdown Value'. Tested again dev and prerelease orgs, and
I also modified NPSP to use this keyword instead of its own and ran against the current release. All tests
passed with both chrome and firefox.

# Critical Changes
* added a new keyword in the modal page objects, 'Select dropdown value'. This keyword will
be available whenever you use the 'Wait for modal' keyword to pull in a modal page object.
# Changes

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/pageobjects/BasePageObjects.py
+++ b/cumulusci/robotframework/pageobjects/BasePageObjects.py
@@ -10,8 +10,8 @@ file.
 """
 
 import re
-from selenium.webdriver.common.action_chains import ActionChains
 from SeleniumLibrary.errors import ElementNotFound
+from selenium.common.exceptions import NoSuchElementException
 from cumulusci.robotframework.pageobjects import pageobject
 from cumulusci.robotframework.pageobjects import BasePage
 from cumulusci.robotframework.utils import capture_screenshot_on_error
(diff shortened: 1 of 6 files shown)


cumulusci/robotframework/pageobjects/BasePageObjects.py
```diff
@@ -136,6 +138,42 @@ def populate_form(self, *args, **kwargs):
...
+        if input_element is None:
+            raise Exception(f"No form element found with the label '{label}'")
+
+        # SeleniumLibrary's scroll_element_into_view doesn't seem to work
+        # for elements in a scrollable div on Firefox. Javascript seems to
+        # work though.
+        self.selenium.driver.execute_script(
+            "arguments[0].scrollIntoView()", input_element
+        )
+        trigger = input_element.find_element_by_class_name("uiPopupTrigger")
```
davisagli, line 160: Fwiw, I suspect these classes are associated with one of the deprecated aura components from the `ui` namespace (https://developer.salesforce.com/docs/atlas.en-us.lightning.meta/lightning/ui_input.htm) and will need to be updated in the future as forms are switched over to LWC implementations.

cumulusci/robotframework/pageobjects/BasePageObjects.py
```diff
@@ -136,6 +138,42 @@ def populate_form(self, *args, **kwargs):
...
+        trigger.click()
+
+        try:
+            popup_locator = (
+                "//div[contains(@class, 'uiPopupTarget')][contains(@class, 'visible')]"
+            )
+            popup = self.selenium.get_webelement(popup_locator)
+        except ElementNotFound:
+            raise ElementNotFound("Timed out waiting for the popup menu")
```
davisagli, line 169: Is it a dropdown or a popup?

cumulusci/robotframework/tests/salesforce/ui.robot
```diff
@@ -49,6 +49,25 @@ Create test data
...
+    FOR  ${i}  IN RANGE  3
+        &{obj} =          Salesforce Get  ${obj name}  ${obj id}
+        ${actual_value}=  Set variable  ${obj}[${field}]
+
+        Return from keyword if  $expected_value == $actual_value
+        log  Retrying API call...  WARN
+        Sleep  1 second
+    END
+    Fail  Expected ${obj name} field ${field} to be '${expected_value}' but it was '${actual_value}'
+
```
davisagli, line 70: I'm surprised you didn't implement this one in Python.

cumulusci/robotframework/pageobjects/BasePageObjects.py
boakley: I used "popup" in this context since I'm waiting for an element with the class `uiPopupTarget` as the result of clicking on a "uiPopupTrigger".

boakley: Though, if you're referring to the error message, yeah, that should probably be "dropdown".  I'll fix it.

cumulusci/robotframework/tests/salesforce/ui.robot
boakley: That might have been smarter. Though since at the moment the keyword is unique to this specific test case, it was simpler to do it here. "simpler" in that if it were in a library I'd have to write a test case for it. I think for now I'll leave it as is. 

It might be useful to make this (or something like it) an official keyword, but I'll save that for a separate work item. I'll ping the other robot users to see if they would find such a keyword useful.

davisagli: That makes sense that this is more of a one-off.

davisagli:
@boakley I want to approve this, but the UI tests failed on Chrome. Does it look related to the change or should we just try re-running it?

boakley:
The failure is unrelated to this change. It failed while trying to edit an opportunity that was created during the test setup. The creation of the opportunity via the API seems to have succeeded, but it didn't show up in the list of opportunities.

PR #1771, opened 2020-06-01
Centralize handling of the ".cumulusci" directory -- mostly.

prescod:
We had this same logic in many places and it will probably continue to proliferate:

```os.path.join(
        os.path.expanduser("~"), BaseGlobalConfig.config_local_dir
    )
```

Now we have a central place for it.


cumulusci/core/utils.py
```diff
@@ -178,3 +179,12 @@ def dictmerge(a, b, name=None):
             config_name=name,
         )
     return a
+
+
+def cumulusci_config_dir():
+    config_dir = Path(Path.home(), ".cumulusci")  # breaks unit tests
```
davisagli, line 185: It looks like you've addressed the comment here?

prescod:
This directory is used in check_latest_version which runs before there is a RUNTIME.

So in a sense, centralization and polymorphism are at odds with each other.

I guess I will leave it in a util and add a a polymorphic method which can be used in the contexts where polymorphism makes sense?

prescod:
Or I can leave it in a util and someone can add polymorphism when they need it.

davisagli:
Possibly the logic for checking the version should move into CliRuntime; it only really makes sense in the context of the CLI. But anyway, you convinced me that we can go ahead with this as a first step and take it further later.

prescod:
Okay I saw another way to do it which may balance the various goals well enough and keep the code reasonably clear.

cumulusci/core/keychain/EncryptedFileProjectKeychain.py
```diff
@@ -15,7 +16,7 @@ def config_local_dir(self):
         except AttributeError:
             # Handle a global config passed as project config
             config_local_dir = self.project_config.config_local_dir
-        return os.path.join(os.path.expanduser("~"), config_local_dir)
+        return str(Path(Path.home(), config_local_dir))
```
davisagli, line 19: I think we want to stop using BaseGlobalConfig.config_local_dir altogether; it's not very useful.

This property should become:

    try:
        config_local_dir = self.project_config.global_config_obj.cumulusci_config_dir
    except AttributeError:
        # Handle a global config passed as a project config
        config_local_dir = self.project_config.cumulusci_config_dir
    return config_local_dir

or else add a cumulusci_config_dir property to BaseProjectConfig as well (delegating to its global_config_obj) so that we don't need to do the try/except here.

prescod: I was afraid to remove the os.path.join because I wasn't sure whether it would sometimes be invoked like this:

```
>>> os.path.join("/foo/bar", "/a/b")
'/a/b'
```

cumulusci/core/config/BaseGlobalConfig.py
```diff
@@ -21,9 +22,24 @@ def __init__(self, config=None):
...
         self._load_config()
 
+    @property
+    def cumulusci_config_dir(self):
+        "The .cumulus directory for storing persistent data"
+        return self.default_cumulusci_dir()
+
+    @staticmethod
+    def default_cumulusci_dir():
+        "The default .cumulus directory for storing persistent data"
```
davisagli, line 32:
```suggestion
        """Get the root directory for storing persistent data (~/.cumulusci)
        
        Creates it if it doesn't exist yet.
        """
```

cumulusci/core/config/BaseGlobalConfig.py
```diff
@@ -21,9 +22,24 @@ def __init__(self, config=None):
         self._init_logger()
         self._load_config()
 
+    @property
+    def cumulusci_config_dir(self):
+        "The .cumulus directory for storing persistent data"
```
davisagli, line 27:
```suggestion
        """Get the root directory for storing persistent data, as an instance property"""
```

cumulusci/core/config/BaseGlobalConfig.py
```diff
@@ -21,9 +22,24 @@ def __init__(self, config=None):
...
+        return self.default_cumulusci_dir()
+
+    @staticmethod
+    def default_cumulusci_dir():
+        "The default .cumulus directory for storing persistent data"
+        config_dir = Path(Path.home(), ".cumulusci")
```
davisagli, line 33:
```suggestion
        config_dir = Path.home() / ".cumulusci"
```

cumulusci/core/keychain/EncryptedFileProjectKeychain.py
davisagli: @prescod I'm not following. The point of this property is to get the cumulusci config dir from the GlobalConfig object.

prescod: I honestly have no idea what the two branches of this function are intended to do, so I'm basically reluctant to touch it.

Why would `self.project_config.global_config_obj.config_local_dir` be sometimes missing and why would it be available at `self.project_config.config_local_dir`. If its supposed to be an override passed as a "project config" (as opposed to a global config) then why not look in the project first?

Everything about how these config objects interact makes me nervous, so I am reluctant to change them more than necessary.

davisagli: @prescod `BaseProjectKeychain.__init__` takes an arg called `project_config`, but in reality it is sometimes passed an instance of BaseProjectConfig (when running inside a project directory) and sometimes an instance of BaseGlobalConfig (when not inside a project directory, because keychain entries like orgs and services can also be configured globally). So the line inside the `try` is for when we have a project config, and the line in the except is for when we actually have a global config.

davisagli: @prescod let me know if that helped. If not, maybe we should discuss in real time.

PR #1772, opened 2020-06-01
Feature/smart get lookup key field

prescod:
Behaviour change:

 * lookups do not anymore default to using snake_case, although they will look for snake_case in a dataset.

 * a few code paths which previously did not use our official mapping parser, now do. Which means that they will give error messages for misplaced, misnamed or missing keys

Internal renamings/refactorings:

 * some classes used "self.mappings" and others "self.mapping". Now they all use "self.mapping".

 * get_lookup_key_field is now a method which takes 0 arguments, not a function which takes 2. The old function remains temporarily for backwards compatibility. (primarily with snowfakery)

### DIFF
--- a/cumulusci/tasks/bulkdata/mapping_parser.py
+++ b/cumulusci/tasks/bulkdata/mapping_parser.py
@@ -2,9 +2,11 @@ from typing import Dict, List, Union, IO, Optional
 from logging import getLogger
 from pathlib import Path
 
-from pydantic import Field, validator, ValidationError
+from pydantic import Field, validator, root_validator, ValidationError
 
 from cumulusci.utils.yaml.model_parser import CCIDictModel
+from cumulusci.utils import convert_to_snake_case
+
 from typing_extensions import Literal
 
 LOGGER_NAME = "MAPPING_LOADER"
@@ -19,6 +21,28 @@ class MappingLookup(CCIDictModel):
     join_field: Optional[str] = None
     after: Optional[str] = None
     aliased_table: Optional[str] = None
+    name: Optional[str] = None  # populated by parent
+
+    def get_lookup_key_field(self, model=None):
+        guesses = []
+        if self.get("key_field"):
+            guesses.append(self.get("key_field"))
+
+        guesses.append(self.name)
+
+        if not model:
+            return guesses[0]
+
+        # CCI used snake_case until mid-2020.
+        # At some point this code could probably be simplified.
+        snake_cased_guesses = list(map(convert_to_snake_case, guesses))
+        for guess in guesses + snake_cased_guesses:
+            if hasattr(model, guess):
+                return guess
+        raise KeyError(
+            f"Could not find a key field for {self.name}.\n"
+            + f"Tried {', '.join(snake_cased_guesses)}"
+        )
 
 
 class MappingStep(CCIDictModel):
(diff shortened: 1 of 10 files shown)


cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -19,6 +21,28 @@ class MappingLookup(CCIDictModel):
...
+
+        # CCI used snake_case until mid-2020.
+        # At some point this code could probably be simplified.
+        snake_cased_guesses = list(map(convert_to_snake_case, guesses))
+        for guess in guesses + snake_cased_guesses:
+            if hasattr(model, guess):
+                return guess
+        raise KeyError(
+            f"Could not find a key field for {self.name}.\n"
+            + f"Tried {', '.join(snake_cased_guesses)}"
```
davisagli, line 44:
```suggestion
            + f"Tried {', '.join(guesses)}"
```

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -19,6 +21,28 @@ class MappingLookup(CCIDictModel):
     join_field: Optional[str] = None
     after: Optional[str] = None
     aliased_table: Optional[str] = None
+    name: Optional[str] = None  # populated by parent
+
+    def get_lookup_key_field(self, model=None):
```
davisagli, line 26: docstring please

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -50,6 +75,12 @@ def oid_as_pk_is_deprecated(cls, v):
         )
         return v
 
+    @root_validator  # not really a validator, more like a post-processor
+    def fixup_lookup_names(cls, v):
```
davisagli, line 79: docstring please

cumulusci/tasks/bulkdata/utils.py
```diff
@@ -9,11 +10,16 @@
...
 from cumulusci.core.exceptions import BulkDataException
-from cumulusci.utils import convert_to_snake_case
 
 
-def get_lookup_key_field(lookup, sf_field):
-    return lookup.get("key_field") or convert_to_snake_case(sf_field)
+def get_lookup_key_field(lookup, model=None):
+    # This should be removed in late 2020.
+    # It just provides some temporary backwards-compatibility
+    # for Snowfakery
```
davisagli, line 18: What do you mean about backwards-compatibility for Snowfakery? This function is not used in Snowfakery.

prescod: I forgot that it is only called indirectly. You think it's a safe bet that nobody is using it externally?

cumulusci/tasks/bulkdata/mapping_parser.py
prescod: 3441a8c8

cumulusci/tasks/bulkdata/mapping_parser.py
prescod: 3441a8c8

cumulusci/tasks/bulkdata/mapping_parser.py
prescod: 3441a8c8

cumulusci/tasks/bulkdata/utils.py
davisagli: I don't see any mention of it in our products and think it's unlikely to be used elsewhere.

davisagli: I removed it: https://github.com/SFDO-Tooling/CumulusCI/pull/1772/commits/766f4612dcb6f99c27ce1f7a9523a4fbee5ac162

PR #1787, opened 2020-06-08
Data manipulation fixes

prescod:
Fix some bugs discovered in integration testing of Data manipulation features.

### DIFF
--- a/cumulusci/tasks/bulkdata/extract.py
+++ b/cumulusci/tasks/bulkdata/extract.py
@@ -109,7 +109,7 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
         fields = self._fields_for_mapping(mapping)
         soql = f"SELECT {', '.join(fields)} FROM {sf_object}"
         if mapping.record_type:
-            soql += f" WHERE RecordType.DeveloperName = '{mapping['record_type']}'"
+            soql += f" WHERE RecordType.DeveloperName = '{mapping.record_type}'"
 
         return soql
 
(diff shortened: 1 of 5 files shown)


cumulusci/tasks/bulkdata/extract.py
```diff
@@ -108,7 +108,7 @@ def _soql_for_mapping(self, mapping):
         sf_object = mapping["sf_object"]
         fields = self._fields_for_mapping(mapping)
         soql = f"SELECT {', '.join(fields)} FROM {sf_object}"
-        if "record_type" in mapping:
+        if mapping.record_type:
             soql += f" WHERE RecordType.DeveloperName = '{mapping['record_type']}'"
```
davisagli, line 112: Need to use mapping.record_type inside the f-string too, no?

cumulusci/tasks/bulkdata/tests/test_extract.py
```diff
@@ -524,19 +524,23 @@ def test_soql_for_mapping(self):
         task = _make_task(
             ExtractData, {"options": {"database_url": "sqlite:///", "mapping": ""}}
         )
-        mapping = {
-            "sf_object": "Contact",
-            "oid_as_pk": True,
-            "fields": {"Id": "sf_id", "Test__c": "Test"},
-        }
+        mapping = MappingStep(
+            **{
```
davisagli, line 528: This is a strange way of constructing the keyword arguments. I would prioritize readability and make the switch to passing these as individual keyword args.

cumulusci/tasks/bulkdata/extract.py
prescod: The MappingStep object is largely backwards compatible with the dict but I didn't implement __contains__ behaviour. Nevertheless, code consistency is a good thing so I'll change that.

PR #1794, opened 2020-06-10
Refresh oauth token more reliably

prescod:
Refresh oauth token more reliably in the face of Salesforce timeouts.

# Changes

Previously, after 10 minutes of inactivity, Salesforce would close an HTTP connection and  oauth token refresh would fail with a connection error. Now each token refresh should generate a new HTTP connection.

### DIFF
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -29,6 +29,7 @@ class OrgConfig(BaseConfig):
         super(OrgConfig, self).__init__(config)
 
     def refresh_oauth_token(self, keychain, connected_app=None):
+        self.refresh_client()
         if not SKIP_REFRESH:
             SFDX_CLIENT_ID = os.environ.get("SFDX_CLIENT_ID")
             SFDX_HUB_KEY = os.environ.get("SFDX_HUB_KEY")
@@ -83,6 +84,9 @@ class OrgConfig(BaseConfig):
             )
 
         return self._client
+
+    def refresh_client(self):
+        self._client = None
 
     @property
     def latest_api_version(self):


davisagli:
@banderson5144 MetaCI requires setting both SFDX_CLIENT_ID and SFDX_HUB_KEY, which does indeed mean that CumulusCI will use the JWT auth flow when running in the context of MetaCI. What is your reason for not wanting to use it?

banderson5144:
> @banderson5144 MetaCI requires setting both SFDX_CLIENT_ID and SFDX_HUB_KEY, which does indeed mean that CumulusCI will use the JWT auth flow when running in the context of MetaCI. What is your reason for not wanting to use it?

I want it to use the org configuration (permanent QA sandbox / prod)  that I have defined with cci org info --json which contains a good refresh token. I guess my question is, why is there an If condition check on those variables it is always going to run a JWT flow

davisagli:
@banderson5144 CumulusCI supports authenticating with a refresh token, but MetaCI does not. We found that using refresh tokens made it cumbersome to replace the connected app if multiple orgs were connected. The supported way to connect an org to MetaCI is:
1. Create a connected app representing your instance of MetaCI. Make sure that it has a certificate for digital signatures (which is needed for the JWT flow). Make sure that it includes the web, offline_access and full OAuth scopes.
2. Set SFDX_CLIENT_ID to the connected app's consumer key, and SFDX_HUB_KEY to the private key that was used to sign the certificate.
3. Approve use of this connected app by the persistent org user. Option 1: First connect cci to the same connected app (using "cci service connect connected_app" with the connected app's consumer key and secret) and then connect the org and get its json using "cci org info" like you mentioned. Option 2: Give the persistent org user a profile or permission set which includes pre-approved permission for the connected app.
4. Create the Org record in MetaCI. In the JSON for the org config, you only need to include the username and instance_url.

In the future we'd like to add a streamlined way to initiate the login flow to connect an org from within MetaCI, but that will still likely continue to use the JWT flow for subsequent authentication.

banderson5144:
I guess I'm not understanding the code at line 56 then, could you please explain https://www.github.com/SFDO-Tooling/CumulusCI/tree/master/cumulusci%2Fcore%2Fconfig%2FOrgConfig.py

davisagli:
@banderson5144 It shouldn't be possible to run MetaCI without them being defined. These lines in MetaCI's production settings file should throw an error on startup if they are not found in the environment: https://github.com/SFDO-Tooling/MetaCI/blob/master/config/settings/production.py#L293

banderson5144:
> @banderson5144 It shouldn't be possible to run MetaCI without them being defined. These lines in MetaCI's production settings file should throw an error on startup if they are not found in the environment: https://github.com/SFDO-Tooling/MetaCI/blob/master/config/settings/production.py#L293

Fair enough I am seeing that issue, so are you saying I should be able to use the same KEY / Client ID across multiple orgs correct?

davisagli:
Yes, users in different orgs can approve access by the same connected app.

banderson5144:
Also my SFDX_HUB_USERNAME shouldn't matter right

davisagli:
MetaCI uses SFDX_HUB_USERNAME to authenticate sfdx to a Dev Hub org (using the same connected app) in order to create scratch orgs. If you aren't using scratch orgs, you can set it to a bogus value.  (If you've included https://github.com/SalesforceFoundation/simple-salesforce-dx-buildpack.git in the app's buildpacks, which is also necessary to create scratch orgs, then it will probably log an error when the app starts about not being able to authenticate to the dev hub if SFDX_HUB_USERNAME is a bogus value -- but this shouldn't prevent startup.)

banderson5144:
Excellent, thank you very much @davisagli

banderson5144:
@davisagli I am seeing an issue when I connect a Sandbox org to MetaCI. I am getting an Audience Invalid error when the JWT authentication tries to use the My Domain from the `instance_url`. If I switch the `instance_url` to `https://test.salesforce.com` it works. However, it seems at the end of my flow, it updates it back to the My Domain instance. This causes me to fix it every time. Is this intended functionality or is this a bug?

PR #1800, opened 2020-06-12
[CCI] MetadataETL task to replace/insert help text for a custom field

jofsky:
# Critical Changes

# Changes

- The Metadata ETL task `set_field_help_text` allows setting Help Text values on existing fields.

# Issues Closed


cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,86 @@
...
+from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.tasks.metadata_etl import MetadataSingleEntityTransformTask
+from cumulusci.utils.xml.metadata_tree import MetadataElement
+from cumulusci.core.utils import process_list_arg
+
+
+class AddHelpText(MetadataSingleEntityTransformTask):
+    entity = "CustomObject"
+    task_options = {
+        "entries": {
```
davidmreed, line 12: I'm not sure about "entries" here. It makes sense in a picklist context, but it may not be very clear here. How about `"fields"`?

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,86 @@
...
+            raise TaskOptionsError(
+                "The 'entries' option is required, please pass a dictionary with the object_field and help_text keys."
+            )
+        if type(self.options["entries"]) != list or len(self.options["entries"]) == 0:
+            raise TaskOptionsError(
+                "Please populate the entries field with a list of dictionaries containing at minimum one entry with an 'object_field' and 'help_text' keys"
+            )
+        if not all(["object_field" in entry for entry in self.options["entries"]]):
+            raise TaskOptionsError(
+                "The 'object_field' key is required on all entry values."
```
davidmreed, line 39: Can I suggest `"api_name"` rather than `"object_field"`?

cumulusci/tasks/metadata_etl/tests/test_help_text.py
```diff
@@ -0,0 +1,405 @@
...
+        <description>Indicates that the Contact is allowed to receive information protected by FERPA and other privacy laws, regulations, and policies.</description>
+        <externalId>false</externalId>
+        <inlineHelpText>Indicates that the Contact is allowed to receive information protected by FERPA and other privacy laws, regulations, and policies.</inlineHelpText>
+        <label>FERPA Approved</label>
+        <trackHistory>false</trackHistory>
+        <trackTrending>false</trackTrending>
+        <type>Checkbox</type>
+    </fields>
+    <fields>
+        <fullName>Bluth__c</fullName>
```
davidmreed, line 23: Now I am a huge Arrested Development fan, but in the context of corporate OSS, I'm not sure we should theme our unit tests. Can we use neutral identifiers like `Foo__c`?

cumulusci/tasks/metadata_etl/tests/test_help_text.py
```diff
@@ -0,0 +1,405 @@
...
+        test_elem = result.find("fields", fullName="Buster__c")
+
+        assert test_elem is not None
+        assert test_elem.inlineHelpText.text == "buster"
+        # Validate that the sObject alters only the custom field listed
+        test_elem = result.find("fields", fullName="Bluth__c")
+        assert test_elem is not None
+        assert (
+            test_elem.inlineHelpText.text
+            == "Indicates that the Contact is allowed to receive information protected by HIPPA and other privacy laws, regulations, and policies."
```
davidmreed, line 167: You might want to trim down the text values in your unit tests to help make them more readable/accessible.

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,86 @@
...
+        # Locate the <fields> entry for this field entry.
+        field = metadata.find("fields", fullName=custom_field)
+        if not field:
+            raise TaskOptionsError(
+                f"The field {api_name}.{custom_field} was not found."
+            )
+        try:
+            field.inlineHelpText.text = help_text
+        except AttributeError:
+            print("there")
```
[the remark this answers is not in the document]
jofsky: ah good catch.

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,86 @@
...
+    ):
+        # Locate the <fields> entry for this field entry.
+        field = metadata.find("fields", fullName=custom_field)
+        if not field:
+            raise TaskOptionsError(
+                f"The field {api_name}.{custom_field} was not found."
+            )
+        try:
+            field.inlineHelpText.text = help_text
+        except AttributeError:
```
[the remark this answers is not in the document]
jofsky: Yeah set this because was experiencing the situation where there was no help text whatsoever. Was slightly concerned about setting it on an object that isn't actually supposed to have help text. With that said my knowledge of which objects should actually have that field is not totally there.

cumulusci/tasks/metadata_etl/tests/test_help_text.py
jofsky: I was also wondering if there were TM concerns.

cumulusci/tasks/metadata_etl/help_text.py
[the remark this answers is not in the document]
davidmreed: Any custom field can have it set. I think this is fine.

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,86 @@
...
+            "required": True,
+        },
+        **MetadataSingleEntityTransformTask.task_options,
+    }
+
+    def _init_options(self, kwargs):
+        self.task_config.options["api_names"] = "dummy"
+        super()._init_options(kwargs)
+
+        try:
```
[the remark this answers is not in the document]
jofsky: i took out the version minimum but still check to make sure a float is passed in for the api_version was unsure if this was useless.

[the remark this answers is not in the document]
davisagli: I think validating that the API version is the correct format is a good idea, but it belongs in MetadataSingleEntityTransformTask._init_options rather than here, so that the validation happens for all tasks that are based on MetadataSingleEntityTransformTask

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,80 @@
...
+            float(self.api_version)
+        except ValueError:
+            raise TaskOptionsError(f"Invalid API version {self.api_version}")
+
+        if "fields" not in self.options:
+            raise TaskOptionsError(
+                "The 'fields' option is required, please pass a dictionary with the api_name and help_text keys."
+            )
```
davisagli, line 31: BaseTask takes care of validating that options with `"required": True` are provided, so I don't think this check is needed.

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,80 @@
...
+            except ValueError:
+                raise TaskOptionsError(
+                    f"api_name {entry} is not a valid Object.Field reference"
+                )
+
+        self.api_names = set(self.api_name_list.keys())
+
+    def _transform_entity(self, metadata: MetadataElement, api_name: str):
+        for entry in self.api_name_list[api_name]:
+            self._modify_help_text(metadata, api_name, entry[0], entry[1])
```
davisagli, line 61: I would write it like this to make it clearer what the parts of `entry` are:

```suggestion
        for field, help_text in self.api_name_list[api_name]:
            self._modify_help_text(metadata, api_name, field, help_text)
```

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,80 @@
...
+        help_text: str,
+    ):
+        # Locate the <fields> entry for this field entry.
+        field = metadata.find("fields", fullName=custom_field)
+        if not field:
+            raise TaskOptionsError(
+                f"The field {api_name}.{custom_field} was not found."
+            )
+        try:
+            field.inlineHelpText.text = help_text
```
davisagli, line 78: I'm concerned that this could overwrite help text customizations that the admin has made in the org.

cumulusci/cumulusci.yml
```diff
@@ -6,6 +6,12 @@ tasks:
         group: Metadata Transformations
         description: Activates Flows identified by a given list of Developer Names
         class_path: cumulusci.tasks.salesforce.activate_flow.ActivateFlow
+    add_field_help_text:
```
davisagli, line 9: Should we rename this to `set_field_help_text` to match the Python class name?

cumulusci/tasks/metadata_etl/help_text.py
davidmreed: This is a good point. I was envisioning the use case of this task as being in an explicit opt-in installer, used by admins to update their orgs to the latest shipped help text - i.e., that possibility would be disclosed to them and they'd accept before using it.

I wonder if there's a safer way to do that, though, that still provides utility. One option could be to have the task accept the last-shipped Help Text and check if it is still set before updating it. But there may be situations in NPSP where the help text has changed multiple times, and the end-user org may not have the text we expect even if it has not been modified by the end-user admin.

davisagli: Yeah, I'd be less concerned with it in that context. Maybe the task should have an option for whether it will overwrite existing help text or not?

davidmreed: I like that idea. @jofsky How about it - overwrite existing text only if an option is set to True, and default the option to False?

davidmreed: That will force authors of flows and installers to be explicit about their intentions.

cumulusci/cumulusci.yml
jofsky: sure

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,88 @@
...
+        field = metadata.find("fields", fullName=custom_field)
+        if not field:
+            raise TaskOptionsError(
+                f"The field {api_name}.{custom_field} was not found."
+            )
+        try:
+            if self.options["overwrite"] is True:
+                field.inlineHelpText.text = help_text
+            else:
+                raise TaskOptionsError(
```
davidmreed, line 84: I think the use case of this task dictates that this should be a `logger.warn()` rather than an exception.

We're potentially going to be running this on customer orgs, and we want the default mode of operation to be safe. In that mode, we'd set the help text if and only if it is blank, and otherwise do nothing.

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,87 @@
...
+
+class SetFieldHelpText(MetadataSingleEntityTransformTask):
+    entity = "CustomObject"
+    task_options = {
+        "fields": {
+            "description": "List of object fields to affect, in Object__c.Field__c form.",
+            "required": True,
+        },
+        "overwrite": {
+            "description": "List of object fields to affect, in Object__c.Field__c form."
```
davisagli, line 16: @jofsky can you update this option's description please?

Also run `make docs` to regenerate docs/tasks.rst, and make sure that the section for this task is accurate and complete.

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,87 @@
...
+        help_text: str,
+    ):
+        # Locate the <fields> entry for this field entry.
+        field = metadata.find("fields", fullName=custom_field)
+        if not field:
+            raise TaskOptionsError(
+                f"The field {api_name}.{custom_field} was not found."
+            )
+        try:
+            if self.options["overwrite"] is True:
```
davisagli, line 80:
```suggestion
            if self.options["overwrite"]:
```

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,87 @@
...
+        self.options["overwrite"] = process_bool_arg(
+            self.options.get("overwrite", False)
+        )
+
+        try:
+            float(self.api_version)
+        except ValueError:
+            raise TaskOptionsError(f"Invalid API version {self.api_version}")
+
+        if type(self.options["fields"]) != list or len(self.options["fields"]) == 0:
```
davisagli, line 34:
```suggestion
        if not isinstance(self.options.get("fields"), list) or len(self.options["fields"]) == 0:
```

Changed to use isinstance which is more idiomatic in Python (it also matches subclasses, though that doesn't matter here) and to use .get() in case the user forgot to provide the fields option (validation for "required" options doesn't take place until after _init_options)

cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -0,0 +1,87 @@
...
+            raise TaskOptionsError(
+                f"The field {api_name}.{custom_field} was not found."
+            )
+        try:
+            if self.options["overwrite"] is True:
+                field.inlineHelpText.text = help_text
+            else:
+                self.logger.warning(
+                    f"Skipping over help text field: {field.inlineHelpText.text}. Please set the overwrite option to True to overwrite this help text field."
+                )
```
davisagli, line 85: Should we allow setting the value without a warning if the existing value is blank, or happens to already have the same value that would be set?

davidmreed: I think that's a good suggestion (@jofsky) for situations like a hypothetical NPSP RD2 installer.

davidmreed: Added in 52bd579c

cumulusci/tasks/metadata_etl/help_text.py
davidmreed: Docs revised in 52bd579c

davidmreed:
Pushed a commit that I believe resolves outstanding issues with this PR.

PR #1808, opened 2020-06-14
Better error messaging for dependency management

prescod:
Improve error messaging and add assertions to dependency management features.

### DIFF
--- a/cumulusci/salesforce_api/metadata.py
+++ b/cumulusci/salesforce_api/metadata.py
@@ -98,8 +98,8 @@ class BaseMetadataApiCall(object):
             return self.soap_envelope_result.format(process_id=self.process_id)
 
     def _build_envelope_start(self):
-        if self.soap_envelope_start:
-            return self.soap_envelope_start.format(api_version=self.api_version)
+        assert self.soap_envelope_start, "soap_envelope_start should be supplied"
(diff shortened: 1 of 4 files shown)


cumulusci/salesforce_api/metadata.py
```diff
@@ -426,6 +426,10 @@ def _build_envelope_start(self):
                 run_tests=run_tests,
                 api_version=self.api_version,
             )
+        else:
+            # if there is a valid use-case for this branch, please
+            # add a unit test
+            assert self.package_zip, "package_zip should be specified"
```
davisagli, line 432: I don't think there is a valid use case. I think it would make more sense to assert that package_zip is not empty in ApiDeploy's constructor.

cumulusci/tasks/salesforce/UpdateDependencies.py
```diff
@@ -317,7 +317,8 @@ def _install_dependency(self, dependency):
                     dependency["version"],
                     securityType=self.options["security_type"],
                 )()
-
+        if not package_zip:
+            raise TaskOptionsError(f"Could not find package for {dependency}")
```
davisagli, line 321: Please make sure this line has test coverage.

PR #1809, opened 2020-06-14
Add success message for flow

prescod:
Add a success message for flow completion and slight code cleanup.

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -350,7 +350,8 @@ class FlowCoordinator(object):
         try:
             for step in self.steps:
                 self._run_step(step)
-            self.logger.info(f"Completed flow '{self.name}' successfully! 🎉")
+            flow_name = f"'{self.name}' " if self.name else ""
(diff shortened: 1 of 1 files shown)


cumulusci/core/flowrunner.py
```diff
@@ -349,42 +349,46 @@ def run(self, org_config):
...
-                self.callbacks.post_task(step, result)
-
-                self.results.append(
-                    result
-                )  # add even a failed result to the result set for the post flow
-
-                if result.exception and not step.allow_failure:
-                    raise result.exception  # PY3: raise an exception type we control *from* this exception instead?
+                self._run_step(step)
+            self.logger.info(f"Completed flow '{self.name}' successfully! 🎉")
```
davisagli, line 353: - I would avoid saying "successfully" -- we know the flow completed without an exception, but we don't know if it did what the user intended.
- self.name may be None
- I would avoid the emoji unless you've tested to make sure it works okay in Windows and when the logs are stored in Django models and rendered in a web browser.

PR #1810, opened 2020-06-14
Test data dictionary in tempdir

prescod:
Avoid generating files in repo during pytest.

### DIFF
--- a/cumulusci/tasks/tests/test_datadictionary.py
+++ b/cumulusci/tasks/tests/test_datadictionary.py
@@ -1,6 +1,5 @@
 import io
 import unittest
-
 from unittest.mock import Mock, call, patch, mock_open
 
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/tests/test_datadictionary.py
```diff
@@ -910,7 +912,9 @@ def test_run_task__additional_dependencies(self):
         task._get_repo_dependencies = Mock(return_value=[1, 2])
         task._walk_releases = Mock()
 
-        task._run_task()
+        with TemporaryDirectory() as t:
+            with cd(t):
+                task._run_task()
```
davisagli, line 917: I'd suggest using cumulusci.utils.temporary_dir which creates a dir _and_ cds to it.

PR #1812, opened 2020-06-15
Improve sandboxing in test classes that make pervasive use of tempdirs.

prescod:
Fix a bug where $HOME was leaking beyond a single test case.

### DIFF
--- a/cumulusci/core/tests/test_config_expensive.py
+++ b/cumulusci/core/tests/test_config_expensive.py
@@ -9,7 +9,7 @@ import shutil
 
 import yaml
 
-from cumulusci.utils import temporary_dir
+from cumulusci.utils import temporary_dir, cd
 from cumulusci.core.config import ScratchOrgConfig
 from cumulusci.core.config import BaseGlobalConfig
 from cumulusci.core.config import BaseProjectConfig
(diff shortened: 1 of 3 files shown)


cumulusci/core/tests/test_config_expensive.py
```diff
@@ -23,8 +24,13 @@
...
 class TestBaseGlobalConfig(unittest.TestCase):
-    def setUp(self):
-        self.tempdir_home = tempfile.mkdtemp()
+    @classmethod
+    def setUp(cls):
+        cls.tempdir_home = tempfile.mkdtemp()
+
+    @classmethod
+    def tearDownClass(cls):
+        shutil.rmtree(cls.tempdir_home)
```
davisagli, line 33: This doesn't look right; it is creating a new temp dir before every test case but only removing one of them (did you mean to use setUpClass instead of setUp?)

davisagli:
@prescod still confused about what you're trying to do here since you are now using both pytest and unittest.TestCase hooks.

prescod:
There are two classes in that file and I noticed that the second one also does not clean up its tempfiles. It was written in UnitTest style so I followed it. Now I've changed it to be the same as the other class.

PR #1815, opened 2020-06-16
Add task to create unlocked package versions

davisagli:
# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/package_2gp.py
```diff
@@ -0,0 +1,537 @@
...
+            zip_src, options=dependency, logger=self.logger
+        )
+
+        package_config = PackageConfig(
+            name="{repo_owner}/{repo_name} {subfolder}".format(**dependency),
+            version_name="Auto",
+            package_type="Unlocked",
+            # Ideally we'd do this without a namespace,
+            # but it needs to match the dependent package
+            namespace=self.package_config.namespace,
```
davidmreed, line 468: This _shouldn't_ break package builds (although it's possible if someone did something a bit foolish and created dependencies between unpackaged/pre metadata in different packages), but I think there's a good chance that we'll have some broken setup automation. For example, packages downstream of NPSP that do further customization may depend on the unpackaged Record Types not being namespaced. `load_dataset` is _probably_ resilient to that because of how it resolves record type names, but other automation may not be.

What's the source of the need for the namespaces to match?

cumulusci/salesforce_api/package_zip.py
```diff
@@ -13,61 +29,269 @@
...
+        # otherwise we hit a bug in Windows where ZipInfo objects have the wrong path separators.
+        fp = self.zf.fp
+        self.zf.close()
+        self.zf = zipfile.ZipFile(fp, "r")
+
+        self._process_namespace_tokens()
+        self._clean_meta_xml()
+        self._bundle_staticresources()
+
+    def _process_namespace_tokens(self):
```
davidmreed, line 177: Took me a sec to discover that this is preserving behavior from `update_dependencies`. I wonder if anyone is using this feature; I don't think I really understand the use case for tokenizing during this operation.

cumulusci/tasks/tests/test_package_2gp.py
```diff
@@ -0,0 +1,493 @@
...
+            },
+        )
+        responses.add(  # get dependencies from SubscriberPackageVersion (main package)
+            "GET",
+            f"{self.base_url}/tooling/query/",
+            json={"size": 1, "records": [{"Dependencies": ""}]},
+        )
+
+        task.options["dependency_org"] = "2gp_dependencies"
+        task()
```
davidmreed, line 288: Is there a meaningful assert to be added here?

cumulusci/salesforce_api/package_zip.py
```diff
@@ -13,61 +29,269 @@
...
+        zipf = self.zf
+        if self.options.get("namespace_tokenize"):
+            self.logger.info(
+                f"Tokenizing namespace prefix {self.options['namespace_tokenize']}__"
+            )
+            zipf = process_text_in_zipfile(
```
davidmreed, line 183: This is really elegant.

cumulusci/tasks/package_2gp.py
```diff
@@ -0,0 +1,537 @@
...
+            # Add the package.zip
+            package_hash = package_zip_builder.as_hash()
+            version_info.writestr("package.zip", package_zip_builder.as_bytes())
+
+            # Check for an existing package with the same contents
+            res = self.tooling.query(
+                "SELECT Id "
+                "FROM Package2VersionCreateRequest "
+                f"WHERE Package2Id = '{package_id}' "
+                "AND Status != 'Error' "
```
jlantz, line 205: We should add a where clause for SkipValidation here since if you want to promote a package that had been previously uploaded with SkipValidation you'd need to re-upload the package with SkipValidation set to false

davisagli: And writing this down because we talked about it but I haven't done yet: re-add the option to force re-uploading.

cumulusci/salesforce_api/package_zip.py
```diff
@@ -13,61 +29,269 @@
...
+        """
+        for root, dirs, files in os.walk(path):
+            root_parts = pathlib.Path(root).relative_to(path).parts
+            if self._include_directory(root_parts):
+                for f in files:
+                    if self._include_file(root_parts, f):
+                        yield pathlib.Path(root, f)
+
+    def _include_directory(self, root_parts):
+        """Return boolean for whether this directory should be included in the package."""
```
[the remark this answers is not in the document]
davisagli: The details of this logic aren't really the focus of this PR -- it is moving over from the Deploy task unchanged.

cumulusci/salesforce_api/package_zip.py
davisagli: It's centralizing logic that was previously done separately by both Deploy and UpdateDependencies. We clearly need namespace injection; I'm not sure why namespace_strip is a separate operation (isn't that the same as namespace injection with managed=False?); and I agree tokenizing while building a package zip doesn't make any sense. @jlantz can you provide any historic context?

cumulusci/tasks/package_2gp.py
```diff
@@ -0,0 +1,537 @@
...
+        3. Wait for completion.
+        4. Collect package information as return values.
+        """
+        # find existing package in Dev Hub, or create one if necessary
+        self.package_id = self._get_or_create_package(self.package_config)
+        self.return_values["package_id"] = self.package_id
+
+        # submit request to create package version
+        package_zip_builder = MetadataPackageZipBuilder(
+            path=self.project_config.default_package_path,
```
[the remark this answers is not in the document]
davisagli: I think sfdx is working on a change so that multiple package directories get deployed in multiple metadata API operations in a well-defined order, rather than a combined one. I'm not sure how that will translate to package creation.

cumulusci/tasks/package_2gp.py
```diff
@@ -0,0 +1,537 @@
...
+        Package2 = self._get_tooling_object("Package2")
+        package = Package2.create(
+            {
+                "ContainerOptions": package_config.package_type,
+                "Name": package_config.name,
+                "Description": package_config.description,
+                "NamespacePrefix": package_config.namespace,
+            }
+        )
+        return package["id"]
```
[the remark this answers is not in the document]
davisagli: I guess we'll find out! And can translate it into a more helpful exception once we know how it manifests.

cumulusci/tasks/package_2gp.py
```diff
@@ -0,0 +1,537 @@
...
+        """
+        new_dependencies = []
+        for dependency in dependencies:
+            if dependency.get("dependencies"):
+                new_dependencies.extend(
+                    self._convert_project_dependencies(dependency["dependencies"])
+                )
+
+            new_dependency = {}
+            if dependency.get("version_id"):
```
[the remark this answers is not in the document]
davisagli: Good catch!

cumulusci/tasks/package_2gp.py
davisagli: Yeah, it's not ideal. If you try to use an un-namespaced package as a dependency of another 2gp package, you get an error saying you can't do that. It doesn't need to match but it needs to have a namespace, and using the same one as the primary package being created is simplest because we know it's already in the devhub's namespace registry.

cumulusci/tasks/tests/test_package_2gp.py
davisagli: Fair question. I should at least make sure that all the fake responses were actually used.

docs/features.rst
```diff
@@ -772,3 +772,54 @@ will not show you exception tracebacks because they are seldom
...
+
+To create a new unlocked package version, run the ``create_package_version`` task against the Dev Hub org:
+
+.. code-block:: console
+
+    $ cci task run create_package_version --org devhub -o package_type Unlocked
+
+This task will look for an unlocked package with the name and namespace specified in the task options (defaulting to the name and namespace from the ``project__package`` section of ``cumulusci.yml``). If a matching package doesn't exist yet, it will be created. Then the task will start creating a new version of this package.
+
+Once creation of the package version has completed (which can take some time), the task will output some information including the SubscriberPackageVersion Id, which can be used to install the package in another org.
```
Br4nd0R, line 804:
```suggestion
This task will look for an unlocked package with the name and namespace specified in the task options (defaulting to the name and namespace from the ``project__package`` section of ``cumulusci.yml``). If a matching package doesn't exist yet, it will be created. Then the task will start creating a new version of this package.

The task submits a request to create the package version, and once completed (which can take some time), the task will output some information including the SubscriberPackageVersion Id, which can be used to install the package in another org.
```

docs/features.rst
```diff
@@ -772,3 +772,54 @@ will not show you exception tracebacks because they are seldom
...
+-------------
+
+In order to create unlocked package versions, you need to have a few things set up:
+
+1. `Enable Dev Hub in Your Org <https://developer.salesforce.com/docs/atlas.en-us.sfdx_setup.meta/sfdx_setup/sfdx_setup_enable_devhub.htm>`_
+2. `Enable Unlocked and Second-Generation Managed Packaging <https://developer.salesforce.com/docs/atlas.en-us.sfdx_setup.meta/sfdx_setup/sfdx_setup_enable_secondgen_pkg.htm>`_
+3. Connect the Dev Hub org to the CumulusCI keychain by running ``cci org connect devhub`` (this is necessary even if sfdx has already authenticated to the Dev Hub).
+4. If you want to create an unlocked package with a namespace, you must also create a new Developer Edition org to `Create and Register Your Namespace <https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_unlocked_pkg_create_namespace.htm>`_, and link the namespace to your Dev Hub.
+
+Create a package version
```
Br4nd0R, line 793:
```suggestion
Create a Package Version
```

docs/features.rst
```diff
@@ -772,3 +772,54 @@ will not show you exception tracebacks because they are seldom
...
+
+    $ cci task run create_package_version --org devhub -o package_type Unlocked
+
+This task will look for an unlocked package with the name and namespace specified in the task options (defaulting to the name and namespace from the ``project__package`` section of ``cumulusci.yml``). If a matching package doesn't exist yet, it will be created. Then the task will start creating a new version of this package.
+
+Once creation of the package version has completed (which can take some time), the task will output some information including the SubscriberPackageVersion Id, which can be used to install the package in another org.
+
+If a package version already exists with the exact same contents, its Id will be returned instead of creating a new package version.
+
+Handling dependencies
```
Br4nd0R, line 808:
```suggestion
Handling Dependencies
```

docs/features.rst
```diff
@@ -772,3 +772,54 @@ will not show you exception tracebacks because they are seldom
...
+---------------------
+
+If your project has dependencies configured in the ``project`` section of ``cumulusci.yml``, CumulusCI will try to convert them into a Subscriber Package Version Id (``04t`` key prefix), which is the format required for dependencies in the API for creating a package version.
+
+For dependencies that are specified as a managed package namespace and version, or dependencies specified as a GitHub repository with releases that can be resolved to a namespace and version, CumulusCI needs an org with the dependencies installed in order to do this conversion. By default, it will create a new scratch org named ``2gp_dependencies`` and run the ``dependencies`` flow in order to get an org where these ids can be looked up. If you want to use an existing scratch org rather than creating a new one, set the ``dependency_org`` option for the ``create_package_version`` task.
+
+For dependencies that are an unpackaged bundle of metadata, CumulusCI will create an additional unlocked package to contain them.
+
+Promote a package version
```
Br4nd0R, line 817:
```suggestion
Promote a Package Version
```

docs/features.rst
```diff
@@ -772,3 +772,54 @@ will not show you exception tracebacks because they are seldom
...
+-------------------------
+
+In order to be installed in a production org, an unlocked package version must be
+`promoted <https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_unlocked_pkg_create_pkg_ver_promote.htm>`_
+to mark it as released.
+
+CumulusCI does not yet provide any tools to help with this, so for now you must use the ``sfdx force:package:version:promote`` command.
```
Br4nd0R, line 824: @davisagli - Is there a reason we don't do this? Seems like it would be trivial to make a couple`sfdx` calls, so I'm guessing could be something else that's a blocker here.

Br4nd0R (approved):

docs/features.rst
davisagli: @Br4nd0R Just haven't gotten to it yet. It shouldn't be terribly tricky, but:
- if force:package:version:promote depends on package ids being in sfdx-project.json, we'll probably want to do it directly by updating Package2Version via the Tooling API instead
- if the package version depends on additional unlocked packages that were created for pre/post metadata, we should promote them too

jlantz (approved):

davidmreed (approved):

PR #1820, opened 2020-06-18
Automatic Namespace Injection in Bulk Data

davidmreed:
TODO:
- [x] Rewrite documentation to match
- [x] Revise `generate_dataset_mapping` to match.
- [x] Unit test
- [x] Integration test
- [x] Test on real projects
  - [x] Volunteers for Salesforce
    - [x] Data load in namespaced org
    - [x] Data load in managed org
    - [x] Data extraction from managed org
    - [x] Mapping generation from managed org
    - [x] Data deletion in managed org
    - [x] Legacy data load in managed org
    - [x] Legacy data extraction in managed org
    - [x] Legacy data delete in managed org
  - [x] CaseMan
    - [x] Data load in NPSP org
    - [x] Data load in non-NPSP org

# Critical Changes

- The `load_dataset`, `extract_dataset`, and `delete_data` tasks now support automatic namespace injection. When mapping files and options (for `delete_data`) are specified _without_ namespaces, and the target org only contains the given schema with the project's namespace attached, CumulusCI automatically resolves the schema reference to the namespaced field or object. This makes it easier for projects to use a single mapping file for unmanaged orgs, namespaced scratch org, and managed orgs.
 
   This behavior is on by default, and may be disabled with an option. This option is believed to be backwards-compatible. However, projects that subclass built-in data loading classes, or which use data loading tasks in very unusual ways, might be impacted. 

# Changes

- The `load_dataset` and `extract_dataset` commands support a `drop_missing_schema` option. This option causes CumulusCI to silently ignore elements in a dataset or mapping that are not present in the target org. This option is useful when building datasets that support additional, optional managed packages, which may or may not be installed.
- The `generate_dataset_mapping` task generates mappings in line with the latest revisions of load/extract functionality: fields are specified as a list; the `table` tag is omitted; and namespaces are stripped.

# Issues Closed


cumulusci/tasks/bulkdata/load.py
```diff
@@ -83,6 +87,13 @@ def _init_options(self, kwargs):
         if self.bulk_mode and self.bulk_mode not in ["Serial", "Parallel"]:
             raise TaskOptionsError("bulk_mode must be either Serial or Parallel")
 
+        self.options["inject_namespaces"] = process_bool_arg(
+            self.options.get("inject_namespaces", False)
```
prescod, line 91: What are the risks of injecting by default?

docs/bulk_data.rst
```diff
@@ -164,7 +203,9 @@ If no such mapping is provided, CumulusCI will remove the Salesforce Id from ext
 data and replace it with an autoincrementing integer primary key.
 
 Use of integer primary keys may help yield more readable text diffs when storing data in SQL
-script format. However, it comes at some performance penalty when extracting data.
+script format. However, it comes at some performance penalty when extracting data. It's
+recommended that most mappings do not map the Id field and allow CumulusCI to utilize
+the automatic primary key.
 
 Handling Namespaces
```
prescod, line 210: Is this out of date now? The goal of the PR is to avoid having to maintain two mapping files, right?

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -51,7 +53,7 @@ class MappingStep(CCIDictModel):
     "Step in a load or extract process"
     sf_object: str
     table: Optional[str] = None
-    fields_: Dict[str, str] = Field(..., alias="fields")
+    fields_: Optional[Union[Dict[str, str], List[str]]] = Field({}, alias="fields")
```
prescod, line 56: Not sure that we need the keyword Optional here if we are also supplying a default. Perhaps the default implies the Optional?

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -78,18 +80,167 @@ def oid_as_pk_is_deprecated(cls, v):
...
+        """Process the schema elements in this step.
+
+        First, we inject the namespace into object and field names where applicable.
+        Second, we validate that all fields are accessible by the running user with the
+        correct permission level for this operation.
+        Lastly, if drop_missing is True, we strip any fields that are not present (namespaced
+        or otherwise) from the target org.
+
+        Return True if this object should be processed. If drop_missing is True, a False return
+        value indicates we should skip this object. If drop_missing is False, a Fakse return
```
prescod, line 127:
```suggestion
        value indicates we should skip this object. If drop_missing is False, a False return
```

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -78,18 +80,167 @@ def oid_as_pk_is_deprecated(cls, v):
...
+        ret = True
+
+        def inject(element: str):
+            return f"{namespace}__{element}"
+
+        # Determine whether we need to inject our sObject.
+        if inject_namespaces and self._is_injectable(self.sf_object):
+            global_describe = {
+                entry["name"]: entry
+                for entry in org_config.salesforce_client.global_describe()["sobjects"]
```
prescod, line 139: Funny that I can't find this method:

```
 $  cci org shell --python "print(org_config.salesforce_client.global_describe())"
2020-06-18 12:29:15: Getting scratch org info from Salesforce DX
Error: 'SFType' object is not callable
Run this command for more information about debugging errors: cci error --help
```

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -78,18 +80,167 @@ def oid_as_pk_is_deprecated(cls, v):
...
+                    self.fields[inject(f)] = db
+                    del self.fields[f]
+                    f = inject(f)
+
+            # Do we have the right permissions for this field, or do we need to drop it?
+            if f not in describe or (
+                not describe[f]["createable"] and operation is DataOperationType.LOAD
+            ):
+                logger.warning(
+                    f"Field {self.sf_object}.{f} is not present or does not have the correct permissions."
```
prescod, line 193: Might be worth the small extra effort to tell the user which case is the case.

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -78,18 +80,167 @@ def oid_as_pk_is_deprecated(cls, v):
...
+                )
+                if drop_missing:
+                    del self.fields[f]
+                else:
+                    ret = False
+
+        # Validate, inject, and drop (if configured) lookups.
```
prescod, line 200: So..close to the above. Could it be factored into a method and take the dictionary to work with as a param?

prescod:
Overall looks very nice. Looking forward to it!!!

docs/bulk_data.rst
davidmreed: Yes, good point - this will need to be rewritten, as will the mapping generator.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: That's because it is a mistake! Fixed that.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: Unfortunately we can't tell the difference between a field that is not readable and a field that does not exist without calling the Tooling API separately. The latest version (just pushed) factors this code in a way that might make that a little trickier but I'll see if I can clarify.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: Yep. Apologies, I failed to push a revision this morning that factors that out.

cumulusci/tasks/bulkdata/load.py
davidmreed: That's a really good question. To be honest, I'm leery of it primarily because the implications are so hard to reason out, across complex scenarios like multi-package assemblages with unmanaged metadata in namespaced orgs.

It is probably safe. There's an additional check I want to add so that we don't attempt to namespace-inject a field if its namespace-injected version is also present in the file, to support the corner case where a field is present both namespaced and non-namespaced. That would make it even safer.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: Pydantic's documentation is a little on the thin side. I don't have a good understanding of how type constraints, defaults, and validators are ordered or interact with each other.

cumulusci/tasks/bulkdata/load.py
davidmreed: In the current version, I have adopted on-by-default injection.

cumulusci/tasks/bulkdata/extract.py
```diff
@@ -36,6 +42,14 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
             "description": "If set, an SQL script will be generated at the path provided "
             + "This is useful for keeping data in the repository and allowing diffs."
         },
+        "inject_namespaces": {
+            "description": "If set, CumulusCI automatically injects the project's namespace if schema is managed in the org. "
+            "Defaults to True. Set to False to deactivate automatic namespace injection."
+        },
+        "drop_missing_schema": {
+            "description": "Set to True to have CumulusCI transparently drop any missing schema elements."
```
davisagli, line 50:
```suggestion
            "description": "Set to True to skip any missing objects or fields instead of stopping with an error."
```

cumulusci/tasks/bulkdata/load.py
```diff
@@ -57,6 +55,14 @@ class LoadData(BaseSalesforceApiTask, SqlAlchemyMixin):
...
             "description": "Set to Serial to force serial mode on all jobs. Parallel is the default."
         },
+        "inject_namespaces": {
+            "description": "If set, CumulusCI automatically injects the project's namespace if schema is managed in the org. "
+            "Defaults to True. Set to False to deactivate automatic namespace injection."
+        },
+        "drop_missing_schema": {
+            "description": "Set to True to have CumulusCI transparently drop any missing schema elements."
+            "This can support mappings that include data for optional packages."
+        },
```
davisagli, line 65: Ditto suggestions for option descriptions in the other task modules.

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -96,18 +103,180 @@ def set_default_table(cls, values):
...
+        return sobject in global_describe and global_describe[sobject][perm]
+
+    def _check_field_permission(
+        self, describe: Dict, field: str, operation: DataOperationType
+    ):
+        perm = self._get_permission_type(operation)
+        # Fields don't have "queryable" permission.
+        return field in describe and (
+            describe[field].get(perm) if perm in describe[field] else True
+        )
```
davisagli, line 138: No need to do this now, but at some point it might make sense to have our own DescribeSObjectResult class which adapts the JSON result to provide helper methods for checking things like this.

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -124,3 +293,36 @@ def validate_mapping(cls, values):
...
+    if not drop_missing and not all(should_continue):
+        raise BulkDataException("One or more permissions errors blocked the operation.")
+
+    if drop_missing:
+        # Drop any steps with sObjects that are not present.
+        for (include, step_name) in zip(should_continue, list(mapping.keys())):
+            if not include:
+                del mapping[step_name]
+
+        # Remove any remaining lookups to dropped objects.
```
davisagli, line 323: Should we try to detect and error if one of these lookups is required?

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -124,3 +293,36 @@ def validate_mapping(cls, values):
 def parse_from_yaml(source: Union[str, Path, IO]) -> Dict:
     "Parse from a path, url, path-like or file-like"
     return MappingSteps.parse_from_yaml(source)
+
+
+def validate_mapping(
```
davisagli, line 298: "validate" for me implies that it will check status but not mutate. Can we give it a name that makes it clearer that mapping may be changed by running this?

cumulusci/tasks/bulkdata/tests/test_delete.py
```diff
@@ -10,12 +13,15 @@
...
 from cumulusci.tasks.bulkdata.tests.utils import _make_task
+from cumulusci.tasks.bulkdata.tests.test_utils import mock_describe_calls
 
 
 class TestDeleteData(unittest.TestCase):
+    @responses.activate
     @mock.patch("cumulusci.tasks.bulkdata.delete.BulkApiQueryOperation")
     @mock.patch("cumulusci.tasks.bulkdata.delete.BulkApiDmlOperation")
     def test_run(self, dml_mock, query_mock):
+        mock_describe_calls()
```
davisagli, line 24: Doesn't have to happen now (it requires refactoring to not use unittest.TestCase) but this looks like a good candidate for a class-scoped pytest fixture

docs/bulk_data.rst
```diff
@@ -258,6 +255,26 @@ pattern, but in reverse.
...
+data sets, but load that data if and only if B is installed.
+
+To support this use case, the ``load_dataset`` and ``extract_dataset`` tasks offer a ``drop_missing_schema``
+option. When enabled, this option results in CumulusCI ignoring any mapped fields, sObjects, or lookups that
+correspond to schema that is not present in the org.
+
+Projects that require this type of conditional behavior can build their datasets in an org that contains managed
+package B, capture it, and then load it safely in orgs that both do and do not contain B. However, it's important
+to always capture from an org with B present, or B data will not be preserved in the dataset.
+
```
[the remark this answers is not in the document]
davidmreed: Yes, great point. I will add a note to that effect.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: If the child object is a master-detail child, the parent must be visible or the child won't be. If it's a lookup relationship, though, that's a risk. We can check for database-level required fields, and I'll add that.

davisagli:
@davidmreed should we adjust generate_dataset_mapping to strip the package namespace if present?

davidmreed:
@davisagli I think so.

I'm trying to come up with a scenario where that would break something, but I think the key is that if the mapping starts with no namespaces and the logic implemented here is applied (which does nothing in case of ambiguity), it's safe in all three contexts save for one circumstance. 

If the target org is managed and contains a field that is packaged and one not packaged with the same API name, a mapping needs to explicitly reference them both, and stripping the namespace would be wrong. But if that happens, the project is doing something profoundly strange in the first place (or, more likely, has broken unpackaged metadata), and it's a pretty easy situation to detect.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: Renaming to `validate_and_inject_mapping()`. This method is admittedly a bit of a chimera. It's efficient to do this in a single operation, though, because the describe logic is in common.

davidmreed:
I'm building the new version of `generate_dataset_mapping` in a child branch to this one, to be pushed shortly.

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: Added!

cumulusci/tasks/bulkdata/tests/test_utils.py
```diff
@@ -31,6 +32,44 @@ def create_db_memory():
     return engine, metadata
 
 
+def mock_describe_calls():
```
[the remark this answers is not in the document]
davidmreed: Yeah, I can do that. Having that extra data in the global describe doesn't add any value without the object describe also being present.

davidmreed:
Added revision of `generate_dataset_mapping` to this branch.

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -161,36 +168,32 @@ def _build_mapping(self):
...
         stack = self._split_dependencies(objs, self.refs)
 
-        field_sort = (
-            lambda f: "  " + f
-            if f == "Id"
-            else (" " + f if f in self.core_fields else f)
-        )
+        def strip_namespace(element):
+            ns = self.project_config.project__package__namespace
+            if self.options["strip_namespace"] and ns and element.startswith(ns):
```
davisagli, line 173: We should be checking if it starts with ns + "__" since there could be multiple namespaces in the org where one is a substring of another.

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -161,36 +168,32 @@ def _build_mapping(self):
...
+                return element
 
         self.mapping = {}
         for obj in stack:
-            key = f"Insert {obj}"
+            key = f"Insert {strip_namespace(obj)}"
             self.mapping[key] = {}
-            self.mapping[key]["sf_object"] = f"{obj}"
-            self.mapping[key]["table"] = f"{obj}"
+            self.mapping[key]["sf_object"] = strip_namespace(obj)
```
davisagli, line 182: Call strip_namespace once and assign it to something so it doesn't need to be called twice

cumulusci/tasks/bulkdata/generate_mapping.py
davisagli: Also while there's no particular need to micro-optimize this, the assignment of `ns` could be moved outside the closure.

docs/tasks.rst
```diff
@@ -3907,6 +3932,16 @@ Options
...
+
+	 If True, the package namespace prefix will be automatically added to objects and fields for which it is present in the org. Defaults to True.
+
+``-o drop_missing_schema DROPMISSINGSCHEMA``
+	 *Optional*
+
+	 Set to True to skip any missing objects or fields instead of stopping with an error.
+
 **load_custom_settings**
 ==========================================
```
davisagli, line 3947: Need to update this with the new generate_dataset_mapping option.

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -160,60 +167,82 @@ def _build_mapping(self):
...
+            stripped_obj = (
+                strip_namespace(orig_obj)
+                if strip_namespace(orig_obj) not in stack
+                else orig_obj
+            )
+            key = f"Insert {stripped_obj}"
             self.mapping[key] = {}
-            self.mapping[key]["sf_object"] = f"{obj}"
-            self.mapping[key]["table"] = f"{obj}"
+            self.mapping[key]["sf_object"] = stripped_obj
```
davisagli, line 188: My suggestion:

    stripped_obj = strip_namespace(orig_obj)
    obj = stripped_obj if stripped_obj not in stack else orig_obj
    key = f"Insert {obj}"
    self.mapping[key] = {}
    self.mapping[key]["sf_object"] = obj

davidmreed: Adopted!

docs/tasks.rst
davidmreed: Done!

davidmreed:
@davisagli Yes, I am. I will aim to do that today.

davidmreed:
All hands-on tests passed successfully. @davisagli

PR #1823, opened 2020-06-19
Feature/resumable generate and load

prescod:
Description TBD.

Start reading the code here:

https://github.com/SFDO-Tooling/CumulusCI/compare/feature/resumable-generate-and-load?expand=1#diff-8f454c91a54a0fcab96adfd4173aa84bR243


cumulusci/utils/fileutils.py
```diff
@@ -86,6 +88,69 @@ def load_from_source(
...
+    def func(self, *args, **kwargs):
+        real_func = getattr(self.fs, funcname)
+        return real_func(self.filename, *args, **kwargs)
+
+    func.__doc__ = getattr(base.FS, funcname).__doc__
+    return func
+
+
+class FSResource(str):
+    """Generalization of pathlib.Path to support S3, FTP, etc
```
davisagli, line 101: It makes me somewhat nervous to make this so flexible, because it makes it so much more important to ensure that the execution environment only has access to resources that the people with commit access to the project can be trusted with.

cumulusci/tasks/util.py
```diff
@@ -68,15 +68,33 @@ def _run_task(self):
...
+
+    def _resume(self):
+        self.logger.info(f"Continuing sleeping for {self.seconds_remaining} seconds")
+
+    def _run_step(self):
+        time.sleep(1)
+        self.seconds_remaining -= 1
+
+    def _is_finished(self):
+        return self.seconds_remaining <= 0
```
davisagli, line 95: This is useful for demonstrating the API, but I would expect an actual resumable sleep task to compare to the original start time rather than tracking how many individual seconds it has slept for. If it's supposed to sleep for 10 seconds, and actually sleeps for 2 before getting interrupted, and then gets resumed 8 seconds later, it should recognize that it has already achieved the goal of waiting 10 seconds.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -246,7 +245,7 @@ def _query_db(self, mapping):
         columns = [getattr(model, id_column)]
 
         for name, f in fields.items():
-            if name not in ("RecordTypeId", "RecordType"):
+            if name not in ("RecordTypeId", "RecordType", "record_type"):
```
davisagli, line 248: This diff in this module looks unrelated.

cumulusci/core/tasks.py
```diff
@@ -226,6 +236,115 @@ def freeze(self, step):
...
+    def _start(self):
+        "Override this: Do any kind of task setup or logging that is necessary."
+        self.logger.info(f"Starting task {self.__class__.name}")
+
+    def _run_steps(self):
+        "Run multiple steps until finished. Usually do not override this."
+        while not self._is_finished():
+            self._run_step()
+            self.resumption_fs_resource.state_data = self.dict()
+            self.resumption_fs_resource.save()
```
davisagli, line 323: Won't an interruption in _run_step generally raise an exception that prevents this code from running to save the state?

cumulusci/core/tasks.py
```diff
@@ -226,6 +236,115 @@ def freeze(self, step):
...
+        self.__dict__[key] = value
+        self.excluded_fields.add(key)
+
+    def dict(self):
+        return {k: v for k, v in self.__dict__.items() if k not in self.excluded_fields}
+
+    def __setattr__(self, name, value):
+        if name not in self.excluded_fields and not hasattr(self.__class__, name):
+            raise AttributeError(f"Undeclared attribute {name}")
+        self.__dict__[name] = value
```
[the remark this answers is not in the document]
prescod: My basic feeling is that almost all state should be in the resumption file or on the stack. Neither of the first two tasks converted to this model use set_nonpersistent_value at all.

So the fact that it is inconvenient is fine because it is rare, and hopefully the fact that it is inconvenient will help ensure it IS rare. I think that mixing a lot of persistent and non-persistent data is just likely to confuse the programmer. The base class uses that function a few times because its just the nature of frameworks that they use a bit more magic than the stuff built on top of them.

cumulusci/utils/fileutils.py
prescod: I'm not sure it does make any security difference, really. If you have the auth information for an FTP site or an S3 bucket, you don't need this class to save your data there. And if you don't, it won't help you. (I have so far not done anything at all about providing access to S3 as a service)

cumulusci/tasks/util.py
prescod: Yeah as long as you trust distributed system clocks. My way is a bit more conservative in that it really counts that all of the time passed no matter what the clocks say. For the practical uses we might use sleep for (a few seconds waiting for a process to finish?), either is probably fine?

prescod: If sleeping were important it might make sense to bake it into a deeper layer so MetaCI could just kill the process and forget about it until the sleeping is done. Reconstituting a process to check sleep progress is kind of wasteful.

cumulusci/core/tasks.py
prescod: I think that's the right design. Remember we had used the metaphor of a "transaction". Saving after an exception would save some the state halfway through a "transaction" with variables in who knows what state?

I am inspired by the concept of "crash only software." You write the software on the presumption that it could just die anywhere anytime with no warning and then it will automatically handle cases where it dies WITH warning as a side effect.

https://en.wikipedia.org/wiki/Crash-only_software

If you do have state that you want to save instantly, arguably a relational database would be a better place for it and I do think that our tasks should have some form of easy access to an RDB.

cumulusci/core/tasks.py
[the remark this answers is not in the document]
davisagli: Okay, I am on board with the argument for making it a bit more work to store nonpersistent state in instance attributes. (It helps that there are a number of existing tasks where I think instance attributes are overused.)

But you didn't answer the other part of my question, about the possibility of keeping persistent state more explicitly contained in a separate `state` object. My arguments for this are:
- It makes it clearer which values are being persisted; instead of having to keep track of which names are being stored which way, you can just look at where the value is being set.
- It would remove the need to manage a list of excluded fields for each task, and clean up the ResumableTask implementation a bit since it wouldn't need to filter based on what's in excluded_fields
- It makes resuming more straightforward: The task can work with a state object reconstructed in its entirety from the resumption file, instead of needing to transfer specific state values to specific task attributes.

cumulusci/tasks/util.py
davisagli: The sleep task is mostly for testing, so it's not particularly important; most tasks that actually do need a pause just call time.sleep directly.

cumulusci/core/tasks.py
davisagli: Got it, I was thinking about this wrong.

It does mean we should be sure to call out some warnings in the docs about writing resumable tasks that do non-idempotent actions. (And for the generate/load task, we should document that resuming it has the potential to insert extra data.)

prescod:
Do you mean that the programmer would need to refer to all persistent state
as `self.state.xyz`

Of do you mean that the base-class would override setattr/getattr to
transparently store its state elsewhere?

cumulusci/core/tasks.py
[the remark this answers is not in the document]
davisagli: @prescod copying your comment here since github didn't attach it to the right thread.

> Do you mean that the programmer would need to refer to all persistent state
as `self.state.xyz`

yes, this one. explicit is better than implicit / namespaces are one honking great idea

> Of do you mean that the base-class would override setattr/getattr to
transparently store its state elsewhere?

No.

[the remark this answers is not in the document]
prescod: I actually implemented it that way at first, but then it seemed like a bunch of extra typing to no benefit for the task creator.

```
self.state.this = 
self.state.that =
```

Plus two classes per task (the task and its strong-typed state). And two base classes to import (the BaseResumableTask and its BaseResumptionState).

It seemed like a lot of boilerplate for not much (if any) benefit (for the Task programmer) so I combined them.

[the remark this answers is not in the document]
davisagli: I still disagree. To me, it feels like a worthwhile separation of concerns.

jstvz:

PR #1825, opened 2020-06-20
Add a save method for clarity and to follow Python conventions

prescod:
Personally, I think it is clearer if org_configs should have a .save() method like Django models. Maybe just a matter of taste. We'll see.

I left instances of `runtime.keychain.set_org()` in those cases where the org is new.


cumulusci/core/config/OrgConfig.py
```diff
@@ -248,3 +248,6 @@ def installed_packages(self):
 
     def reset_installed_packages(self):
         self._installed_packages = None
+
+    def save(self, global_org: bool = False):
```
davisagli, line 252: docstring

prescod:
Do you prefer me to add the docstring and ask for a re-review or just abandon the idea because you don't think it helps anything?

Another option would be improving the naming of the set_org method which to me sounds like a setter and not a saver. set_whatever implies that you're saying whatever = X. But set_org is not doing anything like that.

davisagli:
> save_org would be a bit better.

I agree. Although set_org is part of the public interface for keychains and I suspect there are custom keychain subclasses in our web apps that override it. So we probably need a phased approach:
1. add save_org
2. turn set_org into a deprecated alias for save_org and convert any uses within CumulusCI
3. make a release
4. update the web apps to use save_org
5. maybe remove set_org

davisagli:
Superceded by #1876

PR #1826, opened 2020-06-20
Clear instance URL

prescod:
# Changes

Deleting a scratch org now clears the instance URL from cci org list.


davisagli:
@prescod It would be helpful if you could put a draft release note in the appropriate section of the pull request template rather than deleting the release note headings.

PR #1829, opened 2020-06-23
bulkdata supports Person Accounts

spelak-salesforce:
# Critical Changes

# Changes
## bulkdata supports Person Accounts:
- `bulkdata.ExtractData` queries and persists `IsPersonAccount` field for steps on the Account and Contact objects if Person Accounts is enabled.   This information is used when loading the bulkdata.
- `bulkdata.LoadData`:
    - When loading Account records:
        - Before loading Account records, sets Account `Name` as blank for Person Accounts.
        - Include Person Account Contact fields in the mapping's Account's step., e.g. `LastName`, `PersonBirthdate`.
    - When inserting Contact records:
        - Only inserts non-Person Account Contact records.
        - Fields for Person Account Contact records are NOT sent for updates.   Include Person Account Contact fields in the mapping's Account's step.
        - After inserting Contact records, queries for Person Account Contact record IDs so downstream objects can look up to Person Account Contact records.

### Recommendations
- Make sure your org has person accounts enabled.  For scratch orgs, add the [PersonAccounts](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file_config_values.htm#so_personaccounts) feature. in the scratch org definition.
- For Person Account **Account** records, all **Contact** fields must be included in **Account** object steps.
- If an Account mapping step does not specify a `RecordTypeId` field mapping, person account **Account** records get assigned the first Person Account **Account** record type, and business **Account** records get assigned the default **Account** record type.
- Example:
   ```yml
  Account:
    sf_object: Account
    table: Account
    fields:
      # We want our records to get the specified record types
      RecordTypeId: RecordTypeId
      # Account Fields
      Name: Name
      AccountNumber: AccountNumber
      # ...
      Description: Description
      # Contact Fields
      Salutation: Salutation
      FirstName: FirstName
      LastName: LastName
      PersonTitle: PersonTitle
      PersonEmail: PersonEmail
      PersonBirthdate: PersonBirthdate
      # ...
  ```

# Issues Closed


cumulusci/tasks/bulkdata/load.py
```diff
@@ -485,3 +522,128 @@ def _expand_mapping(self):
...
+                table = self.models[mapping["table"]].__table__
+                self.session.connection().execute(
+                    table.update()
+                    .where(table.columns.get("IsPersonAccount") == "true")
+                    .values(**{column_name: ""})
+                )
+                self.session.flush()
+                return
+
+    def _filter_out_person_account_records(self, query, model):
```
spelak-salesforce, line 569: This is its own method so it's easy to test when this is called in `_query_db` method.

spelak-salesforce:
GUS: [W-7867779: Add Person Account storytelling data[(https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008WVqfIAG/view)

cumulusci/tasks/bulkdata/tests/test_extract.py
```diff
@@ -374,24 +503,112 @@ def test_convert_lookups_to_id__sqlite(self):
...
+        self, mapper_mock, create_mock
+    ):
+        for sf_object, table in [
+            ("Account", "accounts"),
+            ("ACCOUNT", "accounts"),
+            ("account", "accounts"),
+            ("Contact", "contacts"),
+            ("CONTACT", "contacts"),
+            ("contact", "contacts"),
+        ]:
```
davisagli, line 516: I'm not a fan of this way of cyling through different parameters for a test, because when it breaks the test output does not make it clear which input failed. A better way to do this is with pytest's parametrization feature: https://docs.pytest.org/en/stable/parametrize.html

cumulusci/tasks/bulkdata/load.py
```diff
@@ -508,3 +548,128 @@ def _expand_mapping(self):
...
+        for api_name, lookup in mapping.get("lookups", {}).items():
+            if api_name.lower() == "accountid":
+                return lookup
+
+    def _generate_contact_id_map_for_person_accounts(
+        self, contact_mapping, account_id_lookup, conn
+    ):
+        """
+        Yields (local_id, sf_id) for Contact records where IsPersonAccount
+        is true that can handle large data volumes.
```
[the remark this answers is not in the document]
davisagli: I don't think that it's critical for us to support large data volumes with person accounts for our immediate use cases though, so I don't think that making that change is a blocker for this PR. (And perhaps I'm overlooking something about it that wouldn't work.)

davidmreed:
@spelak-salesforce  Do you have a small/relatively simple Person Accounts data set that I can steal to build an integration test for this functionality? We have a suite of (fairly expensive) integration tests for Bulk Data that I'd like to get this added to.

cumulusci/tasks/bulkdata/extract.py
```diff
@@ -286,6 +290,13 @@ def _create_table(self, mapping):
         mapper_kwargs = {}
         self.models[mapping["table"]] = type(model_name, (object,), {})
 
+        # Track IsPersonAccount for Account and Contact tables if person accounts is enabled.
+        # IsPersonAccount field should never exist in a mapping because the field is not creatable.
```
davidmreed, line 294: With FLS-checking changes recently merged, the operation will fail if the user tries to include this field. (Just an FYI, not a change request).

cumulusci/tasks/bulkdata/extract.py
```diff
@@ -286,6 +290,13 @@ def _create_table(self, mapping):
...
+        if (
+            mapping["sf_object"].lower() in ["account", "contact"]
+        ) and self._is_person_accounts_enabled():
+            mapping["fields"]["IsPersonAccount"] = "IsPersonAccount"
```
davidmreed, line 298: Open to discussion on the best place, but I'd prefer to have this kind of mapping mutation go in the `mapping_parser` module along with the FLS checker and namespace injector. (Those have been added since you started this work). Keeping as much of that logic in one place as possible will help reduce the level of magic required to work on this code and let us keep `_create_table` minimal.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -508,3 +548,128 @@ def _expand_mapping(self):
...
+    def _is_person_account_column_exists(self, mapping):
+        """
+        Returns if "IsPersonAccount" is a column in mapping's table.
+        """
+        return (
+            self.models[mapping.get("table")].__table__.columns.get("IsPersonAccount")
+            is not None
+        )
+
+    def _is_person_accounts_enabled(self, mapping):
```
davidmreed, line 561: Should we detect that a Person Accounts data store is being loaded into a non-Person Accounts org and stop the operation? I'm not entirely sure what would happen in that situation.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -249,6 +256,12 @@ def _query_db(self, mapping):
         """
         model = self.models[mapping.get("table")]
 
+        # Update Account.Name as blank for IsPersonAccount Account records.
+        if mapping[
+            "sf_object"
+        ].lower() == "account" and self._is_person_accounts_enabled(mapping):
+            self._update_person_account_name_as_blank(mapping)
```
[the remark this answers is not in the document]
davidmreed: I agree with David - it could be done cleanly, I think, by adding another generator in the chain within ExtractData's `_import_results()`, and then we don't have to touch the SQL database later. There's some legacy code that still does this, but we'd like to avoid mutating the data store during a load operation.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -508,3 +548,128 @@ def _expand_mapping(self):
...
+        Caches is_person_accounts_enabled response which consumes a describe
+        call.
+        """
+        if self._is_person_account_column_exists(mapping):
+            if self._person_accounts_enabled is None:
+                self._person_accounts_enabled = is_person_accounts_enabled(self)
+            return self._person_accounts_enabled
+        return False
+
+    def _update_person_account_name_as_blank(self, mapping) -> None:
```
davidmreed, line 575: cf. discussion thread above.

cumulusci/tasks/bulkdata/load.py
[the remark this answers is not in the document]
davidmreed: We could split the difference by using the [sObject Collections](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_retrieve.htm) API to retrieve 2,000 records at a go.

On balance though I'm inclined to have this stand as-is and do a light refactor when we write REST API primitive `Operation` classes to let the loader pick its API throughout, so we can be smart about using either the Bulk or REST APIs. Let's just document that Person Accounts should not be used in large data volume loads for now.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -508,3 +548,128 @@ def _expand_mapping(self):
...
+            # While we have a chunk to process
+            chunk = query_result.fetchmany(200)
+            if not chunk:
+                break
+
+            # Collect Map: Account SF ID --> Contact ID
+            contact_ids_by_account_sf_id = {record[1]: record[0] for record in chunk}
+
+            # Query Map: Account SF ID --> Contact SF ID
+            # It's safe to use query_all since the chunk size to 200.
```
davidmreed, line 665: The WHERE clause limit is 4,000 characters. Each Id takes 20 characters (18 + two quotes), plus N-1 commas, plus the 44 static characters = 44 + 199 + (20 * 200) = 4,243 characters. The maximum we can fetch is 188 records, if my math's right, unless we switch to using the Collections API.

cumulusci/tasks/bulkdata/load.py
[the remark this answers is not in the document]
spelak-salesforce: 👍   Sounds great and will update.

PS, I merged #1885 into this branch.

cumulusci/tasks/bulkdata/load.py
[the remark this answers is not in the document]
spelak-salesforce: I ran a "high volume" test with 25,000 person accounts (with 25,000 children), and this step to retrieve person account Contact IDs is significantly shorter than the Bulk API to insert the person account records.

I think this does support high volumes.

- `Account` Bulk API insert
    - Start: `13:41:15`
    - Complete: `13:43:03`
    - Time: `1:48`
- Person Account Contact ID queries.  Runs during the Contact step.
    - Start: `13:43:03`
    - Complete: `13:43:19`
    - Time: `0:16`

```
2020-07-27 13:41:06: Getting scratch org info from Salesforce DX
2020-07-27 13:41:09: Beginning task: LoadData
2020-07-27 13:41:09:        As user: test-xf0wb2lazvsl@example.com
2020-07-27 13:41:09:         In org: 00D0R000000AMmL
2020-07-27 13:41:09: 
2020-07-27 13:41:15: Using in-memory SQLite database
2020-07-27 13:41:16: Running step: Account
2020-07-27 13:41:16: Extracting Record Types for Account
2020-07-27 13:41:18: Uploading batch 1
2020-07-27 13:41:20: Uploading batch 2
2020-07-27 13:41:22: Prepared 25000 rows for insert to Account
2020-07-27 13:41:22: Uploading batch 3
2020-07-27 13:41:24: Waiting for job 7500R000003g2POQAY (0/3 batches complete)
2020-07-27 13:41:35: Waiting for job 7500R000003g2POQAY (0/3 batches complete)
2020-07-27 13:41:46: Waiting for job 7500R000003g2POQAY (0/3 batches complete)
2020-07-27 13:41:57: Waiting for job 7500R000003g2POQAY (0/3 batches complete)
2020-07-27 13:42:07: Waiting for job 7500R000003g2POQAY (0/3 batches complete)
2020-07-27 13:42:18: Waiting for job 7500R000003g2POQAY (0/3 batches complete)
2020-07-27 13:42:29: Waiting for job 7500R000003g2POQAY (1/3 batches complete)
2020-07-27 13:42:40: Waiting for job 7500R000003g2POQAY (1/3 batches complete)
2020-07-27 13:42:51: Waiting for job 7500R000003g2POQAY (1/3 batches complete)
2020-07-27 13:43:02: Waiting for job 7500R000003g2POQAY (3/3 batches complete)
2020-07-27 13:43:02: Job 7500R000003g2POQAY finished with result: Success
2020-07-27 13:43:02: Downloaded results for batch 7510R000003dxacQAA
2020-07-27 13:43:03: Downloaded results for batch 7510R000003dxVsQAI
2020-07-27 13:43:03: Downloaded results for batch 7510R000003dxahQAA
2020-07-27 13:43:03: Running step: Contact
2020-07-27 13:43:04: Prepared 0 rows for insert to Contact
2020-07-27 13:43:05: Waiting for job 7500R000003g2PnQAI (0/0 batches complete)
2020-07-27 13:43:05: Job 7500R000003g2PnQAI finished with result: Success
2020-07-27 13:43:17: Running post-load step: Update Contact Dependencies After Contact
2020-07-27 13:43:17: Prepared 0 rows for update to Contact
2020-07-27 13:43:18: Waiting for job 7500R000003g2PsQAI (0/0 batches complete)
2020-07-27 13:43:19: Job 7500R000003g2PsQAI finished with result: Success
2020-07-27 13:43:19: Running step: outfunds__Funding_Program__c
2020-07-27 13:43:19: Prepared 1 rows for insert to outfunds__Funding_Program__c
2020-07-27 13:43:19: Uploading batch 1
2020-07-27 13:43:20: Waiting for job 7500R000003g2PxQAI (1/1 batches complete)
2020-07-27 13:43:20: Job 7500R000003g2PxQAI finished with result: Success
2020-07-27 13:43:20: Downloaded results for batch 7510R000003dxbBQAQ
2020-07-27 13:43:20: Running post-load step: Update outfunds__Funding_Program__c Dependencies After outfunds__Funding_Program__c
2020-07-27 13:43:20: Prepared 0 rows for update to outfunds__Funding_Program__c
2020-07-27 13:43:21: Waiting for job 7500R000003g2QCQAY (0/0 batches complete)
2020-07-27 13:43:21: Job 7500R000003g2QCQAY finished with result: Success
2020-07-27 13:43:21: Running step: outfunds__Funding_Request__c
```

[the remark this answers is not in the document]
davidmreed: @spelak-salesforce That is good insight - this should be 100% usable for any scratch org-based build.

When we say "high volume", we're often looking to the kinds of LDV loads that Paul does for performance engineering. That's tens of millions of records, and can't be done in a scratch org at all.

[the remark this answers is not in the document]
davisagli: @spelak-salesforce I agree with @davidmreed's assessment: keep this as you've implemented for now, and refactor it to use the BaseDataOperation API once we've implemented support for doing those operations using either the REST API or bulk API depending on what is most appropriate for the scale of dataset being loaded.

cumulusci/tasks/bulkdata/load.py
spelak-salesforce: It's an interesting idea that the task automatically suppresses loading Person Account `Account` and `Contact` records if person accounts is not enabled in the org.

davidmreed: I was thinking more of throwing an exception.

spelak-salesforce:
> @spelak-salesforce Do you have a small/relatively simple Person Accounts data set that I can steal to build an integration test for this functionality? We have a suite of (fairly expensive) integration tests for Bulk Data that I'd like to get this added to.

@davidmreed , `feature/submit-application__person-accounts` branch for [PR 105](https://github.com/SalesforceFoundation/SFDO-Grants/pull/105) in SFDO-Grants.  There are a few data sets in `dev/datasets`.   Please ignore the poor choice in directory names, plus some mappings are commented out because I was in the middle of tests.

cumulusci/tasks/bulkdata/extract.py
spelak-salesforce: That's good the operation will fail.  `IsPersonAccount` should never exist in the user's mapping.   cci then adds `IsPersonAccount` behind the scenes.

cumulusci/tasks/bulkdata/extract.py
spelak-salesforce: This seems like a great approach.  Will dig into change request.

cumulusci/tasks/bulkdata/load.py
spelak-salesforce: Will add check in load in `_run_task` after `_init_db` and before `_expand_mapping` to check if both are true and raise a Bulk Data Exception:
- person accounts is not enabled in the org
- There exists records in the DB where `IsPersonAccount` is true

This means an Exception is raised if person accounts is not enabled in the org but load will try to insert person account records.

cumulusci/tasks/bulkdata/extract.py
spelak-salesforce: This is done.

cumulusci/tasks/bulkdata/load.py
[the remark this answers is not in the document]
spelak-salesforce: Setting Account.Name for person account `Account` records is done in the extract task instead of the load task.

davidmreed:
Existing integration test suite passes with this branch.

davidmreed:
Person Account-specific integration tests are also passing.

cumulusci/tasks/bulkdata/extract.py
davisagli: @spelak-salesforce The new validation code in mapping_parser.py does its own describe call for each object in order to validate FLS and inject namespaces. Now that that is happening, that would be a natural place to check if Account and Contact have the IsPersonAccount field and update the mapping accordingly. The actual tasks could then use presence of IsPersonAccount in the mapping fields as a trustworthy indicator for whether the org has person accounts enabled, and we could get rid of OrgInfoMixin and the _org_has_person_accounts_enabled method.

But -- things are working as is, the release is soon, and I don't think it's worth holding this up to do that refactoring. It can happen later.

cumulusci/tasks/bulkdata/load.py
spelak-salesforce: This is done.

cumulusci/tasks/bulkdata/tests/test_extract.py
spelak-salesforce: I refactored the tests.   Also, I was unable to https://docs.pytest.org/en/stable/parametrize.html because `test_load.py` uses `UnitTest.TestCase`, and pytest.mark.paramertize [is not supported](https://docs.pytest.org/en/3.0.1/unittest.html) for `UnitTest.TestCase`.

cumulusci/tasks/bulkdata/load.py
davidmreed: Per @spelak-salesforce this works fine with volumes of 25k contacts in batches of 200, so I'm resolving - unclear why this is not tripping a SOQL limit.

spelak-salesforce: For reference, the WHERE clause length is 4,247 using `WHERE IsPersonAccount = true AND AccountId IN ('0018A00000VFVsIQAX','0018A00000VFVsJQAX',...)` as the length of the string.  (Maybe the `WHERE` is not counted?).

Either way, integration tests pass, so it seems we don't need to worry about this limit in query filter characters.

PR #1834, opened 2020-06-24
Allow user to select specific objects to extract

prescod:
# Changes

Add a feature to generate_dataset_mapping to allow users to nominate objects to include in the mapping regardless of our usual heuristics.


cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -61,7 +65,7 @@ def _run_task(self):
 
     def _collect_objects(self):
         """Walk the global describe and identify the sObjects we need to include in a minimal operation."""
-        self.mapping_objects = []
+        self.mapping_objects = self.options["include"]
```
davisagli, line 68: We probably should validate that the specified names are in the describe result so the user doesn't just get a KeyError at https://github.com/SFDO-Tooling/CumulusCI/pull/1834/files#diff-d1647e48e48295d91b884a9874b6ad9fR95 if they made a typo.

prescod: 4349fb8

davisagli:
@prescod This is primarily to provide a way to include standard objects, right? Is it useful even though we're still only including Id, Name, and custom fields for those objects?

prescod:
> @prescod This is primarily to provide a way to include standard objects, right? Is it useful even though we're still only including Id, Name, and custom fields for those objects?

I don't know every detail about DReed's heuristics, but I did an experiment, and I was able to generate some Task fields, some  CustomBrandAsset fields and interestingly, hed_Affiliation__c was discarded by the heuristic but I could force it to output with the option.

So I think this feature is useful and will become more so if I add a feature to let you pick fields to include (which is on my todo list). The 'include' option can also be used as a tie-breaker in output objects that otherwise have no natural order.

I ran cci like this:

cci task run generate_dataset_mapping -o path datasets/extra.yml -o include Task,hed__Affiliation__c,CustomBrandAsset 

Then I did a diff to generate this: 

```
 $ diff datasets/extra.yml datasets/noextras.yml 
1,15d0
< Insert Task:
<   sf_object: Task
<   table: Task
<   fields:
<     Id: sf_id
<     IsRecurrence: IsRecurrence
<     IsReminderSet: IsReminderSet
<     Priority: Priority
<     Status: Status
< Insert CustomBrandAsset:
<   sf_object: CustomBrandAsset
<   table: CustomBrandAsset
<   fields:
<     Id: sf_id
<     AssetCategory: AssetCategory
41,58d25
< Insert hed__Affiliation__c:
<   sf_object: hed__Affiliation__c
<   table: hed__Affiliation__c
<   fields:
<     Id: sf_id
<     hed__Description__c: hed__Description__c
<     hed__EndDate__c: hed__EndDate__c
<     hed__Primary__c: hed__Primary__c
<     hed__Role__c: hed__Role__c
<     hed__StartDate__c: hed__StartDate__c
<     hed__Status__c: hed__Status__c
<   lookups:
<     hed__Account__c:
<       table: Account
<       after: Insert Account
<     hed__Contact__c:
<       table: Contact
<       after: Insert Contact
```

davisagli:
@prescod ah yes, it includes required fields, so that makes more sense. The example I tested with was Account, which doesn't have any required fields other than Name!

PR #1835, opened 2020-06-24
Automate picking the serialization order

prescod:
Background: I find that 90% of the time I choose to break cycles randomly or just pick "Account" because it "seems right". Although we've talked about "smarter" solutions, this small feature at least automates what I personally already do. In the future we can make it smarter based on heuristics or inputs.

# Changes

The generator has new heuristics for picking a serialization order in the face of reference cycles:

1. If one or either direction in a reference loop is nillable, then the system can choose to load the target sObject after the pointing sObject. This means that in Master-Detail relationships, the Parent is loaded first because the Child->Parent relationship is NOT nillable but the Parent->Child relationship (if such a thing exists) is (almost always) nillable.

For example, Location.VistorAddressId is a nillable reference to Address but Address.ParentId is a non-nillable reference to Location. Therefore Location must be loaded first.

2. If both objects are standard objects and one is Account, load Account first.

3. If the `break_cycles` option is set to `auto`, and the rules above do not resolve the issue, the mapping generator will break dependency cycles "randomly" (which is to say the choice mechanism is not documented and may change over time).

Based on testing with Vanilla Salesforce orgs, rule 3 seems to be extremely rarely needed.

Cycles of more than 2 sObjects can also be broken with the principles described above.

# Issues Closed


davidmreed:
I think a couple of simple heuristics will resolve 90% of the cycles without needing random choices:

1. I believe we currently don't privilege Master-Detail relationships over Lookups, and we should. Detail objects cannot be loaded prior to parents, so we should always break cycles in favor of master-detail parents going before master-detail children.
2. If both objects in a relationship are standard and one of them is Account, Account should go first.

prescod:
I don't see anything in the "describe" as obvious as "reltype: 'MasterDetail'". Do you mean that we would check for nillability or cascadedelete or something like that?

I observe the following properties that might be relevant:

cascadeDelete
nillable
permissionable
relationshipOrder
updateable

I guess if the field IS a reference and it IS NOT updateable then you'd probably need the referenced object to be loaded before the referencing object.

So maybe that's what would be the API level feature forcing the ordering?

davisagli:
@prescod according to https://salesforce.stackexchange.com/questions/73405/how-to-identify-relationship-type-lookup-master-detail-from-the-field-api-na/73407 the relationshipOrder property is an integer for master-detail fields but null for lookups.

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -355,3 +390,36 @@ def _is_lookup_to_included_object(self, field):
...
+    def __eq__(self, other: "FieldData"):
+        return self.__dict__ == other.__dict__
+
+
+def is_standard(obj: str):
+    return not obj.endswith("__c")
+
+
+def soft_dependency(sobj, target_obj: str, field_data: FieldData):
+    return field_data.nillable or (sobj == "Account" and is_standard(target_obj))
```
davidmreed, line 410: I think this is pleonastic here. Account doesn't have any non-nillable standard parent relationships other than `OwnerId`, which we ignore.

I previously wanted to propose cycle breaking by making Account go first over any standard object that is acting as its child, through either a standard or custom relationship. Maybe that is more complexity than is needed. Looks like a reasonable way to do something similar would be to do a check in `find_free_object()` that returns Account if it is present in the list. That seems likely to me to produce optimal results the vast majority of the time. 

What do you think?

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -355,3 +390,36 @@ def _is_lookup_to_included_object(self, field):
...
+
+    def __init__(self, describe_data: dict):
+        self.nillable = describe_data.get("nillable")
+
+    def __eq__(self, other: "FieldData"):
+        return self.__dict__ == other.__dict__
+
+
+def is_standard(obj: str):
+    return not obj.endswith("__c")
```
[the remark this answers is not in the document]
davidmreed: (Per below comment, though, I think we can probably remove it instead).

cumulusci/tasks/bulkdata/generate_mapping.py
prescod: Do you have specific cases in mind? 

If I delete the special case and run the code below on a clean org, it generates 800 steps without asking anything. It generates 54 "After" statements and only 2 of them are "after: Insert Account". One is an Account self-reference. The other is NetworkSelfRegistration. So it seems to guess where to put Account okay without any special code. The code does use lexicography sorting so Account ends up early just because of the English alphabet.

When I add some special code it inserts Account earlier but the total number of after: statements increases from 54 to 56.

So is there any particular use-case that I should test with demonstrate the benefit of boosting Account early in the process?

davidmreed: @prescod  I'm fine with the pragmatic approach with no special code. My suggestion was not informed by data - thanks for doing the experiment.

cumulusci/tasks/bulkdata/generate_mapping.py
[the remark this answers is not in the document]
davisagli: https://github.com/SFDO-Tooling/CumulusCI/pull/1835/commits/d6eb1d1faf529097bb8deeb7345daf56d1280223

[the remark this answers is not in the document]
davisagli: https://github.com/SFDO-Tooling/CumulusCI/pull/1835/commits/fa8e1c19b749dae4396f1ffd0295cc7f6c291b97

cumulusci/tasks/bulkdata/tests/test_generatemapping.py
```diff
@@ -748,14 +758,187 @@ def test_split_dependencies__interviews_for_cycles(self, prompt):
...
+                                "Primary_Contact__c": FieldData({"nillable": True})
+                            },
+                        },
+                    },
+                )
             ),
         )
+
+
+@pytest.mark.skip()  # this is slow even with VCR
```
davisagli, line 900: I tried commenting this line to run the tests. The first 2 errored with:

```
E               vcr.errors.CannotOverwriteExistingCassetteException: Can't overwrite existing cassette ('/Users/dglick/Work/CumulusCI/cumulusci/tasks/bulkdata/tests/cassettes/TestIntegrationGenerateMapping.test_simple_generate.yaml') in your current record mode ('none').
E               No match for the request (<Request (GET) https://podname.salesforce.com/services/data/v48.0/sobjects>) was found.
E               No similar requests, that have not been played, found.
```

and the third failed with:

    TypeError: 'method' object is not subscriptable

What is the goal of these tests? Can it be achieved in a way that allows us to run them routinely so they don't end up broken and/or forgotten?

prescod: They are integration tests. They actually run the task against a real org and ensure it doesn't crash or go into an infinite loop. They should be run routinely in CI but not necessarily on every commit that is pushed. Perhaps once a day per open PR? Scheduled github action?

In any case I added some pytest markers, command line flags, comments and documentation to make it all clearer.

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -291,6 +291,49 @@ def _split_dependencies(self, objs, dependencies):
...
+        # peeking into a generator consumes it
+        free_objs = (
+            sobj
+            for sobj in objs_remaining
+            if only_has_soft_dependencies(sobj, dependencies[sobj])
+        )
+        first_free_obj = next(free_objs, None)
+        # first_free_obj = (
+        #     "Account" if "Account" in objs_remaining else next(free_objs, None)
+        # )
```
davisagli, line 305: What's the purpose of this comment? A reminder of an alternative we can try if we need to?

cumulusci/tasks/bulkdata/tests/test_generatemapping.py
davisagli: @prescod Thanks, this makes more sense now that they are identified as integration tests rather than skipped entirely.

Could you please follow up soon with a PR to set up the scheduled github action to run the integration tests regularly?

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -382,3 +425,31 @@ def _is_lookup_to_included_object(self, field):
...
+
+    def __init__(self, describe_data: dict):
+        self.nillable = describe_data.get("nillable", False)
+
+    def __eq__(self, other: "FieldData"):
+        return self.__dict__ == other.__dict__
+
+
+def is_standard(obj: str):
+    return "__" not in obj
```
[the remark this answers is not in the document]
prescod: Removed

cumulusci/tasks/bulkdata/generate_mapping.py
prescod: Removed.

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
```diff
@@ -7,6 +8,12 @@
 
 def pytest_addoption(parser, pluginmanager):
     parser.addoption("--org", action="store", default=None, help="org to use")
+    parser.addoption(
+        "--use-cached-org-output",
```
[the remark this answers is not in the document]
prescod: cea69a7b

cumulusci/tasks/bulkdata/generate_mapping.py
```diff
@@ -34,6 +38,10 @@ class GenerateMapping(BaseSalesforceApiTask):
         "ignore": {
             "description": "Object API names, or fields in Object.Field format, to ignore"
         },
+        "break_cycles": {
+            "description": "If the generator is unsure of the order to load, what to do? "
+            "Set to `ask` (the default) to allow the user to choose or `auto` to pick randomly."
+        },
```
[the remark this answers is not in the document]
prescod: 886fa996f

cumulusci/tasks/bulkdata/tests/test_generatemapping.py
prescod: https://github.com/SFDO-Tooling/CumulusCI/pull/1894

Not tested yet.

PR #1836, opened 2020-06-24
Add cli__show_stacktraces config option

prescod:
# Changes

If you would like to investigate bugs in CumulusCI when you find
them, you can set the config option `show_stacktraces` to `True`
in the `cli` section of `~/.cumulusci/cumulusci.yml` and stacktraces
will no longer be suppressed when they are thrown within CumulusCI.
Usage Errors (wrong command line arguments, missing files, etc.)
will not show you exception tracebacks because they are seldom
helpful in that case.


jstvz:
Yeah, #1136. It was held up while waiting on a Click release that exposed `get_parameter_source`, which was needed to detect whether an option was passed via CLI so that we could set the correct order of precedence.

The basic idea was that we have a naming convention for these options and allow them to be set via yaml, env vars, or the CLI. 

Probably worth shaving that yak sooner or later.

cumulusci/cli/cci.py
```diff
@@ -178,6 +179,8 @@ def new_func(*args, **kw):
     """Run this command for more information about debugging errors: cci error --help"""
 )
 
+UsageErrors = (CumulusCIUsageError, click.UsageError)
```
davisagli, line 182:
```suggestion
USAGE_ERRORS = (CumulusCIUsageError, click.UsageError)
```

to match our convention and make sure it doesn't look like it's a class

cumulusci/cli/cci.py
```diff
@@ -223,9 +227,12 @@ def main(args=None):
...
         except Exception as e:
-            show_debug_info() if debug else handle_exception(
-                e, is_error_command, tempfile_path
-            )
+            if debug:
+                show_debug_info()
+            elif should_show_exceptions and not isinstance(e, UsageErrors):
+                raise
+            else:
+                handle_exception(e, is_error_command, tempfile_path)
```
davisagli, line 235: I think Brandon's intention was that `handle_exception` contains the logic for deciding what to display after an exception. Let's put the new conditional in there.

docs/features.rst
```diff
@@ -761,3 +761,12 @@ The gist command creates a gist comprised of:
 
 The URL for the gist is displayed on the terminal of the user as output, and a web browser will automatically open a tab to the gist.
 
+Seeing Stack Traces Automatically
+---------------------------------
+If you would like to investigate bugs in CumulusCI when you find them, you can set the config 
+option `show_exceptions` to `True` in the `cli` section of `~/.cumulusci/cumulusci.yml` and exceptions will no 
+longer be surpressed when they are thrown within CumulusCI. Usage Errors (wrong command
```
davisagli, line 768: I think the option should be called something like `show_exception_stack_traces`; we are never suppressing the exception message.

Also: spelling of suppressed

prescod: d3b626236dff6c3f40a2e13b9f5115d46607d21a

cumulusci/cli/cci.py
prescod: d3b626236dff6c3f40a2e13b9f5115d46607d21a

PR #1850, opened 2020-06-30
[Robot] Adding in the objectmanager pageobject and a test

rjanjanam:
- ObjectManager page object
- Added Tests to test the functionality of adding a custom field
- Generated keywords file

# Critical Changes

# Changes

# Issues Closed


cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+		self.selenium.click_element(object)
+		self.selenium.wait_until_location_contains("Details/view", timeout=90)
+		self.selenium.click_element(leftnavoption)
+		
+	@capture_screenshot_on_error
+	def create_currency_field(self,field_name):
+		"""Creates a currency field by taking in the field name"""
+		currency_locator=object_manager['input'].format("dtypeC")
+		next_button=object_manager['button'].format("Next")
+		save_button=object_manager['button'].format("Save")
```
boakley, line 48: I see this sort of code repeated in every method. It might be better to just have a "next" and "save" locator, so that you can skip these lines and just do `self.selenium.click_button(object_manager['next'])`. Or, create a custom keyword called `click_action_button` that takes a label and does the lookup (eg: `self.click_action_button('Save')`). 

While that only saves a couple lines of code, when you multiply that times 20+ keywords for all of the different field types, it starts to add up.

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+		self.selenium.wait_until_page_contains_element(currency_locator,timeout=60)
+		self.selenium.click_element(currency_locator)
+		time.sleep(1)
```
boakley, line 51: Can we get rid of this sleep? Can you identify what you are waiting for, and wait for that instead (eg: wait for field to be enabled, etc). If you absolutely must have a sleep, please add a comment explaining why sleep was required.

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+	@capture_screenshot_on_error
+	def create_custom_field(self, **kwargs):
+		"""Ensure that the custom field does not exist prior and Creates a custom field based on type paramenter and the field_name
+		   IF the custom field exists it will not create the custom field and exits out of object manager
+		 """
+		search_button = object_manager['input'].format("globalQuickfind")
+		self.selenium.wait_until_page_contains_element(search_button,60)
+		self.selenium.get_webelement(search_button).send_keys(kwargs['Field_Name'])
+		self.selenium.get_webelement(search_button).send_keys(Keys.ENTER)
+		time.sleep(1)
```
boakley, line 125: same comment about avoiding sleep if possible

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+		self.salesforce.wait_until_loading_is_complete()
+		search_results = object_manager['search_result'].format(kwargs['Field_Name'])
+		count = len(self.selenium.get_webelements(search_results))
+		if not count == 1:
```
boakley, line 129: I think this would be slightly easier to read as `if count != 1`

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+		self.selenium.click_element(next_button)
+		self.selenium.click_element(next_button)
+		self.selenium.click_element(next_button)
+		self.selenium.click_element(save_button)
+		self.selenium.wait_until_location_contains("FieldsAndRelationships/view", timeout=90,
+												   message="Detail page did not load in 1 min")
+
+	@capture_screenshot_on_error
+	def create_custom_field(self, **kwargs):
+		"""Ensure that the custom field does not exist prior and Creates a custom field based on type paramenter and the field_name
```
boakley, line 118: I recommend trying to keep documentation line length to 80 characters or less. Really wide lines are hard to read.

cumulusci/robotframework/tests/salesforce/pageobjects/objectmanager.robot
```diff
@@ -0,0 +1,36 @@
...
+Library         cumulusci.robotframework.PageObjects
+...  ${CURDIR}/ObjectManagerPageObject.py
+
+
+Suite Setup     Run keywords  Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Keywords ***
+Create Customfield In Object Manager
+    [Documentation]        Reads key value pair arguments.
```
boakley, line 13: Personally I feel that long lines are hard to read. What do you think about removing some of the whitespace? Ultimately it's a judgement call so I'll let you have the final word. Here's how I would format this to make it easier to read:

```
Create Custom field In Object Manager
    [Documentation]
    ...  Reads key value pair arguments.
    ...
    ...  Navigates to Object Manager page and load fields and relationships for the specific object
    ...  Runs keyword to create custom field
    ...
    ...  Example:
    ...
    ...  | Create custom field in object manager
    ...  | ...  Object=Payment
    ...  | ...  Field_Type=Formula
    ...  | ...  Field_Name=Is Opportunity From Prior Year
    ...  | ...  Formula=YEAR( npe01__Opportunity__r.CloseDate ) < YEAR( npe01__Payment_Date__c )
```

cumulusci/robotframework/tests/salesforce/pageobjects/objectmanager.robot
```diff
@@ -0,0 +1,36 @@
...
+    ...                            Object=Payment
+    ...                            Field_Type=Formula
+    ...                            Field_Name=Is Opportunity From Prior Year
+    ...                            Formula=YEAR( npe01__Opportunity__r.CloseDate ) < YEAR( npe01__Payment_Date__c )
+    [Arguments]            &{fields}
+    Load Page Object                                     Custom                           ObjectManager
+    Open Fields And Relationships                        &{fields}[Object]
+    Create Custom Field                                  &{fields}
+
+*** Test Cases ***
```
boakley, line 26: I would like to see one test case that does something like `Go to page  Object Manager  Contact  Detail` and then calls `Current page should be  Object Manager  Contact  Detail` as a sort of smoke test.

cumulusci/robotframework/tests/salesforce/pageobjects/objectmanager.robot
```diff
@@ -0,0 +1,36 @@
...
+
+Create Custom Lookup Field Using Object Manager
+    [Documentation]     To test the ability of creating a custom lookup field
+
+    Create Customfield In Object Manager
+    ...                                                    Object=Contact
+    ...                                                    Field_Type=Lookup
+    ...                                                    Field_Name=Last Soft Credit Opportunity
+    ...                                                    Related_To=Opportunity
+
```
boakley, line 36: Is there a way to add a verification step that makes sure the custom field actually got added?

rjanjanam: Added a verification step

cumulusci/robotframework/tests/salesforce/pageobjects/objectmanager.robot
rjanjanam: Added

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
rjanjanam: Done

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+				  "button-with-text":"//button[contains(text(),'{}')]",
+				  "frame_new":"//iframe[contains(@name, '{}') or contains(@title, '{}')]"
+				  }
+
+
+@pageobject("Custom", "ObjectManager")
+class ObjectManagerPage(BasePage):
+	
+	@capture_screenshot_on_error
+	def open_fields_and_relationships(self, object_name):
```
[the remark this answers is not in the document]
rjanjanam: Implemented as suggested

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+		formula_locator = object_manager['input'].format("dtypeZ")
+		next_button = object_manager['button'].format("Next")
+		save_button = object_manager['button'].format("Save")
+		checkbox_option = object_manager['input'].format("fdtypeB")
+		formula_txtarea = object_manager['formula_txtarea'].format("CalculatedFormula")
+		check_syntax = object_manager['button'].format("Check Syntax")
+		
+		self.selenium.wait_until_page_contains_element(formula_locator,60)
+		self.selenium.click_element(formula_locator)
+		time.sleep(1)
```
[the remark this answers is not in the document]
rjanjanam: I tried removing lot of sleeps and instead replaced them with wait for a specific button. But there were only two scenarios were sleep for 1 sec is required and being used.

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,144 @@
...
+		related = object_manager['select_related'].format("DomainEnumOrId")
+		self.selenium.wait_until_page_contains_element(lookup_locator,60)
+		self.selenium.click_element(lookup_locator)
+		time.sleep(1)
+		self.selenium.click_element(next_button)
+		self.selenium.wait_until_page_contains_element(related,60)
+		self.selenium.scroll_element_into_view(related)
+		self.selenium.get_webelement(related).click()
+		self.selenium.click_element(option)
+		time.sleep(2)
```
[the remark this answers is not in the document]
rjanjanam: removed this

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
rjanjanam: Moved all the common locators and declared above for common reusability

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
rjanjanam: removed it

cumulusci/robotframework/tests/salesforce/pageobjects/ObjectManagerPageObject.py
rjanjanam: removed it

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -28,13 +28,42 @@ class ObjectManagerPage(BasePage):
         | Go to page   ObjectManager  Contact
         """
 
+    global search_button
```
boakley, line 31: You can delete all of these `global` statements if you move the definition of the variables outside of the class. For example:

```
object_manager = {
    "button": "//input[@title='{}']",
   ...
}
search_button = object_manager["input"].format("globalQuickfind")
currency_locator = object_manager["input"].format("dtypeC")

```

However, I don't see the point of both having the locators in a dictionary and then creating variables from the dictionary. Instead of creating the dictionary you can just define the shortcuts to have the xpath and  eliminate the need of having both the locators and shortcuts.

```
search_button = "//input[@title='globalQuickfind']"
currency_locator = "////input[@id='dtypeC']"
...
```
This might be a good subject to talk about in one of the robot checkins - what is the preferred way to reference locators in a page object? Maybe your solution of using individual variables is good. Though, if we ever make it easier to re-use locators between page objects, having a single dictionary would make that easier. 

I don't want this issue to prevent this code from getting merged, but at the very least you should remove all of the global statements, and then move the definition of the variables outside of the class.

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -50,6 +79,7 @@ def _is_current_page(self):
 
     @capture_screenshot_on_error
     def switch_tab_to(self, tab):
+        """Switches tab to a specified tab"""
```
boakley, line 82: This document isn't quite enough if all you have to go on is the description (ie: if you're reading it on a web page of keyword definitions). You should describe what the "tab" argument is. Is it a number? The label of the tab? A tab locator? It would also be good to include an example. 

You can include an example by using a vertical bar at the start of a line to create block of preformatted code. For example:
```
""" blah blah blah

Example:

| switch tab to  Fields & Relationships
```
You can also use a | to create a table if the line also ends with a |

```
| switch to tab  | Fields & Relationships |
```

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -82,15 +109,30 @@ def create_currency_field(self, field_name):
             message="Fields And Relationships page did not load in 1 min",
         )
 
+    @capture_screenshot_on_error
+    def create_text_field(self, field_name):
+        """Creates a text field by taking in the field name"""
+        self.selenium.wait_until_page_contains_element(text_locator, timeout=60)
+        self.selenium.click_element(text_locator)
+        self.selenium.wait_until_page_contains_element(next_button, 60)
```
boakley, line 117: I don't think this wait is necessary, is it? Since you're already waiting for the text field, isn't it safe to assume that if the text field is on the page, so is the button?

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,239 @@
...
+        search_results = object_manager["search_result"].format(field_name)
+        assert (
+            len(self.selenium.get_webelements(search_results)) == 1
+        ), "Expected result count to be 1 but found {}".format(
+            len(self.selenium.get_webelements(search_results))
+        )
+
+    @capture_screenshot_on_error
+    def delete_custom_field(self):
+        """Deletes the added custom field"""
```
boakley, line 201: There's not enough documentation for this keyword. It's not clear how or when you would use this. How does it know which field to delete? If this is only ever called from a certain page or dialog, you might want to mention that.

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,239 @@
...
+        self.selenium.click_element(actions_menu)
+        self.selenium.wait_until_page_contains_element(action_item_delete, 60)
+        self.selenium.click_element(action_item_delete)
+        self.selenium.wait_until_page_contains_element(confirm_delete, 60)
+        self.selenium.click_element(confirm_delete)
+        self.selenium.wait_until_location_contains("/view", timeout=90)
+
+    @capture_screenshot_on_error
+    def create_custom_field(self, **kwargs):
+        """Ensure that the custom field does not exist prior and Creates a custom field based
```
boakley, line 211: This description doesn't sound quite right. If the field already exists, what does this keyword do? Will it throw an error, or will it silently return since the field already exists? The documentation implies it throws an error ("Ensures that...") but it looks like it just returns if it already exists.

Would it be better to say "this creates a custom field if one doesn't already exist by the given name"?

This documentation should also include an example, since it's not clear just from the documentation what "kwargs" should be. An example would quickly clear up any misunderstanding.

Maybe something like:

```
Example:

| Create custom field
| ...  Object=Payment
| ...  Field_Type=Formula
| ...  Field_Name=Is Opportunity From Prior Year
| ...  Formula=YEAR( npe01__Opportunity__r.CloseDate ) < YEAR( npe01__Payment_Date__c )

File: cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py [hunk 17]
```diff
@@ -0,0 +1,239 @@
...
+        on type paramenter and the field_name if the custom field exists it will not create the
```
boakley, line 212: you mention `field_name` here, but in a test you wrote you use `Field_Name`. Does case matter here?

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: updated

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: update to be uniform with the convention

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: Added

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: removed the wait

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: updated

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: For now, changed as per your suggestion

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,230 @@
...
+class ObjectManagerPage(BasePage):
+    """A page object representing the Object Manager of an object.
+        Example
+        | Go to page   ObjectManager  Contact
+        """
+
+    def _go_to_page(self):
+        url_template = "{root}/lightning/setup/ObjectManager/home"
+        url = url_template.format(
+            root=self.cumulusci.org.lightning_base_url, object_name=self.object_name
```
davisagli, line 47: No need to include object_name here, since it's not part of url_template

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
```diff
@@ -0,0 +1,230 @@
...
+        self.selenium.click_element(next_button)
+        self.selenium.click_element(save_button)
+        self.selenium.wait_until_location_contains(
+            "FieldsAndRelationships/view",
+            timeout=90,
+            message="Detail page did not load in 1 min",
+        )
+
+    def create_lookup_field(self, field_name, related):
+        """Creates a Lookpup field by taking in the inputs field_name and related field"""
```
davisagli, line 142:
```suggestion
        """Creates a Lookup field by taking in the inputs field_name and related field"""
```

cumulusci/robotframework/tests/salesforce/pageobjects/objectmanager.robot
```diff
@@ -0,0 +1,74 @@
+*** Settings ***
+
+Resource        cumulusci/robotframework/Salesforce.robot
+Library         cumulusci.robotframework.PageObjects
+Suite Setup     Run keywords  Open Test Browser
+Suite Teardown  Delete Records and Close Browser
+
+*** Keywords ***
+Create Customfield In Object Manager
```
davisagli, line 9:
```suggestion
Create Custom Field In Object Manager
```

(and then the lines which use the keyword also need to be updated)

cumulusci/robotframework/pageobjects/ObjectManagerPageObject.py
rjanjanam: Agree, done

davisagli: @rjanjanam It's still here -- do you need to push a commit?

cumulusci/robotframework/tests/salesforce/pageobjects/objectmanager.robot
davisagli: @rjanjanam don't you also need to update lines 41, 54, and 67 which use this keyword? Actually it's weird that the tests passed -- @boakley can you make sure they are included in CI?

davisagli: The tests are running -- I guess the keyword is found even though it is spelled with different whitespace. I'll fix the spaces and then approve this.

PR #1853, opened 2020-07-02
Include package install links in release notes

jlantz:
Example output: https://github.com/SFDO-Tooling/CCI-Food-Bank/releases/tag/beta%2F1.6-Beta_6

# Critical Changes

# Changes

* The `github_release_notes` supports the `version_id` option for passing in the PackageVersionId.  If passed, a section for installation links to install the package is automatically added to the release notes.  This functionality is auto-configured by default in the `release_beta` and `release_production` flows
* The `git -> release_notes -> parsers` section in `cumulusci.yml` now supports the ability to remove a parser from a step by setting its `class_path` to `None`

# Issues Closed


cumulusci/tasks/release_notes/parser.py
```diff
@@ -283,3 +284,35 @@ def _add_issue_comment(self, issue):
...
+        "https://test.salesforce.com/packaging/installPackage.apexp?p0={}"
+    )
+
+    def __init__(self, release_notes_generator, title):
+        super().__init__(title)
+        self.release_notes_generator = release_notes_generator
+
+    def parse(self, change_note):
+        # There's no need to parse lines, this parser gets its values from task options
+        pass
```
davisagli, line 304: What about this scenario:
1. release notes are generated during the release_beta flow, including the installation links
2. a release engineer manually runs the task to update the release notes, without specifying the version_id option.

In this case, shouldn't we preserve the existing content rather than removing it?

cumulusci/cumulusci.yml
```diff
@@ -1000,6 +1002,10 @@ project:
                 5:
                     class_path: cumulusci.tasks.release_notes.parser.GithubLinesParser
                     title: Deleted Metadata
+                6:
+                    class_path: cumulusci.tasks.release_notes.parser.InstallLinkParser
+                    title: Installation Links
+                    skip_parent_pr: True
```
davisagli, line 1008: Why is this needed? There won't be a version_id when parent release notes are generated, so the InstallLinkParser won't produce any content.

cumulusci/tasks/release_notes/parser.py
```diff
@@ -283,3 +284,35 @@ def _add_issue_comment(self, issue):
...
+
+
+class InstallLinkParser(BaseChangeNotesParser):
+    prod_url_template = (
+        "https://login.salesforce.com/packaging/installPackage.apexp?p0={}"
+    )
+
+    test_url_template = (
+        "https://test.salesforce.com/packaging/installPackage.apexp?p0={}"
+    )
```
davisagli, line 296: We should use PROD_LOGIN_URL and SANDBOX_LOGIN_URL from cumulusci.oauth.salesforce, and then the URLs will be generated correctly for non-prod environments too.

cumulusci/tasks/release_notes/parser.py
```diff
@@ -283,3 +284,35 @@ def _add_issue_comment(self, issue):
...
+
+    def parse(self, change_note):
+        # There's no need to parse lines, this parser gets its values from task options
+        pass
+
+    def render(self):
+        pv_id = self.release_notes_generator.version_id
+        if pv_id is not None:
+            pv_id = urllib.parse.quote_plus(pv_id)
+            return "\n\r".join(
```
davisagli, line 310:
```suggestion
            return "\r\n".join(
```

(but why are we using Windows line endings anyway?)

cumulusci/tasks/release_notes/tests/test_parser.py
```diff
@@ -549,3 +550,26 @@ def _create_expected_render(self, issue_number, issue_title, link_pr):
         if link_pr:
             render += " [[PR{}]({})]".format(self.pr_number, self.pr_url)
         return render
+
+
+class TestGithubLinesParser(unittest.TestCase):
```
davisagli, line 555:
```suggestion
class TestInstallLinkParser:
```

davisagli: Please make sure your editor is set up to run flake8 for Python files. It should have warned you that you were re-defining an existing class.

cumulusci/tasks/release_notes/tests/test_parser.py
```diff
@@ -549,3 +550,26 @@ def _create_expected_render(self, issue_number, issue_title, link_pr):
...
+    def setUp(self):
+        self.title = "Title"
+
+    def test_no_package_version(self):
+        generator = mock.Mock(link_pr=True)
+        generator.version_id = None
+        parser = InstallLinkParser(generator, self.title)
+        parser.parse("abc")
+        self.assertEqual(None, parser.render())
```
davisagli, line 564:
```suggestion
        assert parser.render() is None
```

cumulusci/cumulusci.yml
jlantz: I don't remember clearly, but vaguely remember it was something to do with how the two tasks are initializing the generator and parsers.  There's likely a better way to implement this.

cumulusci/tasks/release_notes/parser.py
jlantz: Good point.  Yes, we should preserve the install links section if it's already there and no version id is specified

cumulusci/tasks/release_notes/parser.py
jlantz: I honestly don't remember as these parsers were written years ago.  We could test not using Windows line endings and see if GitHub handles them the way we expect.

PR #1856, opened 2020-07-06
Feature/tooling api

prescod:
# Changes

`cci org shell` now has access to the `tooling` API through the keyword `tooling`

### DIFF
--- a/cumulusci/tasks/salesforce/BaseSalesforceApiTask.py
+++ b/cumulusci/tasks/salesforce/BaseSalesforceApiTask.py
@@ -12,15 +12,16 @@ class BaseSalesforceApiTask(BaseSalesforceTask):
     def _init_task(self):
         self.sf = self._init_api()
         self.bulk = self._init_bulk()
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/salesforce/BaseSalesforceApiTask.py
```diff
@@ -12,15 +12,16 @@ class BaseSalesforceApiTask(BaseSalesforceTask):
     def _init_task(self):
         self.sf = self._init_api()
         self.bulk = self._init_bulk()
-        self.tooling = self._init_api("tooling/")
+        self.tooling = self._init_api("tooling")
         self._init_class()
 
-    def _init_api(self, base_url=None):
+    def _init_api(self, api_name=None):
```
davisagli, line 18: Can we keep calling this base_url throughout? I find that a more helpful reminder of what it does, and that way we don't have to check and make sure no one is calling this as a named argument.

PR #1857, opened 2020-07-07
Add OWD and Communities preflight checks

davidmreed:
# Critical Changes

# Changes

- A `check_communities_enabled` preflight check allows installers to determine if Communities can be created.
- A `check_sobjects_available` preflight check allows installers to determine if specific sObjects exist.
- A `check_org_wide_defaults` preflight check allows installers to determine if Org-Wide Defaults are set to expected values.

# Issues Closed


cumulusci/tasks/salesforce/communities_preflights.py
```diff
@@ -0,0 +1,21 @@
+import requests
```
davisagli, line 1: What would you think about organizing tasks that are primarily meant as preflight checks (i.e. gathering info rather than mutating the org) into a cumulusci.tasks.preflight package, rather than continuing to grow the already-large cumulusci.tasks.salesforce?

cumulusci/tasks/salesforce/communities_preflights.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def _run_task(self):
+        s = requests.Session()
+        s.get(self.org_config.start_url).raise_for_status()
+        r = s.get(
+            "{}/sites/servlet.SitePrerequisiteServlet".format(
+                self.org_config.instance_url
+            )
+        )
+        self.return_values = r.status_code == 200
```
davisagli, line 16: What status code is returned if communities are not enabled? Should we distinguish between that specific status code and other codes like 503 which could indicate inability to complete the check?

cumulusci/cumulusci.yml
```diff
@@ -165,6 +169,10 @@ tasks:
         description: Retrieves a list of the currently available Permission Set License definition keys
         class_path: cumulusci.tasks.salesforce.license_preflights.GetAvailablePermissionSetLicenses
         group: Salesforce Preflight Checks
+    get_is_communities_enabled:
```
davisagli, line 172: This is awkward wording. Maybe `check_communities_enabled`?

cumulusci/tasks/salesforce/sobject_preflights.py
```diff
@@ -0,0 +1,78 @@
...
+                else ""
+            )
+            self.return_values = all(
+                self._check_owds(
+                    obj,
+                    self.sf.query(
+                        f"SELECT InternalSharingModel{ext} "
+                        "FROM EntityDefinition "
+                        f"WHERE QualifiedApiName = '{obj}'"
+                    )["records"][0],
```
davisagli, line 69: Is it possible to do a single query for all the api names?

cumulusci/tasks/salesforce/sobject_preflights.py
```diff
@@ -0,0 +1,78 @@
...
+                )
+                for obj in self.owds.keys()
+            )
```
davisagli, line 72: Can you give me a sense of how this check will be used? Would it make sense for it to return more detailed information instead of just a boolean so that preflight checks can show more specific messages about what is wrong?

cumulusci/tasks/salesforce/communities_preflights.py
davidmreed: I'm in favor of the idea. I didn't do it here because I wanted to keep the scope of the PR limited (e.g., regression-testing to make sure existing tasks aren't impacted in MetaDeploy after reorganization). Would you be comfortable with creating the package here with these new tasks and doing the reorganization as a separate PR?

davisagli: Yep.

cumulusci/tasks/salesforce/sobject_preflights.py
davidmreed: Seems you can! The Tooling API is often fussy about multi-row queries, but it works on this entity.

cumulusci/tasks/salesforce/sobject_preflights.py
davidmreed: For target product, we need to check whether the OWDs of two specific Outbound Funds objects are set to Private. If and only if they're set to Private, we can deploy a Sharing Set to customize sharing for the Community, but we won't set the OWDs if the user has configured them otherwise.

I think in many cases it will make more sense to direct the user to a configuration guide section, since there may be a large number of differences between the org shape and the target configuration. Is there a way for the message of a preflight check to include the literal return value from the task?

cumulusci/tasks/salesforce/communities_preflights.py
davidmreed: It returns a 500 when Communities are not enabled, but that could potentially mean other things too (this API is not documented).

As far as we know, any status code other than 200 means that a first Community-create call won't be successful - this is the endpoint that the UI calls when an org's first Community creation is initiated, and it continues to return 200 for subsequent calls.

I might have found a better proxy for Community availability, though. It looks like the `Network` sObject doesn't get added to the schema until Communities are explicitly turned on, even if the licenses are available in the org. I'll revise to use that strategy instead.

cumulusci/tasks/salesforce/sobject_preflights.py
davisagli: > Is there a way for the message of a preflight check to include the literal return value from the task?

Not currently. And that would probably make it more difficult to handle translations for preflight check messages. (Which, to be fair, already needs work -- I think we are currently extracting them for translation but not actually applying those translations when we display the messages.)

cumulusci/tasks/salesforce/communities_preflights.py
davidmreed: Canceling the above; I went ahead and did the whole move, leaving behind compatibility aliases (eyes appreciated - I _think_ I hit what's needed).

davisagli: I immediately hit an import cycle (trying to import GetInstalledPackages imports BaseSalesforceApiTask from the salesforce package, which itself tries to import GetInstalledPackages), so I bit the bullet and rewrote cumulusci.tasks.salesforce to import things lazily. @davidmreed @prescod want to see if you can break it?

cumulusci/tasks/preflight/sobjects.py
```diff
@@ -0,0 +1,101 @@
...
+            "required": True,
+        }
+    }
+
+    def _run_task(self):
+        all_objects = {entry["name"] for entry in self.sf.describe()["sobjects"]}
+        self.return_values = all(
+            sobject in all_objects
+            for sobject in process_list_arg(self.options["sobjects"])
+        )
```
davisagli, line 23: I'd suggest setting return_values to a list of all the sobjects. That way the return value can be cached and used for multiple preflight checks (e.g. `when: "Network" not in tasks.check_sobjects_available()`). If multiple checks need to call the task with different options, it can't be cached.

davidmreed: Done!

cumulusci/tasks/salesforce/communities_preflights.py
prescod: You've updated it since I started reviewing this morning and addressed most of my concerns.

I guess the only remaining one is whether this mechanism should move to a Utility class to be used anywhere we need to do lazy module loading (either for backwards compatibility or just because we want shorter moodule paths)

davidmreed:
@davisagli  I did find one import that works on master and does not work in this refactor - 

    from cumulusci.tasks.salesforce.GetInstalledPackages import GetInstalledPackages

It throws

> ImportError: bad magic number in 'cumulusci.tasks.salesforce.GetInstalledPackages': b'\x03\xf3\r\n'

davisagli:
@davidmreed I think it's finding an old .pyc file which it doesn't like -- but also I forgot to re-add the GetInstalledPackages module alias. Good catch!

davisagli:
> I guess the only remaining one is whether this mechanism should move to a Utility class to be used anywhere we need to do lazy module loading (either for backwards compatibility or just because we want shorter moodule paths)

@prescod probably, but I figured that could wait until there are two locations that use it.

PR #1858, opened 2020-07-08
Feature/robot lint defaults

boakley:
This change explicitly sets the ignore/warn/error status of every rule for the robot_lint task, rather than relying on rflint defaults. These defaults were provided to us by QE.

There is also a change in how the `list` option is processed. Prior to this change it would only ever list rflint defaults. With this change it now displays our own defaults along with any command line options. It can now accurately represents the error level of each rule as if you were running the task. 

# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/robotframework/lint_defaults.args
```diff
@@ -0,0 +1,25 @@
...
+--error InvalidTableInResource
+--error PeriodInSuiteName
+--error PeriodInTestName
+--error RequireKeywordDocumentation
+--error RequireTestDocumentation
+--error TagWithSpaces
+--error TooFewKeywordSteps
+--error TooFewTestSteps
+--error TrailingBlankLines
+--error TrailingWhitespace
```
davisagli, line 25: Please either rename this to use a .txt extension or add *.args to MANIFEST.in, to make sure the file gets included in the Python distribution.

boakley: I've renamed it to .txt. Thanks for catching this!

PR #1859, opened 2020-07-09
Add record_count func and make table format the default

prescod:
# Changes

Create a function for interactive use called record_count() which outputs an estimate of a count of records.

Table format is the default for query(), describe(), record_count()


cumulusci/cli/cci.py
```diff
@@ -1285,6 +1286,7 @@ def org_shell(runtime, org_name, script=None, python=None):
         "help": CCIHelp(),
         "query": sf_helpers.query,
         "describe": sf_helpers.describe,
+        "record_count": sf_helpers.record_count,
     }
```
davisagli, line 1290: Not a blocker for this PR, but I'm starting to think we should attach all this stuff to SimpleSalesforceUIHelpers, rename it to something like ShellKeywords, and use its `__dict__` as the globals.

cumulusci/cli/ui.py
```diff
@@ -243,18 +244,52 @@ def describe(self, sobj_name, detailed=False, format="pprint"):
...
+    def record_count(self, *sobject_list: str, format="table"):
+        """Get approximate record counts for sObjects
+
+        If the approximate count is 0, this function will return nothing.
+        If the sObject doesn't exist in the org, the same is true
+        (as per the Salesforce REST API)
+
+        Arguments:
+
+            sobject_list - list of sobjects to display, or a single sObject
```
davisagli, line 263: It looks like it's also possible to get the counts for all objects by not specifying any sobject_list, but I don't see that mentioned here.

cumulusci/cli/ui.py
```diff
@@ -243,18 +244,52 @@ def describe(self, sobj_name, detailed=False, format="pprint"):
...
-            rc = data
+        if not detailed:
+            data = [_summarize(field) for field in data["fields"]]
+
+        return self._format(data, format)
+
+    def record_count(self, *sobject_list: str, format="table"):
+        """Get approximate record counts for sObjects
+
+        If the approximate count is 0, this function will return nothing.
```
davisagli, line 257: What does "nothing" mean? None? Not included in the result list?

I think including it with a value of 0 would be more intuitive.

cumulusci/cli/ui.py
```diff
@@ -243,18 +244,52 @@ def describe(self, sobj_name, detailed=False, format="pprint"):
...
+            """
+        if sobject_list:
+            params = {"sObjects": ",".join(sobject_list)}
+
         else:
-            rc = dict(_summarize(field) for field in data["fields"])
+            params = None
+        data = self._sf.restful("limits/recordCount", params=params)["sObjects"]
+
+        data = sorted(data, key=lambda v: -v["count"])
```
davisagli, line 281:
```suggestion
        data = sorted(data, key=lambda v: v["count"], reverse=True)
```

davisagli:
@prescod waiting for you to respond to feedback on this one.

prescod:
It seemed like it was low-urgency and I wanted to think through the refactoring you suggested and also a deeper refactoring that you had suggested in a releng meeting.

davisagli:
@prescod yeah, no worries, just checking on status while I went through the PR list this morning

jstvz:

PR #1870, opened 2020-07-14
Add delete task to cumulusci.yml

prescod:
# Changes

Add a task called delete_data to cumulusci. Previously this task was unavailable for command line use.


cumulusci/cumulusci.yml
```diff
@@ -84,6 +84,10 @@ tasks:
             path: src
             revert_path: src.orig
         group: Salesforce Metadata
+    delete_data:
+        description: Delete matching records in an org
+        class_path: cumulusci.tasks.bulkdata.DeleteData
+        group: Test Data
```
davisagli, line 90:
```suggestion
    delete_dataset:
        description: Delete matching records in an org
        class_path: cumulusci.tasks.bulkdata.DeleteData
        group: 'Data Operations'
```

for consistency with the other bulkdata tasks

davisagli:
Could you also add a section about this to docs/bulk_data.rst explaining and showing how to use it?

cumulusci/cumulusci.yml
prescod: I agree with the changed group but I think that delete_data is a better name than delete_dataset.

This task is different than the _dataset tasks because it doesn't have anything to do with the files in the datasets/ directory.

davisagli: I kind of buy that argument, but also think it's likely to contribute to people remembering it wrong and getting a "did you mean delete_data?" message. Any preference here, @davidmreed?

prescod:
> Could you also add a section about this to docs/bulk_data.rst explaining and showing how to use it?

I started writing that but it was so repetitive with the generated documentation that I started to question whether it was wise and helpful. I'd suggest adding a paragraph saying that the feature is available but not documenting the individual options.

cumulusci/cumulusci.yml
davidmreed: I think I'm with Paul on this.

davisagli:
@prescod I had this Hub comment on my mind when asking for docs: https://powerofus.force.com/s/feed/0D51E00005MFFPMSA5 (see first comment on my post). Our users want examples.

prescod:
> @prescod I had this Hub comment on my mind when asking for docs: https://powerofus.force.com/s/feed/0D51E00005MFFPMSA5 (see first comment on my post). Our users want examples.

I agree with the goal, but the examples should probably end up in the Task Reference, which means that they should start out in the Python code. do you agree?

For complex tasks that relate to each other (extract_dataset and load_dataset), the tutorial information would be more important. For this one, I think that bolstering the "task info" would be a better approach than having two different places to go for information. If you used "cci task info" to learn about the task you wouldn't know that you should go to a website to find examples.

davisagli:
@prescod I think my recommendation would be to mention the availability of the task in the narrative bulkdata task docs, show one straightforward example, and link to the task reference for more details on the options.

PR #1871, opened 2020-07-14
week 29 dependency updates

davisagli:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/.github/workflows/release_test.yml
+++ b/.github/workflows/release_test.yml
@@ -7,7 +7,7 @@ on:
     - master
 
 env:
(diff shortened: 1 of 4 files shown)


davisagli:
@boakley well I'm still trying to get the tests to pass (so you'll need to re-approve), but yes they are

davisagli:
@boakley we're in the phase where the CumulusCI-Test packaging org has upgraded to the new release but not all instances have, so we need to be sure to use a prerelease scratch org for testing installation of the package. But for some reason the `pub` package dependency is failing to install on some instances.

PR #1872, opened 2020-07-15
Use working directory if provided

prescod:
# Changes

generate_and_load and generate_and_load_from_yaml tasks now respect the working_directory option and store their tempfiles in the working directory. This is primarily helpful for debugging.

debug_dir was another name for working_directory and has been removed.


davisagli:
@prescod what is the distinction between debug_dir and working_directory? Do we need both?

PR #1875, opened 2020-07-16
Normalized Option Syntax

Br4nd0R:
# Changes
* CumulusCI now accepts a normalized task option syntax in the form of: `--opt-name value`. This can be used in place of the old task option syntax: `-o opt-name value`


cumulusci/cli/cci.py
```diff
@@ -1240,83 +1240,133 @@ def task_info(runtime, task_name):
...
+                RUNTIME.alert(f"Task complete: {task_name}")
+
+        return click.Command(task_name, params=params, callback=run_task)
+
+    def resolve_command(self, ctx, args):
+        """We override this method to allow us access to the actual
+        command line args being passed in. This allows us to convert
+        the old option syntax to the new option syntax"""
+
+        # TODO: What's a good way to account for scenarios where
```
prescod, line 1299: Don't for get this todo.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+            # Overwrite any existing configured options with those
+            # being passed in as long as their value != None.
+            # Values come in as None when they are specified via a cumulusci.yml file
+            for name, value in kwargs.items():
+                if value and name in task_options.keys():
+
+                    # Add options if they aren't present on this task
+                    # they are defined on the parent class
+                    if "options" not in task_config.config:
+                        task_config.config["options"] = {}
```
davisagli, line 1464: Not sure I follow why you moved this into the loop. It's just to make sure the task config has an options dict. If the merged task config has no options set, there's no harm in adding options as an empty dict.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+                    task_config.config["options"][name] = value
+
+            try:
+                task = task_class(
+                    task_config.project_config, task_config, org_config=org_config
                 )
 
-            # Override the option in the task config
-            task_config.config["options"][name] = value
+                if kwargs.pop("debug_before", None):
```
davisagli, line 1472:
```suggestion
                if kwargs.get("debug_before"):
```

I don't see a reason to remove it from kwargs

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+                    self._import_pdb_and_set_trace()
```
davisagli, line 1473: Let's keep the call to pdb inline. It'll make it easier to see where the execution is at from the pdb prompt; i.e. you can see whether it's before or after calling the task, rather than just being dropped in the _import_pdb_and_set_trace function and having to figure out where it was called from.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+
+        Raises:
+            CumulusCIUsageError: if there is an error in parsing the old option syntax,
+            or the option doesn't exist for the given task.
+        """
+        duplicate = self._has_duplicate_options(args)
+        if duplicate:
+            raise CumulusCIUsageError(
+                f"Found duplicate option `{duplicate}` in given command:\n cci task run {' '.join(args)}"
+            )
```
davisagli, line 1514: Good call to check for this. Can we phrase the error in a way that suggests what the user should do next rather than just telling them it didn't work? (i.e. "found duplicate option {duplicate}. please make sure you only specify it once")

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+
+        task_name = args[0]
+        while "-o" in args:
+            idx = args.index("-o")
+            opt_name = args[idx + 1]
+            opt_value = args[idx + 2]
```
davisagli, line 1520: I tried running the util_sleep task with just `-o seconds` (i.e. forgot the value), and got an IndexError here.

davisagli: I notice you have code for looping through the args and identifying the options both here and in _parse_option_names. Why not have one method which does the parsing and yields name/value pairs, and another which iterates through that, identifies duplicates (by adding the names to a set and checking for existence in that set), and adds the modified args to a new list.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+
+        return opt_names
+
+    def _option_in_task(self, opt_name, task_name):
+        """
+        Returns True if opt_name is the name of an
+        option defined in the given task, else False.
+        """
+        task = RUNTIME.project_config.get_task(task_name)
+        task_class = import_global(task.config["class_path"])
```
davisagli, line 1571: Do this in _convert_old_option_syntax and pass the task_class into _option_in_task as an arg so it doesn't need to be imported multiple times.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+
+        if "task_options" not in task_class.__dict__.keys():
+            task_option_names = self._get_task_options_in_hierarchy(task_class)
```
davisagli, line 1574: You don't have to do this. `task_class.task_options` will inherit it from the task class's base classes if it's not defined on task_class itself.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+            f"Unable to find `task_options` in base classes:\n{task_class.__bases__}"
+        )
+
+    def _get_click_options_for_task(self, task_options):
+        """
+        Given a dict of options in a task, constructs and returns the
+        corresponding list of click.Option instances
+        """
+        click_options = []
+        if task_options:
```
davisagli, line 1605: I don't think this conditional is needed; we should be able to rely on task_options always being a dict.

cumulusci/cli/cci.py
```diff
@@ -1425,83 +1425,235 @@ def task_info(runtime, task_name):
...
+            for name, properties in task_options.items():
+                # NOTE: When task options aren't explicitly given via the command line
+                # click complains that there are no values for options. We set required=False
+                # to mitigate this error. Task option validation should be performed at the
+                # task level via task._validate_options() or Pydantic models.
+                click_options.append(
+                    click.Option(
+                        param_decls=(f"--{name}",),
+                        required=False,  # don't enforce option values in Click
+                        help=properties.pop("description", ""),
```
davisagli, line 1615: Don't use pop! This is probably mutating the task config within the actual project_config!

cumulusci/cli/tests/test_run_task.py
```diff
@@ -0,0 +1,343 @@
...
+            }
+        },
+    )
+    return runtime
+
+
+def test_task_run(runtime):
+    DummyTask._run_task = Mock()
+    multi_cmd = cci.RunTaskCommand()
+    with patch("cumulusci.cli.cci.RUNTIME", runtime):
```
davisagli, line 38: You could make the `runtime` fixture take care of doing this patching.

cumulusci/cli/tests/test_cci.py
```diff
@@ -1813,55 +1792,10 @@ def test_task_info(self, doc_task, rst2ansi):
...
-                org=None,
-                o=[("bogus", "blue")],
-                debug=False,
-                debug_before=False,
-                debug_after=False,
-                no_prompt=True,
-            )
-
     @mock.patch("pdb.set_trace")
     def test_task_run_debug_before(self, set_trace):
```
davisagli, line 1796: It looks like you re-created some of these in test_run_task.py and they can probably be removed from here?

davisagli (changes requested):

cumulusci/cli/cci.py
Br4nd0R: @davisagli - If I move the call to `import_global()` to `_convert_old_option_syntax()` how do I get the value for `task_class` back to the `get_command()` function?

cumulusci/cli/tests/test_run_task.py
Br4nd0R: if I set the end of the `runtime` fixture to:
```
with patch('cumulusci.cli.cci.RUNTIME', runtime):
    return runtime
```
I'm not seeing the patch take place when I execute the test. Anything wrong with this approach?

cumulusci/cli/utils.py
```diff
@@ -0,0 +1,28 @@
...
+        else runtime.universal_config.list_tasks()
+    )
+
+
+def get_available_flows(runtime):
+    return (
+        runtime.project_config.list_flows()
+        if runtime.project_config is not None
+        else runtime.universal_config.list_flows()
+    )
```
davisagli, line 17: get_available_tasks and get_available_flows look like they would make sense as methods of the Runtime class, rather than standalone functions.

cumulusci/cli/cci.py
davisagli: @Br4nd0R Both of them have access to `ctx`, which is a context object provided by click. We are allowed to store whatever we want in `ctx.obj`

cumulusci/cli/tests/test_run_task.py
davisagli: @Br4nd0R You need to do `yield runtime` instead of `return runtime` so that the patch context manager exits after pytest is done using the fixture, rather than immediately before returning runtime.

Br4nd0R:
@davisagli - Fingers crossed, this should be ready to go.

cumulusci/cli/cci.py
```diff
@@ -1449,83 +1442,178 @@ def task_info(runtime, task_name):
...
-)
-@click.option(
-    "--no-prompt",
-    is_flag=True,
-    help="Disables all prompts.  Set for non-interactive mode use such as calling from scripts or CI systems",
-)
-@pass_runtime(require_keychain=True)
-def task_run(runtime, task_name, org, o, debug, debug_before, debug_after, no_prompt):
+class RunTaskCommand(click.MultiCommand):
+    # options hat are not task specific
```
davisagli, line 1446:
```suggestion
    # options that are not task specific
```

cumulusci/cli/cci.py
```diff
@@ -1449,83 +1442,178 @@ def task_info(runtime, task_name):
...
+    global_options = {
+        "org": {
+            "help": "Specify the target org. By default, runs against the current default org.",
+            "is_flag": True,
```
davisagli, line 1450: I guess this is overwritten on line 1604, but it's not a boolean flag.

cumulusci/cli/cci.py
```diff
@@ -1449,83 +1442,178 @@ def task_info(runtime, task_name):
...
+                    required=False,  # don't enforce option values in Click
+                    help=properties.get("description", ""),
+                )
+            )
+        return click_options
+
+    def _get_default_command_options(self, is_salesforce_task):
+        click_options = []
+        for opt_name, config in self.global_options.items():
+            if opt_name == "org":
```
davisagli, line 1593: I think you can just leave `org` out of global_options, and include its help inline below. It doesn't need to be in global_options to be excluded on line 1499, because it already got popped out of kwargs on line 1493

cumulusci/cli/tests/test_run_task.py
```diff
@@ -0,0 +1,156 @@
...
+    "dummy-derived-task": {
+        "class_path": "cumulusci.cli.tests.test_run_task.DummyDerivedTask"
+    },
+}
+
+
+@pytest.fixture
+def ctx():
+    ctx = Mock()
+    ctx.task_class = DummyTask
```
davisagli, line 27: No longer used, right?

cumulusci/cli/tests/test_run_task.py
```diff
@@ -0,0 +1,156 @@
...
+        assert 4 == echo.call_count
+        assert 0 == len(runtime.universal_config.method_calls)
+
+
+def test_format_help__proj_conf_does_not_exist(runtime):
+    with patch("cumulusci.cli.cci.click.echo") as echo:
+        runtime.universal_config = runtime.project_config
+        runtime.project_config = None
+        RunTaskCommand().format_help(Mock(), Mock())
+        assert 4 == echo.call_count
```
davisagli, line 108: Probably don't need these 2 different tests for format_help now that get_available_tasks takes care of the difference between having a project config or not?

cumulusci/cli/cci.py
Br4nd0R: I'll remove it.

davisagli (approved):
Looking good! 💥

PR #1876, opened 2020-07-16
Config and org config cleanup

prescod:
# Changes

Internal changes:

org_configs now know what keychain they came from and whether they were global or not.

org_configs now save back to the directory they came from. Previously a local definition would be created which would supersede the global one.

The gist of how they do this is that when they are loaded from disk they are "tagged" either GlobalOrg or LocalOrg using a NamedTuple based on where they are loaded. Then, later, after they are decrypted, the tag becomes a property on the OrgConfig object.

The global/local state is not stored redundantly in the encrypted files.

OrgConfig.safe_if_changed now centralizes the logic for whether to save orgs after refreshing.


cumulusci/cli/tests/test_cci.py
```diff
@@ -893,19 +893,27 @@ def test_org_connect(self, oauth):
...
+                org_name="test",
+                sandbox=False,
+                login_url="https://login.salesforce.com",
+                default=True,
+                global_org=False,
+            )
 
         runtime.check_org_overwrite.assert_called_once()
-        runtime.keychain.set_org.assert_called_once()
-        org_config = runtime.keychain.set_org.call_args[0][0]
```
davisagli, line 908: Was there something wrong with this way of doing it? The new code is more complicated and doesn't actually assert that the org was stored in the expected keychain. (But I suppose that is probably covered by the tests of OrgConfig itself...)

cumulusci/core/config/BaseConfig.py
```diff
@@ -2,7 +2,12 @@
 
 
 class BaseConfig(object):
-    """ BaseConfig provides a common interface for nested access for all Config objects in CCI. """
+    """ BaseConfig provides a common interface for nested access for all Config objects in CCI.
+
+    As of 2020 there is a set() method for setting values. Please use that
+    instead of config.config["foo"] = "bar". Directly manipulating the
+    internal config dict will eventually be phased out.
```
davisagli, line 9: I'm not convinced yet:
1. It invents a new non-standard Python protocol for setting values, so it's one more thing consumers of BaseConfig need to get familiar with. Also it looks too much like the `set()` builtin for constructing a set object.
2. This docstring says to use set(), but the docstring for set() mentions a situation where you can't use set().
3. I don't see updates elsewhere to actually use this method.
4. We were brainstorming possible larger changes to config objects such as trying to use pydantic or switching to item-style access. I'd like to have consensus on our end goal before starting to make smaller changes.

I think we should keep this change out of this PR and talk about it separately.

cumulusci/core/flowrunner.py
```diff
@@ -604,7 +604,7 @@ def _init_org(self):
 
         if self.org_config.config != orig_config:
             self.logger.info("Org info has changed, updating org in keychain")
-            self.project_config.keychain.set_org(self.org_config)
+            self.org_config.save()
```
davisagli, line 607: Looks like this could be updated to use OrgConfig._save_if_changed

cumulusci/core/keychain/BaseEncryptedProjectKeychain.py
```diff
@@ -46,6 +46,8 @@ def _set_encrypted_service(self, service, encrypted, project):
         self.services[service] = encrypted
 
     def _set_org(self, org_config, global_org):
+        org_config.keychain = self
+        org_config.global_org = global_org
```
davisagli, line 50: Perhaps we should have some sanity-check assertions to make sure that these are not already set to different values.

cumulusci/core/keychain/BaseProjectKeychain.py
```diff
@@ -98,8 +98,8 @@ def create_scratch_org(self, org_name, config_name, days=None, set_password=True
         scratch_config[
             "sfdx_alias"
         ] = f"{self.project_config.project__name}__{org_name}"
-        org_config = ScratchOrgConfig(scratch_config, org_name)
-        self.set_org(org_config)
+        org_config = ScratchOrgConfig(scratch_config, org_name, self, global_org=False)
```
davisagli, line 101:
```suggestion
        org_config = ScratchOrgConfig(scratch_config, org_name, keychain=self, global_org=False)
```

cumulusci/core/keychain/EnvironmentProjectKeychain.py
```diff
@@ -44,10 +44,12 @@ def _load_orgs(self):
                 org_name = key[len(self.org_var_prefix) :].lower()
                 if org_config.get("scratch"):
                     self.orgs[org_name] = scratch_org_factory(
-                        json.loads(value), org_name
+                        json.loads(value), org_name, self, global_org=False
```
davisagli, line 47:
```suggestion
                        json.loads(value), org_name, keychain=self, global_org=False
```

cumulusci/core/keychain/EnvironmentProjectKeychain.py
```diff
@@ -44,10 +44,12 @@ def _load_orgs(self):
...
                     )
                 else:
-                    self.orgs[org_name] = OrgConfig(json.loads(value), org_name)
+                    self.orgs[org_name] = OrgConfig(
+                        json.loads(value), org_name, self, global_org=False
```
davisagli, line 51:
```suggestion
                        json.loads(value), org_name, keychain=self, global_org=False
```

cumulusci/core/keychain/__init__.py
```diff
@@ -13,6 +13,14 @@
...
+    EncryptedFileProjectKeychain,
+)
+
+shut_up_flake8 = (
+    BaseProjectKeychain,
+    DEFAULT_CONNECTED_APP,
+    BaseEncryptedProjectKeychain,
+    EnvironmentProjectKeychain,
     EncryptedFileProjectKeychain,
 )
```
davisagli, line 26: I prefer to do this by setting `__all__`.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -1,8 +1,10 @@
 import os
```
davisagli, line 1: Regarding renaming this module, have you checked to make sure no one is doing `from cumulusci.core.keychain.EncryptedFileProjectKeychain import EncryptedFileProjectKeychain`?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -33,7 +35,7 @@ def _load_files(self, dirname, extension, key):
                 name = item.replace(extension, "")
                 if key not in self.config:
                     self.config[key] = {}
-                self.config[key][name] = config
+                self.config[key][name] = constructor(config)
```
davisagli, line 38: I would make constructor default to None, and then do

```suggestion
                self.config[key][name] = constructor(config) if constructor else config
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -107,3 +114,24 @@ def _raise_service_not_configured(self, name):
...
             f"Maybe you need to run: cci service connect {name}"
         )
+
+    def _get_org(self, name):
+        org = self._decrypt_config(
+            OrgConfig,
+            self.orgs[name].encrypted_data,
+            extra=[name],
+            context=f"org config ({name})",
+        )
```
davisagli, line 124: It doesn't look like _decrypt_config attaches the keychain to the org config that it constructs.

cumulusci/core/config/OrgConfig.py
```diff
@@ -23,15 +24,18 @@ class OrgConfig(BaseConfig):
...
+
         self.name = name
         self._community_info_cache = {}
         self._client = None
         self._latest_api_version = None
         self._installed_packages = None
         super(OrgConfig, self).__init__(config)
 
-    def refresh_oauth_token(self, keychain, connected_app=None):
+    def refresh_oauth_token(self, keychain, connected_app=None, save=False):
```
davisagli, line 38: It would be a cleaner separation of concerns to keep the save logic out of `refresh_oauth_token`. `_save_changes_if` could be renamed to `save_if_changed` and used like this in cases where saving is desired:

    with org_config.save_if_changed():
        org_config.refresh_oauth_token(...)

cumulusci/tasks/salesforce/tests/test_base_tasks.py
```diff
@@ -35,7 +35,11 @@ def test_run_task(self):
...
     def test_update_credentials(self):
-        def update_config(keychain):
+        update_config_called = False
+
+        def update_config(keychain, save=False):
+            nonlocal update_config_called
+            update_config_called = True
             self.org_config.config["new"] = "new"
 
         self.org_config.refresh_oauth_token = update_config
```
davisagli, line 45: You could potentially use a mock with `side_effect` here to update the config without needing your own logic for keeping track of whether it was called.

cumulusci/tests/util.py
```diff
@@ -44,12 +44,12 @@ def __init__(
 
 
 class DummyOrgConfig(OrgConfig):
-    def __init__(self, config=None, name=None):
+    def __init__(self, config=None, name=None, keychain=None):
```
davisagli, line 47: For completeness, let's make sure this has `global_org` too.

cumulusci/cli/cci.py
```diff
@@ -903,7 +903,7 @@ def org_connect(runtime, org_name, sandbox, login_url, default, global_org):
         scope="web full refresh_token",
     )
     oauth_dict = oauth_capture()
-    org_config = OrgConfig(oauth_dict, org_name)
+    org_config = OrgConfig(oauth_dict, org_name, runtime.keychain, global_org)
```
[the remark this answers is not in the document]
prescod: Good catch: 

8e13c470

cumulusci/core/tests/test_config.py
```diff
@@ -1438,3 +1438,11 @@ def test_has_minimum_package_version(self):
...
         with self.assertRaises(CumulusCIException):
             config.has_minimum_package_version("GW_Volunteers", "1.0")
+
+    def test_set_value(self):
+        config = OrgConfig({}, "test")
+        config.set("Foo", "Bar")
+        assert config.Foo == "Bar"
+
+        with pytest.raises(AssertionError):
+            config.set("Foo__Bar", "5")
```
davisagli, line 1448: Need to remove this too

cumulusci/tasks/salesforce/tests/test_base_tasks.py
prescod: I would probably have done that if I could do the assignment in a lambda, but since I still need to write a full function for the assignment, I think the number of lines of code would be similar.

cumulusci/tests/util.py
prescod: 7e25e3cd

cumulusci/core/keychain/encrypted_file_project_keychain.py
davisagli: I searched github and only see uses of cumulusci.core.keychain.EncryptedFileProjectKeychain, so I think this is probably okay.

cumulusci/core/keychain/encrypted_file_project_keychain.py
davisagli: Okay, I think you're right -- but wouldn't it be cleaner to attach the keychain right when the org config is constructed? That way if org config constructors ever need to do something with the keychain, they can.

cumulusci/tasks/salesforce/tests/test_base_tasks.py
davisagli: I started to make that change to see how it would work out, then realized there's a more significant problem: the test used to verify that running the task saved the org config in the keychain, but now it only asserts that refresh_oauth_token was called.

cumulusci/core/keychain/encrypted_file_project_keychain.py
prescod: Okay, take a look at: c074c621

I left code in BaseProjectKeychain because not every subclass may implement _get_org in a way that sets the keychain. To make the keychain 100% reliably available for org constructors, we'd need to make it mandatory in the initializer which would be a breaking change.

cumulusci/tasks/salesforce/tests/test_base_tasks.py
prescod: Okay I added back that assertion: 546545b5

cumulusci/core/keychain/BaseEncryptedProjectKeychain.py
```diff
@@ -46,6 +46,11 @@ def _set_encrypted_service(self, service, encrypted, project):
         self.services[service] = encrypted
 
     def _set_org(self, org_config, global_org):
+        if org_config.keychain:
+            assert org_config.keychain == self
```
davisagli, line 50:
```suggestion
            assert org_config.keychain is self
```

cumulusci/core/keychain/BaseProjectKeychain.py
```diff
@@ -176,15 +178,18 @@ def unset_default_org(self):
...
         sfdx("force:config:set defaultusername=")
 
     def get_org(self, name):
         """ retrieve an org configuration by name key """
         if name not in self.orgs:
             self._raise_org_not_found(name)
         org = self._get_org(name)
-        org.keychain = self
+        if org.keychain:
+            assert org.keychain == self
```
davisagli, line 190:
```suggestion
            assert org.keychain is self
```

PR #1878, opened 2020-07-17
Org cache directories

prescod:
# Internal Changes

Create a mechanism for tasks and other code to cache data that is org-specific.

Create code to automatically garbage-collect these caches when orgs are removed.

# User-visible Changes

Change the disable_tdtm_trigger_handlers (SetTDTMHandlerStatus) task so that trigger handler state is remembered in the cache directory instead of REPO_ROOT as it previously did.

Updates to the options:
```
-o restore_file RESTOREFILE
	 *Optional*

	 Path to the state file to store or restore the current trigger handler state. 
Set to False to discard trigger state information. 
By default the state is cached in an org-specific directory for later restore.

-o restore RESTORE
	 *Optional*

	 If True, restore the state of Trigger Handlers to that stored in the (specified or default) restore file.
```

# Context

Using the SetTDTMHandlerStatus task as a "demo" for the caching mechanism allows me to get this cache mechanism into review separate from the more complex tasks that it will also be used with (e.g. generate_dataset_mapping).

@davidmreed  you might want to review just the SetTDTMHandlerStatus class in case you want to express an opinion about how I changed the semantics of it with respect to where the restore file is managed. Or at least read the updates above.


davidmreed:
The changes to the trigger state tasks look fine to me, @prescod.

cumulusci/core/config/OrgConfig.py
```diff
@@ -273,3 +274,18 @@ def reset_installed_packages(self):
...
+        instance_url = self.config.get("instance_url", "")
+        return urlparse(instance_url).hostname or ""
+
+    def get_orginfo_cache_dir(self):
+        assert self.keychain, "Keychain should be set"
+        if self.global_org:
+            cache_dir = self.keychain.config_local_dir / "orginfo" / self.get_domain()
+        else:
+            cache_dir = self.keychain.project_cache_dir / "orginfo" / self.get_domain()
+        assert cache_dir
```
davisagli, line 288: What would cause this to be false?

cumulusci/core/config/project_config.py
```diff
@@ -835,3 +835,7 @@ def construct_subproject_config(self, **kwargs):
     def relpath(self, path):
         """Convert path to be relative to the project repo root."""
         return os.path.relpath(os.path.join(self.repo_root, path))
+
+    @property
+    def project_cache_dir(self):
+        return Path(self.repo_root, ".cci")
```
davisagli, line 841: There are a couple of existing bits of code that have ".cci" hardcoded -- let's go ahead and update them to use this.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -126,6 +126,10 @@ def _get_org(self, name):
             org.global_org = True
         return org
 
+    @property
+    def project_cache_dir(self):
+        return self.project_config.project_cache_dir
```
davisagli, line 131: What about other keychains? Is it intentional for code that uses `keychain.project_cache_dir` to break if running with EnvironmentProjectKeychain or a keychain that stores org configs as Django models?

prescod: What do you think about if get_orginfo_cache_dir had this:

```
    def get_orginfo_cache_dir(self):
        assert self.keychain, "Keychain should be set"
        if self.global_org or not hasattr(self.keychain, "project_cache_dir"):
            config_dir = self.keychain.config_local_dir
        else:
            config_dir = self.keychain.project_cache_dir

        cache_dir = config_dir / "orginfo" / self.get_domain()
```

davisagli: @prescod So the proposal is to store in the global location (~/.cumulusci) if the keychain doesn't support a project-specific cache?

I think that makes sense. But what about giving BaseProjectKeychain an implementation of `project_cache_dir` that simply returns config_local_dir, and then we don't need this hasattr check?

Btw, do you think we should we use the pyfilesystem API for interactions with the cache, so that it's easier to switch to alternate storage if we want something that can be made to survive across resumed builds?

cumulusci/core/config/project_config.py
prescod: I'll submit a PR for that but it balloons.

"change this to a call to the project_config.project_cache_dir"

Now the tests need project_config.repo_root to be set, or some of them fail.

The "right" way to set that is with a fixture I created a few weeks ago.

But fixtures can only be used in UnitTest-style tests.

And UnitTest-style tests have a different assertion syntax.

davisagli: You mean pytest fixtures can only be used in test classes that don't extend unittest.TestCase.

cumulusci/core/config/OrgConfig.py
prescod: Removed

cumulusci/core/keychain/encrypted_file_project_keychain.py
prescod: With respect to project_cache_dir, I did that in c2429df

With respect to PyFilesystem. Yeah, I guess that makes sense. I'll try to do that in the next several days.

prescod:
Okay @davisagli, you suggested I add support for storing the cached data in virtual file systems and I did that. I suspect I will find that the internals need to evolve when I add support for S3 (because the abstractions underneath are probably slightly leaky as I've already discovered with Windows support) but the API used by consumers should not need to change.

prescod:
You can see how the whole thing is intended to be used here:

https://github.com/SFDO-Tooling/CumulusCI/pull/1878/files#diff-5f699bf0a5eb7c06f9c51ec680ba92e4R54

One thing that has changed since you last looked is that there are now multiple conceptually distinct cache so that it is now possible e.g. to open the "trigger handler cache" (named `cumulusci.tasks.salesforce.trigger_handlers`) separate from the describe cache (which might be named `cumulusci.utils.org_schema` or something like that). This is a proactive measure to protect performance by not uploading and downloading irrelevant files if they are remote.

cumulusci/tasks/salesforce/tests/util.py
```diff
@@ -27,3 +33,43 @@ def create_task(task_class, options=None, project_config=None, org_config=None):
...
+    to_patch = {
+        "cumulusci.core.config.UniversalConfig.cumulusci_config_dir": temp_root
+        / "fixture_userhome/.cumulusci",
+        "cumulusci.tests.util.DummyKeychain.global_config_dir": temp_root
+        / "fixture_userhome/.cumulusci",
+        "cumulusci.core.config.project_config.BaseProjectConfig.project_cache_dir": temp_root
+        / "fixture_userhome/project_home/.cci",
+        "cumulusci.tests.util.DummyKeychain.project_cache_dir": temp_root
+        / "fixture_userhome/project_home/.cci",
+    }
```
davisagli, line 61: Is it really necessary to patch all these? I think most would follow naturally if we patch Path.home and make sure the project config's repo_root is set? With this approach I'm afraid we'll be forever tracking down things that didn't get patched and are accidentally writing to the user's real homedir...

cumulusci/utils/fileutils.py
```diff
@@ -86,6 +92,206 @@ def load_from_source(
...
+    """Cleanup directories that are not associated with a connected/live org."""
+    domains = set()
+    for org in keychain.list_orgs():
+        org_config = keychain.get_org(org)
+        domain = org_config.get_domain()
+        if domain:
+            domains.add(domain)
+
+    assert project_config.project_cache_dir
+    assert keychain.global_config_dir
```
davisagli, line 105: These could be the same when running `cci org list` outside of a project directory, right? So probably should add the glob results to a set to avoid processing twice.

davisagli: Also I'd be inclined to put this function in cumulusci.core somewhere. cumulusci.utils mostly has standalone util functions that don't have a dependency on cumulusci domain objects

cumulusci/utils/fileutils.py
```diff
@@ -86,6 +92,206 @@ def load_from_source(
...
+
+
+class FSResource:
+    """Generalization of pathlib.Path to support S3, FTP, etc
+
+    Create them through the open_fs_resource module function or static
+    function which will create a context manager that generates an FSResource.
+
+    If you don't need the resource management aspects of the context manager,
+    you can call the `new()` classmethod."""
```
davisagli, line 131: What's the reason for inventing a new way to construct it instead of using `__init__`?

cumulusci/utils/fileutils.py
```diff
@@ -86,6 +92,206 @@ def load_from_source(
...
+        try:
+            yield resource
+        finally:
+            filesystem.close()
+
+
+open_fs_resource = FSResource.open_fs_resource
+
 if __name__ == "__main__":  # pragma: no cover
     import doctest
```
davisagli, line 297: Do the doctests also get discovered and run by pytest?

cumulusci/core/config/OrgConfig.py
```diff
@@ -276,6 +278,22 @@ def save(self):
...
+        return urlparse(instance_url).hostname or ""
+
+    def get_orginfo_cache_dir(self, cachename):
+        "Returns a context managed FSResource object"
+        assert self.keychain, "Keychain should be set"
+        if self.global_org:
+            cache_dir = self.keychain.global_config_dir
+        else:
+            cache_dir = self.keychain.project_cache_dir
+        cache_dir = cache_dir / "orginfo" / self.get_domain() / cachename
```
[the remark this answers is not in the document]
prescod: Do users share repo directories?

The issue with ~/.cumulusci/ProjectName is that we have no mechanism for cleaning up when repos are deleted from the filesystem.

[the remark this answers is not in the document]
prescod: 0ee2aa3

cumulusci/core/config/BaseTaskFlowConfig.py
```diff
@@ -61,3 +63,11 @@ def get_suggested_name(self, name, steps):
             return f'. Did you mean "{match_list[0]}"?'
         else:
             return ""
+
+    @contextmanager
+    def open_project_cache(self, cache_name):
```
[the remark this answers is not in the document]
prescod: open_project_cache is a wrapper around project_cache_dir, which we named a month ago.

I think of it as "the cache for the project". In some contexts (e.g. MetaCI flow runner) there is only one project in scope so it can go essentially anywhere writeable, but it is still the cache for the project.

cumulusci/core/source/github.py
```diff
@@ -107,6 +112,8 @@ def fetch(self, path=None):
                 shutil.rmtree(path)
                 raise
 
+        assert path.is_dir()
```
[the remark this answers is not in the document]
prescod: We only extract if it doesn't exist. So there is nothing to test in that case.

cumulusci/utils/fileutils.py
prescod: Because `__init__`'ing the object is easy but discouraged. I put a small amount of effort in front of it so you're more likely to read the docs before you do it. And raising an exception in `__init__` increases the likelihood of you reading the docs and using the class correctly.

Some libraries rename the class (e.g. lxml.etree._Element) but then isinstance checks and repr get ugly. I thought that this is a better approach.

prescod: What I'm trying to avoid is the situation we're in with SQL Alchemy/SQLite where we should have managed the engine as a resource but it wasn't obvious that we should have done so until we had tons of code that doesn't do so.

cumulusci/core/keychain/BaseProjectKeychain.py
```diff
@@ -269,3 +270,8 @@ def list_services(self):
         services = list(self.services.keys())
         services.sort()
         return services
+
+    @property
+    def project_cache_dir(self):
+        "Helper function to get the project_cache_dir from the project_config"
+        return self.project_config.universal_config_obj.cumulusci_config_dir
```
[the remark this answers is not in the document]
prescod: 9cb8b1b0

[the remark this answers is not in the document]
prescod: ab7fe3e874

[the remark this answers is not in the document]
prescod: On our infinite todo list it might be nice to add some kind of tests for CCI which would emulate things that would break when it is used in MetaCI and MetaDeploy, because I don't have those use cases internalized at all. In other words: the invariants they expect should be in the CCI test suite so that running the suite would give some confidence that they aren't broken.

cumulusci/core/config/project_config.py
```diff
@@ -822,3 +822,10 @@ def construct_subproject_config(self, **kwargs):
     def relpath(self, path):
         """Convert path to be relative to the project repo root."""
         return os.path.relpath(os.path.join(self.repo_root, path))
+
+    @property
+    def project_cache_dir(self):
+        "A project cache which is on the local filesystem. Prefer open_project_cache where possible."
+        project_cache_dir = Path(self.repo_root, ".cci")
+        project_cache_dir.mkdir(exist_ok=True)
+        return project_cache_dir
```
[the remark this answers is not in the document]
prescod: I think that we decided that this is "good enough" unless you can suggest a way to get the "real" repo_root.

cumulusci/tasks/salesforce/tests/util.py
prescod: I pared it back somewhat but DummyKeychain doesn't inherit from real keychain and it broke things when I tried to make it do so.

cumulusci/utils/fileutils.py
prescod: Moved. 216b7482

Also renamed BaseProjectConfig.py so I could do mocking.

cumulusci/utils/fileutils.py
prescod: Not automatically but I have a method called `test_docstrings` in `cumulusci/utils/tests/test_fileutils.py` which does it (for this module).

cumulusci/tasks/salesforce/tests/test_sourcetracking.py
```diff
@@ -219,7 +221,10 @@ def test_run_task(self, sfdx):
...
-    def test_freeze(self):
-        task = create_task(SnapshotChanges)
+    def test_freeze(self, create_task_fixture):
+        task = create_task_fixture(SnapshotChanges)
         steps = task.freeze(None)
-        self.assertEqual([], steps)
+        assert steps == []
+
+
+flake8 = (create_task_fixture,)
```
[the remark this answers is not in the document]
prescod: f19422f8

cumulusci/core/config/BaseTaskFlowConfig.py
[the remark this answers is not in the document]
davisagli: But a BaseTaskFlowConfig is not guaranteed to be used in the context of a project.

This is more of a theoretical concern because the `cci task run` and `cci flow run` commands do currently require being used in the context of a project. But I think that's more of an accident than  intentional design. I could imagine wanting to support running tasks/flows defined in the universal config against a global org without being in a project directory.

cumulusci/core/config/project_config.py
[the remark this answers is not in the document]
davisagli: Yep. Just wanted to flag it as a limitation of the caching to keep in mind.

cumulusci/tasks/salesforce/tests/util.py
davisagli: DummyKeychain could implement global_config_dir and project_cache_dir as properties that make use of Path.home()

cumulusci/utils/fileutils.py
davisagli: You mean BaseProjectKeychain. BaseProjectConfig was renamed to project_config quite a while ago.

cumulusci/utils/fileutils.py
davisagli: Okay, you convinced me.

cumulusci/tasks/salesforce/tests/util.py
prescod: 7b98b4fd

cumulusci/core/config/BaseTaskFlowConfig.py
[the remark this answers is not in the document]
prescod: Do you prefer just open_cache ? Not sure what descriptor to put between "open" and "cache".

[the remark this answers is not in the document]
davisagli: I like that. I think the config object you're calling it on implies something about the intended scope of the cache.

cumulusci/core/config/project_config.py
```diff
@@ -266,15 +266,15 @@ def repo_root(self):
...
             if (path / ".git").is_dir():
                 return str(path)
-            head, tail = os.path.split(path)
-            if not tail:
+            parent = path.parent
+            if parent == path:
                 # reached the root
                 break
-            path = Path(head)
+            path = parent
```
davisagli, line 277: The docs for PurePath.parent say: "If you want to walk an arbitrary filesystem path upwards, it is recommended to first call Path.resolve() so as to resolve symlinks and eliminate “..” components."

cumulusci/core/config/universal_config.py
```diff
@@ -82,3 +82,7 @@ def _load_config(self):
                 "global_config": UniversalConfig.config_global,
             }
         )
+
+    @property
+    def cache_dir(self):
+        return self.config_global_local_path
```
davisagli, line 88: @prescod config_global_local_path was renamed to config_global_path as part of the universal/global/local clarification. And be aware that this is one of the ones that MetaCI sets to None: https://github.com/SFDO-Tooling/MetaCI/blob/master/metaci/cumulusci/config.py#L52

cumulusci/core/config/project_config.py
prescod: 6fad035

cumulusci/core/config/universal_config.py
prescod: This is one of those things that makes my head hurt. What is the circumstance where this code gets invoked and how can I emulate or trigger it?

We look at the universal config if we don't have a project config.

But the universal config is a subclass of BaseTaskFlowConfig, so its presumably intended to be used when invoking Tasks or Flows. But as of August 2020, Tasks and Flows always run within the context of Projects...

Does this method exist just to future-proof in case we one-day allow tasks and flows to run outside of projects?

Or is there such a context right now? MetaDeploy or Metecho or ...

prescod: I'm leaning towards just deleting this method unless you know of a case where the UniversalConfig is the project_config in a context that needs access to the describe schema or other info we might cache about orgs.

prescod: And then open_cache would move into project_config and cache_dirs would be a project_config-only feature.

davisagli: @prescod I'm fine with making it a project_config-only feature. I'm pretty sure that there isn't currently a place where tasks and flows can be run without having a project_config as context (actually I'm quite certain, since project_config is the first thing passed when instantiating a task).

In theory we could make it so that tasks and flows can be run outside the context of a project if they are defined in universal or global config and interact with no org or a global org. But it's probably cleanest to keep this out of UniversalConfig until we're actually considering that possibility closely.

prescod: b3adc74

PR #1879, opened 2020-07-18
Rename global config

prescod:
# Internal Changes

Previously CumulusCI used the word "Global" to mean two different things. The "GlobalConfig" ships with CumulusCI. It can be overrriden by a "Local" Config which goes in the directory ~/.cumulusci.

Orgs that go in that directory are called "Global" orgs. "Local" orgs go in the project directory.

This checkin moves us to a more consistent naming scheme:

Universal -> Ships with CumulusCI. Applies "universally" to all CumulusCI iinstallations.

Global -> Applies to all projects on a single computer.

Local -> Applies to this project.

The Universal adage is Think Globally! Act Locally!


cumulusci/core/config/project_config.py
```diff
@@ -139,11 +139,11 @@ def _validate_package_api_format(self):
 
     @property
     def config_global_local(self):
-        return self.global_config_obj.config_global_local
+        return self.universal_config_obj.config_global_local
 
     @property
     def config_global(self):
-        return self.global_config_obj.config_global
+        return self.universal_config_obj.config_global
```
davisagli, line 146: I think we should also rename `config_global_local` to `config_global`, and `config_global` to `config_universal`, for consistency. It's not backwards-compatible, but should be a pretty easy adjustment to make in downstream code.

davisagli: And actually, given this and other subtle backwards incompatibilities, I'm now thinking maybe it's better to _not_ keep the backwards-compatible BaseGlobalConfig alias.

cumulusci/core/config/universal_config.py
```diff
@@ -0,0 +1,84 @@
...
+__location__ = os.path.dirname(os.path.realpath(__file__))
+
+
+class UniversalConfig(BaseTaskFlowConfig):
+    """ Base class for the global config which contains all configuration not specific to projects """
+
+    config = None
+    config_filename = "cumulusci.yml"
+    project_config_class = BaseProjectConfig
+
```
davisagli, line 19: @prescod btw I got rid of `global_config_dir = ".cumulusci"` here since it was unused.

cumulusci/core/config/project_config.py
prescod: Can you confirm that the names in quotes here are not really used for anything and can/should also be changed?

https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/core/config/project_config.py#L117

prescod: Or if they are used for something, you know what to change in the other apps?

prescod: Just used for error reporting it seems...

PR #1881, opened 2020-07-20
Feature/robot pabot

boakley:
This PR adds a `processes` option to the robot task. If the value is > 1, the task will use pabot to run test suites in parallel using the given number of processes.

At last weeks robot checkin there was a discussion about how to solve the problem that tests are taking a very long time to complete, which is starting to hinder some teams. One of several items discussed was the possibility to run our robot tests in parallel. Just about the only tool available off the shelf today is [pabot](https://github.com/mkorpela/pabot).

QE can't just download and use pabot themselves since they use a cci task rather than the robot command directly. It was decided at the meeting that we could try adding an option to the robot task to use pabot to run the tests. This PR is the result. I've tested it with the CumulusCI robot tests but haven't tested it with the tests of other teams.

There are a couple of pabot options that aren't supported, and I haven't mentioned this new task option in the readthedocs documentation. It may not be worth a lot of engineering time to implement additional features until we can determine if these changes add a measurable improvement, or if it just gives us a different set of problems due to concurrency issues. 

Approval for including pabot: [GUS Third Party Software approval](https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qB0000000ZJvFIAW/view)

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -40,7 +40,6 @@ class Robot(BaseSalesforceTask):
         },
         "processes": {
             "description": "*experimental* Number of processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robotframework. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
-            "type": "integer",
             "default": "1",
         },
     }
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -59,15 +59,10 @@ class Robot(BaseSalesforceTask):
             self.options["options"] = {}
 
         # processes needs to be an integer.
-        if "processes" in self.options:
-            try:
-                self.options["processes"] = int(self.options["processes"])
-            except ValueError:
-                raise TaskOptionsError(
-                    f"'processes' value '{self.options['processes']}' is not an integer"
-                )
-        else:
-            self.options["processes"] = 1
+        try:
+            self.options["processes"] = int(self.options.get("processes", 1))
+        except (TypeError, ValueError):
+            raise TaskOptionsError("Please specify an integer for the `processes` option.")
 
         # There are potentially many robot options that are or could
         # be lists, but the only one we currently care about is the
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -39,7 +39,7 @@ class Robot(BaseSalesforceTask):
             "description": "If true, enable the `breakpoint` keyword to enable the robot debugger"
         },
         "processes": {
-            "description": "*zexperimental* Number of parallel processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robot. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
+            "description": "*experimental* Number of processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robotframework. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
             "type": "integer",
             "default": "1",
         },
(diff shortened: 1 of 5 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -37,6 +38,11 @@ class Robot(BaseSalesforceTask):
         "debug": {
             "description": "If true, enable the `breakpoint` keyword to enable the robot debugger"
         },
+        "processes": {
+            "description": "*zexperimental* Number of parallel processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robot. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
```
davisagli, line 42:
```suggestion
            "description": "*experimental* Number of processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robotframework. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
```

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -37,6 +38,11 @@ class Robot(BaseSalesforceTask):
...
+            "type": "integer",
```
davisagli, line 43:
```suggestion
```

CumulusCI does not currently do anything with this. We want to add declarative option type validation but it will be configured differently.

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -52,6 +58,17 @@ def _init_options(self, kwargs):
...
+        # processes needs to be an integer.
+        if "processes" in self.options:
+            try:
+                self.options["processes"] = int(self.options["processes"])
+            except ValueError:
+                raise TaskOptionsError(
+                    f"'processes' value '{self.options['processes']}' is not an integer"
+                )
+        else:
+            self.options["processes"] = 1
```
davisagli, line 70:
```suggestion
        try:
            self.options["processes"] = int(self.options.get("processes", 1))
        except (TypeError, ValueError):
            raise TaskOptionsError("Please specify an integer for the `processes` option.")
```

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -83,7 +100,29 @@ def _run_task(self):
...
+                "pabot.pabot",
+                "--pabotlib",
+                "--processes",
+                str(self.options["processes"]),
+            ]
+            # We need to convert options to their commandline equivalent
+            for option, value in options.items():
+                if isinstance(value, list):
+                    for item in value:
+                        cmd.extend([f"--{option}", str(item)])
```
davisagli, line 116:
```suggestion
                        cmd.extend(["--{option}", str(item)])
```

Only use f-strings when you're actually substituting something; otherwise flake8 complains (are you not running flake8?)

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -83,7 +100,29 @@ def _run_task(self):
...
+                else:
+                    cmd.extend([f"--{option}", str(value)])
```
davisagli, line 118:
```suggestion
                    cmd.extend(["--{option}", str(value)])
```

cumulusci/tasks/robotframework/robotframework.py
boakley: The type shows up when  you do `cci task info robot`, which I thought was reason enough to include it.

cumulusci/tasks/robotframework/robotframework.py
boakley: Yeah, that makes sense.  Did you purposefully change the wording on the exception that is raised?

cumulusci/tasks/robotframework/robotframework.py
boakley: This is substituting something:  `{option}`. Maybe I should rename that to `{option_name}` to make it more obvious.

davisagli: @boakley sorry, my mistake, not sure what I was thinking

cumulusci/tasks/robotframework/robotframework.py
davisagli: It's a suggestion; take it or leave it -- generally I prefer to word things so we're giving the user a clue about what they need to change rather than just telling them what they did wrong.

cumulusci/tasks/robotframework/robotframework.py
davisagli: Oh huh, I didn't realize that. I guess I have no objection, but it doesn't look like we are actually using that anywhere other than the test for `doc_task`, and we'll likely need to replace it when we finish up #1618

boakley: Ok, I'll remove it.

PR #1884, opened 2020-07-23
metadata_etl supports updating metadata's first child tag's text

spelak-salesforce:
# Critical Changes

# Changes
Adds `cumulusci.tasks.metadata_etl.UpdateFirstAttributeTextTask` supporting updating a metadata's first child tag's text.
- A generic task that supports setting `entity`, `tag`, and `value` options.
- Only updates the metadata's first child tag.
- Does not support tags with complex/object-like values.
- Example:
   ```yml
    assign_account_compact_layout:
        description: "Assigns the Fancy Compact Layout as Account's Compact Layout."
        class_path: cumulusci.tasks.metadata_etl.UpdateFirstAttributeTextTask
        options:
            managed: False
            namespace_inject: $project_config.project__package__namespace
            entity: CustomObject
            api_names: Account
            attribute: compactLayoutAssignment
            value: "%%%NAMESPACE%%%Fancy_Account_Compact_Layout"
   ```

# Issues Closed


cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
                 removed_api_names.add(api_name)
 
         self.api_names = self.api_names - removed_api_names
+
+
+class UpdateFirstAttributeTextTask(MetadataSingleEntityTransformTask):
```
spelak-salesforce, line 311: Change name to `UpdateFirstChildTextTask`

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+class UpdateMetadataFirstChildTextTask(MetadataSingleEntityTransformTask):
+    """
+    Updates the metadata's first child tag's text.
+    Does not support tags with complex/object-like values.
+    """
+
+    task_options = {
+        "entity": {"description": "Metadata Entity", "required": True},
+        "tag": {
+            "description": "Tag of metadata's first child whose text will be updated.   Only supports setting the tag's text.",
```
davidmreed, line 320:
```suggestion
            "description": "Targeted tag. The text of the first instance of this tag within the metadata entity will be updated.",
```

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+            "required": True,
+        },
+        "value": {
+            "description": "Text of metadata's first child with tag.  This value is namespaced injected.",
```
davidmreed, line 324:
```suggestion
            "description": "Desired value to set for the targeted tag's text. This value is namespace-injected.",
```

cumulusci/tasks/metadata_etl/tests/test_base.py
```diff
@@ -379,3 +381,173 @@ def test_transform__xml_parse_error(self):
             test_path.write_text(">>>>>NOT XML<<<<<")
             with pytest.raises(etree.ParseError):
                 task._transform()
+
+
+class TestUpdateMetadataFirstChildTextTask:
+    def test_init_options(self):
+        for options, expected_value in [
```
davidmreed, line 388: This test style is very hard to read (and hard to evaluate when it fails). Please break out these cases into separate unit tests. A utility method or Pytest fixture can help elide the boilerplate while preserving the clarity and independence of each test case.

spelak-salesforce: Working on making tests readable.   Looking into [https://docs.pytest.org/en/stable/parametrize.html](https://docs.pytest.org/en/stable/parametrize.html)

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
 
         self.api_names = self.api_names - removed_api_names
+
+
+class UpdateMetadataFirstChildTextTask(MetadataSingleEntityTransformTask):
+    """
+    Updates the metadata's first child tag's text.
+    Does not support tags with complex/object-like values.
+    """
+
```
davisagli, line 316: When adding a task, part of the DoD is making sure that it is documented.

1. Add a task_docs attribute with a string (reStructuredText format) explaining what the task is used for and at least one example.
2. Run `make docs` to update docs/tasks.rst, build the Sphinx docs, and open them in your browser.
3. Check that everything is rendering as expected.

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+    task_options = {
+        "entity": {"description": "Metadata Entity", "required": True},
```
davisagli, line 318: Entities are called Metadata Types in the Metadata API docs; is there a reason to depart from that here?

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+        "tag": {
+            "description": "Targeted tag. The text of the first instance of this tag within the metadata entity will be updated.",
+            "required": True,
+        },
+        "value": {
+            "description": "Desired value to set for the targeted tag's text. This value is namespace-injected.",
+            "required": True,
+        },
+        **MetadataSingleEntityTransformTask.task_options,
+    }
```
davisagli, line 328: I can imagine a few other options that would make this more versatile (such as the option to target all instances of a tag, or to target a nested tag using xpath). But there's also a strong argument to be made for keeping it simple and creating other tasks for other use cases. Any thoughts, @davidmreed

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+    def _transform_entity(
+        self, metadata: MetadataElement, api_name: str
+    ) -> MetadataElement:
+        """
+        Finds metadata's first child with tag.  If no child is found, appends
+        a new child with tag.  Then updates child's text as the value option.
+        """
+        tag = self.options.get("tag")
+
+        attribute = metadata.find(tag)
```
davisagli, line 344: Call this element or child -- this is not an XML attribute, which could be confusing.

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+        if not attribute:
```
davisagli, line 345:
```suggestion
        if attribute is None:
```

(I don't think it actually matters in this case, but it's a good habit to get into because in Python there can be non-None objects which are considered false when evaluated as a boolean)

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,48 @@ def _transform(self):
...
+        self.options["value"] = self._inject_namespace(self.options.get("value"))
+
+    def _transform_entity(
+        self, metadata: MetadataElement, api_name: str
+    ) -> MetadataElement:
+        """
+        Finds metadata's first child with tag.  If no child is found, appends
+        a new child with tag.  Then updates child's text as the value option.
+        """
+        tag = self.options.get("tag")
```
davisagli, line 342: Within here, it's safe to do `self.options["tag"]` instead of `self.options.get("tag")` because the options are marked as required.

(The framework first calls, _init_options, then BaseTask._validate_options which checks for required fields, then runs the task. So using `get` _is_ necessary in _init_options)

cumulusci/tasks/metadata_etl/tests/test_base.py
davisagli: If I were writing the tests I might do them like this:

    def test_init_options__no_namespace(self):
        task = create_task(UpdateMetadataFirstChildTextTask, {
            "managed": False,
            "namespace_inject": None,
            "entity": "CustomObject",
            "tag": "customObjectAttribute",
            "value": "newAttributeValue",
        })
        assert options["entity"] == task.entity
        assert task.options["value"] == "newAttributeValue"

with a separate function for each test case, and not worry about the repeated assertions. This way there's a hint in the function name about what scenario is being tested. Tests are _usually_ not really the place to abstract things.

That said, I think using pytest parametrize is an acceptable way to go here too (it at least gives more helpful output when one of the sets of parameters fails) and you may use it if you prefer.

cumulusci/tasks/metadata_etl/tests/test_base.py
```diff
@@ -379,3 +381,173 @@ def test_transform__xml_parse_error(self):
...
+        assert tag == task.options.get("tag")
+        assert value == task.options.get("value")
+
+        actual = task._transform_entity(metadata, api_name)
+
+        assert metadata == actual
+
+        metadata.find.assert_called_once_with(tag)
+        metadata.append.assert_not_called()
+        assert metadata.find.return_value.text == value
```
davisagli, line 506: I would suggest not using mocks and treating this as more of a black box test using an actual MetadataElement instance:
1. Construct a MetadataElement tree for the initial XML (possibly using metadata_tree.parse so the XML is readable in the test)
2. Run it through _transform_entity
3. Assert that the result is as expected (possibly after using .tostring() so the expected result is readable in the test)

The point of the test is to make sure the transform made the desired change; we don't need to verify that it was done in a particular way. Also, using mocks means that if we changed something about the real MetadataElement's implementation of these methods that broke _transform_entity, this test would not catch it.

cumulusci/tasks/metadata_etl/base.py
davidmreed: I'm on Team Simple for this one. I prefer to have Metadata ETL tasks that clearly bound/state the operation they are capable of executing. This one is the furthest we've gone so far in the direction of general-purpose MDETL tasks, and I think it's fine, but I'm more comfortable with lots of configurability living in abstract base classes and primarily exposing tasks that do one thing explicitly.

cumulusci/tasks/metadata_etl/tests/test_base.py
spelak-salesforce: @davidmreed  and @davisagli 

I split up `test_base.py`'s `TestUpdateMetadataFirstChildTextTask` tests:  one for each expected outcome using `@pytest.mark.parametrize`.   Thoughts?   I can split out my scenarios into separate tests too.

Question:  My difficulty is testing `_init_options`: `_init_options` is called during the task's `__init__`.   It appears I can only test `_init_options` with functional tests.   Is this correct?   A unit test would be much shorter.

cumulusci/tasks/metadata_etl/base.py
spelak-salesforce: `metadata_etl` uses the `entity` variable, but yes!  I much prefer using the usual terminology

spelak-salesforce: updated

cumulusci/tasks/metadata_etl/base.py
spelak-salesforce: Done.  Will push momentarily

cumulusci/tasks/metadata_etl/base.py
spelak-salesforce: Done.  Will push momentarily.

cumulusci/tasks/metadata_etl/base.py
spelak-salesforce: Done. Will push momentarily.

cumulusci/tasks/metadata_etl/tests/test_base.py
spelak-salesforce: I will add a functional test.

spelak-salesforce: Done

cumulusci/tasks/metadata_etl/base.py
spelak-salesforce: Done.   This is super cool!

spelak-salesforce:
@davidmreed @davisagli 

I believe I made all requested changes.  Please let me know if I should update the `_init_options` test further.

cumulusci/tasks/metadata_etl/tests/test_base.py
davisagli: @spelak-salesforce I think separate tests would be easier to read, but I also think this is good enough and not worth spending more time on. I did make a commit to consolidate the two tests for _init_options since they had exactly the same test code.

You're right that _init_options can't be called without instantiating the task, which runs it implicitly via `__init__`. I would still say this test is closer to the unit test end of the spectrum than functional since it's testing that a portion of the code works properly rather than testing the functionality seen by the user. I suppose the framework could be changed so that option parsing happens in pure functions that get attached to task classes. But I don't think it's worth the backwards incompatible change in the framework unless this current approach becomes a performance bottleneck. Anyway we have some other changes in mind for task options that will make it possible for them to be specified declaratively (and thus not need special code in _init_options) in many more cases...

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,97 @@ def _transform(self):
...
+class UpdateMetadataFirstChildTextTask(MetadataSingleEntityTransformTask):
+    task_docs = """
+MetadataSingleEntityTransformTask to update a single child's text.  Specify:
+
+- ``metadata_type`` as the Metadata's `Metadata Type <https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_types_list.htm>`_
+
+- ``tag`` as the targeted child's tag
+
+- ``value`` as the new value of the child's text
+
```
davisagli, line 320:
```suggestion
Metadata ETL task to update a single child element's text within metadata XML.
```

Task docs should not assume the reader is a Python developer familiar with CumulusCI internals.

And the task options are automatically listed below by the doc generator (based on `task_options`), so there's no need to repeat them.

cumulusci/tasks/metadata_etl/base.py
```diff
@@ -305,3 +306,97 @@ def _transform(self):
...
+  <?xml version="1.0" encoding="UTF-8"?>
+  <CustomObject xmlns="http://soap.sforce.com/2006/04/metadata">
+      ...
+      <label>Our Custom Object</label>
+      <compactLayoutAssignment>OriginalCompactLayout</compactLayoutAssignment>
+      ...
+  </CustomObject>
+
+After running the ``cumulusci.tasks.metadata_etl.UpdateMetadataFirstChildTextTask``, the CustomObject metadata XML is deployed as:
```
davisagli, line 355:
```suggestion
After running ``cci task run assign_compact_layout``, the CustomObject metadata XML is deployed as:
```

cumulusci/tasks/metadata_etl/tests/test_base.py
```diff
@@ -379,3 +381,236 @@ def test_transform__xml_parse_error(self):
...
+        )
+        task.logger = mock.Mock()
+        assert tag == task.options.get("tag")
+        assert value == task.options.get("value")
+
+        actual = task._transform_entity(metadata, api_name)
+
+        assert metadata == actual
+
+        assert actual.tostring() == expected_metadata.tostring()
```
davisagli, line 609: Can't you just compare to EXPECTED_XML directly without needing to fromstring/tostring it?

spelak-salesforce: No.  I tried, but there's weird differences my guess is with indentation or line endings.

davisagli: Here you go: 53c4b7d

docs/tasks.rst
```diff
@@ -231,6 +231,111 @@ Options
...
 	 Metadata API version to use, if not project__package__api_version.
 
+**assign_compact_layout**
+==========================================
+
+**Description:** Assigns the Compact Layout specified in the 'value' option to the Custom Objects in 'api_names' option.
+
+**Class:** cumulusci.tasks.metadata_etl.UpdateMetadataFirstChildTextTask
+
+Metadata ETL task to update a single child element's text within metadata XML.
```
davidmreed, line 241: @davisagli  I suggested to @spelak-salesforce that we publish a preconfigured task in `cumulusci.yml` for the known concrete use of this generic task. That unfortunately results in reduplicating the generic tasks's docs. 

Is there a better solution? I suppose we could ship a separate Python task but part of the virtue of this PR is not requiring additional Python.

davisagli: @davidmreed as of right now we need a separate Python class in order to provide different task_docs. It would be a small tweak of the doc_task function to allow for overriding task_docs in the YAML task_config

davidmreed: I think having the task available outweighs the minor confusion in the task reference. Happy to put that tweak on my backlog.

PR #1893, opened 2020-07-28
Source tracking fixes

davisagli:
1. Use RevisionCounter instead of RevisionNum, because the former doesn't get reset during a push.
2. Use the new force:source:tracking:reset command to reset sfdx's local files, instead of trying to read/write maxrevision.json directly (which was no longer working correctly due to changes in the format).

# Critical Changes

# Changes

# Issues Closed
- Fixed a bug where the list_changes, retrieve_changes, and snapshot_changes tasks could error while trying to reset sfdx sourcetracking.


cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -397,8 +381,7 @@ def _run_task(self):
 
             if changes:
                 self._store_snapshot(changes)
-                maxrevision = max(change["RevisionNum"] or -1 for change in changes)
-                self._store_maxrevision(maxrevision)
+            self._reset_sfdx_snapshot()
```
davidmreed, line 384: Is this outdent intention? It looks like a possible behavior change, albeit probably not a undesirable one.

davisagli: It's intentional -- probably doesn't matter much, but I figured it'll make sure sfdx gets reset even if there's some error in how cci is keeping track of changes.

PR #1894, opened 2020-07-28
Run integration tests on a daily job.

prescod:
Github action to Run integration tests daily.


davidmreed:
This change suggests to me that we need to have more formalization of the CumulusCI release process. If we institute running integration tests nightly, which is a good idea (we could ultimately include Bulk Data's integration tests in that process), I think we need a merge freeze starting at the end of the day prior to the release and continuing through the conclusion of release activities.

@davisagli Do you have thoughts?

davisagli:
@davidmreed I agree it makes sense to require that the integration tests have been run on a commit before releasing it. I don't think that necessarily requires waiting for the automated run of the integration tests; i.e. they could also be triggered manually before making a release.

davidmreed:
That's fair. I suppose it matters significantly how long the integration test suites take to run.

prescod:
Do you think that the button to run this manually will only appear when its merged to master?

PR #1895, opened 2020-07-28
only try to reset sfdx source tracking if the org is actually from sfdx

davisagli:
# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -133,13 +134,15 @@ def _store_snapshot(self, changes):
...
     def _reset_sfdx_snapshot(self):
-        sfdx(
-            "force:source:tracking:reset",
-            args=["-p"],
-            username=self.org_config.username,
-            capture_output=True,
-            check_return=True,
-        )
+        # If org is from sfdx, reset sfdx source tracking
+        if isinstance(self.org_config, ScratchOrgConfig):
```
davidmreed, line 138: If a user `cci org connect`'s someone else's scratch org, is that a `ScratchOrgConfig`?

davisagli: No. And that is one of the cases I'm trying to avoid here, since sfdx would not know about the org and throws an error `No AuthInfo found for name [username]`

PR #1896, opened 2020-07-28
CCI directory cleanup

prescod:
Send all references to ".cci" through "project_config.project_cache_dir"


cumulusci/tasks/salesforce/sourcetracking.py
```diff
@@ -52,10 +52,11 @@ def _init_options(self, kwargs):
 
     @property
     def _snapshot_path(self):
-        parent_dir = os.path.join(".cci", "snapshot")
-        if not os.path.isdir(parent_dir):
-            os.makedirs(parent_dir)
-        return os.path.join(parent_dir, "{}.json".format(self.org_config.name))
+        parent_dir = self.project_config.project_cache_dir / "snapshot"
+        if not parent_dir.exists():
+            parent_dir.mkdir(parents=True)
```
davisagli, line 57:
```suggestion
        parent_dir.mkdir(parents=True, exist_ok=True)
```

cumulusci/tasks/salesforce/tests/test_sourcetracking.py
```diff
@@ -256,4 +257,7 @@ def test_run_task__null_revnum(self):
     def test_freeze(self):
         task = create_task(SnapshotChanges)
         steps = task.freeze(None)
-        self.assertEqual([], steps)
+        assert steps == []
+
+
+flake8 = (create_task_fixture,)
```
davisagli, line 263: One way around this is to define fixtures in conftest.py so that pytest finds them without needing to be imported.

prescod: We have a few create_task, _make_task etc. thingees floating around and I wasn't confident enough to promote this to being the global everyone-should-use-this one until I'd tried it in a few more contexts.

davisagli: Yeah I definitely want to try to consolidate those, but am also not necessarily asking you to do that right now.

PR #1913, opened 2020-08-03
Config file housekeeping

jstvz:
Pytest added support for `pyproject.toml` configuration files in the latest release, so thought it made sense to consolidate there where possible. Not possible yet:

- `.flake8`: They're holding off on support. ([Ticket](https://gitlab.com/pycqa/flake8/-/issues/428))
- `tox.ini`: Has TOML support, but only as inline [INI multiline string](https://tox.readthedocs.io/en/latest/config.html). (So possible, but suboptimal)

# Critical Changes

# Changes

# Issues Closed


davisagli:
@jstvz the Release Test workflow failed with `Couldn't read '../.coveragerc' as a config file`

PR #1915, opened 2020-08-04
Test coverage for cumulusci push tasks

jofsky:
Resolving flake8 issues across multiple files project wide. Creating test coverage to begin to cover the push api library and provide a reliable testing harness.

# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    task.push = mock.MagicMock()
+    task.push.get_package_objs.return_value = [
+        {"push_api": "123", "sf_id": SF_ID, "name": NAME, "namespace": NAMESPACE}
+    ]
+    assert task._get_package(NAMESPACE) == {
+        "push_api": "123",
+        "sf_id": SF_ID,
+        "name": NAME,
+        "namespace": NAMESPACE,
+    }
```
davisagli, line 231: The _real_ SalesforcePushApi.get_package_objs returns a list of MetadataPackage instances, not a list of dicts. For this test really all we need to prove is that BaseSalesforcePushTask._get_package returns the first item from the list that was returned by SalesforcePushApi.get_package_objs. So you could do one of two things:
1. Change this to build a MetadataPackage instance for the faked SalesforcePushApi.get_package_objs return value
2. Create an arbitrary object, `expected_package = object()`. Make SalesforcePushApi.get_package_objs return a list including that object. Assert that BaseSalesforcePushTask._get_package returns the same object (`assert task._get_package(NAMESPACE) is expected_package`)

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+
+
+def test_get_package_error():
+    task = create_task(BaseSalesforcePushTask, options={})
+    task.push = mock.MagicMock()
+    task.push.get_package_objs.return_value = None
```
davisagli, line 237: Again, an empty list would be a more realistic simulation of the API you're mocking

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    with pytest.raises(PushApiObjectNotFound):
+        task._get_package(NAMESPACE)
+
+
+def test_schedule_push_org_list_get_orgs():
```
davisagli, line 242: If this were me writing the tests, I would group the SchedulePushOrgList tests together by putting them inside a class named TestSchedulePushOrgList, just for the sake of organization. I'm fine leaving this up to personal preference though.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+        options={
+            "orgs": ORG_FILE,
+            "version": VERSION,
+            "namespace": NAMESPACE,
+            "start_time": datetime.datetime.now(),
+            "batch_size": 10,
+        },
+    )
+    assert task._get_orgs() == ["00DS0000003TJJ6MAO", "00DS0000003TJJ6MAL"]
+    os.remove(ORG_FILE)
```
davisagli, line 256: This is a non-guaranteed way to do cleanup in a test, because if there's an error or assertion failure in the test, `os.remove` will not run.

You've got a few options for how to make it better:
1. Wrap the entire test in try/finally and put os.remove in the finally block.
2. Wrap the entire test in `with temporary_dir() as path:` and make sure you create the file inside the `path` directory (the temporary_dir context manager takes care of deleting the directory when it exits)
3. Create a separate fixture to handle the file:

```
@pytest.fixture
def org_file():
    with open(ORG_FILE, "w") as file:
        file.write(ORG_FILE_TEXT)
    try:
        yield  # this is where the test using the fixture runs
    finally:
        os.remove(ORG_FILE)
```

I like option 3 because then you don't have to repeat the logic for each test that uses the fixture.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    task = create_task(BaseSalesforcePushTask, options={})
+    task.push = mock.MagicMock()
+    task.push.get_package_objs.return_value = None
+    with pytest.raises(PushApiObjectNotFound):
+        task._get_package(NAMESPACE)
+
+
+def test_schedule_push_org_list_get_orgs():
+    with open(ORG_FILE, "w") as file:
+        file.write(f"{ORG_FILE_TEXT}")
```
davisagli, line 244: There's no benefit to using an f-string here.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+        options={
+            "orgs": ORG_FILE,
+            "version": VERSION,
+            "start_time": datetime.datetime.now(),
+        },
+    )
+    task._init_task()
+    task._init_options(
+        {"orgs": ORG_FILE, "version": VERSION, "start_time": datetime.datetime.now()}
+    )
```
davisagli, line 273: _init_options is called by the task's `__init__` method when it is constructed; you don't need to call it again.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    with pytest.raises(NotImplementedError):
+        task()
+
+
+def test_filter_subs_raises_err():
+    task = create_task(
+        FilterSubscriberList, options={"file_in": "in.csv", "file_out": "out.txt"}
+    )
+    with pytest.raises(NotImplementedError):
+        task()
```
davisagli, line 327: I think we should just remove GetSubscriberList and FilterSubscriberList since they are not implemented.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    task = create_task(BaseSalesforcePushTask, options={})
+    with pytest.raises(NotImplementedError):
+        task()
+
+
+def test_report_push_status_error():
+    task = create_task(BaseSalesforcePushTask, options={})
+    task.sf = mock.MagicMock()
+    task.push_report = mock.MagicMock()
+    task.sf.query_all.return_value = {"totalSize": 1, "records": []}
```
davisagli, line 340: This is a nonsensical query result -- if records is an empty list, then totalSize would be 0

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    task.sf = mock.MagicMock()
+    task.push_report = mock.MagicMock()
+    task.push_request = mock.MagicMock()
+    task.push_request.get_push_job_objs.return_value = [
+        package_push_job_success,
+        package_push_job_failure,
+        package_push_job_cancel,
+    ]
+    task._get_push_request_job_results()
+    task.push_request.get_push_job_objs.assert_called_once()
```
davisagli, line 358: This would be a better test if it actually asserted that _get_push_request_job_results logged info about the failed job.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+        },
+    )
+    task.push = mock.MagicMock()
+    task.push_api = mock.MagicMock()
+    task.sf = mock.Mock()
+    task.push_api.return_query_records = mock.MagicMock()
+    task.push_api.return_query_records.return_value = {"totalSize": "1"}
+    task.sf.query_all.return_value = PACKAGE_OBJS
+    with pytest.raises(ValueError):
+        assert task._get_orgs() == [NAME]
```
davisagli, line 380: What behavior is this trying to test for? I would tend to think a ValueError is a bug since we generally want to show helpful error messages that explain what happened.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+            "namespace": NAMESPACE,
+            "start_time": None,
+            "batch_size": "200",
+            "subscriber_where": "OrgType = 'Sandbox'",
+        },
+    )
+    task.push = mock.MagicMock()
+    task.push_api = mock.MagicMock()
+    task.sf = mock.Mock()
+    task.sf.query_all.return_value = PACKAGE_OBJS
```
davisagli, line 398: Your mock return value doesn't make sense here. The code being tested is doing a query for PackageSubscribers, but PACKAGE_OBJS contains a response with PackagePushRequests.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+            "start_time": "now",
+            "namespace": NAMESPACE,
+        },
+    )
+    task.push = mock.MagicMock()
+    task.sf = mock.MagicMock()
+    task.sf.query_all.return_value = PACKAGE_OBJS
+    task.push.create_push_request.return_value = (task.sf.query_all.return_value, 2)
+    task._run_task()
+    task.sf.query_all.assert_called_with(query)
```
davisagli, line 420: Currently this is only asserting that the task did _some_ query, which is a pretty vague assertion. I think the most important assertion to make is that task.push.create_push_request was called with the expected values.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+    task.push = mock.MagicMock()
+    task.sf = mock.MagicMock()
+    task.sf.query_all.return_value = PACKAGE_OBJS
+    task.push.create_push_request.return_value = (task.sf.query_all.return_value, 2)
+    task._run_task()
+    task.push.create_push_request.assert_called_once()
+    os.remove(ORG_FILE)
+
+
+def test_schedule_push_org_list_run_task_without_orgs():
```
davisagli, line 440: In what way is this test "without orgs"? It's using the same org file as the other tests.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,481 @@
...
+            "namespace": NAMESPACE,
+            "start_time": "now",
+        },
+    )
+    task.push = mock.MagicMock()
+    task.sf = mock.MagicMock()
+    task.sf.query_all.return_value = PACKAGE_OBJS
+    task.push.create_push_request.return_value = (task.sf.query_all.return_value, 1001)
+    task._run_task()
+    task.sf.query_all.assert_called_with(query)
```
davisagli, line 480: Would be good to also assert that create_push_request was called with the expected arguments.

cumulusci/tasks/push/tests/test_push_tasks.py
jstvz: pytest provides the [`tmp_path` fixture](https://docs.pytest.org/en/stable/tmpdir.html#the-tmp-path-fixture), which does what you want.

davisagli: @jstvz TIL!

jstvz:
@jofsky Which PR should we be reviewing? This one or #1933 ?

jofsky:
@jstvz review both #1933 and #1915 I split out the work into two separate PR's since it was so much.

jstvz:
> @jstvz review both #1933 and #1915 I split out the work into two separate PR's since it was so much.

Ah, now I see it, thanks. I was thrown by the list of commits.

cumulusci/tasks/push/tests/test_push_tasks.py
jofsky: The only way Im willing to do this is if we put all the others in classes as well.

cumulusci/tasks/push/tests/test_push_tasks.py
jofsky: Not sure what that would look like.

davisagli: @jofsky pytest has a `caplog` fixture which can be used to capture logs and make assertions about them: https://docs.pytest.org/en/stable/logging.html#caplog-fixture

Here is an example: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/tasks/bulkdata/tests/test_mapping_parser.py#L42

cumulusci/tasks/push/tests/test_push_tasks.py
davisagli: @jofsky can you show me where the ValueError is coming from?

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,518 @@
...
+    task.push = mock.MagicMock()
+    task.sf = mock.MagicMock()
+    task.sf.query_all.return_value = PACKAGE_OBJS
+    task.push.create_push_request.return_value = (task.sf.query_all.return_value, 2)
+    task._run_task()
+    task.push.create_push_request.assert_called_once_with(
+        task.push.get_package_objs()
+        .__getitem__()
+        .get_package_version_objs()
+        .__getitem__(),
```
davisagli, line 428: @jofsky is this just a hack to get the same mock objects that were used for the actual call? Because it looks quite odd to be calling `__getitem__` without passing any argument. Since these are not the point of the assertion, you could pass unittest.mock.ANY instead: https://docs.python.org/3/library/unittest.mock.html#any

cumulusci/tasks/push/tests/test_push_tasks.py
davisagli: @jofsky I unresolved this one; doesn't look like it was done.

jofsky: @davisagli  I think I now addressed lmk.

cumulusci/tasks/push/tests/test_push_tasks.py
jofsky: I believe https://github.com/SFDO-Tooling/CumulusCI/blob/7b2d16741eac2bb9b98f9bdb6404f524703710ab/cumulusci/tasks/push/tasks.py#L363

davisagli:
@jofsky Thanks for your work on this!

PR #1916, opened 2020-08-04
Global orgs had bugs relating to defaulting

prescod:
Previously there were subtle issues where local org files were being created and shadowing global ones. So, for example, removing a global org might not remove it for all projects.

Fixing that bug exposed a logic error because "default" was stored on a boolean in the orgconfig which means that if a global org is defaulted it will be defaulted for all projects. This fixes that bug.

The default is now stored externally in its own file analogous to how Git stores its HEAD.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -918,7 +918,7 @@ def org_connect(runtime, org_name, sandbox, login_url, default, global_org):
 
     org_config.save()
 
-    if default:
+    if default and not global_org:
         runtime.keychain.set_default_org(org_name)
         click.echo(f"{org_name} is now the default org")
 
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -918,7 +918,7 @@ def org_connect(runtime, org_name, sandbox, login_url, default, global_org):
 
     org_config.save()
 
-    if default and not global_org:
+    if default and runtime.project_config is not None:
         runtime.keychain.set_default_org(org_name)
         click.echo(f"{org_name} is now the default org")
 
@@ -1044,8 +1044,9 @@ def org_list(runtime, plain):
         org: runtime.keychain.get_org(org) for org in runtime.keychain.list_orgs()
     }
     rows_to_dim = []
+    default_org_name, _ = runtime.keychain.get_default_org()
     for org, org_config in org_configs.items():
-        row = [org, org_config.default]
+        row = [org, org == default_org_name]
         if isinstance(org_config, ScratchOrgConfig):
             org_days = org_config.format_org_days()
             if org_config.expired:
--- a/cumulusci/core/keychain/BaseProjectKeychain.py
+++ b/cumulusci/core/keychain/BaseProjectKeychain.py
@@ -142,11 +142,6 @@ class BaseProjectKeychain(BaseConfig):
         del self.orgs[name]
         self._load_orgs()
 
-    # TODO: Are these used in any context other than
-    # encrypted_file_project_keychain?
-    #
-    # If not, they can be removed after a reasonable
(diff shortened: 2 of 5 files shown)


cumulusci/core/keychain/BaseProjectKeychain.py
```diff
@@ -142,6 +142,11 @@ def _remove_org(self, name, global_org):
         del self.orgs[name]
         self._load_orgs()
 
+    # TODO: Are these used in any context other than
+    # encrypted_file_project_keychain?
+    #
+    # If not, they can be removed after a reasonable
+    # period of backwards compatibility
```
davisagli, line 149: Not sure I understand the comment. set_org is the standard interface for saving org configs to a keychain, so yeah, this is used by any keychain that doesn't override it.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -122,6 +123,40 @@ def _get_org(self, name):
...
+        org_name, org_config = super().get_default_org()
+        if org_name:
+            self.set_default_org(org_name)  # upgrade to new way
+        return org_name, org_config
+
+    def set_default_org(self, name: str):
+        """ Set the default org for tasks and flows by name """
+        super().set_default_org(name)
+        with open(self.default_org_name, "w") as f:
+            f.write(name)
```
davisagli, line 150:
```suggestion
        self.default_org_name.write_text(name)
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -122,6 +123,40 @@ def _get_org(self, name):
...
+
+    def unset_default_org(self):
+        """Unset the default orgs for tasks and flows """
+        super().unset_default_org()
+        try:
+            os.unlink(self.default_org_name)
+        except FileNotFoundError:
+            pass
```
davisagli, line 158:
```suggestion
        self.default_org_name.unlink(missing_ok=True)
```

cumulusci/core/keychain/BaseProjectKeychain.py
prescod: Sorry the comment should be about 10 lines lower around get_default, set_default, unset_default.

Do we use default orgs in contexts other than the command line/encrypted project?

Basically I'm not thrilled about having two overlapping ways of defining the default org, using the filesystem in some places and using a flag in others. If we don't need the flag, we should get rid of it.

davisagli: The web app keychains don't use default orgs, afaik. It's conceivable that someone using the CLI could use the EnvironmentProjectKeychain, which inherits these from BaseProjectKeychain. But setting a default org when using that keychain wouldn't actually work since it doesn't get persisted anywhere. So it would probably be good to raise a clear error about that anyway.

davisagli:
> Have you tested this when using cci org default outside of a project directory or cci org connect --default --global-org? I think it'll break when it tries to get project_local_dir from the keychain's project_config which is actually a UniversalConfig in that case.

`cci org default` already requires being in a project

I'm adding a conditional to skip setting the default if `cci org connect` is called with both `--default` and `--global-org`

cumulusci/cli/cci.py
```diff
@@ -1044,8 +1044,9 @@ def org_list(runtime, plain):
         org: runtime.keychain.get_org(org) for org in runtime.keychain.list_orgs()
     }
     rows_to_dim = []
+    default_org_name, _ = runtime.keychain.get_default_org()
     for org, org_config in org_configs.items():
-        row = [org, org_config.default]
+        row = [org, org == default_org_name]
```
davisagli, line 1049: @prescod good catch!

PR #1918, opened 2020-08-04
Support $project_config substitution in task options in any position

davisagli:
# Critical Changes

# Changes
- Task options can now use $project_config substitutions in any position, not just at the start of the value.

# Issues Closed


davisagli:
@davidmreed getting an attribute of project_config that is not defined returns None rather than raising an error, so I don't think it's a change in behavior. The `except AttributeError` was to handle values that didn't have a `startswith` method; now we explicitly check if the value is a string instead.

PR #1919, opened 2020-08-04
Case agnostic bulkdata

prescod:
# Changes

A pre-release (?) version of CumulusCI was very strict about requiring the case of sObjects and Fields to be exactly the same as the "canonical" Salesforce name. This puts it back to being case agnostic.

### DIFF
--- a/cumulusci/tasks/bulkdata/mapping_parser.py
+++ b/cumulusci/tasks/bulkdata/mapping_parser.py
@@ -18,17 +18,15 @@ logger = getLogger(__name__)
 
 class CaseInsensitiveDict(RequestsCaseInsensitiveDict):
     def __init__(self, *args, **kwargs):
+        self._canonical_keys = {}
         super().__init__(*args, **kwargs)
-        # there's a more efficient way to do this by poking around the
-        # base class internals but this way may be less fragile
-        self._canonical_keys = {key.lower(): key for key in self.keys()}
 
     def canonical_key(self, name):
         return self._canonical_keys[name.lower()]
 
-    def __setitem__(self, name, value):
-        # self._canonical_keys doesn't get updated if this dict gets mutated
-        raise NotImplementedError()
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+        self._canonical_keys[key.lower()] = key
 
 
 class MappingLookup(CCIDictModel):
--- a/cumulusci/tasks/bulkdata/mapping_parser.py
+++ b/cumulusci/tasks/bulkdata/mapping_parser.py
@@ -13,7 +13,6 @@ from cumulusci.utils import convert_to_snake_case
 
 from typing_extensions import Literal
 
-LOGGER_NAME = "MAPPING_LOADER"
 logger = getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


prescod:
> @prescod I think we need a solution which updates the mapping with normalized schema names rather than just making sure they pass validation as written. Otherwise checks like this (coming in Scott's person account PR) will fail when the case is different: https://github.com/SFDO-Tooling/CumulusCI/pull/1829/files#diff-81c24e397db90aa3b439d7c1753048a1R199

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -123,13 +135,14 @@ def _get_permission_type(self, operation: DataOperationType) -> str:
         return "createable"
 
     def _check_object_permission(
-        self, global_describe: Dict, sobject: str, operation: DataOperationType
+        self, global_describe: Mapping, sobject: str, operation: DataOperationType
     ):
+        assert sobject in global_describe
```
davidmreed, line 140: `global_describe` will not contain `sobject` if the user doesn't have Read permission at the object level.

prescod: That case should be caught on line 235 which is new.

https://github.com/SFDO-Tooling/CumulusCI/pull/1919/files/98a4287b36aa7070c85cfb4896c2baf279f999de#diff-0a8fc3e0cff0ad9e099cacf4f6fb0f41R235

davidmreed: Ah, got it. Thanks for clarifying.

cumulusci/tasks/bulkdata/tests/test_utils.py
```diff
@@ -46,6 +46,12 @@ def mock_sobject_describe(name: str):
             body=read_mock(name),
             status=200,
         )
+        responses.add(
+            method="GET",
+            url=f"https://example.com/services/data/v48.0/sobjects/{name.lower()}/describe",
+            body=read_mock(name),
+            status=200,
+        )
```
davisagli, line 54:
```suggestion
```

This seems to be unnecessary

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -13,7 +14,18 @@
...
+
+class CaseInsensitiveDict(RequestsCaseInsensitiveDict):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # there's a more efficient way to do this by poking around the
+        # base class internals but this way may be less fragile
+        self._canonical_keys = {key.lower(): key for key in self.keys()}
+
+    def canonical_key(self, name):
+        return self._canonical_keys[name.lower()]
```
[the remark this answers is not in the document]
prescod: Sure.

The alternative alluded to in the comment is that Requests internal data-structure has all of the information we need, but its an undocumented internal data-structure.

cumulusci/tasks/bulkdata/tests/test_utils.py
prescod: I think it seems unnecessary on Mac but was added for Linux.

davisagli: But how does this end up lowercase on Linux? The describe call is made using the mapping's sf_object after _validate_sobject was called, so it should be canonicalized based on the objects from the global describe.

davisagli:
@davidmreed Does this look all right to you?

davidmreed:
Yeah, this looks fine to me. @prescod Have you run it through the integration test suite?

prescod:
Integration tests pass.

PR #1921, opened 2020-08-04
bulkdata load validates Contact steps with person account records must have an AccountId Lookup

spelak-salesforce:
# Critical Changes

# Changes
- Bulkdata load task validates **Contact** steps with person account **Contact** records have an `AccountId` lookup so downstream lookups are correctly populated. 

# Issues Closed


spelak-salesforce:
@davidmreed @davisagli 

For person account **Contact** records, load requires an `AccountId` lookup to `_generate_contact_id_map_for_person_accounts`.  This pull request adds pre-load validation asserting this condition.

Do we need to add any validation in the extract task?  If so, I think it we be a post-extract check before the data file is saved.   (I only want to validate if a Contact step has person account records which requires the extract to have happened.)

davidmreed:
> Do we need to add any validation in the extract task? If so, I think it we be a post-extract check before the data file is saved. (I only want to validate if a Contact step has person account records which requires the extract to have happened.)

That makes sense to me. It's valuable to tell the user as early as possible that they've got an unusable artifact on their hands, rather than allowing them to commit it and only later discover that it cannot be loaded.

jstvz:

PR #1930, opened 2020-08-10
Add support for REST Collections API to bulk data tasks

davidmreed:
# Critical Changes

# Changes

- CumulusCI now supports using the REST Collections API in data load, extract, and delete operations. By default, CumulusCI will select an API for you based on data volume (<2000 records uses the REST API, >=2000 uses Bulk); a desired API can be configured via the mapping file.

# Issues Closed


cumulusci/tasks/bulkdata/load.py
```diff
@@ -142,42 +144,58 @@ def _execute_step(
...
+                sobject=mapping["sf_object"],
+                operation=(
+                    DataOperationType.INSERT
+                    if mapping.get("action") == "insert"
+                    else DataOperationType.UPDATE
+                ),
+                api_options={"batch_size": mapping.batch_size},
+                context=self,
+                fields=self._get_columns(mapping),
+            )
```
davisagli, line 180: I would guess this logic for picking the API/operation will be easier to test if you refactor it into a separate method.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -220,6 +231,39 @@ def get_results(self):
...
+    def query(self):
+        pass
+
+    def get_results(self):
+        for chunk in get_batch_iterator(iter(self.ids), 2000):
+            yield from self.sf.restful(
+                f"composite/sobjects/{self.sobject}",
+                method="POST",
+                json={"ids": chunk, "fields": self.fields},
+            )
```
davisagli, line 264: It doesn't look like RestApiRetrieveOperation is used anywhere?

cumulusci/tasks/bulkdata/load.py
```diff
@@ -142,42 +144,58 @@ def _execute_step(
...
 
         statics = self._get_statics(mapping)
-        query = self._query_db(mapping)
 
         total_rows = 0
 
-        # 10,000 is the maximum Bulk API size. Clamping the yield from the query ensures we do not
-        # create more Bulk API batches than expected, regardless of batch size, while capping
-        # memory usage.
-        for row in query.yield_per(10000):
```
davisagli, line 180: I know the bulk API operations now have improved logic for creating batches that fit within limits, but this is still useful for making sure we don't try to load all rows from the SQL db into memory at once, which can be a problem when processing millions of records.

cumulusci/tasks/bulkdata/step.py
davidmreed: I thought I needed it to support Scott's Person Account use case but was mistaken I'll remove it if I don't find a good use for it - it's so nice to be able to pull 2k records at a transaction.

cumulusci/tasks/bulkdata/load.py
davidmreed: I've restored this element.

cumulusci/tasks/bulkdata/delete.py
```diff
@@ -55,6 +56,19 @@ def _init_options(self, kwargs):
...
+        try:
+            self.options["api"] = {
+                "bulk": DataApi.BULK,
+                "rest": DataApi.REST,
+                "smart": DataApi.SMART,
+            }[self.options.get("api", "smart").lower()]
+        except KeyError:
+            raise TaskOptionsError(
+                f"{self.options['api']} is not a valid value for API (valid: bulk, rest, smart"
+            )
```
davisagli, line 68: `api` should be included in `task_options` above (and task docs regenerated)

cumulusci/tasks/bulkdata/load.py
```diff
@@ -138,41 +139,43 @@ def _execute_step(
 
         mapping["oid_as_pk"] = bool(mapping.get("fields", {}).get("Id"))
 
+        local_ids = []
+        query = self._query_db(mapping)
+        operation = (
+            DataOperationType.INSERT
+            if mapping.get("action") == "insert"
+            else DataOperationType.UPDATE
+        )
```
davisagli, line 148: Perhaps the mapping parser should be updated to parse `action` into one of these two types.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -220,6 +232,32 @@ def get_results(self):
...
+        super().__init__(
+            sobject=sobject, api_options=api_options, context=context, query=query
+        )
+        self.fields = fields
+
+    def query(self):
+        # This is not as memory-efficient as we might like.
+        # To preserve the invariant from the Bulk API that
+        # we know the query count before consuming the results,
+        # we use `query_all` instead of `query_all_iter`.
```
davisagli, line 248: Couldn't you get the first batch with `query` and grab the `totalSize`, and then make get_results call `query_more` repeatedly the same way that `query_all_iter` does?

cumulusci/tasks/bulkdata/step.py
```diff
@@ -365,3 +401,153 @@ def get_results(self):
...
+                r[boolean_field] = bool(r[boolean_field])
+
+            # Remove empty fields (different semantics in REST API)
+            # We do this for insert only - on update, any fields set to `null`
+            # are meant to be blanked out.
+            if self.operation is DataOperationType.INSERT:
+                r = {k: r[k] for k in r if r[k] is not None and r[k] != ""}
+
+            r["attributes"] = {"type": self.sobject}
+            return r
```
davisagli, line 436: Consider using longer variable names to make this easier to read.

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -72,6 +72,8 @@ class MappingStep(CCIDictModel):
     static: Dict[str, str] = {}
     filters: List[str] = []
     action: str = "insert"
+    api: DataApi = DataApi.SMART
+    batch_size: int = 200
     oid_as_pk: bool = False  # this one should be discussed and probably deprecated
     record_type: Optional[str] = None  # should be discussed and probably deprecated
     bulk_mode: Optional[
```
davisagli, line 79: Should we add a validator to make sure the batch size can't be set too high (i.e. with different limits based on which api is specified)?

cumulusci/tasks/bulkdata/step.py
```diff
@@ -365,3 +401,153 @@ def get_results(self):
...
+
+def get_query_operation(
+    *,
+    sobject: str,
+    fields: List[str],
+    api_options: Dict,
+    context: Any,
+    query: str,
+    api: DataApi,
+) -> BaseQueryOperation:
```
davisagli, line 497: docstring

cumulusci/tasks/bulkdata/step.py
```diff
@@ -365,3 +401,153 @@ def get_results(self):
                 raise BulkDataException(
                     f"Failed to download results for batch {batch_id} ({str(e)})"
                 )
+
+
+class RestApiDmlOperation(BaseDmlOperation):
```
davisagli, line 406: docstring

cumulusci/tasks/bulkdata/step.py
```diff
@@ -365,3 +401,153 @@ def get_results(self):
...
+def get_dml_operation(
+    *,
+    sobject: str,
+    operation: DataOperationType,
+    fields: List[str],
+    api_options: Dict,
+    context: Any,
+    api: DataApi,
+    volume: int,
+) -> BaseQueryOperation:
```
davisagli, line 534: docstring

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: Great idea. Adding.

cumulusci/tasks/bulkdata/load.py
davidmreed: Yeah, this one is also held back by the legacy code that constructs mapping steps as dicts (to support `after` steps). This ugliness can be refactored out when I fix that.

cumulusci/tasks/bulkdata/step.py
davidmreed: I was shying away from adding more complexity, but turns out to be quite straightforward - thanks for the hint.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -255,7 +251,14 @@ def get_results(self):
...
 
-        yield from (convert(rec) for rec in self.response["records"])
+        while True:
+            yield from (convert(rec) for rec in self.response["records"])
+            if not self.response["done"]:
+                self.response = self.sf.query_more(
+                    self.response["nextRecordsUrl"], identifier_is_url=True
+                )
+            else:
+                return
```
davisagli, line 261:
```suggestion
        while not self.response["done"]:
            yield from (convert(rec) for rec in self.response["records"])
            self.response = self.sf.query_more(
                self.response["nextRecordsUrl"], identifier_is_url=True
            )
```

davidmreed: This would exit early if the query returns a number of results 0<n<batch size, wouldn't it?

davisagli: Yep. I thought it felt too good to be true.

davisagli (approved):

prescod:
@davidmreed  Just some minor bike shedding: did you consider 'rest', 'bulk' or 'auto'?

davidmreed:
@prescod I didn't think about it; I just liked the tone of 'smart', and I think we want to add other features that could be described that way in the future. Willing to reconsider.

prescod:
Auto rings a bit more "standard" to me, but I don't feel strongly about it. Canadian modesty, I guess.

PR #1934, opened 2020-08-11
Add CSV support to push_list

jofsky:
# Critical Changes

# Changes
- We added support for CSV files to the `push_list` task.
# Issues Closed


jofsky:
@jstvz for your implementation idea, is there going to be two options for the csv being passed in?

cumulusci/tasks/push/tasks.py
```diff
@@ -231,13 +237,31 @@ def _init_options(self, kwargs):
...
+            )
         if "namespace" not in self.options:
             self.options["namespace"] = self.project_config.project__package__namespace
         if "batch_size" not in self.options:
             self.options["batch_size"] = 200
+        if "csv" not in self.options and "csv_field_name" in self.options:
+            raise TaskOptionsError("Please provide a csv file for this task to run.")
+        if (
+            "csv" in self.options
+            or "orgs" in self.options
```
jstvz, line 252: I think we only care about `csv_field_name` when `csv` is passed, since `orgs` can only have one org ID per line.

cumulusci/tasks/push/tasks.py
```diff
@@ -194,9 +195,14 @@ def _report_push_status(self, request_id):
 class SchedulePushOrgList(BaseSalesforcePushTask):
 
     task_options = {
+        "csv": {"description": "The CSV file to read.", "required": False},
+        "csv_field_name": {
+            "description": "The CSV field names to parse from the orgs file given. Please provide in string form.",
```
jstvz, line 200: You're using a plural here, when we only want one column, and we'll want to specify the default value. How about:
```suggestion
            "description": "The CSV field name that contains organization IDs. Defaults to 'OrganizationID'",
```

cumulusci/tasks/push/tasks.py
```diff
@@ -1,7 +1,8 @@
 from datetime import datetime
 from datetime import timedelta
 import time
-
+import csv
```
jstvz, line 4: 💭 : We should probably start enforcing isort in pre-commit/lint actions.

cumulusci/tasks/push/tasks.py
```diff
@@ -231,13 +237,31 @@ def _init_options(self, kwargs):
...
+            "csv" in self.options
+            or "orgs" in self.options
+            and "csv_field_name" not in self.options
+        ):
+            self.options["csv_field_name"] = "OrganizationId"
 
     def _get_orgs(self):
-        return self._load_orgs_file(self.options.get("orgs"))
+        if "csv" in self.options:
+            with open(self.options.get("orgs")) as csvfile:
```
jstvz, line 259: You want to open `csv` here instead of `orgs`.

cumulusci/tasks/push/tasks.py
```diff
@@ -231,13 +237,31 @@ def _init_options(self, kwargs):
...
+                reader = csv.DictReader(csvfile)
+                orgs = [row[f"{self.options['csv_field_name']}"] for row in reader]
```
jstvz, line 261: You don't need to wrap `self.options['csv_field_name']` in an fstring.

cumulusci/tasks/push/tasks.py
```diff
@@ -231,13 +237,31 @@ def _init_options(self, kwargs):
...
+                return orgs
```
jstvz, line 262: Since we're not using this variable, you can simply return the list comp.

cumulusci/tasks/push/tasks.py
```diff
@@ -194,9 +195,14 @@ def _report_push_status(self, request_id):
 class SchedulePushOrgList(BaseSalesforcePushTask):
 
     task_options = {
+        "csv": {"description": "The CSV file to read.", "required": False},
```
jstvz, line 198: Let's match the `orgs` description:
```suggestion
        "csv": {"description": "The path to a CSV file to read.", "required": False},
```

cumulusci/tasks/push/tasks.py
```diff
@@ -231,13 +237,31 @@ def _init_options(self, kwargs):
 
         # Set the namespace option to the value from cumulusci.yml if not
         # already set
+        if "orgs" not in self.options and "csv" not in self.options:
```
jstvz, line 240: Seems like you need _either_ `orgs` or `csv`, but not _both_. 

I would expect that, if the user passes both, you'd raise an exception.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+
+def test_schedule_push_org_list_get_orgs_formatted(tmp_path):
+    tmp = tmp_path / "tmp"
+    tmp.mkdir()
+    orgs = tmp / ORG_FILE
+    orgs.write_text(ORG_FILE_TEXT)
+    task = create_task(
+        SchedulePushOrgList,
+        options={
+            "orgs": orgs,
```
jstvz, line 26: 🤔 : Huh, kinda neat that a `Path` object just works here.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+ORG_FILE = "output.csv"
+VERSION = "1.2.3"
+ORG = "00DS0000003TJJ6MAO"
+ORG_FILE_TEXT = """\n00DS0000003TJJ6MAO\n00DS0000003TJJ6MAL"""
+REAL_ORG_FILE_TEXT = """OrganizationId,OrgName,OrgType,OrgStatus,InstanceName,ErrorSeverity,ErrorTitle,ErrorType,ErrorMessage,Gack Id,Stacktrace Id,\n00D5w000004zXXX,,,,,Error,Unexpected Failure,UnclassifiedError,An unexpected failure was experienced during the upgrade. The subscriber's organization was unaffected. Contact salesforce.com Support through your normal channels and provide the following error number: 1351793968-113330 (-1345328791).,1351793968-113330,-1345328791\n00D5w000005VXXX,,,,,Error,Unexpected Failure,UnclassifiedError,An unexpected failure was experienced during the upgrade. The subscriber's organization was unaffected. Contact salesforce.com Support through your normal channels and provide the following error number: 822524189-80345 (-2096886284).,822524189-80345,-2096886284"""
+
+
+def test_schedule_push_org_list_get_orgs_formatted(tmp_path):
+    tmp = tmp_path / "tmp"
+    tmp.mkdir()
```
jstvz, line 20: Do you need a sub-directory here? Can we just use `orgs = tmp_path / ORG_FILE`?

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+        task = create_task(
+            SchedulePushOrgList,
+            options={
+                "version": VERSION,
+                "namespace": NAMESPACE,
+                "start_time": datetime.datetime.now(),
+                "batch_size": 10,
+            },
+        )
+        assert task._get_orgs() == ["00DS0000003TJJ6MAO", "00DS0000003TJJ6MAL"]
```
jstvz, line 111: I don't think this line is executed because of the exception.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+NAMESPACE = "foo"
+NAME = "foo"
+ORG_FILE = "output.csv"
+VERSION = "1.2.3"
+ORG = "00DS0000003TJJ6MAO"
+ORG_FILE_TEXT = """\n00DS0000003TJJ6MAO\n00DS0000003TJJ6MAL"""
+REAL_ORG_FILE_TEXT = """OrganizationId,OrgName,OrgType,OrgStatus,InstanceName,ErrorSeverity,ErrorTitle,ErrorType,ErrorMessage,Gack Id,Stacktrace Id,\n00D5w000004zXXX,,,,,Error,Unexpected Failure,UnclassifiedError,An unexpected failure was experienced during the upgrade. The subscriber's organization was unaffected. Contact salesforce.com Support through your normal channels and provide the following error number: 1351793968-113330 (-1345328791).,1351793968-113330,-1345328791\n00D5w000005VXXX,,,,,Error,Unexpected Failure,UnclassifiedError,An unexpected failure was experienced during the upgrade. The subscriber's organization was unaffected. Contact salesforce.com Support through your normal channels and provide the following error number: 822524189-80345 (-2096886284).,822524189-80345,-2096886284"""
+
+
+def test_schedule_push_org_list_get_orgs_formatted(tmp_path):
```
jstvz, line 18: 💭 : I'm beginning to come around to David's point of view about grouping these tests into a class.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+from cumulusci.tasks.salesforce.tests.util import create_task
+
+SF_ID = "033xxxxxxxxx"
+NAMESPACE = "foo"
+NAME = "foo"
+ORG_FILE = "output.csv"
+VERSION = "1.2.3"
+ORG = "00DS0000003TJJ6MAO"
+ORG_FILE_TEXT = """\n00DS0000003TJJ6MAO\n00DS0000003TJJ6MAL"""
+REAL_ORG_FILE_TEXT = """OrganizationId,OrgName,OrgType,OrgStatus,InstanceName,ErrorSeverity,ErrorTitle,ErrorType,ErrorMessage,Gack Id,Stacktrace Id,\n00D5w000004zXXX,,,,,Error,Unexpected Failure,UnclassifiedError,An unexpected failure was experienced during the upgrade. The subscriber's organization was unaffected. Contact salesforce.com Support through your normal channels and provide the following error number: 1351793968-113330 (-1345328791).,1351793968-113330,-1345328791\n00D5w000005VXXX,,,,,Error,Unexpected Failure,UnclassifiedError,An unexpected failure was experienced during the upgrade. The subscriber's organization was unaffected. Contact salesforce.com Support through your normal channels and provide the following error number: 822524189-80345 (-2096886284).,822524189-80345,-2096886284"""
```
jstvz, line 15: Maybe call it `CSV_FILE_TEXT` to make it a bit clearer?

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,111 @@
...
+            "batch_size": 10,
+            "csv_field_name": "OrganizationId",
+            "csv": ORG_FILE,
+        },
+    )
+    assert task._get_orgs() == ["00D5w000004zXXX", "00D5w000005VXXX"]
+
+
+def test_schedule_push_org_list_get_error(tmp_path):
+    tmp = tmp_path / "tmp"
```
jstvz, line 57: Same here, don't think we need a subdir.

cumulusci/tasks/push/tests/test_push_tasks.py
jofsky: I know I was quite pleased.

cumulusci/tasks/push/tests/test_push_tasks.py
jofsky: Can I address this in a second PR?

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,117 @@
...
+            "version": VERSION,
+            "namespace": NAMESPACE,
+            "start_time": datetime.datetime.now(),
+            "batch_size": 10,
+        },
+    )
+    assert task._get_orgs() == ["00DS0000003TJJ6MAO", "00DS0000003TJJ6MAL"]
+
+
+def test_schedule_push_org_list_get_orgs(tmp_path):
```
jstvz, line 34: I don't think your test name matches what you're testing here. You aren't checking that you can `get_orgs`, you're checking that passing both `csv` and `orgs` as options will raise an exception.

cumulusci/tasks/push/tasks.py
```diff
@@ -229,15 +235,30 @@ def _init_task(self):
...
         # Set the namespace option to the value from cumulusci.yml if not
         # already set
         if "namespace" not in self.options:
             self.options["namespace"] = self.project_config.project__package__namespace
         if "batch_size" not in self.options:
             self.options["batch_size"] = 200
+        if "csv" not in self.options and "csv_field_name" in self.options:
+            raise TaskOptionsError("Please provide a csv file for this task to run.")
+        if "csv" in self.options and "csv_field_name" not in self.options:
+            self.options["csv_field_name"] = "OrganizationId"
```
jstvz, line 253: One last thing:

Since you're also checking whether `csv_field_name` is set in `_get_orgs()` you could dispense with this check by using `self.options.get`:
```python
if "csv" in self.options:
    with open(self.options.get("csv")) as csvfile:
        reader = csv.DictReader(csvfile)
        return [row[self.options.get("csv_field_name", "OrganizationId")] for row in reader]
```

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -0,0 +1,132 @@
...
+            options={
+                "version": VERSION,
+                "namespace": NAMESPACE,
+                "start_time": datetime.datetime.now(),
+                "batch_size": 10,
+            },
+        )
+
+
+# Should set csv_field_name to OrganizationId by default
```
jstvz, line 117:
```suggestion
```

cumulusci/tasks/push/tasks.py
```diff
@@ -229,15 +235,31 @@ def _init_task(self):
...
             self.options["namespace"] = self.project_config.project__package__namespace
         if "batch_size" not in self.options:
             self.options["batch_size"] = 200
+        if "csv" not in self.options and "csv_field_name" in self.options:
+            raise TaskOptionsError("Please provide a csv file for this task to run.")
 
     def _get_orgs(self):
-        return self._load_orgs_file(self.options.get("orgs"))
+        if "csv" in self.options:
+            with open(self.options.get("csv")) as csvfile:
```
davisagli, line 255:
```suggestion
            with open(self.options.get("csv"), newline="", encoding="utf-8") as csvfile:
```

CSV files should always be opened with `newline=""` so that the csv module can handle processing of newlines (otherwise newlines inside quoted fields might not get handled correctly). And it's best to explicitly specify encoding="utf-8" because the default file encoding is different on Windows.

PR #1945, opened 2020-08-14
Make it possible to override the scratch org instance via the org config

davisagli:
This should make it possible to override the instance target of a particular org in MetaCI.

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -187,16 +187,17 @@ def create_org(self):
...
             else "",
             "default": " -s" if self.default else "",
+            "instance": f" instance={self.instance}" if self.instance else "",
             "extraargs": os.environ.get("SFDX_ORG_CREATE_ARGS", ""),
         }
 
         # This feels a little dirty, but the use cases for extra args would mostly
         # work best with env vars
-        command = "force:org:create -f {config_file}{devhub}{namespaced}{days}{alias}{default}{wait} {email} {extraargs}".format(
+        command = "force:org:create -f {config_file}{devhub}{namespaced}{days}{alias}{default}{wait}{email}{instance}{extraargs}".format(
```
davidmreed, line 200: I think we still need a space prior to `extraargs`.

```suggestion
        command = "force:org:create -f {config_file}{devhub}{namespaced}{days}{alias}{default}{wait}{email}{instance} {extraargs}".format(
```

cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -187,16 +187,17 @@ def create_org(self):
             "alias": sarge.shell_format(' -a "{0!s}"', self.sfdx_alias)
             if self.sfdx_alias
             else "",
-            "email": sarge.shell_format('adminEmail="{0!s}"', self.email_address)
+            "email": sarge.shell_format(' adminEmail="{0!s}"', self.email_address)
             if self.email_address and not org_def_has_email
             else "",
             "default": " -s" if self.default else "",
+            "instance": f" instance={self.instance}" if self.instance else "",
             "extraargs": os.environ.get("SFDX_ORG_CREATE_ARGS", ""),
```
davidmreed, line 195: I don't know of any uses cases we've applied for this parameter other than instance. I checked and SFDX will throw an error if `instance` should be specified twice (if we had a MetaCI global override as well as an org-level override). Perhaps ignore `instance` if the environment variable is non-empty, or if they both contain `instance`?

cumulusci/core/config/ScratchOrgConfig.py
davisagli: Good catch.

cumulusci/core/config/ScratchOrgConfig.py
davisagli: Ah, good point. Or we could add a new environment variable for specifying the instance, so that we can do something like:
    
    instance = self.instance or os.environ.get("SFDX_SIGNUP_INSTANCE")

But really I think we probably want to get to a place where the instance can be configured in the db without restarting MetaCI, and so always passed into CumulusCI as part of the org config.

davisagli: @davidmreed I updated it so that it uses self.instance with a fallback to the SFDX_SIGNUP_INSTANCE env var

cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -187,16 +188,17 @@ def create_org(self):
             "alias": sarge.shell_format(' -a "{0!s}"', self.sfdx_alias)
             if self.sfdx_alias
             else "",
-            "email": sarge.shell_format('adminEmail="{0!s}"', self.email_address)
+            "email": sarge.shell_format(' adminEmail="{0!s}"', self.email_address)
             if self.email_address and not org_def_has_email
             else "",
             "default": " -s" if self.default else "",
+            "instance": f" instance={self.instance}" if instance else "",
```
davidmreed, line 195:
```suggestion
            "instance": f" instance={instance}" if instance else "",
```

davisagli: Eesh. I think I need a vacation.

PR #1947, opened 2020-08-17
Feature/robot get locator module

boakley:
This updates the Salesforce.py library to get locators from the highest-numbered locator file if a locator file with the current API can't be found. Prior to this fix, Salesforce.py will fail to load if someone runs in a prerelease before we have an opportunity to create a version-specific lcoator file. This results in all robot tests failing due to keywords not being available.

# Critical Changes

# Changes
* fixed a bug that caused the Salesforce keyword library to not load if it can't find the locators for a prerelease org

# Issues Closed


cumulusci/robotframework/utils.py
```diff
@@ -229,3 +232,31 @@ def wrapper(*args, **kwargs):
...
+    else:
+
+        def by_num(filename):
+            """Pull out the number from a filename"""
+            nums = re.findall(r"_(\d+)", filename)
+            if nums:
+                return int(nums[-1])
+            return 0
+
+        files = sorted(glob.glob1(here, "locators_*.py"), key=by_num)
```
davidmreed, line 258: `glob1()` is [not documented](https://docs.python.org/3/library/glob.html) in the current standard library and I believe is considered private. Can we use `Pathlib`'s glob instead (https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob)?

cumulusci/robotframework/utils.py
```diff
@@ -229,3 +232,31 @@ def wrapper(*args, **kwargs):
...
+        locator_filename = files[-1]
+
+    locator_module_name = f"cumulusci.robotframework.{locator_filename[:-3]}"
```
davidmreed, line 261: Should we log a warning-level message here that we're using a fallback locator library, so it'll be clear in the logs?

cumulusci/robotframework/tests/test_utils.py
```diff
@@ -31,3 +35,50 @@ def example_function():
...
+    @classmethod
+    def setUpClass(cls):
+        # get_locator_module uses __file__ to locate the locator
+        # module. We'll point it to a temporary directory so that
+        # we can control what files we test against.
+
+        cls.tempdir = Path(tempfile.mkdtemp())
+        # sorted alphabetically, 5 should come after 40. The
+        # code should be sorting numerically so that 5 comes
+        # before 40. That's why "locators_5.py" is here.
```
davidmreed, line 50: 👍

cumulusci/robotframework/utils.py
boakley: Hmmm. I don't recall why I switched to glob1. Good catch. I'll use the code from pathlib.

cumulusci/robotframework/utils.py
boakley: Thank you very much for this comment! I started to add it earlier, but logging with robot doesn't always play nice with pytest. However, this forced me to take a step back and look at the problem with fresh eyes, and was reminded that robot has a little-used logging API that handles the case of logging outside of the context of a running robot test. I'm glad I was forced to think a bit harder about this problem.

cumulusci/robotframework/utils.py
```diff
@@ -229,3 +233,34 @@ def wrapper(*args, **kwargs):
...
+        def by_num(filename):
+            """Pull out the number from a filename, or return zero"""
+            nums = re.findall(r"_(\d+)", str(filename))
+            if nums:
+                return int(nums[-1])
+            return 0
+
+        files = sorted(here.glob("locators_*.py"), key=by_num)
+        locator_name = files[-1].stem
+        robot_api.logger.warn(
```
davisagli, line 261: @boakley no big deal, but assuming this follows the normal Python logging interface, `warn` is deprecated; `warning` is preferred: https://docs.python.org/3/library/logging.html#logging.Logger.warning

boakley:
It's a library custom to robot, and `warn` is the documented interface.
https://robot-framework.readthedocs.io/en/3.0/_modules/robot/api/logger.html

PR #1948, opened 2020-08-17
Feature/add pod property to org config

prescod:
Capture the instance_name in the org_config and show it in CCI Org Info.

First step towards showing it in MetaCI.


prescod:
Some tests were failing for legitimate reasons (the tests were out of sync with the code) and I fixed them.

Some Robot tests are also failing but those are almost certainly unrelated to my code.

PR #1952, opened 2020-08-18
Adding all ip address allowed in security settings in Org

jofsky:
# Critical Changes

# Changes

# Issues Closed


davidmreed:
Upon consideration, let's talk through use cases before we move towards shipping this task. It _is_ a security risk to make this change to an org, and one that we have very clear reasons to do, but I want to make sure we can message that clearly or consider the option of keeping this task internal.

davisagli:
Yeah, I don't think this is a good practice. What is the reason for wanting to do this?

jofsky:
When capturing screenshots for metadeploy translations robot plan the ip security settings need to be enabled to allow for the robot automation to properly capture the installation flow otherwise it hangs waiting for a confirmation code, which is unique and only sent by email @davisagli . I don't intend for this to be used on production orgs but for testing this is needed.

davisagli:
Closing per discussion with Josh -- we'll create an internal version of this that only allows the internal IP ranges.

PR #1964, opened 2020-08-20
Fix tests to match SF pre-release

prescod:
Platform changed its error format and the conditions under which it throws exceptions. This updates our tests to match.


davisagli:
@prescod what was the platform change?

cumulusci/robotframework/tests/salesforce/api.robot
```diff
@@ -101,22 +101,13 @@ Collection API Errors Test
         ...  FirstName=User {{number}}
         ...  LastName={{fake.last_name}}
         ...  Xyzzy=qwertz
-    Run Keyword And Expect Error   SalesforceMalformedRequest*   Salesforce Collection Insert  ${objects}
+    Run Keyword And Expect Error   *Xyzzy*   Salesforce Collection Insert  ${objects}
```
boakley, line 104: This modification seems rather imprecise. It's essentially saying we don't care what error is thrown as long as it mentions the bad field. We could instead modify it to do accept a regexp , unless you think this is suffienct

For example, this syntax should work if we care to check for either the old or new error:

```
     Run Keyword And Expect Error   REGEXP: ^(SalesforceMalformedRequest|Error on Object 0):*
     ...  Salesforce Collection Insert  ${objects}
```

prescod: The reason the test broke in the first place was because the regexp was too precise and I'm trying not to repeat that mistake. What value does it offer to prove that the error has the syntax that I expect? The could change the syntax again in the next version. It's just text designed for a human being to read...

cumulusci/robotframework/tests/salesforce/api.robot
```diff
@@ -101,22 +101,13 @@ Collection API Errors Test
...
 
     @{objects} =  Generate Test Data  Contact  20
         ...  FirstName=User {{number}}
         ...  LastName=
     Run Keyword And Expect Error   Error*  Salesforce Collection Insert  ${objects}
 
-    @{objects} =  Generate Test Data  Contact  20
```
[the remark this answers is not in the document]
prescod: The platform semantics seem to changed so that this no longer throws an exception. There is nothing to test, really, anymore. We can't test our exception handling because there is no exception and we shouldn't test that NO exception is thrown because we don't really care if they change it back to start throwing exceptions again in the future.

cumulusci/robotframework/tests/salesforce/api.robot
davisagli: @prescod "SalesforceMalformedRequest" is an exception class in simple-salesforce, not text from the platform. So I'm still not sure I understand what was the change that precipitated this: did the platform start returning a different response status such that simple-salesforce raises a different exception class, or did simple-salesforce change how it reports exceptions?

prescod:
> @prescod what was the platform change?

1. The syntax of the error message for collection insert of a field that doesn't exist changed.

2. Doing a collection update on a field that doesn't exist seems not to give any error at all anymore.

The second one seems a bit fishy but SFDO hardly uses these APIs and I was not feeling altruistic enough to verify the details, make a minimally reproducible test-case in a Salesforce-standard language and make a bug report. Hopefully someone who actually uses and cares about the API regularly will get to it. In any case, it is only turning an exception into a silent ignore which perhaps is a purposeful change and would seldom be a harmful change.

cumulusci/robotframework/tests/salesforce/api.robot
prescod: Yes, the platform went from returning 400, which SimpleSalesforce interprets as an error/exception to returning 200.

It turns out that on deeper investigation, the platform does return the error in a different way in the 200 result, so I've added code to look for it.

davisagli:
@prescod here's the relevant release note: https://releasenotes.docs.salesforce.com/en-us/winter21/release-notes/rn_api_rest.htm#rn_api_rest

prescod:
> @prescod here's the relevant release note: https://releasenotes.docs.salesforce.com/en-us/winter21/release-notes/rn_api_rest.htm#rn_api_rest

I wonder why that i

cumulusci/robotframework/Salesforce.py
```diff
@@ -920,6 +920,12 @@ def salesforce_collection_update(self, objects):
         for record, obj in zip(records, objects):
             obj[STATUS_KEY] = record
 
+        for idx, (record, obj) in enumerate(zip(records, objects)):
+            if record["errors"]:
+                raise AssertionError(
+                    "Error on Object {idx}: {record} : {obj}".format(**vars())
```
davisagli, line 926: Is there a reason you didn't use an f-string here?

PR #1969, opened 2020-08-23
Prevent update_dependencies from attempting to install the same package twice

davidmreed:
# Critical Changes

# Changes

- `update_dependencies` deduplicates its package install list, making it possible to handle situations where the same beta package is reached by two dependency paths.

# Issues Closed


cumulusci/tasks/salesforce/tests/test_UpdateDependencies.py
```diff
@@ -197,6 +197,41 @@ def test_run_task__no_dependencies(self):
...
+        project_config.get_github_api = mock.Mock()
+        project_config.get_static_dependencies = mock.Mock()
+        project_config.get_static_dependencies.return_value = [
+            {
+                "namespace": "dep1",
+                "version": "1.1",
+                "dependencies": [{"namespace": "dep2", "version": "1.1"}],
+            },
+            {"namespace": "dep2", "version": "1.1"},
+        ]
```
davisagli, line 217: This is okay, but since these are all namespace+version dependencies I think you could have just passed them in as the task's `dependencies` option without needing a mock.

PR #1985, opened 2020-09-01
Update related list locator for Winter '21.

boakley:
This has changes to all of the related list locators for winter '21, along with a few other changes that cropped up during prerelease testing. 

Related lists may be rendered in a couple of different ways, this new locator should work for either way. I also took the liberty of refactoring the 'load related list' keyword to use some features in SeleniumLibrary that weren't available when the keyword was written nearly two years ago. I also added some tests for this keyword, which we didn't previously have.

I'm still working on some automated tests. A couple of the issues can be reproduced via an NPSP test, but I can't yet reproduce them in our CumulusCI tests. 

# Critical Changes

# Changes

* The following robot keywords have been updated to work with Winter '21: 
  - `Load related list`
  - `Click related list button`
  - `Click related item link`
  - `Click related item popup link`
  - `Go to object home`
  - `Go to object list`
  - `Go to record home`
* The keyword `Load related list` has been rewritten to be slightly more efficient. It also has a new parameter `tries` which can be used if the target is more than 1000 pixels below the bottom of the window. 

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/Salesforce.py
+++ b/cumulusci/robotframework/Salesforce.py
@@ -567,6 +567,7 @@ class Salesforce(object):
                 break
         self.selenium.set_focus_to_element(menu_locator)
         self._jsclick(menu_locator)
+        self.wait_for_aura()
 
     def _get_input_field_locator(self, name):
         """Given an input field label, return a locator for the related input field
@@ -604,8 +605,7 @@ class Salesforce(object):
         self._focus(field)
         if field.get_attribute("value"):
             self._clear(field)
-        actions = ActionChains(self.selenium.driver)
-        actions.send_keys_to_element(field, value).perform()
+        field.send_keys(value)
 
     def _focus(self, element):
         """Set focus to an element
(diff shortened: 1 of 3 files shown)


cumulusci/robotframework/Salesforce.py
```diff
@@ -238,26 +241,29 @@ def click_object_button(self, title):
...
-    def load_related_list(self, heading):
+    @capture_screenshot_on_error
+    def load_related_list(self, heading, tries=10):
         """Scrolls down until the specified related list loads.
+
+        If the related list isn't found, the keyword will scroll down
+        in 100 pixel increments to trigger lightning into loading the
+        list. This process of scrolling will be repeated until the
+        related list has been loaded or we've tried several times
+        (the default is 10 tries)
```
davisagli, line 252: Are you sure 10 is going to be enough? It was previously 50. There can be pages with quite a few related lists.

boakley: No, I'm not 100% sure. Though I think for our testing scenarios it probably is. However, now that it's parameterized, if a tester knows they are testing a page with a lot of data then they can set it to a very large number. I can easily bump the default number back up if you think I should. 50 just seemed like overkill, at least for the tests I was running. 

Ideally the code would scroll until it discovers there's nothing left to scroll rather than scrolling a fixed number of times. That was going to require more work and more testing, and I didn't want to hold up this fix for that.

Issue #2001, opened 2020-09-05
CumulusCI incorrectly persists extracted data when using Autoincrement Primary Keys and a step contains an `after` lookup


davidmreed:
CumulusCI does not correctly rewrite lookups to autoincrement PKs for `after` steps. This is because rewriting is done after each step, meaning that `after` references are not resolvable at the time CumulusCI attempts to rewrite.

The outcome is an extracted data set that incorrectly contains Salesforce object Ids in relationship fields which are not resolvable at load time, breaking the data set.

To fix this issue, we will move rewriting object Ids to the end of the extraction, instead of occurring with each step.

Reference W-7898122.

davidmreed:
Fixed in #2002.

PR #2002, opened 2020-09-06
Fix bug with after steps in autopk extracts

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue (#2001) that caused CumulusCI to extract invalid data sets when using `after:` steps with autoincrement primary keys.

### DIFF
--- a/cumulusci/tasks/bulkdata/extract.py
+++ b/cumulusci/tasks/bulkdata/extract.py
@@ -83,7 +83,7 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
             soql = self._soql_for_mapping(mapping)
             self._run_query(soql, mapping)
 
-        self._drop_sf_id_columns()
+        self._map_autopks()
 
         if self.options.get("sql_path"):
             self._sqlite_dump()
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/bulkdata/extract.py
```diff
@@ -251,8 +249,18 @@ def strip_name_field(record):
 
         self.session.commit()
 
-        if lookup_keys and not mapping["oid_as_pk"]:
-            self._convert_lookups_to_id(mapping, lookup_keys)
+    def _map_autopks(self):
+        # Convert Salesforce Ids to autopks
+        for m in self.mapping.values():
+            lookup_keys = list(m.get("lookups", {}).keys())
```
davisagli, line 255: You should be able to assume that the mapping has a `lookups` key and that it is a dict, thanks to https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/tasks/bulkdata/mapping_parser.py#L71

davisagli (approved):

cumulusci/tasks/bulkdata/extract.py
davidmreed: We should, but: `after` step mappings are synthesized in code and are still `dict`s. It is on my list to fix that and then go through and remove all `dict`-ish code throughout bulk data.

PR #2008, opened 2020-09-09
Refactor Bulk Data to remove dict code

davidmreed:
All Bulk Data code is converted to directly use Pydantic models. Some older code refactored to suit this new strategy.

# Critical Changes

- The `oid_as_pk` key is no longer supported in bulk data mappings. (This key was already deprecated). Select object Id mode by including the `Id` field in mappings.

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata/extract.py
+++ b/cumulusci/tasks/bulkdata/extract.py
@@ -130,7 +130,7 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
     def _fields_for_mapping(self, mapping):
         """Return a flat list of fields for this mapping."""
         fields = []
-        if not mapping["oid_as_pk"]:
+        if not mapping.get_oid_as_pk():
             fields.append("Id")
         fields += [field["sf"] for field in fields_for_mapping(mapping)]
 
@@ -138,7 +138,7 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
 
     def _soql_for_mapping(self, mapping):
         """Return a SOQL query suitable for extracting data for this mapping."""
-        sf_object = mapping["sf_object"]
+        sf_object = mapping.sf_object
         fields = self._fields_for_mapping(mapping)
         soql = f"SELECT {', '.join(fields)} FROM {sf_object}"
         if mapping.record_type:
@@ -149,8 +149,8 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
     def _run_query(self, soql, mapping):
         """Execute a Bulk or REST API query job and store the results."""
         step = get_query_operation(
-            sobject=mapping["sf_object"],
-            api=mapping["api"],
+            sobject=mapping.sf_object,
+            api=mapping.api,
             fields=self._fields_for_mapping(mapping),
             api_options={},
             context=self,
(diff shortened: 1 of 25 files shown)


cumulusci/tasks/bulkdata/extract.py
```diff
@@ -175,39 +166,27 @@ def _import_results(self, mapping, step):
...
+            and "Name" in field_map
             and self.org_config.is_person_accounts_enabled
         ):
             # Bump indices by one since record's ID is the first column.
-            Name_index = columns.index(mapping["fields"]["Name"]) + 1
-            IsPersonAccount_index = (
-                columns.index(mapping["fields"]["IsPersonAccount"]) + 1
-            )
+            Name_index = columns.index(mapping.fields["Name"])
+            IsPersonAccount_index = columns.index(mapping.fields["IsPersonAccount"])
```
davidmreed, line 189: The Id field is now included so no need for math here.

cumulusci/tasks/bulkdata/tests/test_extract.py
```diff
@@ -339,35 +335,13 @@ def test_import_results__autopk(self, csv_mock):
...
                     connection=task.session.connection.return_value,
-                    table="Opportunity_sf_ids",
+                    table=mapping.get_sf_id_table(),
                     columns=["sf_id"],
                     record_iterable=csv_mock.return_value,
                 ),
             ]
         )
 
-    def test_import_results__no_columns(self):
```
davidmreed, line 349: Removed this test because I changed the behavior: it is perfectly legal, and occasionally desirable, to extract an object with no columns (save the automatically-included `Id`).

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -98,10 +124,9 @@ def record_type_is_deprecated(cls, v):
     @validator("oid_as_pk")
     @classmethod
     def oid_as_pk_is_deprecated(cls, v):
-        logger.warning(
-            "oid_as_pk is deprecated. Just supply an Id column declaration and it will be inferred."
+        raise ValueError(
+            "oid_as_pk is no longer supported. Include the Id field if desired."
         )
```
davisagli, line 129: Do you think there's any risk of still needing to run old metadeploy plans which use oid_as_pk?

davidmreed: Good question. Very few MetaDeploy plans run data loads at all. NPSP's Trailhead installer does, but we shouldn't have a need to use an old version since that would not replicate a customer org. Grants Management does, but is very recent; foundationConnect's partner enablement installer does, but was never rolled out publicly.

@jstvz @jofsky Are you aware of any potential risk to older EDU installers that use data load tasks?

davisagli (commented):
@davidmreed Will this require updates to any custom subclasses of these tasks in product repositories?

davidmreed:
@davisagli  I don't think that we have any custom subclasses left. PMM's has been removed and in a quick review I don't see any on the EDU projects.

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -71,16 +71,42 @@ class MappingStep(CCIDictModel):
...
+
+        fields.update(self.fields)
+        fields.update(
+            {
+                lookup: self.lookups[lookup].get_lookup_key_field()
+                for lookup in self.lookups
+            }
+        )
+
+        return fields
```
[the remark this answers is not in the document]
davidmreed: Done.

davisagli (approved):

cumulusci/tasks/bulkdata/mapping_parser.py
davidmreed: I took a look as well and don't see any clear risks.

PR #2015, opened 2020-09-15
Bulk Data Relative Dates

davidmreed:
# Critical Changes

# Changes

- CumulusCI's `load_dataset` and `extract_dataset` tasks now support relative dates. To take advantage of relative dates, include the `anchor_date` key (with a date in `YYYY-MM-DD` format) in each mapping step you wish to relativize. On extract, dates will be modified to be the same interval from the anchor date as they are from the current date; on load, dates will be modified to be the same interval from today's date as they are from their anchor. Both date and date-time fields are supported.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata/dates.py
+++ b/cumulusci/tasks/bulkdata/dates.py
@@ -69,21 +69,15 @@ def adjust_relative_dates(
     return r
 
 
-# The way Salesforce formats ISO8601 date-times is not quite compatible
-# with Python's datetime. Salesforce returns, e.g., "2020-09-14T20:00:17.000+0000",
-# while Python wants a : character in the timezone: "2020-09-14T20:00:17.000000+00:00".
-
-
 def datetime_from_salesforce(d):
     """Create a Python datetime from a Salesforce-style ISO8601 string"""
-    # Convert Salesforce's `+0000`, which Python would want as `+00:00`
-    return datetime.strptime(d[:-5] + "+00:00", "%Y-%m-%dT%H:%M:%S.%f%z")
+    return datetime.strptime(d, "%Y-%m-%dT%H:%M:%S.%f%z")
 
 
 def salesforce_from_datetime(d):
     """Create a Salesforce-style ISO8601 string from a Python datetime"""
     # Convert microseconds to milliseconds. Salesforce uses 3 decimals for milliseconds;
-    # Python uses 6 for microseconds.
+    # Python uses 6 for microseconds. Truncate here; the final 3 digits should be 0.
     return d.strftime("%Y-%m-%dT%H:%M:%S.{}+0000").format(str(d.microsecond)[:3])
 
 
--- a/cumulusci/tasks/bulkdata/dates.py
+++ b/cumulusci/tasks/bulkdata/dates.py
@@ -69,6 +69,14 @@ def adjust_relative_dates(
     return r
 
 
+# The Salesforce API returns datetimes with millisecond resolution, but milliseconds
+# are always zero (that is, .000). Python does parse this with strptime.
+# Python renders datetimes into ISO8601 with microsecond resolution (.123456),
+# which Salesforce won't accept - we need exactly three digits, although they are
+# currently ignored. Python also right-truncates to `.0`, which Salesforce won't take.
+# Hence this clumsy workaround.
+
+
 def datetime_from_salesforce(d):
     """Create a Python datetime from a Salesforce-style ISO8601 string"""
     return datetime.strptime(d, "%Y-%m-%dT%H:%M:%S.%f%z")
@@ -76,9 +84,9 @@ def datetime_from_salesforce(d):
 
 def salesforce_from_datetime(d):
     """Create a Salesforce-style ISO8601 string from a Python datetime"""
-    # Convert microseconds to milliseconds. Salesforce uses 3 decimals for milliseconds;
-    # Python uses 6 for microseconds. Truncate here; the final 3 digits should be 0.
-    return d.strftime("%Y-%m-%dT%H:%M:%S.{}+0000").format(str(d.microsecond)[:3])
+    return d.strftime("%Y-%m-%dT%H:%M:%S.{}+0000").format(
+        str(d.microsecond)[:3].ljust(3, "0")
+    )
 
 
 # Python 3.6 doesn't support the isoformat()/fromisoformat() methods. Shims.
--- a/cumulusci/tasks/bulkdata/dates.py
+++ b/cumulusci/tasks/bulkdata/dates.py
@@ -1,0 +1,113 @@
+from datetime import date, datetime
+from typing import List, Optional
+from cumulusci.core.config.OrgConfig import OrgConfig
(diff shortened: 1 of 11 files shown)


davidmreed:
Passes end-to-end tests.

docs/bulk_data.rst
```diff
@@ -173,6 +173,30 @@ the same Record Type upon load.
 It's recommended that new datasets use Record Type mapping by including the ``RecordTypeId`` 
 field. Using ``record_type`` will result in CumulusCI issuing a warning.
 
+Relative Dates
```
spelak-salesforce, line 176: @davidmreed , can you add a sample mapping that uses relative dates?   So, as a customer, I am confident I am setting up my mapping correctly.

davidmreed: Yep. Duly added.

cumulusci/tasks/bulkdata/dates.py
```diff
@@ -0,0 +1,115 @@
...
+        target_anchor = mapping.anchor_date
+    else:
+        current_anchor = mapping.anchor_date
+        target_anchor = date.today()
+
+    r = record.copy()
+
+    for index in date_fields:
+        if operation is DataOperationType.QUERY:
+            index += 1  # For the Id field.
```
davisagli, line 51: This arithmetic to target the right column looks error prone. This would be a bigger effort and doesn't need to happen right now, but I wonder if we could use a different internal representation of rows so that they are still stored compactly but can be addressed by name. i.e. something like namedtuple or dataclasses

cumulusci/tasks/bulkdata/dates.py
```diff
@@ -0,0 +1,115 @@
...
+    """Adjust this_date to be relative to target_anchor instead of current_anchor"""
+    return target_anchor + (this_date - current_anchor)
+
+
+def _convert_datetime(target_anchor, current_anchor, this_datetime):
+    """Adjust this_datetime to be relative to target_anchor instead of current_anchor"""
+    return datetime.combine(
+        _convert_date(target_anchor, current_anchor, this_datetime.date()),
+        this_datetime.time(),
+    )
```
davisagli, line 115: How about "offset" instead of "convert"? "Convert" makes me think it's going to change the type.

cumulusci/tasks/bulkdata/dates.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+def salesforce_from_datetime(d):
+    """Create a Salesforce-style ISO8601 string from a Python datetime"""
+    return d.strftime("%Y-%m-%dT%H:%M:%S.{}+0000").format(
+        str(d.microsecond)[:3].ljust(3, "0")
+    )
+
+
+# Python 3.6 doesn't support the isoformat()/fromisoformat() methods. Shims.
```
davisagli, line 92: I think it supports isoformat but not fromisoformat

cumulusci/tasks/bulkdata/dates.py
```diff
@@ -0,0 +1,115 @@
...
+
+def datetime_from_salesforce(d):
+    """Create a Python datetime from a Salesforce-style ISO8601 string"""
+    return datetime.strptime(d, "%Y-%m-%dT%H:%M:%S.%f%z")
+
+
+def salesforce_from_datetime(d):
+    """Create a Salesforce-style ISO8601 string from a Python datetime"""
+    return d.strftime("%Y-%m-%dT%H:%M:%S.{}+0000").format(
+        str(d.microsecond)[:3].ljust(3, "0")
```
davisagli, line 88: TIL about ljust!

davisagli (commented):
I'm excited to finally add this feature!

Do you think we'll ever want an object where some date fields are fixed and others float? Like, I dunno, a table of when historic events that includes both the date when the event happened and the date it was last reviewed?

davidmreed:
> Do you think we'll ever want an object where some date fields are fixed and others float? Like, I dunno, a table of historic events that includes both the date when the event happened and the date it was last reviewed?

Yes, I think that's definitely possible, but it's not one of the first use cases that any of our teams have articulated. For simplicity, I left it at "all or nothing on per mapping step", but I think that extension is quickly add-able.

cumulusci/tasks/bulkdata/dates.py
davidmreed: I like `offset`. I'll make that change.

cumulusci/tasks/bulkdata/dates.py
davidmreed: Ah, you're right. I'll correct the comment. Do you have any concerns about keeping the shim? I ended up liking the explicitness given the other minor mismatches between `datetime` and the Salesforce API's behavior.

cumulusci/tasks/bulkdata/dates.py
```diff
@@ -0,0 +1,115 @@
...
+        for f in mapping.get_fields_by_type("date", org_config)
+        if f in mapping.fields
+    ]
+    date_time_fields = [
+        fields.index(f)
+        for f in mapping.get_fields_by_type("datetime", org_config)
+        if f in mapping.fields
+    ]
+
+    return (date_fields, date_time_fields)
```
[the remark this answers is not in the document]
davidmreed: Good point. I'll also add the current date to the context so that it's retrieved exactly once per operation.

cumulusci/tasks/bulkdata/dates.py
davisagli: I'll allow it

cumulusci/tasks/bulkdata/dates.py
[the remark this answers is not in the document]
davidmreed: Done.

cumulusci/tasks/bulkdata/dates.py
davidmreed: Done

cumulusci/tasks/bulkdata/dates.py
```diff
@@ -0,0 +1,115 @@
...
+    # Determine the direction in which we are converting.
+    # For extracts, we convert the date from today-anchored to mapping.anchor_date-anchored.
+    # For loads, we do the reverse.
+
+    if operation is DataOperationType.QUERY:
+        current_anchor = date.today()
+        target_anchor = mapping.anchor_date
+    else:
+        current_anchor = mapping.anchor_date
+        target_anchor = date.today()
```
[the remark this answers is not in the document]
davidmreed: Done; moved to the relative date context

cumulusci/tasks/bulkdata/dates.py
davidmreed: Agreed. This caused me no end of headaches while implementing REST API support. I'd definitely like to revisit.

On an allied but much larger front, I'd like to seriously look at deprecating oid-as-pk mode. Paul's making autopks a lot more efficient, and having two modes adds a lot of surface area for wrong-expectation mistakes.

davisagli (approved):

PR #2016, opened 2020-09-15
Snowfakery 1.1

prescod:
Support Snowfakery 1.1:

https://github.com/SFDO-Tooling/Snowfakery/releases/tag/v1.1

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -33,7 +33,7 @@ sarge==0.1.6
 selenium==3.141.0
 simple-salesforce==1.10.1
 six==1.15.0
-snowfakery==1.0.1
+snowfakery==1.1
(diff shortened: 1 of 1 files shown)


davisagli:
@prescod looks like you need to add version.txt to Snowfakery's manifest and make a new release

prescod:
It's in there. Do you have a guess as to why it wouldn't be picked up?

https://github.com/SFDO-Tooling/Snowfakery/blob/master/MANIFEST.in

prescod:
I guess MANIFEST.in is some kind of legacy file that's ignored by modern Pythons???

davisagli:
@prescod I'm fairly sure that's not it. What commands are you using to make the release?

PR #2017, opened 2020-09-15
Fix JWT aud for instanceless sandboxes

davisagli:
When the SFDX_CLIENT_ID and SFDX_HUB_KEY environment variables are set, CumulusCI uses the JWT oauth flow to get a fresh access token for an org. In order to do that, it needs to correctly specify `aud` in the JWT as either https://login.salesforce.com or https://test.salesforce.com

Previously, we used the instance URL to pick the `aud` value (based on whether the URL contained a sandbox instance name like cs22, or not). But lately it is possible for there to be sandboxes with "instanceless" instance URLs that do not include the instance name.

As an alternative, this passes the `id` from the org config into the jwt_session function. It will not always be present, but it should be for orgs that were connected using oauth, and it also starts with either the production or sandbox login URL (login.salesforce.com or test.salesforce.com). If it is present, jwt_session uses it as the signal for which of those to use for the `aud`, instead of using the instance_url.

This is the first step in fixing https://github.com/SFDO-Tooling/MetaCI/issues/1497. The remaining step will be to update MetaCI to use jwt_session from CumulusCI instead of its own copy so that we don't have to maintain it in 2 different places.

# Critical Changes

# Changes

# Issues Closed
- Fixed a bug that prevented using JWT auth with sandboxes if the sandbox's instance_url did not include an instance name.

### DIFF
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -57,7 +57,11 @@ class OrgConfig(BaseConfig):
             SFDX_HUB_KEY = os.environ.get("SFDX_HUB_KEY")
             if SFDX_CLIENT_ID and SFDX_HUB_KEY:
                 info = jwt_session(
-                    SFDX_CLIENT_ID, SFDX_HUB_KEY, self.username, self.instance_url
+                    SFDX_CLIENT_ID,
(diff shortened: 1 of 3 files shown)


Br4nd0R (approved):
LGTM!

PR #2019, opened 2020-09-16
Add task for creating Permission Set with specific User Permissions

davidmreed:
# Critical Changes

# Changes

- The task `create_permission_set` allows for creating and assigning a Permission Set that enables specific User Permissions. (Note: other types of permissions are not yet supported).
- The task `create_bulk_data_permission_set` creates a Permission Set with the Hard Delete and Set Audit Fields permissions for use with data load operations. The org permission to allow Set Audit Fields must be turned on. 

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -85,6 +85,20 @@ tasks:
             path: src
             revert_path: src.orig
         group: Salesforce Metadata
+    create_permission_set:
+        group: Metadata Transformations
+        description: Creates a Permission Set with specified User Permissions and assigns it to the running user.
+        class_path: cumulusci.tasks.salesforce.create_permission_sets.CreatePermissionSet
+    create_bulk_data_permission_set:
+        group: Data Operations
+        description: "Creates a Permission Set with the Hard Delete and Set Audit Fields user permissions. NOTE: the org setting to allow Set Audit Fields must be turned on."
+        class_path: cumulusci.tasks.salesforce.create_permission_sets.CreatePermissionSet
+        options:
+            api_name: CumulusCI_Bulk_Data
+            label: CumulusCI Bulk Data
+            user_permissions:
+                - PermissionsBulkApiHardDelete
+                - PermissionsCreateAuditFields
     create_unmanaged_ee_src:
         description: Modifies the src directory for unmanaged deployment to an EE org
         class_path: cumulusci.tasks.metadata.ee_src.CreateUnmanagedEESrc
--- a/cumulusci/tasks/salesforce/create_permission_sets.py
+++ b/cumulusci/tasks/salesforce/create_permission_sets.py
@@ -30,6 +30,6 @@ class CreatePermissionSet(BaseSalesforceApiTask):
                 **{perm: True for perm in self.options["user_permissions"]},
             }
         )
-        result = self.sf.PermissionSetAssignment.create(
+        self.sf.PermissionSetAssignment.create(
             {"AssigneeId": self.org_config.user_id, "PermissionSetId": result["id"]}
         )
--- a/cumulusci/tasks/salesforce/create_permission_sets.py
+++ b/cumulusci/tasks/salesforce/create_permission_sets.py
@@ -5,7 +5,7 @@ from cumulusci.core.utils import process_list_arg
 class CreatePermissionSet(BaseSalesforceApiTask):
     task_options = {
         "api_name": {
-            "description": "API names of generated Permission Set",
+            "description": "API name of generated Permission Set",
             "required": True,
         },
         "label": {"description": "Label of generated Permission Set"},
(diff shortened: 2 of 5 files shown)


cumulusci/tasks/salesforce/create_permission_sets.py
```diff
@@ -0,0 +1,35 @@
+from cumulusci.tasks.salesforce.BaseSalesforceApiTask import BaseSalesforceApiTask
+from cumulusci.core.utils import process_list_arg
+
+
+class CreatePermissionSet(BaseSalesforceApiTask):
+    task_options = {
+        "api_name": {
+            "description": "API names of generated Permission Set",
```
Br4nd0R, line 8:
```suggestion
            "description": "API name of generated Permission Set",
```
Is this suppose to be singular?

cumulusci/tasks/salesforce/create_permission_sets.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _run_task(self):
+        result = self.sf.PermissionSet.create(
+            {
+                "Name": self.options["api_name"],
+                "Label": self.options.get("label") or self.options["api_name"],
+                **{perm: True for perm in self.options["user_permissions"]},
+            }
+        )
+        result = self.sf.PermissionSetAssignment.create(
```
Br4nd0R, line 33: If we aren't referencing `result` do we need to assign it a new value here? Also, if there's an error creating either the PS or PSA I'm guessing that SFDX is bubbling up the message to the user?

cumulusci/tasks/salesforce/create_permission_sets.py
davidmreed: Yep. Thanks for the catch.

docs/tasks.rst
```diff
@@ -693,6 +693,76 @@ Options
...
+
+
+Options
+------------------------------------------
+
+
+``-o api_name APINAME``
+	 *Required*
+
+	 API names of generated Permission Set
```
Br4nd0R, line 716:
```suggestion
	 API name of generated Permission Set
```
Again, just seeing if this should be singular?

cumulusci/tasks/salesforce/create_permission_sets.py
davidmreed: We don't, I'll fix that.

Originally I had written error-handling code here but I quickly found that any mishap in creation just results in simple_salesforce throwing an exception rather than me getting back a message.

docs/tasks.rst
Br4nd0R: Same comment on the additional occurrences below.

Br4nd0R (approved):
Such clean code, I love the readability. Just a couple minor things I had questions on.

cumulusci/cumulusci.yml
```diff
@@ -85,6 +85,20 @@ tasks:
             path: src
             revert_path: src.orig
         group: Salesforce Metadata
+    create_permission_set:
+        group: Metadata Transformations
+        description: Creates a Permission Set with specified User Permissions and assigns it to the running user.
+        class_path: cumulusci.tasks.salesforce.create_permission_sets.CreatePermissionSet
+    create_bulk_data_permission_set:
+        group: Data Operations
+        description: "Creates a Permission Set with the Hard Delete and Set Audit Fields user permissions. NOTE: the org setting to allow Set Audit Fields must be turned on."
```
davisagli, line 94: Is there anything we can link to with an example of that setting?

cumulusci/cumulusci.yml
```diff
@@ -85,6 +85,20 @@ tasks:
...
+        class_path: cumulusci.tasks.salesforce.create_permission_sets.CreatePermissionSet
+        options:
+            api_name: CumulusCI_Bulk_Data
+            label: CumulusCI Bulk Data
```
davisagli, line 98: How about leave CumulusCI out of the name, but set a description for the permission set that explains what its for and that it was created by this task.

cumulusci/cumulusci.yml
davidmreed: Yes. I'll add a short section to the Bulk Data docs to call out that this task can help, and link to the Salesforce help article.

cumulusci/cumulusci.yml
davidmreed: Can you say more? I wanted to put CumulusCI in the name to make its purpose clear at a glance.

davisagli: Open to discussion but my thinking is: There's nothing about the permission set that is inherently tied to CumulusCI once it's in use in an org. We don't label other things that CumulusCI does in an org as CumulusCI-related just because that's how it happened to get there. I can understand wanting to provide some trail to understand how it got there, which is why I suggested putting it in the description, but including it in the developer name in particular seems heavy handed since someone may decide the permission set is also useful for non-CumulusCI-related purposes.

davidmreed: Hmm. I hear you, but at the same time, this Permission Set has a somewhat odd combination of permissions - it's really useful only for performing data operations. Since it's unmanaged, the end user is free to change the API name and label at any time.

cumulusci/cumulusci.yml
davidmreed: Documentation added with link.

cumulusci/cumulusci.yml
davisagli: Ok, fair enough.

davisagli (approved):

PR #2025, opened 2020-09-16
Week 38 dependency updates

Br4nd0R:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ click==7.1.2
 coloredlogs==14.0
 cryptography==3.1
 docutils==0.16
-Faker==4.1.2
+Faker==4.1.3
 fs==2.4.11
 github3.py==1.3.0
 humanfriendly==8.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -33,7 +33,7 @@ sarge==0.1.6
 selenium==3.141.0
 simple-salesforce==1.10.1
 six==1.15.0
-snowfakery==1.0.1
+snowfakery==1.1
```
davisagli, line 36:
```suggestion
snowfakery==1.1.1
```

davisagli:
@Br4nd0R @prescod looks like Faker 4.1.3 removed a datetime_safe function that snowfakery uses (https://github.com/joke2k/faker/pull/1258). I think we should:
1. stick with faker 4.1.2 as the cumulusci dependency for now
2. update snowfakery to no longer use datetime_safe

davisagli (approved):

PR #2026, opened 2020-09-16
Snowfakery 1.1.1

prescod:
Update to Snowfakery 1.1.1


davisagli:
@Br4nd0R is going to do this as part of the combined dependency update.

PR #2028, opened 2020-09-16
CumulusCI Release Branches

Br4nd0R:
* The MergeBranch class has been refactored to detect and propagate commits made to release branches to other existing release branches that correspond to a future release.
* The children_only option has been removed from MergeBranch.
* MergeBranch now has an update_future_releases option that defaults to False. If set to true, commit propagation between release branches occurs (see updated docs for more info).
* The new headers Release Branches and Release to (Future) Release Merges in cumulusci_flow.rst covers this piece of functionality and why teams might have a need for it.
* The old MergeBranch class has been renamed to MergeBranchOld. It is being kept in-place should we need to revert (by editing the class associated with the task) upon discovering a bug. A work item has been created to ensure this code is removed once everything is confirmed working in production.

# Changes
We've made some updates to the tasks for automatically merging changes between branches:
*The `github_master_to_feature` task has been renamed to `github_automerge_main`. It still merges changes from the default branch to feature branches. In the case of an orphaned feature branch (a branch with a name like feature/parent__child where feature/parent does not exist as its own branch), the `github_automerge_main` branch will no longer merge to the orphaned branch.
* The `github_parent_to_children` task has been renamed to `github_automerge_feature`. It still merges changes from feature branches to their children (e.g. feature/parent would be merged to feature/parent__child). It is now possible to use multiple double-underscores to create more deeply nested children, and the task will only merge to the next level (e.g. feature/parent would merge to feature/parent__child which would merge to feature/parent__child__grandchild).
* The `children_only` option for these tasks has been removed. The strategy for picking which branches to target for merging is now determined by the source branch.
* Feature branches named with an integer (e.g. feature/228) are considered to be "release" branches where a future release is under development. The `github_automerge_feature` task now supports merging from a release branch to other release branches with a higher number. Enable this feature using the `update_future_releases` option, which is off by default.


cumulusci/cumulusci.yml
```diff
@@ -219,7 +219,6 @@ tasks:
         description: Merges the latest commit on a parent feature branch into all child feature branches
         class_path: cumulusci.tasks.github.MergeBranch
         options:
-            children_only: True
             source_branch: $project_config.repo_branch
         group: GitHub
```
davisagli, line 223: Should we rename this task to `github_automerge_feature` like I suggested in the first automerge RFC? It's no longer just handling parent-to-child.

docs/cumulusci_flow.rst
```diff
@@ -36,36 +36,56 @@ When a Pull Request is approved and passing build, it is merged using the Merge
 
 Once the Pull Request is merged, the feature branch is deleted.
 
-CumulusCI and Feature Branches
-------------------------------
+Branch Configuraiton
```
davisagli, line 39:
```suggestion
Branch Configuration
```

docs/cumulusci_flow.rst
```diff
@@ -36,36 +36,56 @@ When a Pull Request is approved and passing build, it is merged using the Merge
...
 
-Auto-Merging Main to Feature Branches
-=======================================
+Auto Merging
+============
+CumulusCI helps to keep large diffs and merge conflicts from being the norm. CumulusCI's auto-merge functionality helps teams:
+
+   * Keep feature branches up-to-date with the ``main`` branch (main to feature merges)
+   * Manage long-lived feature branches for larger features worked on by multiple developers (parent to child merges)
+   * Mange large releases that occur several times a year (release to future release merges).
```
davisagli, line 67:
```suggestion
   * Manage large releases that occur several times a year (release to future release merges).  
```

docs/cumulusci_flow.rst
```diff
@@ -36,36 +36,56 @@ When a Pull Request is approved and passing build, it is merged using the Merge
...
 
 To understand the benefit of auto-merging to feature branches, consider the following scenario: A developer starts work on a feature branch, puts in a few weeks on it, and then has to leave unexpectedly for a few months.  While they are on leave, their feature branch gets automatically updated with any new commits on main and rebuilt.  A few weeks into their leave, a new commit on main gets merged to their feature branch and breaks the build.  When the developer returns after their leave, they can look at the build history to find which commit from main broke their feature branch.
 
 Without auto-merging, the developer would return, merge main into their feature branch, and then have to sift through all the commits to main during their leave to figure out which one broke their feature branch.  More testing and build history is always a good thing in addition to the other benefits we gain from auto-merging.
 
-CumulusCI and Auto-Merging to Feature Branches
-----------------------------------------------
+CumulusCI facilitates the auto-merge to feature branches via the ``github_master_to_feature`` task which is included by default in the ``release_beta`` flow.
```
davisagli, line 83: Should we rename this task to github_automerge_main?

docs/cumulusci_flow.rst
```diff
@@ -78,12 +98,60 @@ If this combination of named parent and child branches exist, the auto-merging f
...
+
+CumulusCI facilitates parent to child auto-merges via the `github_parent_to_children` task, which is included by deault in the `ci_feature` flow.  If a parent feature branch passes the build, it is automatically merged into all child branches.
+
+Release Branches
+----------------
+Some teams deliver large releases several times a year.
+For this type of release cadence, Salesforce.org uses a special type of branch referred to as a release branch. These long-lived branches are created off of the ``main`` branch, serve as the target branch for all features associated with that release and are eventually merged back to the ``main`` branch when a release occurs.
+To be able to clearly track what work is associated with a specific release, release branches adhere to the following:
+   * They are the parent branches of ALL feature work associated with a release. Put another way; all feature branches use the parent-child naming convention with its target release branch.
+   * Use a strict naming format: ``feature/release_num`` where ``release_num`` is a valid integer.
```
davisagli, line 111: Have you checked the rendered docs? With reStructuredText bulleted lists should usually have a blank line before them and no indent, or else they'll be rendered with an extra level of indentation.

docs/cumulusci_flow.rst
```diff
@@ -78,12 +98,60 @@ If this combination of named parent and child branches exist, the auto-merging f
...
+
+Using ``feature/`` branch prefix for the release branch names allow those branches to stay in sync with our main branch (they are just another feature branch to CumulusCI).
+The release number immediately after the ``feature/`` prefix allows CumulusCI to perform yet another type of auto-merge for your convenience.
```
davisagli, line 114: I think showing an example of a release branch name and feature branches under it would be helpful, rather than just describing the format. You do a good job of this in the next section so maybe part of that example just needs to be moved up here.

docs/cumulusci_flow.rst
```diff
@@ -78,12 +98,60 @@ If this combination of named parent and child branches exist, the auto-merging f
...
+
+
+Release to (Future) Release Merges
+----------------------------------
+Because release branches are so long-lived, and so much work goes into them, their diffs can get quite large.
+This means headaches are inevitable the day after a major release, and you need to pull down all of the changes from the new release into the next release branch (which has likely been in development for months already).
+To alleviate this pain point, CumulusCI can ensure that all release branches propogate commits they receive to other existing release branches that correspond to future releases.
```
davisagli, line 121:
```suggestion
To alleviate this pain point, CumulusCI can ensure that all release branches propagate commits they receive to other existing release branches that correspond to future releases.
```

docs/cumulusci_flow.rst
```diff
@@ -78,12 +98,60 @@ If this combination of named parent and child branches exist, the auto-merging f
...
+   * ``feature/002__large_feature`` - A large feature associated with release ``002``
+   * ``feature/002__large_feature__child1`` - First chunk of work for the large feature
+   * ``feature/002__large_feature__child2`` - Second chunk of work for the large feature
+   * ``feature/003`` - The release that comes after ``002``
+   * ``feature/003__feature1`` - A single feature associated with release ``003``
+
+In this scenario, CumulusCI ensures that when ``feature/002`` receives a commit, that that commit is also merged into ``feature/003``.
+This kicks off tests in our CI system and ensures that funcitonality going into ``feature/002`` doesn't break work being done for future releases.
+Once those tests pass, the commit on ``feature/003`` is merged to ``feature/003__feature1`` because they adhere to the parent/child naming convention described above.
+Commits **never** propogate in the opposite direction. (A commit to ``feature/002`` would never be merged to ``feature/001`` if it was an existing branch in the GitHub repository).
```
davisagli, line 137:
```suggestion
Commits **never** propagate in the opposite direction. (A commit to ``feature/002`` would never be merged to ``feature/001`` if it was an existing branch in the GitHub repository).
```

docs/cumulusci_flow.rst
```diff
@@ -78,12 +98,60 @@ If this combination of named parent and child branches exist, the auto-merging f
...
+
+**Propogating commits to future release branches is turned off by default.**
```
davisagli, line 139:
```suggestion
**Propagating commits to future release branches is turned off by default.** 
```

docs/cumulusci_flow.rst
```diff
@@ -78,12 +98,60 @@ If this combination of named parent and child branches exist, the auto-merging f
...
+      options:
+         update_future_releases: True
 
-CumulusCI and Parent/Child Feature Branches
--------------------------------------------
+Orphaned Branches
+-----------------
+If you have both a parent and a child branch, and the parent is deleted, this creates an orphaned branch.
+Orphaned branches do not receive any auto-merges from any branches.
+You can rename an orphaned branch to include the ``feature/`` prefix and contain no double underscores ('__') to begin receiving merges from the main branch again.
```
davisagli, line 153: An example would help clarify.

davisagli (commented):
copy-editing and comments on docs; I still need to review the code

cumulusci/tasks/github/merge.py
```diff
@@ -218,3 +217,199 @@ def _merge(self, branch, source, commit, children=None):
...
+        """A release branch begins with the given prefix"""
+        prefix = self.options["branch_prefix"]
+        if not branch_name.startswith(prefix):
+            return False
+        parts = branch_name[len(prefix) :].split("__")
+        return len(parts) == 1 and parts[0].isdigit()
+
+    def _get_release_num(self, release_branch_name):
+        """Given a release branch, returns an integer that
+        corresponds toithe release number for that branch"""
```
davisagli, line 360:
```suggestion
        corresponds to the release number for that branch"""
```

cumulusci/tasks/github/tests/test_merge.py
```diff
@@ -335,162 +347,462 @@ def test_main_parent_with_child_pr(self):
...
+                    f"Merged 1 commits into branch: {child_branches[0]}",
+                ),
                 (
                     "INFO",
-                    "Merged 1 commits into parent branch {}".format(parent_branch_name),
+                    f"Merged 1 commits into branch: {child_branches[1]}",
+                ),
+                (
+                    "INFO",
+                    f"Merged 1 commits into branch: {child_branches[2]}",
```
davisagli, line 436: In general in tests I would aim for abstracting things out into functions and variables less. It would be easier for me to read this and verify that the output makes sense if this had the expected branch names inline instead of referring to a variable I have to jump around to look at.

davisagli: Also perhaps we could simplify all these tests whose purpose is to verify that the correct branches are selected, by checking the result of task._get_branches_to_merge() rather than running the entire task and looking at what was logged.

cumulusci/cumulusci.yml
Br4nd0R: @davisagli - I'm not sure I agree with the name change. This task's primary purpose is to facilitate the auto-merging from "Parent to Child" branches. We've added an option which allows for updating of future releases, but this is feels more secondary to the parent-child automerges. I would prupose: `github_automerge_main_to_feature` and `github_automerge_parent_to_child` for the names of the two flows. Let me know what you think.

cumulusci/tasks/github/merge.py
```diff
@@ -218,3 +217,199 @@ def _merge(self, branch, source, commit, children=None):
...
+            to_merge = child_branches
+        elif not self.source_branch_is_default:
+            self.logger.debug(
+                f"No children found for branch {self.options['source_branch']}"
+            )
+
+        if release_branches:
+            self.logger.debug(
+                f"Found future release branches to update: {[branch.name for branch in release_branches]}"
+            )
```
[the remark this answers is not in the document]
Br4nd0R: Correct. My logic was; actual merges matter, everything else is likely only needed when troubleshooting an issue.

cumulusci/tasks/github/merge.py
```diff
@@ -218,3 +217,199 @@ def _merge(self, branch, source, commit, children=None):
...
+            to_merge = to_merge + release_branches
+
+        if main_descendents:
+            to_merge = to_merge + main_descendents
```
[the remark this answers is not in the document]
Br4nd0R: Ya, this would be a good idea. Then debug messages will always allow for inspection of what this method returns.

docs/cumulusci_flow.rst
Br4nd0R: I was thinking `github_automerge_maint_to_feature` and `github_automerge_parent_to_children`.

cumulusci/cumulusci.yml
davisagli: I thought it might be a bit confusing to call it parent_to_child when it does more than that. But I'm okay with it.

If we're going in that direction, I'd suggest github_main_to_feature and github_parent_to_children to minimize changes from the status quo.

docs/cumulusci_flow.rst
Br4nd0R: Have definitely checked rendered docs, but may have not noticed this. Checking again.

davisagli: It may vary depending on what system is rendering it. It's not a big deal.

Br4nd0R:
@davisagli - Made code changes, revamped tests (only a couple check the full log output now, and the others assert only on what is returned by `task._get_branches_to_merge()`, and updated docs. I ended up using the names that you suggested for the tasks. I think that 'automerge' in the name is a better fit.

Br4nd0R:
@davisagli - Thanks for revamping my release notes. It reminded me that I didn't include a statement that multi-level parent child branches are supported. I've included a statement, example branching structure, and walkthrough of how a commit can trickle down through the structure.

Let me know if:
* You think this is too similar to the example I give under the "Release to (Future) Release Merge" section and I can remove it.
* If we want to keep it, I'm not very happy of how it renders on screen; it's a wall of text. Long branch names don't help either. Not sure if there's a better way to format that may help improve readability. Maybe a graphic would work better 🤔 

fyi @davidmreed - docs work you may be interested in looking at.

docs/cumulusci_flow.rst
```diff
@@ -95,17 +95,38 @@ Parent/Child Feature Branches are created using a simple naming format for branc
 If this combination of named parent and child branches exist, the auto-merging functionality changes a bit:
 
 * Child branches never receive the auto-merge from main
-* Parent branches do receive the merge from main which kicks off a Feature Test build
+* Parent branches do receive the merge from main which kicks off a Feature Test build. (This assumes the parent branch is not iteself a child.)
```
davisagli, line 98:
```suggestion
* Parent branches do receive the merge from main which kicks off a Feature Test build. (This assumes the parent branch is not itself a child.)
```

davisagli (commented):
@Br4nd0R I think it's fine, and we can always edit later.

Overall I think this is just about ready. Should we update version.txt to 3.20.0.dev0 so that once this is merged it creates a pre-release so that we can update MetaCI staging and do some functional testing with CumulusCI-Test?

davisagli (approved):

PR #2031, opened 2020-09-18
Enhance the Insert Record task

davidmreed:
# Critical Changes

# Changes

- The `insert_record` task can now be used against the Tooling API. We clarified that this task can accept a dict of values if configured in `cumulusci.yml`.

# Issues Closed


prescod (approved):
LGTM! Glad this task is useful to someone other than me!

cumulusci/tasks/salesforce/insert_record.py
```diff
@@ -13,18 +13,27 @@ class InsertRecord(BaseSalesforceApiTask):
...
+            "description": "Field names and values in the format 'aa:bb,cc:dd', or a YAML dict",
             "required": True,
         },
+        "tooling": {"description": "If True, use the Tooling API instead of REST API."},
     }
 
     def _init_options(self, kwargs):
         super()._init_options(kwargs)
-        self.values = process_list_of_pairs_dict_arg(self.options["values"])
+        if type(self.options["values"]) is str:
```
davisagli, line 24: process_list_of_pairs_dict_arg already handles values of type dict and str...is there something else you're trying to handle here?

cumulusci/tasks/salesforce/insert_record.py
```diff
@@ -13,18 +13,27 @@ class InsertRecord(BaseSalesforceApiTask):
...
+            self.values = self.options["values"]
         self.object = self.options["object"]
+        self.use_tooling = process_bool_arg(self.options.get("tooling", False))
 
     def _run_task(self):
-        object_handler = getattr(self.sf, self.object)
+        if not self.use_tooling:
+            object_handler = getattr(self.sf, self.object)
+        else:
+            object_handler = getattr(self.tooling, self.object)
```
davisagli, line 35:
```suggestion
        api = self.tooling if self.use_tooling else self.sf
        object_handler = getattr(self.sf, self.object)
```

cumulusci/tasks/salesforce/insert_record.py
davidmreed: There is not. I just didn't know that. I'll update the task doc instead of making this change.

davisagli (approved):

PR #2033, opened 2020-09-18
Fix  handling of stored Booleans for REST DML

### LINKED ISSUE #2032: Dataset checks all checkbox fields regardless of true/false insert
**Describe the bug**
Datasets that are loaded seem to be setting all checkboxes to true (checked).

**Reproduction steps**
1. cci flow run dev_org using previously working [datasets for Summit Events App](https://github.com/SFDO-Community-Sprints/Summit-Events-App/tree/master/datasets/dev)
2. Checked org and all checkboxes regardless of true/false insert are checked.
3. Created a new mapping/dataset in [branch currently here](https://github.com/SFDO-Community-Sprints/Summit-Events-App/tree/cookie_key/datasets/dev) and tried again with the same results

**Context**
- CumulusCI version: 3.19.0
- Python version: 3.8.5
- Windows 10.0.19041 Build 19041
- Using Powershell
- Cumulus installed/updated with pipx.

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- CumulusCI now correctly handles converting stored Booleans for REST API data loads (fixes #2032)

### DIFF
--- a/cumulusci/tasks/bulkdata/step.py
+++ b/cumulusci/tasks/bulkdata/step.py
@@ -429,8 +429,10 @@ class RestApiDmlOperation(BaseDmlOperation):
         def _convert(rec):
             result = dict(zip(self.fields, rec))
             for boolean_field in self.boolean_fields:
-                # This is clumsy but required since Booleans are stored as text.
-                result[boolean_field] = result[boolean_field].lower() == "true"
+                # This is clumsy but required since Booleans are stored as.
+                result[boolean_field] = (
+                    result[boolean_field] or "false"
+                ).lower() == "true"
 
             # Remove empty fields (different semantics in REST API)
             # We do this for insert only - on update, any fields set to `null`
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/bulkdata/step.py
```diff
@@ -429,7 +429,8 @@ def load_records(self, records):
         def _convert(rec):
             result = dict(zip(self.fields, rec))
             for boolean_field in self.boolean_fields:
-                result[boolean_field] = bool(result[boolean_field])
+                # This is clumsy but required since Booleans are stored as text.
+                result[boolean_field] = result[boolean_field].lower() == "true"
```
davisagli, line 433: Just to check, are we sure result[boolean_field] is always a string and not None?

davisagli (commented):
Is there a test that would be worth adding for this case in _convert?

cumulusci/tasks/bulkdata/step.py
davidmreed: It can't be `null`/`None` unless the user manually edited the SQL. Booleans can be `null` in Apex but are always stored and returned to SOQL as `False`.

davidmreed:
I'll have a test case in a couple of minutes.

cumulusci/tasks/bulkdata/step.py
davidmreed: I suppose if we're loading from a Postgres dataset that was not generated from a Salesforce extract it could be `None`. I'll add a guard.

davisagli (approved):
Thanks for the quick fix.

PR #2010, opened 2020-09-10
Feature/extract and load use tempfiles

prescod:
# Changes

Previously, large data loads and extracts would use enormous amounts of memory. Now they should roughly constant amounts of memory.

Along the way:

 * backported `nullcontext` to Python 3.6
 * use a tempfile database instead of in-memory database 
 * remove the use of two CSV tempfiles on extract (because the data can be streamed to the database)
 * ADD a CSV tempfile on load to replace an in-memory list
 * created new test mocking infrastructure

PR #2035, opened 2020-09-18
Make all tests independent of home directory

prescod:
Create a fake HOME directory to insulate all tests from the user's real HOME directory (and vice versa).

### DIFF
--- a/cumulusci/conftest.py
+++ b/cumulusci/conftest.py
@@ -85,7 +85,7 @@ create_task_fixture = fixture(create_task_fixture, scope="function")
 #       test case changes.
 @pytest.fixture(autouse=True, scope="session")
 def patch_home(request):
-    with TemporaryDirectory() as home, mock.patch(
+    with TemporaryDirectory(prefix="fake_home_") as home, mock.patch(
         "pathlib.Path.home", lambda: Path(home)
     ), mock.patch.dict(
         os.environ,
--- a/cumulusci/conftest.py
+++ b/cumulusci/conftest.py
@@ -2,6 +2,9 @@ import io
 import os
 from http.client import HTTPMessage
 from unittest import mock
+import pytest
+from pathlib import Path
+from tempfile import TemporaryDirectory
 
 from pytest import fixture
 from cumulusci.core.github import get_github_api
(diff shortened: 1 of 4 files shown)


cumulusci/conftest.py
```diff
@@ -75,3 +78,19 @@ def fallback_orgconfig():
 vcr = fixture(salesforce_vcr, scope="module")
 
 create_task_fixture = fixture(create_task_fixture, scope="function")
+
+
+# TODO: This should also chdir to a temp directory which
+#       can represent the repo-root, but that will require
+#       test case changes.
```
davisagli, line 85: Not sure I agree. I think we should try to standardize how we're building project configs for tests and make sure that includes creating a temporary repo root, but I don't think we need one for tests that don't use a project config.

cumulusci/conftest.py
```diff
@@ -75,3 +78,19 @@ def fallback_orgconfig():
...
+@pytest.fixture(autouse=True, scope="session")
+def patch_home(request):
+    with TemporaryDirectory(prefix="fake_home_") as home, mock.patch(
+        "pathlib.Path.home", lambda: Path(home)
+    ), mock.patch.dict(
+        os.environ,
+        {"HOME": home, "USERPROFILE": home, "CUMULUSCI_KEY": "0123456789ABCDEF"},
```
davisagli, line 92: I like isolating the test environment, but:
- Call this `patch_home_and_env` to make it more obvious that this is part of what it does
- What's the argument for including CUMULUSCI_KEY in the default environment? Running cci with it set is relatively uncommon.

cumulusci/conftest.py
prescod: If we chdir to a random directory on pytest-startup then we know for sure that we are neither reading or writing from the CumulusCI REPO_ROOT by accident. Any test that wants to opt-in to reading or writing from the CumulusCI REPO_ROOT should do so explicitly.

cumulusci/conftest.py
prescod: It was necessary because CCI doesn't run without a key so you either supply one or use one from the user's environment. I got error messages before I added this. Of course setting this environment variable is not the only way to fake a key, but it was one easy way.

prescod:
It may or may not have been obvious, but this is not a fixture in the sense of a function that tests opt-into. It's more or less initialization code for the whole testing infrastructure. I added a docstring to clarify.

cumulusci/conftest.py
davisagli: Oh I see you don't mean the repo root of an arbitrary test project, you mean changing the current directory so the repo root of CumulusCI itself doesn't accidentally get treated as a project in tests. Yeah, good idea.

cumulusci/conftest.py
davisagli: @prescod I guess this is fine, but there are a couple of tests which do `del os.environ["CUMULUSCI_KEY"]` to make sure we get coverage of the code which uses keyring to generate a key when one is missing. We should update them to make sure they only remove it temporarily.

davisagli:
@prescod This is now showing a bunch of changes from #2010 which was already merged -- what happened?

cumulusci/conftest.py
prescod: I've ensured they use a nested patch now.

davisagli (approved):

PR #2047, opened 2020-09-23
[CCI] Create automation to remove manual github release notes step

jofsky:
Aims to remove the https://salesforce.quip.com/HknxAPzYNOi1 `Create Releases` manual readme creation.

# Critical Changes

# Changes

# Issues Closed


davisagli (changes requested):
@jofsky I think we should take a different approach on this for a few reasons:
1. The release notes framework uses a concept of configurable "parsers". A project can override project__git__release_notes__parsers to change which sections are rendered, and in what order.  The release info should be handled by a parser so that it's consistent with how other sections of the release notes are handled.
2. There is already a parser called InstallLinkParser which adds the install links if a version id was set as a task option (like it is when the task is called as part of the release flows). We should use or modify this parser rather than creating a new way of doing the same thing.
3. You've hardcoded the assumptions that the Sandbox push will happen on the same day the release is made and the Production push will happen 6 days later. This is usually the case for Salesforce.org, but may not be valid for other users of CumulusCI or if we change our release cadence in the future. I think we discussed this a while back and I suggested that the task should take options for specifying both the sandbox date and production date.

cumulusci/tasks/release_notes/generator.py
```diff
@@ -22,6 +22,12 @@ def __init__(self):
         self.init_parsers()
         self.init_change_notes()
         self.version_id = None
+        self.release_info = False
+        self.trial_info = False
+        self.sandbox_date = datetime.date.today().isoformat()
+        self.production_date = (
+            datetime.date.today() + datetime.timedelta(days=6)
+        ).isoformat()
```
davisagli, line 30: I still don't like hardcoding default dates. Using today as the default for sandbox_date is usually okay for our normal release process. But remember that the github_release_notes task can be run again manually in the future to re-generate the release notes if corrections were made on the pull requests. With default dates here, that would change the dates in the release notes!

I think the path we need to follow is:
- Make the task accept options for sandbox_date and production_date, but don't pick any dates by default.
- Make MetaCI pick good default dates when creating a Release object in MetaCI. Then make sure those get passed by MetaCI when running the release_production flow in a way that can be used by this task.

cumulusci/tasks/release_notes/generator.py
```diff
@@ -187,6 +193,12 @@ def __init__(
         has_issues=True,
         include_empty=False,
         version_id=None,
+        release_info=False,
+        trial_info=False,
+        sandbox_date=datetime.date.today().isoformat(),
+        production_date=(
+            datetime.date.today() + datetime.timedelta(days=6)
+        ).isoformat(),
```
davisagli, line 201: See above re defaults

cumulusci/tasks/release_notes/generator.py
```diff
@@ -293,7 +309,13 @@ def _update_release_content(self, release, content):
         # add empty PR section
         if self.include_empty_pull_requests:
             new_body.extend(render_empty_pr_section(self.empty_change_notes))
-        content = u"\r\n".join(new_body)
+        # # add release installation information section
+        # if self.release_info:
+        #     new_body.extend(render_release_installation_section(self))
+        # # add trial template information section
+        # if self.trial_info:
+        #     new_body.extend(render_release_installation_template())
```
davisagli, line 317: Remove commented section

cumulusci/tasks/release_notes/task.py
```diff
@@ -65,6 +78,15 @@ def _run_task(self):
...
             version_id=self.options.get("version_id"),
+            release_info=self.options.get("release_info", False),
+            trial_info=self.options.get("trial_info", False),
+            sandbox_date=self.options.get(
+                "sandbox_date", datetime.date.today().isoformat()
+            ),
+            production_date=self.options.get(
+                "production_date",
+                (datetime.date.today() + datetime.timedelta(days=6)).isoformat(),
+            ),
```
davisagli, line 89: See above re defaults

cumulusci/tasks/release_notes/parser.py
```diff
@@ -295,9 +295,12 @@ def parse(self, change_note):
 
     def render(self, existing_content=""):
         version_id = self.release_notes_generator.version_id
+        release_info = self.release_notes_generator.release_info
+        trial_info = self.release_notes_generator.trial_info
+
         if version_id:
             version_id = urllib.parse.quote_plus(version_id)
-            return "\r\n".join(
+            existing_content += "\r\n".join(
```
davisagli, line 303: `existing_content` is the existing text of this section if the task is run against a release that already has release notes with this section. The way you've coded this with +=, it will add a new duplicate copy of this information each time the task is run. Instead, we should:
1. re-regenerate the entire section content if any of version_id, sandbox_date, production_date, or trial_info were specified.
2. otherwise return existing_content untouched (so it's possible to run github_release_notes manually without specifying any of the options, but not wipe out the existing content of this section)

cumulusci/tasks/release_notes/parser.py
```diff
@@ -307,4 +310,24 @@ def render(self, existing_content=""):
                     f"{SANDBOX_LOGIN_URL}/packaging/installPackage.apexp?p0={version_id}",
                 ]
             )
+        if release_info is True:
+            existing_content += "\r\n".join(
+                [
+                    self._render_header(),
+                    "\r\n# Installation Info",
```
davisagli, line 317: This parser already has the title "Installation Links" (look in cumulusci.yml where the parser is configured) which is rendered by the call to _render_header(). That title should be changed to Installation Info and this line should be left out.

cumulusci/tasks/release_notes/tests/test_parser.py
```diff
@@ -562,12 +563,77 @@ def _create_expected_render(self, issue_number, issue_title, link_pr):
...
+    def test_package_release_info_no_version_id(self):
+        generator = mock.Mock(link_pr=True, version_id=None, release_info=True)
+        parser = InstallLinkParser(generator, "Title")
+        parser.parse("abc")
+        assert "# Title" in parser.render()
+        assert "# Installation Info" in parser.render()
+        assert "## Push Schedule" in parser.render()
+        assert f"Sandbox orgs: {generator.sandbox_date}" in parser.render()
+        assert f"Production orgs: {generator.production_date}" in parser.render()
+        assert "## Installation URL" not in parser.render()
```
davisagli, line 580: I think it would be better to assert that the entire rendered text matches what we expect (as a multi-line string). Doing it this way doesn't assert that the lines are in the correct order.

cumulusci/tasks/release_notes/parser.py
jofsky: So I changed https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/cumulusci.yml#L1121 to Install Info, but not seeing this show up. Slightly unsure about this, will push a proposed way to handle see what you think.

jofsky: This part has me quite confused, I've made the change locally but my tests are not giving the desired result.

jofsky:
@davisagli should I make it so you can pass sandbox and production dates and trial_info even when theres no version_id?

davisagli:
@jofsky 

> should I make it so you can pass sandbox and production dates and trial_info even when theres no version_id?

Yes, I think that would be ideal.

cumulusci/tasks/release_notes/parser.py
```diff
@@ -295,16 +295,57 @@ def parse(self, change_note):
...
+                    f"Production orgs: {self.release_notes_generator.production_date}",
+                )
+            if trial_info is True:
+                result += ["", "## Trialforce Template ID", "`TBD`"]
+            return "\r\n".join(result)
+        elif trial_info is True:
+            result += ["## Trialforce Template ID", "`TBD`"]
+            return "\r\n".join(result)
+        else:
+            return existing_content
```
davisagli, line 351: I think you're making this more complicated than it needs to be by adding the links just below the push dates when both are available. I don't think that really makes sense (orgs which are getting the push will not be using the links).

I would stick to the current formatting (NPSP's Getting Started page parses the release notes to get the dates, so it's easiest if we don't change the format) so I think the logic can be something like this (in pseudocode):

```
if we don't have any of the options:
    return existing_content
start result with the header "Installation Info"
if production_date or sandbox_date:
    add section titled "Push Schedule" with the specified dates
if version_id:
    add section titled "Installation Links" with the links
if trial_info:
    add section titled "Trialforce Template ID"
return result
```

cumulusci/tasks/release_notes/parser.py
```diff
@@ -295,16 +295,50 @@ def parse(self, change_note):
...
-                    "Sandbox & Scratch Orgs:",
-                    f"{SANDBOX_LOGIN_URL}/packaging/installPackage.apexp?p0={version_id}",
-                ]
-            )
-        return existing_content
+            if (
+                self.release_notes_generator.sandbox_date
+                or self.release_notes_generator.production_date
+            ):
+                result.append("")
```
davisagli, line 327: Rather than conditionally adding these empty lines at the start of the next subsection, I'd add them at the end of each subsection, without a condition. There's no harm to having an extra blank line at the end.

jofsky: Oh for some reason I wasn't a fan of the way it looked tbh.

jofsky: This was intentional butt will do.

davisagli (approved):

PR #2048, opened 2020-09-24
Doc reorg: Flow Reference Automation

Br4nd0R:
# Changes
* CumulusCI has a new task: `cci flow doc`. The task outputs information (description, additional info, and flow steps) about all flows grouped by their `group` attributes as specified in `cumulusci.yml`.
* For each group found, a rst formatted heading is output before the flows that correspond to that group.
* There is a new `doc/flows.yml` where we can include the following information which is spliced into the output of the task in various places:
     * `intro_blurb` - A small summary about flows in general that is displayed at the top of task reference page.
    * `groups` - Under which you can include a group name (that exists) with a `description` attribute, where we can store content 
    * `flow` - where we can list a specific flow with a corresponding `rst_text` attribute. This attribute will display after the flows name and description in documentation, and before the flow's steps.

* `make docs` has been updated to include this task and outputs to `docs/flows.rst` which has been added to `index.rst`.
* Fixed a small bug with alignment of when clauses in the output of flow steps.

TODO: Can't get headers to work in the flow.rst file, but can probably get it to work with a little more tinkering.
TODO: Finalize `docs/flows.rst` content


davisagli:
@Br4nd0R I think this is a good start. We may want to include a bit more context for each flow about what situations and kinds of orgs it's appropriate to use with.

There is a danger in listing out the flow steps because individual projects may have customized flows. We should be very clear that this is documenting the defaults and `cci flow info` is the best way to check what's happening for your project.

I think I prefer the existing formatting of the flow steps rather than your proposed change. That's partly because there are some details which are shown after the flow/task name in certain situations, so moving the "flow"/"task" designation there could make it harder to read. @jstvz was also pondering some updates to the `flow info` format and may have opinions here.

davisagli:
In the future we'll probably want to link from the flow reference to where the flows are discussed in the narrative docs.

jstvz:
I don't have a strong opinion on how these should be formatted here, but I lean towards matching the default `flow info` format to make it easier for users to compare the defaults against their project-level customizations.

Context for the updates David mentioned: The eventual changes for flow and task info I'm considering have to do with generating machine-readable (JSON for clients, possibly CCI YAML for cut-and-paste development) and verbose (including `options` as set, etc.) outputs.

docs/flow_reference.rst
```diff
@@ -0,0 +1,659 @@
...
+The suite of continuous integration flows are intended to execute in the context of a CI system.
+They typically involve the creation of an org to have tests (Apex and Robot) run against it.
+    
+``ci_beta``
+^^^^^^^^^^^
+This flow installs the latest beta version available and runs apex tests from the managed package. 
+
+.. code-block:: console
+
+    Flow Steps
```
davidmreed, line 16: This header looks a bit odd here as part of the code block, but I can't quite see how the current code produces it. Was this render from an older version?

davidmreed: Actually, is this whole file an older version?

Br4nd0R: @davidmreed - This whole file was hand typed... well, vim macros helped me with the repetition :)

Br4nd0R:
@davisagli - Needs your eyes on this again next week; added some new functionality :)

cumulusci/cli/cci.py
```diff
@@ -1353,11 +1354,62 @@ def task_doc(runtime):
...
-        doc = doc_task(name, task_config)
+        doc = document_task(name, task_config)
         click.echo(doc)
         click.echo("")
 
 
+@flow.command(name="doc", help="Exports RST format documentation for all flows")
+@pass_runtime(require_project=False)
+def flow_doc(runtime):
+    # config_src = runtime.universal_config
```
Br4nd0R, line 1365: I need to remove this.

docs/flows.rst
```diff
@@ -0,0 +1,645 @@
...
+	1) flow: install_beta
+	    1) flow: dependencies
+	        1) task: update_dependencies
+	        2) task: deploy_pre
+	    2) task: install_managed_beta
+	    3) flow: config_managed
+	        1) task: deploy_post
+	        2) task: update_admin_profile
+	    4) task: snapshot_changes
+	2) task: run_tests
```
davisagli, line 39: Would it be more readable if get_flow_steps produced a nested structure which could then be formatted as an rST bulleted list? This is good enough though.

docs/flows.rst
```diff
@@ -0,0 +1,645 @@
...
+
+|
```
davisagli, line 41: Are you trying to add extra whitespace? That should be handled in the styles for the sphinx theme rather than making semantic changes to the document structure.

cumulusci/cli/cci.py
```diff
@@ -1353,11 +1354,62 @@ def task_doc(runtime):
...
+    flows = (
+        runtime.project_config.list_flows()
+        if runtime.project_config is not None
+        else runtime.universal_config.list_flows()
+    )
+    flows_by_group = group_items(flows)
+
+    current_group = ""
+    for group, flows in flows_by_group.items():
+        if group != current_group:
```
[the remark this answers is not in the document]
Br4nd0R: No need for tracking current group, not sure what I was thinking there. I also just changed it to output groups (as well as the flows in each group) in alphabetical order. Let me know if how that sits with you.

docs/flows.rst
Br4nd0R: Ya, I didn't like how closely the flow steps were showing next to the heading of the next flow. I'll remove these from being produced in the output.

docs/flows.rst
Br4nd0R: I was thinking that originally but realized this was easier. Maybe change in the future if we want to.

Br4nd0R:
@davisagli - I got the `test_cci.test_flow_doc__with_flows_rst_file()` working with minimal mocks; however, I cannot for the life of me figure out how to pass properly formatted `steps` to the `FlowConfig` instance. It leads me down a rabbit whole of supplying other things (project configs, step specs, etc) that I couldn't figure out. 

I attempted to reference what was in `test_flowrunner.py` but found that the differences in context (running FlowRunner methods in isolation vs. setting up the runtime and executing the full command) made it hard to determine the minimum data requirement. Anyway, spent way to long on it, let me know if there's some easy thing I'm missing to provide flow steps to `FlowConfig`.

cumulusci/cli/cci.py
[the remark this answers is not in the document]
davisagli: I'd still prefer keeping the order of groups as specified in flows.yml rather than alphabetical, but that can be a future improvement.

docs/flows.rst
davisagli: Fair enough.

davisagli:
@Br4nd0R can you show me something you tried that didn't work? It should just be something like FlowConfig({"steps": {1: {"task": "some_task"}, 2: {"flow": "some_flow"}}}}) -- but if you want to construct a FlowCoordinator from it, the tasks and flows that are referenced also need to be defined in the project_config.

Br4nd0R:
@davisagli - So I've tweaked `test_cci.test_flow_doc__with_flows_rst_file()` to try to create an instance of `FlowRunner` while passing both a project and flow config. 

I must be missing a step to connect the configs, because I'm seeing it fail in `FlowCoordinator._check_infinite_flows()` because `flow_config.project_config` is returning none.

davisagli:
@Br4nd0R The project_config is usually attached to the FlowConfig in ProjectConfig.get_flow: https://github.com/SFDO-Tooling/CumulusCI/blob/06198872e52a71654a9bc61bb8bc7c901bd66ba8/cumulusci/core/config/project_config.py#L793

Since you're mocking the runtime's get_flow (which calls the ProjectConfig's get_flow), that is bypassed and you need to set flow_config.project_config yourself.

However if the runtime has a real project config attached and that project config contains a valid flow definition, I'm not sure there's a good reason to mock get_flow.

davisagli:
@Br4nd0R I updated the test to use less mocking.

Br4nd0R:
Thank you @davisagli for your help on this!

davisagli (approved):

PR #2049, opened 2020-09-24
Flow Groups

### LINKED ISSUE #1833: Add 'Group' to Flows
**Please describe the nature of this feature request**
- I'm interested in categorizing my repositories' flows at a glance.

**Does your feature request solve a current problem? Please describe.**
I am attempting to communicate the categories of tasks in my repositories, such as those that perform an all-in-one install, those that do pre-package config, those that perform a post-package config, and so on. I want to have an easily-consumable list of my flows by general function.

**Describe the solution you'd like**
I would like flows to receive groupings in `cci flow list` similar to how tasks receive groupings in `cci task list`.

**Describe alternatives you've considered**
I've considered using naming conventions, prefixes, suffixes, and so on, but they fall short of the mark or create complexities that insufficiently solve the issue.

**Additional context**
This feels consumable.
tasks with groups

This feels less consumable.
flows without groups

Br4nd0R:
Closes #1833

Please comment on initial group names / groupings, they are malleable.

# Changes
* `cci flow list` now displays flows in groups (like `cci task list`)
* Removed the `retrieve_scratch` flow. This predated the `retrieve_changes` task which is the current recommendation for retrieving source-tracked changes.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1316,7 +1316,6 @@ def org_shell(runtime, org_name, script=None, python=None):
 @click.option("--json", "print_json", is_flag=True, help="Print a json string")
 @pass_runtime(require_project=False)
 def task_list(runtime, plain, print_json):
-    task_groups = {}
     tasks = (
         runtime.project_config.list_tasks()
         if runtime.project_config is not None
@@ -1328,6 +1327,7 @@ def task_list(runtime, plain, print_json):
         click.echo(json.dumps(tasks))
         return None
 
+    task_groups = {}
     for task in tasks:
         group = task["group"] or "Other"
         if group not in task_groups:
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -12,11 +12,11 @@ import shutil
 import sys
 import time
 import traceback
-from datetime import datetime
+import runpy
 import webbrowser
 import contextlib
 from pathlib import Path
-import runpy
+from datetime import datetime
 
 import click
 import github3
@@ -1637,11 +1637,9 @@ def gist(runtime):
 
 
 def group_items(items):
-    groups = {}
+    groups = defaultdict(list)
     for item in items:
-        group = item["group"] or "Other"
-        if group not in groups:
-            groups[group] = []
-        groups[group].append([item["name"], item["description"]])
+        group_name = item["group"] or "Other"
+        groups[group_name].append([item["name"], item["description"]])
 
     return groups
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1473,11 +1473,18 @@ def flow_list(runtime, plain, print_json):
         click.echo(json.dumps(flows))
         return None
 
-    data = [["Name", "Description"]]
-    data.extend([flow["name"], flow["description"]] for flow in flows)
+    flow_groups = {}
+    for flow in flows:
+        group = flow["group"] or "Other"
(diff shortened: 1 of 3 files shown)


cumulusci/cli/cci.py
```diff
@@ -1639,3 +1634,14 @@ def gist(runtime):
...
         click.echo(f"Gist created: {gist.html_url}")
         webbrowser.open(gist.html_url)
+
+
+def group_items(items):
+    groups = {}
+    for item in items:
+        group = item["group"] or "Other"
+        if group not in groups:
+            groups[group] = []
```
davisagli, line 1644: This would be a good place to use a collections.defaultdict instead of a plain dict

cumulusci/cumulusci.yml
```diff
@@ -553,6 +553,7 @@ tasks:
 
 flows:
     ci_beta:
+        group: Continous Integration
```
davisagli, line 556:
```suggestion
        group: Continuous Integration
```
(and same for the other copies)

cumulusci/cumulusci.yml
```diff
@@ -650,6 +656,7 @@ flows:
                     managed: True
 
     config_apextest:
+        group: Org Configuraiton
```
davisagli, line 659: Let's call this group Post-Install Configuration, because that makes it clearer that it's meant to be run after the package is installed. (also note misspelled configuration)

cumulusci/cumulusci.yml
```diff
@@ -781,6 +798,7 @@ flows:
                 task: uninstall_packaged_incremental
 
     dev_org:
+        group: Org Creation
```
davisagli, line 801: This is nitpicking but how about Org Setup. Technically the org is created before we run any flows on it.

cumulusci/cumulusci.yml
```diff
@@ -854,6 +877,7 @@ flows:
                 task: uninstall_managed
 
     install_beta:
+        group: Install / Uninstall
```
davisagli, line 880: install_beta belongs in the Org Setup group because it does the full setup including dependencies, installation, and post-install config. It should have been called beta_org

cumulusci/cumulusci.yml
```diff
@@ -866,6 +890,7 @@ flows:
                 task: snapshot_changes
 
     install_prod:
+        group: Install / Uninstall
```
davisagli, line 893: install_prod belongs in the Org Setup group because it does the full setup including dependencies, installation, and post-install config. It should have been called prod_org, release_org, or customer_org

cumulusci/cumulusci.yml
```diff
@@ -900,6 +927,7 @@ flows:
                 task: install_managed_beta
 
     release_beta:
+        group: Release Management
```
davisagli, line 930: How about Release Operations

davisagli (changes requested):
Notes for other flows:
build_feature_test_package: could arguably be either Continuous Integration or Release Operations
retrieve_scratch: We should probably remove this flow, actually. It predated the retrieve_changes task.

cumulusci/cumulusci.yml
Br4nd0R: I was having it match the tasks group for this one. But I like Operations better. Should we change the tasks group name in a separate PR to match?

davisagli: @Br4nd0R Sure. It can go in this PR if you want.

davisagli (approved):

PR #2053, opened 2020-09-28
Feature/robot pabot fixes

boakley:
This fixes a bug related to the recently released `--processes` option which prevented tests from running if they used relative imports. This was caused by spawned processes not picking up the correct PYTHONPATH before executing robot.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -123,7 +123,9 @@ class Robot(BaseSalesforceTask):
                     cmd.extend([f"--{option}", str(value)])
 
             cmd.append(self.options["suites"])
-            self.logger.info(f"pabot command: {shlex.join(cmd)}")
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -97,13 +99,20 @@ def _run_task(self):
         )
 
         if self.options["processes"] > 1:
+            # Since pabot runs multiple processes, we have to use the
+            # --pythonpath option to make sure the current working
+            # directory is included when pabot spawns each
+            # process. Otherwise, robot can't file libraries and resource
+            # files via relative paths.
```
davisagli, line 106: @boakley Can you show me the symptom you're trying to address?

At first glance, it doesn't make sense to me that this would be the solution. Python typically does include the current directory in sys.path without needing to be explicit about it.

PR #2059, opened 2020-09-30
Snowfakery 1.1.2

prescod:
Snowfakery 1.1.2

Minor improvements in error reporting.

Update to Faker 4.1.3


davisagli:
Needs Faker updated at the same time, right?

davisagli (approved):

PR #2065, opened 2020-10-01
CumulusCI Release: 3.20.0

Br4nd0R:


HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+- The ``github_parent_to_children`` task has been renamed to ``github_automerge_feature``. It still merges changes from feature branches to their children (e.g. ``feature/parent`` would be merged to ``feature/parent__child``). It is now possible to use multiple double-underscores to create more deeply nested children, and the task will only merge to the next level (e.g. ``feature/parent`` would merge to ``feature/parent__child`` which would merge to ``feature/parent__child__grandchild``).
+- The ``children_only`` option for these tasks has been removed. The strategy for picking which branches to target for merging is now determined by the ``source_branch``.
+- Previously, large data loads and extracts would use enormous amounts of memory. Now they should use roughly constant amounts of memory.
+- Adjusted tasks: ``install_managed`` and ``update_dependencies`` can now install packages from just a version id, using ``PackageInstallRequest``.
+- New task: ``github_package_data gets`` a package version id from a github commit status
```
davidmreed, line 18:
```suggestion
- New task: ``github_package_data`` gets a package version id from a GitHub commit status
```

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+- New task: ``create_package_version``. Builds a 2gp package (managed or unlocked) via a Dev Hub org. Includes some automated handling of dependencies:
```
davidmreed, line 19: We should be consistent in capitalizing `1GP` and `2GP`, and `GitHub`.

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+
+  - Resolve 1gp managed package dependencies using a scratch org with the dependencies installed (creating it if necessary)
+  - Build an additional unlocked package for unpackaged dependencies from github and for the project's own unpackaged/pre metadata
+
+- New flows for use in MetaCI to run feature tests using 2gp packages:
+
+  - ``build_feature_test_package``: Runs the create_package_version task, and in the context of MetaCI it will set a commit status with the package version id.
```
davidmreed, line 26:
```suggestion
  - ``build_feature_test_package``: Runs the ``create_package_version task``, and in the context of MetaCI it will set a commit status with the package version id.
```

davidmreed (changes requested):

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
 History
 =======
 
+3.20.0 (2020-09-30)
+-------------------
+Tasks, Flows, and Automation:
+
+- ``cci flow list`` now displays flows in different groups that are organized by functional area. (This is similar to how ``cci task list`` currently works).
+- The ``insert_record`` task can now be used against the Tooling API. We clarified that this task can accept a dict of values if configured in ``cumulusci.yml``.
+- Added support for newer metadata components to the ``update_package_xml`` task.
```
davisagli, line 11:
```suggestion
- Added support for newer metadata types to the ``update_package_xml`` task.
```

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+- We've deprecated the ``retrieve_scratch`` standard flow. ``retrieve_changes`` is the current recommended way for retrieving source-tracked changes.
```
davisagli, line 12: Say removed instead of deprecated. Deprecated sounds like it's still available.

Since this is a backwards-incompatible change, it should be called out under a "critical changes" heading at the beginning of the release notes

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+- The ``github_master_to_feature`` task has been renamed to ``github_automerge_main``. It still merges changes from the default branch to feature branches. In the case of an orphaned feature branch (a branch with a name like ``feature/parent__child`` where ``feature/parent`` does not exist as its own branch), the ``github_automerge_main`` branch will no longer merge to the orphaned branch.
+- The ``github_parent_to_children`` task has been renamed to ``github_automerge_feature``. It still merges changes from feature branches to their children (e.g. ``feature/parent`` would be merged to ``feature/parent__child``). It is now possible to use multiple double-underscores to create more deeply nested children, and the task will only merge to the next level (e.g. ``feature/parent`` would merge to ``feature/parent__child`` which would merge to ``feature/parent__child__grandchild``).
```
davisagli, line 14: These should also go under critical changes since the old flow names no longer work.

davisagli: I would also use a top-level bullet point "Changes to automatic merging:" with the various related bullet points indented further.

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+- The ``children_only`` option for these tasks has been removed. The strategy for picking which branches to target for merging is now determined by the ``source_branch``.
+- Previously, large data loads and extracts would use enormous amounts of memory. Now they should use roughly constant amounts of memory.
+- Adjusted tasks: ``install_managed`` and ``update_dependencies`` can now install packages from just a version id, using ``PackageInstallRequest``.
```
davisagli, line 17: Use of PackageInstallRequest is an implementation detail, I wouldn't include that in the release notes.

I'd say either SubscriberPackageVersion id or "version id (04t prefix)" to clarify.

HISTORY.rst
davisagli: I'd add something like "This task is intended primarily for use as part of the ci_feature_2gp flow." to save people the time of trying to figure out what it does.

HISTORY.rst
davisagli: I'd also be inclined to spell it out as "second-generation packaging (2GP)" in the first use.

And I would group create_package_version, the 2gp flows, and the version_id option together under a "Added support for creating 2gp packages (experimental)" heading

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+- New task: ``github_package_data gets`` a package version id from a github commit status
+- New task: ``create_package_version``. Builds a 2gp package (managed or unlocked) via a Dev Hub org. Includes some automated handling of dependencies:
+
+  - Resolve 1gp managed package dependencies using a scratch org with the dependencies installed (creating it if necessary)
+  - Build an additional unlocked package for unpackaged dependencies from github and for the project's own unpackaged/pre metadata
```
davisagli, line 22: Rather than getting into the details of 2gp dependency handling in the release notes it might be better to link to the docs for details.

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+
+- New flows for use in MetaCI to run feature tests using 2gp packages:
+
+  - ``build_feature_test_package``: Runs the create_package_version task, and in the context of MetaCI it will set a commit status with the package version id.
+  - ``ci_feature_2gp``: Retrieves the package version from the commit status set by build_feature_test_package, installs dependencies and the package itself in a scratch org, and runs Apex tests. (There is another new task, github_package_data, which is used by this flow.)
+
+User Experience:
+
+- Improved error messaging when encountering errors during bulk data mapping validation.
+- Improvements to various documentation sections: Why CumulusCI?, Features
```
davisagli, line 32: We haven't typically mentioned doc changes in the release notes.

HISTORY.rst
```diff
@@ -2,6 +2,43 @@
...
+  - ``ci_feature_2gp``: Retrieves the package version from the commit status set by build_feature_test_package, installs dependencies and the package itself in a scratch org, and runs Apex tests. (There is another new task, github_package_data, which is used by this flow.)
+
+User Experience:
+
+- Improved error messaging when encountering errors during bulk data mapping validation.
+- Improvements to various documentation sections: Why CumulusCI?, Features 
+
+Issues Closed:
+
+- Fixed bug to prevent null SubscriberPackageVersion in customer orgs from breaking CumulusCI.
```
davisagli, line 36: It's always good to include some context about when/where an issue takes place so that the reader can determine if it is relevant to them. In this case, adding something like "when checking what packages are installed" would help

davidmreed: In this case, it's hard to know - it's an issue that shouldn't ever happen (there's something wrong with the org that caused this), and could manifest in a couple of different places that aren't all clearly linked to one another.

We could say something like "Fixed a very rare bug that caused CumulusCI to fail to retrieve installed packages from an org when running package-related tasks or evaluating `when` conditional expressions."

davisagli (approved):

davidmreed (approved):

PR #2067, opened 2020-10-02
upload doc artifact

davisagli:
This makes the built docs available for download. It's not super obvious (you have to click into the Details for the Build Docs action, then look for the Artifacts menu, download the docs, unzip and open index.html) but at least it's available.

# Critical Changes

# Changes

# Issues Closed


Br4nd0R (approved):
Thanks for doing this!

PR #2069, opened 2020-10-02
Doc Reorg: Unpackaged Metadata & Metadata ETL

davidmreed:
Rendered:
https://cumulusci.readthedocs.io/en/feature-doc-reorg__unpackaged/unpackaged.html and https://cumulusci.readthedocs.io/en/feature-doc-reorg__unpackaged/metadata_etl.html

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/docs/metadata_etl.rst
+++ b/docs/metadata_etl.rst
@@ -24,8 +24,8 @@ become out-of-sync or entail more destructive deployment operations.
 A primary example use case for Metadata ETL is deployment of Standard Value Sets.
 Standard Value Sets, which define the picklist values available on standard fields
 like ``Opportunity.StageName``, are not packageable, and as such must be part of an
-application's unpackaged metadata. They're critical to many applications: Business
-Processes, for example, will fail to deploy if the requisite values are not available.
+application's unpackaged metadata. They're critical to many applications: A Business
+Process, for example, will fail to deploy if the Stage values it includes are not available.
 And lastly, they come with a serious danger for deployment into subscriber orgs:
 deploying Standard Value Sets is an overwrite operation, so all existing values in the
 target org that aren't part of the deployment are deactivated. This means that it's
@@ -33,7 +33,7 @@ neither safe nor maintainable to store static Standard Value Set metadata in a p
 and deploy it.
 
 These three facets - non-packageability, application requirements, and deployment safety -
-all militate for a Metadata ETL approach. Rather than attempting to deploy static metadata
+all support a Metadata ETL approach. Rather than attempting to deploy static metadata
 stored in the repository, the product's automation should *extract* the Standard Value Set 
 metadata from the org, *transform* it to include the desired values (as well as all existing
 customization), and *load* the transformed metadata back into the org. CumulusCI now ships
@@ -70,7 +70,7 @@ CumulusCI includes several out-of-the-box Metadata ETL tasks:
  - ``set_organization_wide_defaults`` sets the Org-Wide Defaults for one or more sObjects,
    and waits until sharing recalculation completes before proceeding.
 
-Most Metadata ETL tasks accept the option ``api_names``, which specifies the scope of the operation.
+Most Metadata ETL tasks accept the option ``api_names``, which specifies the developer names of the specific metadata components which should be included in the operation.
 In most cases, more than one entity may be transformed in a single operation. Each task is a single
 atomic Metadata API deployment operation. Consult the Task Reference or use the ``cci task info``
 command for more information on the usage of each task.
--- a/docs/metadata_etl.rst
+++ b/docs/metadata_etl.rst
@@ -180,7 +180,7 @@ API names. (The base class will generate a corresponding ``package.xml``). Subcl
 also implement ``_transform()``, as with ``BaseMetadataETLTask``.
 
 ``UpdateFirstAttributeTextTask`` is a base class and generic concrete task that makes it easy to
-perform a specific, common transformation: setting the value of a the first instance of a specific 
+perform a specific, common transformation: setting the value of the first instance of a specific 
 top-level tag in a given metadata entity. Subclasses (or tasks defined in ``cumulusci.yml``)
 must define the ``entity``, targeted ``attribute``, and desired ``value`` to set. Example:
 
--- a/docs/metadata_etl.rst
+++ b/docs/metadata_etl.rst
@@ -1,0 +1,198 @@
+.. _metadata-etl:
+
+============
+Metadata ETL
+============
+
+Introduction to Metadata ETL
+----------------------------
+
+"ETL" refers to "extract, transform, and load" operations, usually applied
+to data. CumulusCI offers a suite of functionality we call *Metadata ETL*.
+Metadata ETL makes it easy to define automation that executes targeted
+transformations of metadata that already exists in an org.
+
+Metadata ETL is particularly useful for building automation in projects
+that extend other managed packages or that perform complex setup operations
+during installations, such as through MetaDeploy. By using Metadata ETL
+tasks, projects can often avoid storing and deploying unpackaged metadata
+by instead extracting metadata from the target org, making changes, and
+then re-deploying. This mode of configuration is lower-risk and lower-
+maintenance than storing extensive unpackaged metadata, which may
+become out of sync or entail more destructive deployment operations.
+
+A primary example use case for Metadata ETL is deployment of Standard Value Sets.
+Standard Value Sets, which define the picklist values available on standard fields
+like ``Opportunity.StageName``, are not packageable, and as such must be part of an
+application's unpackaged metadata. They're critical to many applications: Business
+Processes, for example, will fail to deploy if the requisite values are not available.
+And lastly, they come with a serious danger for deployment into subscriber orgs:
+deploying Standard Value Sets is an overwrite operation, so all existing values in the
+target org that aren't part of the deployment are deactivated. This means that it's
+neither safe nor maintainable to store static Standard Value Set metadata in a project
+and deploy it.
+
+These three facets - non-packageability, application requirements, and deployment safety -
+all militate for a Metadata ETL approach. Rather than attempting to deploy static metadata
+stored in the repository, the product's automation should *extract* the Standard Value Set 
+metadata from the org, *transform* it to include the desired values (as well as all existing
+customization), and *load* the transformed metadata back into the org. CumulusCI now ships
+with a task, ``add_standard_value_set_entries``, that makes it easy to do just this:
+
+  .. code-block:: yaml
+
+    add_standard_value_set_entries:
+        options:
+            entries:
+                - fullName: "New_Value"
+                  label: "New Value"
+                  closed: False
+            api_names:
+                - CaseStatus
+
+This task would retrieve the existing ``Case.Status`` picklist value set from the org,
+add the ``New_Value`` entry to it, and redeploy the modified metadata - ensuring that
+the application's needs are met with a safe, minimal intervention in the target org.
+
+Standard Metadata ETL Tasks
+---------------------------
+
+CumulusCI includes several out-of-the-box Metadata ETL tasks:
+
+ - ``activate_flow`` activates Flows (or Processes).
+ - ``add_page_layout_related_lists`` adds Related Lists, with given fields, to Page Layouts.
+ - ``add_permission_set_perms`` adds FLS and Apex class access entries to Permission Sets.
+ - ``add_picklist_entries`` adds picklist entries to a custom picklist field.
+ - ``add_standard_value_set_entries`` adds entries to a Standard Value Set.
+ - ``assign_compact_layout`` assigns a Compact Layout to a given sObject.
+ - ``set_duplicate_rule_status`` activates and deactivates Duplicate Rules.
+ - ``set_field_help_text`` sets the Help Text of a given field (a non-push-upgradeable attribute).
+ - ``set_organization_wide_defaults`` sets the Org-Wide Defaults for one or more sObjects,
+   and waits until sharing recalculation completes before proceeding.
+
+Most Metadata ETL tasks accept the option ``api_names``, which specifies the scope of the operation.
+In most cases, more than one entity may be transformed in a single operation. Each task is a single
+atomic Metadata API deployment operation. Consult the Task Reference or use the ``cci task info``
+command for more information on the usage of each task.
+
+More Metadata ETL tasks are being added regularly, and the Metadata ETL framework
+makes it easy to add more tasks.
+
+Namespace Injection
+-------------------
+
+All out-of-the-box Metadata ETL tasks accept a Boolean ``managed`` option. If ``True``, CumulusCI
+will replace the token ``%%%NAMESPACE%%%`` in API names and in values used for transforming metadata
+with the project's namespace; if ``False``, the token will simply be removed.
+
+
+Implementation of Metadata ETL Tasks
+------------------------------------
+
+This section covers internals of the Metadata ETL framework, and is intended for
+users who wish to build their own Metadata ETL tasks.
+
+The Metadata ETL framework, and out-of-the-box Metadata ETL tasks, are part of the
+``cumulusci.tasks.metadata_etl`` package. The ``cumulusci.tasks.metadata_etl.base``
+module contains all of the base classes inherited by Metadata ETL classes.
+
+The easiest way to implement a Metadata ETL class that extracts, transforms, and loads
+a specific entity, such as ``CustomObject`` or ``Layout``, is to subclass
+``MetadataSingleEntityTransformTask``.
+
+This abstract base class has two override points: the class attribute ``entity`` should
+be defined to the Metadata API entity that this class is intended to transform, and the
+method ``_transform_entity(self, metadata: MetadataElement, api_name: str)`` must be 
+overridden. This method should make any desired changes to the supplied ``MetadataElement``,
+and either return a ``MetadataElement`` for deployment, or ``None`` to suppress deployment
+of this entity. Classes may also opt to include their own options in ``task_options``, but
+generally should also incorporate the base class's options, and override ``_init_options()``
+(``super``'s implementation should also be called to ensure that supplied API names are
+processed appropriately).
+
+The ``SetDuplicateRuleStatus`` class is a simple example of implementing a 
+``MetadataSingleEntityTransformTask`` subclass, presented here with additional comments:
+
+  .. code-block:: python
+
+    from typing import Optional
+
+    from cumulusci.tasks.metadata_etl import MetadataSingleEntityTransformTask
+    from cumulusci.utils.xml.metadata_tree import MetadataElement
+    from cumulusci.core.utils import process_bool_arg
+
+
+    class SetDuplicateRuleStatus(MetadataSingleEntityTransformTask):
+        # Subclasses *must* define `entity`
+        entity = "DuplicateRule"
+
+        # Most subclasses include the base class's options via
+        # **MetadataSingleEntityTransformTask.task_options. Further
+        # options may be added for this specific task. The base class
+        # options include in particular the standard `api_names` option,
+        # which base class functionality requires.
+        task_options = {
+            "active": {
+                "description": "Boolean value, set the Duplicate Rule to either active or inactive",
+                "required": True,
+            },
+            **MetadataSingleEntityTransformTask.task_options,
+        }
+
+        # The `_transform_entity()` method must be overriden.
+        def _transform_entity(
+            self, metadata: MetadataElement, api_name: str
+        ) -> Optional[MetadataElement]:
+            # This method modifies the supplied `MetadataElement`, using methods
+            # from CumulusCI's metadata_tree module, to match the desired configuration.
+            status = "true" if process_bool_arg(self.options["active"]) else "false"
+            metadata.find("isActive").text = status
+
+            # Always return the modified `MetadataElement` if deployment is desired.
+            # To not deploy this element, return `None`.
+            return metadata
+
+Advanced Metadata ETL Base Classes
+++++++++++++++++++++++++++++++++++
+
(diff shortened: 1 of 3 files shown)


docs/metadata_etl.rst
```diff
@@ -0,0 +1,198 @@
+.. _metadata-etl:
```
davidmreed, line 1: This is copied verbatim from the current docs. I thought there was too much material to go with `unpackaged`.

docs/metadata_etl.rst
```diff
@@ -0,0 +1,198 @@
...
+
+``BaseMetadataTransformTask`` can be used as the base class for ETL tasks that require
+more flexibility than is permitted by ``MetadataSingleEntityTransformTask``, such as tasks
+that must mutate multiple Metadata API entities in a single operation. Subclasses must
+override ``_get_entities()`` to return a dict mapping Metadata API entities to collections of
+API names. (The base class will generate a corresponding ``package.xml``). Subclasses must
+also implement ``_transform()``, as with ``BaseMetadataETLTask``.
+
+``UpdateFirstAttributeTextTask`` is a base class and generic concrete task that makes it easy to
+perform a specific, common transformation: setting the value of a the first instance of a specific
```
Br4nd0R, line 183:
```suggestion
perform a specific, common transformation: setting the value of the first instance of a specific 
```

Br4nd0R (changes requested):
Just a minor tweak on the bottom there. Great write-up, I learned lots :)

docs/metadata_etl.rst
davisagli: I like positioning it right after `unpackaged` and making sure that both topics contextualize themselves in relation to the other. They are two different approaches to the same problem space of post-install configuration.

docs/unpackaged.rst
```diff
@@ -1,14 +1,161 @@
...
+
+For example, a testing-oriented scratch org may wish to deploy a customized set of Page Layouts to help testers easily visualize data under test. Such page layouts could be stored in ``unpackaged/config/qa``.
+
+
+Unpackaged Metadata Folder Structure
+------------------------------------
+
+All unpackaged metadata is stored in the ``unpackaged`` directory tree. Within this directory are the three top-level directories, ``unpackaged/pre``, ``unpackaged/post``, and ``unpackaged/config``.
+
+These trees contain metadata bundles in Metadata API format: that is, a directory containing a ``package.xml`` manifest and Metadata API-format source code. CumulusCI does not support Salesforce DX format for unpackaged bundles.
```
davisagli, line 52: Actually since the merge of the 2gp branch, I believe if you point `deploy` at a folder that doesn't have a package.xml, it will use sfdx to convert it in a tempdir to deploy. But this may not work well with namespace tokens in some locations if they get in the way of the conversion. There is potential to make it work with a scheme like this:
1. Copy sfdx format to a tempdir.
2. Do namespace injection.
3. Convert to mdapi.

docs/unpackaged.rst
```diff
@@ -1,14 +1,161 @@
...
+
+.. _namespace-injection:
+
+Namespace Injection
 -------------------
 
-Customize config flows
+Projects that build managed packages often must construct their unpackaged metadata to be deployable in multiple contexts: unmanaged deployments, such as developer orgs, unmanaged namespaced scratch orgs, and also managed contexts, such as a beta test org or a demo org created with ``install_prod``.
+
+  Projects that are building an org implementation (not a package with a namespace) do not need to use namespace injection.
```
davisagli, line 61: You haven't defined what namespace injection is yet...

docs/unpackaged.rst
```diff
@@ -1,14 +1,161 @@
...
+The task ``deploy_post``, which is part of the ``config_dev``, ``config_qa``, and ``config_managed`` flows, is responsible for deploying these bundles.
+
+Metadata that's not intended to be delivered to all installations of the product should *not* be included in ``unpackaged/post``. It's critical for managed package projects that this metadata include namespace tokens (see :ref:`namespace-injection`).
+
+``unpackaged/config``: Tailoring Orgs
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Projects may come with more than one supported configuration in their CumulusCI automation. For example, projects may support distinct, tailored ``dev_org``, ``qa_org``, and ``install_prod`` flows, each of which performs different setup for their specific use case. Unpackaged metadata stored in ``unpackaged/config`` is a tool to support operational needs to tailor orgs to different configurations. 
+
+For example, a testing-oriented scratch org may wish to deploy a customized set of Page Layouts to help testers easily visualize data under test. Such page layouts could be stored in ``unpackaged/config/qa``.
```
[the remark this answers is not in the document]
davisagli: Oh, I see that's what is in the "Customize Config Flows" section below.

davisagli (changes requested):
Nice work.

docs/metadata_etl.rst
```diff
@@ -0,0 +1,198 @@
...
+ - ``assign_compact_layout`` assigns a Compact Layout to a given sObject.
+ - ``set_duplicate_rule_status`` activates and deactivates Duplicate Rules.
+ - ``set_field_help_text`` sets the Help Text of a given field (a non-push-upgradeable attribute).
+ - ``set_organization_wide_defaults`` sets the Org-Wide Defaults for one or more sObjects,
+   and waits until sharing recalculation completes before proceeding.
+
+Most Metadata ETL tasks accept the option ``api_names``, which specifies the scope of the operation.
+In most cases, more than one entity may be transformed in a single operation. Each task is a single
+atomic Metadata API deployment operation. Consult the Task Reference or use the ``cci task info``
+command for more information on the usage of each task.
```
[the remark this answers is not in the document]
davidmreed: I revised to be more clear. Do you think there's a need for an explicit disclaimer?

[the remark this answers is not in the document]
davisagli: Not sure... I think it's a bit incongruent with how we're emphasizing the safety of the Metadata ETL approach. On the other hand, it's pretty unlikely to occur. But it's probably worth mentioning in a note callout as fine print, so there's something we can point to if someone tries to run multiple concurrent operations and has trouble.

docs/metadata_etl.rst
```diff
@@ -0,0 +1,198 @@
...
+
+Namespace Injection
+-------------------
+
+All out-of-the-box Metadata ETL tasks accept a Boolean ``managed`` option. If ``True``, CumulusCI
+will replace the token ``%%%NAMESPACE%%%`` in API names and in values used for transforming metadata
+with the project's namespace; if ``False``, the token will simply be removed.
+
+
+Implementation of Metadata ETL Tasks
```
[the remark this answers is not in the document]
davidmreed: I moved it to a new PR, https://github.com/SFDO-Tooling/CumulusCI/pull/2101

docs/metadata_etl.rst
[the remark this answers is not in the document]
davidmreed: That's a good point. I added a note to that effect.

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+Not everything that's part of an application can be part of a package.
 
-Namespace injection
+CumulusCI implements the Product Delivery Model by offering support for complex applications -- applications that may include multiple managed packages as well as unpackaged metadata, and setup automation that configures org settings or makes precise changes to existing configuration.
+
+The tools used to implement that support are *unpackaged metadata* and *Metadata ETL*. 
+
+Unpackaged metadata refers to metadata that is not delivered as part of a package, and can include both support metadata delivered to users as well as metadata that operationally configures orgs used by the product. 
+
+Metadata ETL is a suite of tasks that support surgically altering existing metadata in an org. It's a powerful technique that changes the unpackaged configuration in an org without risking damage by overwriting existing customizations with stored metadata. Metadata ETL is relevant for delivering applications to customers safely, and is often a superior alternative to unpackaged metadata.
```
justinvholt, line 12: Would we call it a feature rather than a technique? And the end of this sentence is a bit confusing: Do the existing customizations have stored metadata? Or does Metadata ETL use stored metadata to not damage existing customizations? (I’m  95% certain it’s the former, but I would like confirmation.)

justinvholt, line 12: Would we call it a feature rather than a technique? And the end of this sentence is a bit confusing: Do the existing customizations have stored metadata? Or does Metadata ETL use stored metadata to not damage existing customizations? (I’m 95% certain it’s the former, but I would like confirmation.)

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+
+The ``deploy_pre`` task, which is part of the ``dependencies`` flow, is responsible for deploying these bundles.
+
+.. important:: Do not include metadata in ``unpackaged/pre`` unless it is intended to be delivered to *all* installations of the product.
+
+
+``unpackaged/post``: Configuration After Package Install
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Projects often include metadata that is genuinely part of the application, but either cannot be delivered as part of a managed package for operational reasons. This metadata must be deployed *after* the project's own code and metadata are deployed first and the org is configured.
```
justinvholt, line 39: Changed up some of the wording in the first sentence and added the second one. Please confirm that everything is correct.

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
 -------------------
 
-Customize config flows
+Projects that build managed packages often construct their unpackaged metadata to be deployable in multiple contexts.
+
+* Unmanaged deployments, such as developer orgs
+* Unmanaged namespaced scratch orgs
+* Managed contexts, such as a beta test org or a demo org created with ``install_prod``
+
+Because projects that are building an org implementation or a non-namespaced package do not have a namespace, or a distinction between managed and unmanaged contexts, these projects must also build namespaced packages.
```
justinvholt, line 91: Why must they build namespaced packages? I would like to cap this sentence with just a few words.

"These projects must also build namespaced packages to __________________."

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+
+    Example: Metadata located in ``unpackaged/post`` is deployed after the application code in both unmanaged and managed contexts. If that metadata contains references to the application components, it must be deployable when that metadata is nampespaced (in a managed context or namespaced scratch org) *and* when it is not (in an unmanaged context).
```
justinvholt, line 93: This is getting hard to follow simply because I don’t fully know why things are required to be namespaced. Which is okay because I'm still quite the novice when it comes to building anything with ``cci``. That said, is it safe to presume that most devs this deep into the material know about namespaces, why they're needed and will follow this material with (relative) ease?

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+
+CumulusCI uses a strategy called *namespace injection* to support this use case. Namespace injection is very powerful, and requires care from application implementors to ensure that metadata remains deployable in all contexts.
```
justinvholt, line 95: "application implementors"

Just confirming this is what we want to call them because this is the first time I've heard this label being used.

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+* ``%%%NAMESPACED_ORG_OR_C%%%`` is replaced with the package's namespace in a namespaced org *only*, not in a managed installation. Otherwise, it is replaced with ``c``, the generic namespace used in Lightning components.
+* ``%%%NAMESPACE_DOT%%%`` is replaced with the package’s namespace in any context with a namespace (such as a namespaced org or managed org) followed by a period (``.``) rather than two underscores.
+    .. note:: This token is used to construct references to packaged Record Types.
+
+..
+
+    Example: A portion of metadata from the Nonprofit Success Pack (NPSP) is stored in a subdirectory under ``unpackaged/post``, meaning it's deployed after the application metadata. This metadata updates a Compact Layout on the ``Account`` object, and references packaged metadata from the application as well as from other managed packages. To complete these tasks, this metadata requires the use of namespace tokens to represent the ``npsp`` namespace, letting CumulusCI automatically adapt the metadata to deploy into managed and unmanaged contexts.
```
justinvholt, line 109: “To complete these tasks...”

Added this for clarity, so I’m simply confirming that the updating and referencing are indeed tasks. And if not, is there a better word? Steps?

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+            <fields>npo02__TotalOppAmount__c</fields>
+            <fields>%%%NAMESPACE%%%Number_of_Household_Members__c</fields>
+            <label>NPSP Household Account</label>
+        </compactLayouts>
+    </CustomObject>
+
+..
+
+    Note that only the reference to the NPSP field ``Number_of_Household_Members__c`` is tokenized. (This field is called ``npsp__Number_of_Household_Members__c`` when installed as part of the managed package.) References to NPSP's own managed package dependency, ``npo02``, are not tokenized because this metadata is always namespaced when installed.
+
```
justinvholt, line 127: Why this specific field? Also, is this detail in parentheses relevant/helpful for a dev to know for this example to make sense? I honestly can't make heads or tails of it.  To be honest, it feels like we’re currently missing one or two details here about why only this field is the one to get tokenized.

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+    If this metadata isn't tokenized, it fails to deploy into an org containing NPSP as a beta or released managed package.
+
```
justinvholt, line 129: Why? This example might need more clarification, or it might need to be swapped out with one easier to comprehend.

(Or it might be completely fine for an advanced dev. Your call.)

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+.. note:: The resolution of component references in namespaced scratch orgs and in managed installations of the same metadata are not identical. Metadata that is tokenized and deploys cleanly in a namespaced scratch org can still fail in a managed context.
```
justinvholt, line 130: This seems like real inside baseball. Is it reasonable to presume that advanced devs will understand this? (Again, it's completely fine if I don't because I presume at this point that only advanced devs would be this far into the written materials.)

docs/unpackaged.rst
```diff
@@ -1,14 +1,212 @@
...
+* Whenever possible, build your unpackaged metadata in an org that contains a beta or released managed package. By doing so, the metadata contains namespaces when extracted, which CumulusCI easily replaces with tokens when retrieving metadata. It's difficult to manually tokenize metadata that's retrieved from an unmanaged org without namespaces. 
+
+After building changes to unpackaged metadata in a managed org, retrieve it using ``retrieve_changes`` with the additional ``namespace_tokenize`` option, and use the ``path`` option to direct the retrieved metadata to your desired unpackaged directory.
+
+    Example: Run ``retrieve_changes`` to retrieve metadata changes into the ``unpackaged/config/qa`` subdirectory, and replace references to the namespace ``npsp`` with the appropriate token.
+
+.. code-block:: console
+
+    $ cci task run retrieve_changes --path unpackaged/config/qa --namespace_tokenize npsp
```
justinvholt, line 151: Swapped out the old -o flags. Please confirm this updated command is correct.

docs/unpackaged.rst
Br4nd0R: [ETL](https://en.wikipedia.org/wiki/Extract,_transform,_load) is an established technique/procedure/methodology. I wouldn't refer to it as a feature.

Ya, this sentence could be better. Maybe: "It's a powerful technique that alters the unpackaged configuration in an org without risking damage to existing customization by overwriting them with incoming metadata."

Let me know if that's any better.

docs/unpackaged.rst
Br4nd0R:
```suggestion
Projects often include metadata that is genuinely part of the application, but either cannot be delivered as part of a managed package for operational reasons. This metadata must be deployed *after* the package's own code and metadata are deployed first and the org is configured.
```

docs/unpackaged.rst
Br4nd0R: @davidmreed - Is a non-namespaced package just another way of saying 'unamangeg/unlocked' package? Also wondering about Justin's question here. The following example outlines how `unpackaged/post` needs to work in both managed and unmanaged contexts; however, if I'm building an org implementation in a package without a namespace, what's the reason for needing to build namespaced packages?

docs/unpackaged.rst
Br4nd0R: This example is clear to me (not sure how helpful this is), but I think it should be safe to assume that people reading this are going to be interested in, or familiar with, Salesforce packages in general. We might want to change this example based on feedback from David R. regarding your prior question.

docs/unpackaged.rst
Br4nd0R: We could maybe use "application developers and architects" as a substitute?

docs/unpackaged.rst
davidmreed: Yes to the Brandon edit!

docs/unpackaged.rst
davidmreed, line 39:
```suggestion
Projects often include metadata that is genuinely part of the application, but cannot be delivered as part of a managed package for operational reasons. This metadata must be deployed *after* the project's own code and metadata are deployed first and the org is configured.
```

docs/unpackaged.rst
davidmreed: I think I wrote this sentence, but it's total nonsense. 

```suggestion
Because projects that are building an org implementation or a non-namespaced package do not have a namespace, or a distinction between managed and unmanaged contexts, these projects typically don't need to use namespace injection.
```

docs/unpackaged.rst
davidmreed, line 93:
```suggestion
    Example: Metadata located in ``unpackaged/post`` is deployed after the application code in both unmanaged and managed contexts. If that metadata contains references to the application components, it must be deployable when that metadata is namespaced (in a managed context or namespaced scratch org) *and* when it is not (in an unmanaged context).
```

davidmreed: I'm happy to expand on or clarify this example but I think it's an important one because it is very common.

docs/unpackaged.rst
davidmreed:
```suggestion
CumulusCI uses a strategy called *namespace injection* to support this use case. Namespace injection is very powerful, and requires care from the project team to ensure that metadata remains deployable in all contexts.
```

docs/unpackaged.rst
Br4nd0R: Hm. Not sure that we need that there. The "update" and "referencing" in the paragraph are referring to characteristics of particular piece of metadata, not things that will be done in the future. When deployed, the metadata updates a Compact Layout on the Account object, and it also has references to another managed package (in the form of the `npo02__TotalOppAmount__c` field).

davidmreed: "Steps" would be fine. "Tasks" is definitely confusing and ambiguous.

docs/unpackaged.rst
davidmreed: There's no specific importance to this field, it's just a good example.

But if it's not clear why this field is tokenized, the explanation so far has failed. Should we talk through it? n.b. this are is _notoriously_ difficult to understand for users and even release engineers.

docs/unpackaged.rst
davidmreed:
```suggestion
    If this metadata isn't tokenized, it fails to deploy into an org containing NPSP as a beta or released managed package — because in that context, the field `Number_of_Household_Members__c` is namespaced, as `npsp__ Number_of_Household_Members__c`, and we must refer to it as such.

```

docs/unpackaged.rst
davidmreed: This is Deep Magic. I think it's an important note but it's okay that not everyone gets it. Those who are operating with these tools (namespaced scratch orgs + managed packages) will hopefully get it.

docs/unpackaged.rst
davidmreed: This looks correct to me.

docs/unpackaged.rst
Br4nd0R: Salesforce has a convention of using `namespace__` in front of the field name to indicate which package a particular field belongs to. This particular field is the only one in the compact layout that is a member of the `npsp` package (this is why it's the only one getting the token). We know this only because we were told so here. 

The detail in the parens does help I believe, it's letting us know that in a "managed" context this field will be display as `nspsp__Number_of_Household_Members__c` in an org. This is useful because then I can see how the `%%%NAMESPACE%%%` token is getting replaced with `npsp__` in a "managed" context.

docs/unpackaged.rst
Br4nd0R: We are pretty deep in the weeds here. We should assume that users reading this section are likely a release engineer familiar with packaging basics. But I think we could maybe make this a little more explicit in how the error actually occurs.

@davidmreed - I need you help on how to finish the example. I believe I'm correct up until "the field would be deployed" part. Is the error on deploy, or is this more related with a problem that occurs in the `package.xml` file, or something else?

```suggestion
    If this metadata isn't tokenized, it will fail to deploy into an org containing NPSP as a beta or released managed package. Since all metadata is retrieved in an unmanaged context (i.e. without a namespace prefix) during development, if there is no `%%%NAMESPACE%%%` token present, the field would be deployed without a namespace, and thus result in an error.
```

docs/unpackaged.rst
Br4nd0R: I think this should be fine. There's a lot of jargon, but a release engineer or advanced dev is going to be the target role for this content.

docs/unpackaged.rst
Br4nd0R: 10-4

docs/unpackaged.rst
Br4nd0R: Let's keep this then.

PR #2070, opened 2020-10-05
Stop memoizing the salesforce_client on OrgConfig

davisagli:
Constructing a new one is cheap (it doesn't make a request when instantiated with an existing access token) and avoids the potential for using an old connection that was closed by the server.

# Critical Changes

# Changes

# Issues Closed
- Fixed a connection error which could occur if uploading a package took over 10 minutes.

### DIFF
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -116,12 +116,9 @@ class OrgConfig(BaseConfig):
         else:
             return self.instance_url.split(".")[0] + ".lightning.force.com"
 
-    # to simplify mocking for tests
-    _Salesforce = Salesforce
-
     @property
     def salesforce_client(self):
-        return self._Salesforce(
+        return Salesforce(
             instance=self.instance_url.replace("https://", ""),
             session_id=self.access_token,
             version=self.latest_api_version,
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -38,7 +38,6 @@ class OrgConfig(BaseConfig):
 
         self.name = name
         self._community_info_cache = {}
-        self._client = None
         self._latest_api_version = None
         self._installed_packages = None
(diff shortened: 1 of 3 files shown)


cumulusci/core/config/OrgConfig.py
```diff
@@ -120,16 +116,16 @@ def lightning_base_url(self):
         else:
             return self.instance_url.split(".")[0] + ".lightning.force.com"
 
+    # to simplify mocking for tests
+    _Salesforce = Salesforce
```
prescod, line 120: Where in the tests is _Salesforce referenced?

cumulusci/core/config/SfdxOrgConfig.py
```diff
@@ -137,7 +137,6 @@ def force_refresh_oauth_token(self):
```
prescod, line 137: While you are in this code can you look at the function above? A) The comment on Line 125 doesn't seem right in several ways,
B) the stdout_list should only be parsed in the "if" block and C) I'm just taking it "on faith" that this function which doesn't write to self.anything at all is nevertheless doing everything it needs to do...just forcing a refresh at the SFDX level. Is it properly named?

cumulusci/core/config/OrgConfig.py
davisagli: This was vestigial from an earlier attempt. Fixed.

prescod (approved):

cumulusci/core/config/SfdxOrgConfig.py
davisagli: @prescod

A and B -- I agree, but not in the scope of this PR
C - Yes, it just forces a refresh at the sfdx level. It is only called from `SfdxOrgConfig.refresh_oauth_token`, which first deletes `self._sfdx_info` so that future attempts to get the access token will re-fetch info from sfdx. It wouldn't hurt to move clearing _sfdx_info into `force_refresh_oauth_token` so that it can't be missed.

PR #2074, opened 2020-10-06
Fix handling of Boolean in Person Account loads

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue causing non-Person-Account Contacts to be omitted during Person Account data loads when the dataset was extracted using the REST API.


davisagli (approved):
I sort of think we should be lowercasing boolean values during extract so that they are stored in a format that makes sense to Salesforce. But this is fine as a quick fix.

PR #2079, opened 2020-10-07
Allow update_admin_profile to use a custom package with extra Profile names

davidmreed:
# Critical Changes

# Changes

- The `update_admin_profile` task now accepts the `api_names` option to target extra Profiles, even when using a custom `package.xml`.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/update_profile.py
+++ b/cumulusci/tasks/salesforce/update_profile.py
@@ -137,7 +137,7 @@ class ProfileGrantAllAccess(MetadataSingleEntityTransformTask, BaseSalesforceApi
 
     def _generate_package_xml(self, operation):
         if operation is MetadataOperation.RETRIEVE:
-            with open(self.package_xml_path, "r") as f:
+            with open(self.package_xml_path, "r", encoding="utf-8") as f:
                 package_xml_content = f.read()
 
             package_xml_content = package_xml_content.format(**self.namespace_prefixes)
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/update_profile.py
```diff
@@ -148,27 +142,20 @@ def _generate_package_xml(self, operation):
...
-                # which `fromstring()` doesn't like.
-                package_xml = metadata_tree.fromstring(
-                    package_xml_content.encode("utf-8")
-                )
+            # We need to rewrite the package.xml for one or two reasons.
+            # Either we are using packaged-object expansion, or we're using
+            # a package.xml and need to substitute in profile API names.
+
+            # Convert to bytes because stored `package.xml`s typically have an encoding declaration,
+            # which `fromstring()` doesn't like.
```
davisagli, line 150: This is a tangent from this PR, but we should be sure to specify encoding='utf-8' when opening the file on line 140.

cumulusci/tasks/salesforce/update_profile.py
```diff
@@ -180,8 +167,13 @@ def _expand_profile_members(self, package_xml):
             raise CumulusCIException(
                 "The package.xml does not contain a Profiles member."
             )
```
davisagli, line 169: Can we add the Profiles section if it's missing instead of raising an exception?

davisagli (changes requested):

davisagli (approved):

PR #2082, opened 2020-10-13
merge non-feature source into feature branches

Br4nd0R:
# Changes
* The task `github_automerge_main` now merges the most recent commit into branches that begin with the `feature_prefix` when the source branch is a non-feature branch (includeing main/master).


davisagli (approved):
I made a couple small adjustments.

PR #2095, opened 2020-10-15
Snowfakery 1.2

prescod:
Upgrade to Snowfakery 1.2


davisagli:
Adding this to https://github.com/SFDO-Tooling/CumulusCI/pull/2091 instead

PR #2097, opened 2020-10-15
Project-specific task docs

jofsky:
# Critical Changes

# Changes
Creates a command `cci project doc` which docs up all project specific tasks to an RST file.

# Issues Closed


cumulusci/cli/cci.py
```diff
@@ -722,6 +722,33 @@ def project_dependencies(runtime):
...
+@project.command(
+    name="doc", help="Exports RST format documentation for all project tasks"
+)
+@pass_runtime(require_project=False)
+def project_doc(runtime):
+
+    try:
+        os.mkdir("./docs")
+    except Exception:
+        pass  # skipping as not to freak end user out since directory exists.
```
davisagli, line 734: A better way to do this is to use pathlib:

Path("docs").mkdir(exist_ok=True)

cumulusci/cli/cci.py
```diff
@@ -722,6 +722,33 @@ def project_dependencies(runtime):
...
+    finally:
+        project_config = runtime.project_config.config
+        project_name = project_config["project"]["name"].capitalize()
```
davisagli, line 737:
```suggestion
        project_name = runtime.project_config.project__name.capitalize()
```

davisagli: (The .config dict of Config objects is intended to be for internal use)

cumulusci/cli/cci.py
```diff
@@ -722,6 +722,33 @@ def project_dependencies(runtime):
...
+
+        with open("./docs/project_tasks.rst", "w") as f:
```
davisagli, line 739: Be sure to specify encoding="utf-8"; there's a different default on Windows

cumulusci/cli/cci.py
```diff
@@ -722,6 +722,33 @@ def project_dependencies(runtime):
...
+            f.write("==========================================\n")
+            f.write(f"{project_name} Tasks Reference\n")
+            f.write("==========================================\n")
+            f.write("\n")
+            for name, options in project_config["tasks"].items():
+                task_config = TaskConfig(options)
+                doc = doc_task(name, task_config)
```
davisagli, line 746: Put this inside the `if` so we aren't spending time building docs for tasks that we're not going to include in the output.

cumulusci/cli/cci.py
```diff
@@ -722,6 +722,33 @@ def project_dependencies(runtime):
...
+                if name in runtime.project_config.config_project["tasks"]:
+                    f.write(f"{doc}")
+                    f.write("\n\n")
```
davisagli, line 749: So just to confirm here, the goal is to include any task that was added or changed in the project's cumulusci.yml, but to write documentation using the full merged task config which may include pieces from the project's cumulusci.yml merged with the universal cumulusci.yml? That's what I see here and I think it makes sense, but just wanted to spell it out.

davisagli (changes requested):
Why do you prefer a new `cci project doc` command instead of adding a `--project-only` flag to the `cci task docs` command?

cumulusci/cli/tests/test_cci.py
```diff
@@ -715,6 +715,18 @@ def test_project_dependencies(self):
...
+    def test_project_doc(self, doc_task):
+        runtime = mock.Mock()
+        runtime.universal_config.tasks = {"test": {}}
+        runtime.project_config.config = {
+            "project": {"name": "Test"},
+            "tasks": {"option": {"a": "b"}},
+        }
+        runtime.project_config.config_project = {"tasks": {"option": {"a": "b"}}}
+        run_click_command(cci.project_doc, runtime=runtime)
+        doc_task.assert_called()
```
Br4nd0R, line 728: I think it would be a good idea to also assert that we are receiving the right output to `click.echo` for a given project configuration.

cumulusci/cli/cci.py
jofsky: Correct should it be another way?

jofsky:
> Why do you prefer a new `cci project doc` command instead of adding a `--project-only` flag to the `cci task docs` command?

@davisagli I did this because I want it to be documentation at the project level, task doc implies that it will only document the tasks but in reality this project doc task will have more than tasks.

cumulusci/cli/tests/test_cci.py
jofsky: Isn't this covered by `test_project_info` , especially since this doesn't create any output. Maybe it should, I'm open to change.

davisagli:
@jofsky I wasn't necessarily asking you to go ahead and remove "cci project doc". That makes sense that if there are multiple types of things to be documented, it would be helpful to have one command to do it all instead of several.

But before you make more changes, there's one other thing I want to make sure I understand, which is: How do you anticipate using this project docs feature? Is it something that individual users of cci will run? Is it something that you as the release engineer will run and then commit the generated docs to the project repository? If the latter, will it be a challenge to keep the docs up to date over time?

cumulusci/cli/tests/test_cci.py
Br4nd0R: Ya, that's fine with me.

cumulusci/cli/cci.py
```diff
@@ -1345,30 +1343,59 @@ def task_list(runtime, plain, print_json):
...
+        click.echo("==========================================")
+        click.echo("Tasks Reference")
+        click.echo("==========================================")
         click.echo("")
 
+        for name, options in config_src.tasks.items():
+            task_config = TaskConfig(options)
+            doc = doc_task(name, task_config)
+            click.echo(doc)
+            click.echo("")
```
davisagli, line 1390: Let's refactor this to avoid repeating the code that is currently repeated:
1. Determine the dict of tasks to be documented, based on the `project` option
2. Loop over the tasks, get their docs, and add to a result variable (list or string).
3. Do the actual output, by echoing the result or writing it to the file, based on the `write` option.

cumulusci/cli/cci.py
```diff
@@ -1345,30 +1343,59 @@ def task_list(runtime, plain, print_json):
...
-    click.echo("Tasks Reference")
-    click.echo("==========================================")
-    click.echo("")
-
-    for name, options in config_src.tasks.items():
-        task_config = TaskConfig(options)
-        doc = doc_task(name, task_config)
-        click.echo(doc)
+@click.option("--project", "project", is_flag=True, help="Write project specific tasks")
+@click.option("--write", "write", is_flag=True, help="Print a json string")
```
davisagli, line 1347: This help needs to be updated

davisagli (changes requested):

cumulusci/cli/cci.py
```diff
@@ -1345,30 +1343,55 @@ def task_list(runtime, plain, print_json):
...
+    for name, options in tasks:
         task_config = TaskConfig(options)
         doc = doc_task(name, task_config)
-        click.echo(doc)
-        click.echo("")
+        result += [f"{doc}", ""]
+    result = "\r\n".join(result)
+    if write:
+        Path("docs").mkdir(exist_ok=True)
+        with open(file_name, "w", encoding="utf-8") as f:
```
Br4nd0R, line 1382: We could reduce clutter at the top by moving the `file_name` var down here closer to where it's used. It's then also easier to reference in the code.
```suggestion
        Path("docs").mkdir(exist_ok=True)
        file_name = "project_tasks" if project else "cumulusci_tasks"
        with open(f"./docs/{file_name}.rst", "w", encoding="utf-8") as f:
```

cumulusci/cli/cci.py
```diff
@@ -1345,30 +1343,55 @@ def task_list(runtime, plain, print_json):
...
+    tasks = config.tasks.items()
+    file_name = "./docs/cumulusci_tasks.rst"
+    # setting to read for project specific configuration
+    if project:
+        config = runtime.project_config.config
+        project_name = runtime.project_config.project__name.capitalize()
+        tasks = config["tasks"].items()
+        file_name = "./docs/project_tasks.rst"
+    # handling for general and project specific documentation
+    result.append(f"{project_name} Tasks Reference") if project else result.append(
```
Br4nd0R, line 1368: We can also move the `result` list down here to keep it closer to where it's being utilized.
```suggestion
    result = ["=========================================="]
    result.append(f"{project_name} Tasks Reference") if project else result.append(
```

Br4nd0R:
@jofsky - We will also want to ensure that running tests doesn't have any side effects. Currently these tests would overwrite  any existing `project_tasks.rst` file.

jofsky:
> @jofsky - We will also want to ensure that running tests doesn't have any side effects. Currently these tests would overwrite any existing `project_tasks.rst` file.

@Br4nd0R how do you suggest we deal with this?

Br4nd0R:
@jofsky - Easiest route would probably be to `@patch(cumulusci.cli.cci.open)` on any tests that are writing output.  You may have to explicitly define the context manager `__enter__` to return a mock object too (which is what `f` will get the value of in calls to `f.write()`.

You can then assert the `open()` was called with the correct filename as well.

Br4nd0R:
I'll let @davisagli give this the final rubber stamping. Thanks for making those changes!

cumulusci/cli/cci.py
```diff
@@ -1345,30 +1343,56 @@ def task_list(runtime, plain, print_json):
...
+)
+def task_doc(runtime, project=False, write=False):
+    config = runtime.universal_config
+    tasks = config.tasks.items()
+    file_name = "./docs/cumulusci_tasks.rst"
+    # setting to read for project specific configuration
+    if project:
+        config = runtime.project_config.config
+        project_name = runtime.project_config.project__name.capitalize()
+        tasks = config["tasks"].items()
```
davisagli, line 1364: I tried it and it's still including all tasks when I use the `--project` option. project_config.config contains the _merged_ config combined from all sources. I think you want project_config.config_project here.

davisagli (dismissed):

davisagli (approved):

PR #2104, opened 2020-10-18
Allow integer/decimal values in $project_config

Br4nd0R:
# Changes
* Allow for `$project_config` to reference integer or decimal values in the `cumulusci.yml` file.

Example:
```
variables:
    seconds: 32.88

tasks:
    util_sleep:
        options:
            seconds: $project_config.variables__seconds
```


davisagli:
@Br4nd0R what's the use case that's motivating this?

Br4nd0R:
@davisagli - While working on a doc section regarding the `$project_config` var, I just happened to notice this behavior and submit a fix. If you don't think it fits, or will have unintended consequences I'm not set on this needing to be merged. 

I'm thinking that allowing users to be able to provide string, decimal, and integer values through the `$project_config` var in their `cumulusci.yml` files is more flexible.

davisagli:
@Br4nd0R It does mean that the non-string value is getting converted to a string and then the task using the option is responsible for converting it back into the format it needs, so there's potential for weird edge cases. But generally tasks need to handle string input already (from options passed as CLI parameters) so I think it'll be okay.

davisagli (approved):

PR #2103, opened 2020-10-17
Add "Salesforce" to Keyword Documentation link

Julian88Tex:
# Critical Changes

# Changes
- Add the word "Salesforce" since it includes both CumulusCI and Salesforce Keyword documentation.

# Issues Closed

PR #2105, opened 2020-10-19
Doc fix

davisagli:
This incorporates @Julian88Tex's doc fix from #2103, and also fixes the release test which needs to use dev.json instead of prerelease.json now that we're out of the prerelease period.

# Critical Changes

# Changes

# Issues Closed


Br4nd0R (approved):
Thanks @Julian88Tex!

PR #2106, opened 2020-10-19
Workaround for slow SPV query by splitting it apart

davisagli:
We won't want this in the long term, because it adds more queries, but I think it will help avoid the query timeout we're currently seeing with InstalledSubscriberPackage

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/.github/workflows/release_test.yml
+++ b/.github/workflows/release_test.yml
@@ -7,7 +7,7 @@ on:
     - master
 
 env:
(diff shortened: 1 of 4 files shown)


Br4nd0R (approved):

davidmreed (approved):

davidmreed:
I ran a build that uses relevant `when` clauses using this branch and it completed successfully.

PR #2110, opened 2020-10-19
Release Branch Fix

Br4nd0R:
# Changes
* `github_automerge_main` now only merges to the lowest numbered release branch when multiple are detected.

### DIFF
--- a/cumulusci/tasks/github/merge.py
+++ b/cumulusci/tasks/github/merge.py
@@ -243,6 +243,15 @@ class MergeBranch(BaseGithubTask):
         },
     }
 
+    def __init__(self, *args, **kwargs):
+        self._repo = None
+        self._next_release = None
+        self._repo_branches = None
+        self._update_future_releases = False
+        self._existing_prs = None
+
+        super().__init__(*args, **kwargs)
+
     def _init_options(self, kwargs):
         super()._init_options(kwargs)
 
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/github/merge.py
```diff
@@ -286,6 +288,49 @@ def _validate_source_branch(self):
...
 
+    def _set_next_release(self):
+        # release_nums = []
+        # for b in self.repo.branches():
+        # if self._is_release_branch(b.name):
+        # release_nums.append(int(b.name.split("/")[1]))
+
+        release_nums = [
+            int(branch.name.split("/")[1])
+            for branch in self.repo.branches()
```
davisagli, line 299: Let's be sure to fetch the list of branches from github once so we use fewer API calls. It's currently happening here as well as in _get_branches_to_merge.

Also, filtering the release branches seems like something that should be considered part of getting the branches to merge, conceptually. Would it make sense to move the calls to _set_next_release and _set_should_update_future_releases into _get_branches_to_merge?

I'd also encourage getting next_release as the return value from _set_next_release and then passing it as an argument to _set_should_update_future_releases (and remove set from their names) because it makes it clearer what the inputs and outputs are. Storing intermediate values as instance attributes means the instance needs to be in a particular state in order to run the method successfully.

Br4nd0R: @davisagli - Looking at this again, using more `@property` decorators on the task helped clean this up a little and also aided with making fewer API calls. I'm still not completely satisfied with how `_get_branches_to_merge()` looks.

I feel like it could be refactored to pass in a list of 'checks' which are just functions that adhere to an interface. We iterate through all branches and for each branch we iterate over all checks and `if any([check(branch) for check in merge_checks])` then we add the branch to the list to be merged.

Br4nd0R: @davisagli - Not sure why I sent the route of making setter methods. I switched several items over to be `@property`s of the task. This helped to reduce the number of API calls and help improve readability. A couple of the suggestions you mentioned no longer apply because the methods no longer exist.

davisagli (approved):
@Br4nd0R I forget if you wanted to do more work on this before merging and testing -- I'm fine either way

PR #2111, opened 2020-10-20
Feature/robot pabot fixes

boakley:
This explicitly adds the repo root to `PYTHONPATH` in the case where we are running robot tests with `pabot`. We do this by passing the `--pythonpath` option along with the value we get from `self.project_config.repo_root` to `pabot`.

Without this change, when we use `pabot` to run tests in parallel most tests will be unable to load any libraries or keyword files specified relative to the root, causing catastrophic failures as tests try to use keywords that were unable to be imported. 

**Background**

By default, robot does not look in the current directory or repo root to resolve relative paths. Robot assumes relative paths are relative to the directory containing the test being run. If it can't find a file relative to that directory, it searches the directories in PYTHONPATH. 

When a task runs, we add the repo root to `sys.path`, and thus robot is able to find our files relative to the repo root since robot runs in the same process as the task. However, when we run tests in parallel, pabot spawns multiple processes which do not inherit `sys.path` from the task process. Therefore we have to explicitly add the repo root to PYTHONPATH when calling pabot.

I tested this by creating a simple test case in the `Insights-UI` repository which looked like this:

```
*** Settings ***
Library        robot/Insights-UI/resources/Insights.py

*** Test Cases ***
Example
    # from Insights.py
    get start time
```

To trigger the pabot code path I ran this test with the `-o processes 2` option. Without the fix the test would fail because it couldn't find `robot/Insights-UI/resources/Insights.py`. With this change the test runs and passes. 

# Critical Changes

# Changes

# Issues Closed


davisagli (approved):
Thanks.

PR #2115, opened 2020-10-21
Add assign_permission_sets task

davidmreed:
# Critical Changes

# Changes

- CumulusCI offers an `assign_permission_sets` task, which assigns one or more Permission Sets to the running user if not already assigned. This task may replace invocations of `dx`.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/users/permsets.py
+++ b/cumulusci/tasks/salesforce/users/permsets.py
@@ -1,0 +1,53 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from cumulusci.core.exceptions import CumulusCIException
+from cumulusci.core.utils import process_list_arg
+
+
+class AssignPermissionSets(BaseSalesforceApiTask):
+    task_options = {
+        "api_names": {
+            "description": "API names of desired Permission Sets, separated by commas.",
+            "required": True,
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+
+        self.options["api_names"] = process_list_arg(self.options["api_names"])
+
+    def _run_task(self):
+        # Determine existing assignments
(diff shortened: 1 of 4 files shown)


Br4nd0R (dismissed):
LGTM!

cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -0,0 +1,53 @@
...
+
+        self.options["api_names"] = process_list_arg(self.options["api_names"])
+
+    def _run_task(self):
+        # Determine existing assignments
+        query = f"""SELECT Id,
+                           (SELECT PermissionSetId
+                            FROM PermissionSetAssignments)
+                    FROM User
+                    WHERE Id = '{self.org_config.user_id}'"""
```
davisagli, line 25: For scratch orgs, `user_id` is a property which does its own query of User filtering by username. Would it make more sense to just filter by username in this query?

cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -0,0 +1,53 @@
...
+
+        # Assign all not already assigned
+        for api_name in self.options["api_names"]:
+            if permsets[api_name] not in assigned_permsets:
+                self.sf.PermissionSetAssignment.create(
+                    {
+                        "AssigneeId": self.org_config.user_id,
+                        "PermissionSetId": permsets[api_name],
+                    }
+                )
```
davisagli, line 53: Log which permsets are actually getting assigned

davisagli (changes requested):

davisagli (dismissed):

davidmreed:
After discussion with @bethbrains added ability to target a user by alias.

bethbrains:
Hooray! Thank you!

davisagli (approved):

PR #2118, opened 2020-10-22
Tighten up handling of boolean in data loads, CLIs and YAMLs.

prescod:
For dataloads: previously only a "True" and "true" were considered true and everything else was false. Now, we use the same set of true and false strings, ints and bools that the Salesforce Data Loader uses. Anything else throws an exception. One additional argument for this open stance is that the "SQLite standard" string representation of True is "1" and False is "0" and in postgres it is "true" and "false".

For command lines: previously it could return True for known-true values, False for known-false values and None for anything else. Of course the "None" was generally interpreted as "False" so, for example, "yes" was "False". Now it uses the same rules as above.

None is considered a known-false value because many tasks are coded that way. I am open to fixing the ones in our control and adding a deprecation warning to the others but I didn't know if that would be the team consensus so I pulled back from that plan.

### DIFF
--- a/cumulusci/core/utils.py
+++ b/cumulusci/core/utils.py
@@ -10,6 +10,7 @@ import glob
 import pytz
 import time
 from shutil import rmtree
+from typing import Union
 
 from cumulusci.core.exceptions import ConfigMergeError, TaskOptionsError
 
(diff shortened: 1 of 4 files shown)


cumulusci/core/utils.py
```diff
@@ -33,15 +34,27 @@ def parse_datetime(dt_str, format):
...
+
+    None is considered false.
+    Similar to parts of the Salesforce API, there are a few true-ish and false-ish strings,
+        but "True" and "False" are the canonical ones."""
+    if isinstance(arg, (int, bool)):
+        return bool(arg)
+    elif arg is None:
+        # backwards compatible behaviour that some tasks
+        # rely upon.
+        return False
```
davisagli, line 48: I'm not quite following how this is the backwards-compatible behavior. Previously `None` would have failed the isinstance checks for bool and str, and returned None, like the test above showed.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -420,10 +420,12 @@ def load_records(self, records):
         def _convert(rec):
             result = dict(zip(self.fields, rec))
             for boolean_field in self.boolean_fields:
-                # This is clumsy but required since Booleans are stored as.
-                result[boolean_field] = (
-                    result[boolean_field] or "false"
-                ).lower() == "true"
+                try:
+                    result[boolean_field] = process_bool_arg(
+                        result[boolean_field] or False
```
davisagli, line 425: What purpose is served by the `or False`?

cumulusci/core/utils.py
prescod: It's not a backwards-identical behaviour but I think it is close enough. I prefer that now this function will definitively return True or False, but since the return value for f(None) is still false-y, I doubt it will break any code.

cumulusci/tasks/bulkdata/step.py
prescod: It avoids passing None given that accepting None is just a quirk of process_bool_arg() and not a feature we'd prefer to support forever.

cumulusci/core/utils.py
davisagli: I like the change and agree it's pretty low risk. Just remove the comment here, which confused me because it seems to claim that this is the same return value as before.

davisagli: Oh, your other comment helped me figure out what you mean. You want to reject None as input, but are keeping it backwards compatible by interpreting it as falsy. Okay.

davisagli (approved):

PR #2120, opened 2020-10-23
Doc reorg:  Robot

Br4nd0R:
Rendered: https://cumulusci.readthedocs.io/en/feature-doc-reorg__robot/

This is just the copied current documentation that we have for Robot Framework. Not sure if we want to break it up into multiple sections or if there's a better way to organize it. It's pretty long when it's in one lump.

### DIFF
--- a/docs/robot.rst
+++ b/docs/robot.rst
@@ -1,11 +1,1362 @@
-Automation using Robot Framework
+Automation Using Robot Framework
 ================================
+
+
 
 Introduction to Robot Framework
 -------------------------------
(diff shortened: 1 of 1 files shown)


davisagli:
@Br4nd0R I think it would make sense to keep this as the 3 pages it already is (intro, tutorial, debugger), but arranged so they will all appear under a top-level "Robot Framework" header in the left-side navigation. You should be able to make sphinx do this by turning robot.rst into a file that only includes the heading "Robot Framework", a brief line introducing the section, and its own `toctree` directive that lists the other pages.

davisagli:
@Br4nd0R Also, I see two other open pull requests that are making adjustments to the robot framework docs, targeting master. For sections like this that aren't going to be changed much, we might want to wait until we're close to done with the docs that involve more writing and editing, to minimize the chances of missing changes from master when we merge the docs reorg.

PR #2121, opened 2020-10-23
Feature/robot sources

boakley:
This adds the ability to use keywords and tests from other repositories. It piggybacks off of the `sources` global configuration option along with a new `sources` option in the robot task. 

In short, the cached directories of one or more of the "sources" is added to PYTHONPATH when we run the robot task. Robot uses PYTHONPATH to find libraries and resource files. In addition, I've modified the page object code to also use PYTHONPATH to find page object files.

Sravani has tested this in the spc-integration-npsp repository and verified that it does what she needs it to do.

# Critical Changes

# Changes
* The "suites" option of the robot task can now take a list of suite paths. Paths can include a prefix representing a remote repository as defined by the "sources" configuration option (eg: `-o suites npsp:robot/Cumulus/tests/api`)
* The robot task has a new `sources` option to work in conjunction with the global `sources` option to allow the use of keywords and tests from other repositories
* When running the robot task, the folder containing downloaded repositories via the `sources` option are added to PYTHONPATH so that robot tests can find library and resource files in those repositories

# Issues Closed


cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -79,11 +80,16 @@ def __init__(self, *args):
...
         for file_path in args:
-            try:
-                importer.import_class_or_module_by_path(os.path.abspath(file_path))
-                logger.debug("imported page object {}".format(file_path))
-            except Exception as e:
-                logger.warn(str(e))
+            path = self._find_file_in_pythonpath(file_path)
+            if path:
+                try:
+                    importer.import_class_or_module_by_path(os.path.abspath(path))
```
davisagli, line 86:
```suggestion
                    importer.import_class_or_module_by_path(str(path.resolve()))
```

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -79,11 +80,16 @@ def __init__(self, *args):
...
+                    logger.debug(f"imported page object {path}")
+                except Exception as e:
+                    logger.warn(str(e))
```
davisagli, line 89:
```suggestion
                    logger.warning(str(e))
```

(`warn` is deprecated)

Also, why are we using a warning here instead of letting the exception be raised? Doesn't that just hide the fact that the intended page object couldn't be imported, which could lead to confusing errors if something tries to use that page object later?

cumulusci/robotframework/pageobjects/PageObjects.py
```diff
@@ -79,11 +80,16 @@ def __init__(self, *args):
...
+            else:
+                logger.warn(f"Unable to find page object file '{file_path}'")
```
davisagli, line 91:
```suggestion
                logger.warning(f"Unable to find page object file '{file_path}'")
```

And same question about warning vs exception

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -86,6 +95,8 @@ def _init_options(self, kwargs):
         if process_bool_arg(self.options.get("pdb")):
             patch_statusreporter()
 
+        self.options.setdefault("sources", {})
```
davisagli, line 98: From what I see on lines 60-62 and 115, it looks like `sources` is meant to be a list rather than a dict.

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -97,6 +108,23 @@ def _run_task(self):
...
+                source_config = self.project_config.get_namespace(source)
+                source_paths[source] = source_config.repo_root
+            except NamespaceNotFoundError:
+                raise TaskOptionsError(f"robot source '{source}' could not be found")
+
+        # replace namespace prefixes with path to cached folder
+        for i, path in enumerate(self.options["suites"]):
+            prefix, _, path = path.rpartition(":")
+            if prefix in source_paths:
+                self.options["suites"][i] = f"{source_paths[prefix]}/{path}"
```
davisagli, line 126: Does this work in Windows, or do we need to join with os.sep?

Side note, I've never heard of rpartition before!

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -122,15 +150,27 @@ def _run_task(self):
                 else:
                     cmd.extend([f"--{option}", str(value)])
 
-            cmd.append(self.options["suites"])
+            # Add each source to pythonpath. Use --pythonpath since
+            # pybot will need to use that option for each process that
+            # it spawns.
+            for path in source_paths.values():
+                cmd.append("--pythonpath", path)
```
davisagli, line 157:
```suggestion
                cmd.extend(["--pythonpath", path])
```

davisagli (changes requested):

cumulusci/tasks/robotframework/robotframework.py
boakley: good catch

cumulusci/tasks/robotframework/robotframework.py
boakley: Forward slashes work on windows, but I'm not sure what the behavior is if part of the path includes forward slashes and part has backward slashes (eg: if they used forward slashes in cumulusci.yml to define sources). 

I'll use `os.path.join`. Good catch.

cumulusci/robotframework/pageobjects/PageObjects.py
boakley: That's a good question.  I don't remember why I thought a warning was more appropriate. Now that you bring it up, it does seem like this ought to throw an error. I've changed it to throw errors, and I've written a couple of tests.

cumulusci/tasks/robotframework/robotframework.py
boakley: done

davisagli (approved):

PR #2122, opened 2020-10-23
Deprecation warning for process_bool_arg(None)

prescod:
Clean up all uses of process_bool_arg(None) in our codebase and deprecate it officially.


cumulusci/tasks/bulkdata/delete.py
```diff
@@ -53,9 +53,11 @@ def _init_options(self, kwargs):
...
                 "Criteria cannot be specified if more than one object is specified."
             )
-        self.options["hardDelete"] = process_bool_arg(self.options.get("hardDelete"))
+        self.options["hardDelete"] = process_bool_arg(
+            self.options.get("hardDelete", False)
+        )
         self.options["ignore_row_errors"] = process_bool_arg(
-            self.options.get("ignore_row_errors")
+            self.options.get("ignore_row_errors", False)
         )
```
davisagli, line 61: Better to use `.get("option_name") or False` for all of these. The option could be explicitly set to null in yaml.

cumulusci/core/utils.py
```diff
@@ -37,14 +38,20 @@ def parse_datetime(dt_str, format):
...
+    """
     if isinstance(arg, (int, bool)):
         return bool(arg)
     elif arg is None:
         # backwards compatible behaviour that some tasks
         # rely upon.
+        warnings.warn(
+            "Future versions of CCI will not accept 'None' as an argument to process_bool_arg",
+            DeprecationWarning,
+        )
```
davisagli, line 54: This warning is not very actionable unless we show what called this using None as an argument.

davisagli (changes requested):

davisagli (approved):

PR #2123, opened 2020-10-24
Correct documentation and add a feature to Salesforce Query keyword

prescod:
Most urgently I was fixing documentation but while I was in there I added a feature that allows more flexible WHERE clauses (!=, LIKE etc.)


cumulusci/robotframework/Salesforce.py
```diff
@@ -935,21 +935,31 @@ def salesforce_query(self, obj_name, **kwargs):
...
         By default the results will only contain object Ids. You can
-        specify a SOQL SELECT clase via keyword arguments by passing
+        specify a SOQL SELECT clause via keyword arguments by passing
         a comma-separated list of fields with the ``select`` keyword
         argument.
 
-        Example:
+        You can supply keys and values to match against
+        in keyword arguments, or a full SQOL where-clause
+        in a keyword named ``where``.
```
boakley, line 944: This should probably be _"in a keyword **argument** named ..."_

cumulusci/robotframework/Salesforce.py
```diff
@@ -958,10 +968,13 @@ def salesforce_query(self, obj_name, **kwargs):
...
         query += " FROM {}".format(obj_name)
         where = []
-        for key, value in kwargs.items():
-            if key == "select":
-                continue
-            where.append("{} = '{}'".format(key, value))
+        if "where" in kwargs:
+            where = [kwargs["where"]]
+        else:
+            for key, value in kwargs.items():
```
boakley, line 974: It doesn't look like this code is prepared to handle the case where someone accidentally (or on purpose...) uses both keyword arguments and the "where" keyword. Should that be spelled out in the docs and/or handled explicitly in the code here?

For example:  `Salesforce query  FirstName="Bob"  where=FirstName="James"`

In this case, `FirstName="Bob"` is completely ignored. Is that intentional, or should the keyword throw a usage error?

boakley (commented):
Since we have a more general purpose "soql query" keyword, I'm not sure I see why we need this. If you need a complex WHERE clause, why not just use `SOQL Query`?

prescod:
@boakley 

soql_query returns "the keys as documented for the raw API call" as per the docs.

salesforce_query returns the records you ask for.

salesforce_query does the string interpolation for you.

soql_query requires you to do it yourself.

Personally, I find data manipulation in Robot code to be extremely non-intuitive and verbose, and Robot's target market is "non-programmers", so I think it is justified to try and minimize the data munging they need to do on input or output.

What even is the shortest, most idiomatic way to write:

``` @{records}=  Salesforce Query  Contact  select=Id,Name
 ...          where=FirstName!='Eleanor'
```

in CCI as it exists today?

boakley:
> Personally, I find data manipulation in Robot code to be extremely non-intuitive and verbose

Yeah, that's because it's not designed to do data manipulation. It's close enough to a programming language to trick you into thinking it is, but then it completely pulls the rug out from under you when you try to do anything complex.

boakley (dismissed):

cumulusci/robotframework/Salesforce.py
```diff
@@ -967,16 +968,15 @@ def salesforce_query(self, obj_name, **kwargs):
...
         if "where" in kwargs:
-            where = [kwargs["where"]]
-        else:
-            for key, value in kwargs.items():
-                if key == "select":
-                    continue
-                where.append("{} = '{}'".format(key, value))
-        if where:
-            query += " WHERE " + " AND ".join(where)
+            where_clauses = [kwargs["where"]]
```
boakley, line 973: I wasn't expecting you to support both, I was thinking you would allow one way or the other. <shrug> I guess this is fine, though I'm not sure anyone will use it.

boakley (approved):

cumulusci/robotframework/Salesforce.py
prescod: It was approximately just as easy to support both and it enables one to add an extra != or < or > WHERE clause to code that already exists without rewriting it.

PR #2124, opened 2020-10-25
Never share option dicts between tasks

prescod:
Task instances could share options dicts before if they shared a task config. Options dicts are mutable and frequently mutated, so tasks could interfere with each other, depending on how they were instantiated.

# Issues closed

- Fixed a bug where a task could accidentally mutate its options in a way that would affect other uses of the same task in the same flow.


Br4nd0R (approved):
Good catch!

Br4nd0R:
Looks like the test failures aren't specific to one version. I re-ran the jobs and have seen it fail on 3.7 and 3.8. Will need to tweak things tomorrow.

Br4nd0R (dismissed):

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -69,7 +69,14 @@ def _init_options(self, kwargs):
...
         if not path:
-            path = self.task_config.options__path
+            path = self.options.get("path")
+
+            if not path:
+                # Hopefully this is never triggered and we can delete it.
+                self.logger.warn(
+                    "Getting self.task_config.options__path in _get_api is deprecated"
+                )
+                path = self.task_config.options__path
```
davisagli, line 79: I don't think this is needed. If "path" was set in the task_config's options, then it will also be there in self.options.

davisagli (changes requested):

cumulusci/tasks/salesforce/Deploy.py
prescod: Do you have any guess why we looked it up in self.task_config.options__path ? Just a misunderstanding of how the framework works? I wasn't sure if there was a subtlety that I missed.

davisagli: I dug around in git blame and it looks like it was just written a long time ago when self.options was always a reference to self.task_options and there was no _init_options method.

davisagli: So not so much a misunderstanding as that the difference between the two didn't matter.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -69,7 +69,8 @@ def _init_options(self, kwargs):
 
     def _get_api(self, path=None):
         if not path:
-            path = self.task_config.options__path
+            path = self.options.get("path")
+            assert path, "Path should be specified for task org_settings"
```
davisagli, line 73: org_settings???

davisagli (changes requested):

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -69,7 +69,8 @@ def _init_options(self, kwargs):
 
     def _get_api(self, path=None):
         if not path:
-            path = self.task_config.options__path
+            path = self.options.get("path")
+            assert path, f"Path should be specified for {self.__class__.name}"
```
davisagli, line 73: I'll go ahead an approve since this is a nitpick about an edge case that we don't think should happen, but the task config name (`self.name`) would probably be more helpful here.

davisagli (approved):

cumulusci/tasks/salesforce/Deploy.py
prescod: In my 2 minutes of testing, self.name was None. Perhaps because it was run in the context of a unit test.

```
(Pdb) self.name
(Pdb) self.__class__.name
'BaseSalesforceMetadataApiTask'
```

prescod:
A robot test calls this with path not set, so I guess that is actually allowed and it can will just use a tempdir?

davisagli:
@prescod using a tempdir doesn't make sense; the option is specifying the source of the metadata to deploy. where's the robot test you mentioned?

prescod:
> @prescod using a tempdir doesn't make sense; the option is specifying the source of the metadata to deploy. where's the robot test you mentioned?

https://github.com/SFDO-Tooling/CumulusCI/blob/8e5722f3458528bf82efd69196abe5753611d3bb/cumulusci/robotframework/tests/cumulusci/base.robot#L25

More context in this picture.

image

davisagli:
@prescod You could move the assert into Deploy's `_get_package_zip`

davisagli (approved):

PR #2131, opened 2020-10-28
suggested changes for doc_task

davisagli:
@jofsky Your code is working well now, but I have a number of minor suggestions and thought it would be easiest to show in a pull request. I'll add comments to explain the changes on specific lines.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1377,7 +1377,7 @@ def task_doc(runtime, project=False, write=False):
             continue
         task_config = TaskConfig(task_config_dict)
         doc = doc_task(name, task_config)
-        result += [f"{doc}", ""]
+        result += [doc, ""]
     result = "\r\n".join(result)
 
     if write:
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
 
 
 @task.command(name="doc", help="Exports RST format documentation for all tasks")
-@click.option("--project", "project", is_flag=True, help="Write project specific tasks")
+@click.option(
+    "--project", "project", is_flag=True, help="Include project-specific tasks only"
+)
 @click.option(
     "--write",
     "write",
     is_flag=True,
-    help="If project flag is toggled will write to ./docs/project_tasks.rst otherwise will write by default to ./docs/cumulusci_tasks.rst",
+    help="If true, write output to a file (./docs/project_tasks.rst or ./docs/cumulusci_tasks.rst)",
 )
 @pass_runtime(
     require_project=False,
 )
 def task_doc(runtime, project=False, write=False):
-    config = runtime.project_config.config if project else runtime.universal_config
-    tasks = config["tasks"].items() if project else config.tasks.items()
-    file_name = "./docs/project_tasks.rst" if project else "./docs/cumulusci_tasks.rst"
-    # setting to read for project specific configuration
+    if project and runtime.project_config is None:
+        raise click.UsageError(
+            "The --project option can only be used inside a project."
+        )
     if project:
-        project_name = runtime.project_config.project__name.capitalize()
-    # handling for general and project specific documentation
-    result = ["=========================================="]
-    result.append(f"{project_name} Tasks Reference") if project else result.append(
-        "Tasks Reference"
-    )
-    result += [
-        "==========================================",
-        "",
-    ]  # extra space for formatting
-    for name, options in tasks:
-        task_config = TaskConfig(options)
+        full_tasks = runtime.project_config.tasks
+        selected_tasks = runtime.project_config.config_project.get("tasks", {})
+        file_name = "project_tasks.rst"
+        project_name = runtime.project_config.project__name
+        title = f"{project_name} Tasks Reference"
+    else:
+        full_tasks = selected_tasks = runtime.universal_config.tasks
+        file_name = "cumulusci_tasks.rst"
+        title = "Tasks Reference"
+
+    result = ["=" * len(title), title, "=" * len(title), ""]
+    for name, task_config_dict in full_tasks.items():
+        if name not in selected_tasks:
+            continue
+        task_config = TaskConfig(task_config_dict)
         doc = doc_task(name, task_config)
-        if project:
-            if name in runtime.project_config.config_project["tasks"]:
-                result += [f"{doc}", ""]
-        else:
-            result += [f"{doc}", ""]
+        result += [f"{doc}", ""]
     result = "\r\n".join(result)
+
     if write:
         Path("docs").mkdir(exist_ok=True)
-        file_name = "project_tasks" if project else "cumulusci_tasks"
-        with open(f"./docs/{file_name}.rst", "w", encoding="utf-8") as f:
-            f.write(f"{result}")
-        return  # return to not echo output
-
-    click.echo(f"{result}")
+        (Path("docs") / file_name).write_text(result, encoding="utf-8")
+    else:
+        click.echo(result)
 
 
 @flow.command(name="doc", help="Exports RST format documentation for all flows")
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
 )
 def task_doc(runtime, project=False, write=False):
-    config = runtime.project_config.config if project else runtime.universal_config
-    tasks = config["tasks"].items() if project else config.tasks.items()
-    file_name = "./docs/project_tasks.rst" if project else "./docs/cumulusci_tasks.rst"
-    # setting to read for project specific configuration
+    if project and runtime.project_config is None:
+        raise click.UsageError(
+            "The --project option can only be used inside a project."
+        )
```
davisagli, line 1362: I added this check in case someone tries to run the command with --project from outside of a project directory.

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
     if project:
```
davisagli, line 1363: You had "if project" in a bunch of different places throughout the function; I centralized the differences into this one longer conditional. I think it makes it easier to see at a glance what's being done differently when the project flag is set.

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
-        "Tasks Reference"
-    )
-    result += [
-        "==========================================",
-        "",
-    ]  # extra space for formatting
-    for name, options in tasks:
-        task_config = TaskConfig(options)
+        full_tasks = runtime.project_config.tasks
+        selected_tasks = runtime.project_config.config_project.get("tasks", {})
```
davisagli, line 1365: I changed `config_project["tasks"]` to `config_project.get("tasks", {})` because it's possible the "tasks" key isn't included in the (unmerged) project config at all.

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
+        file_name = "project_tasks.rst"
+        project_name = runtime.project_config.project__name
+        title = f"{project_name} Tasks Reference"
+    else:
+        full_tasks = selected_tasks = runtime.universal_config.tasks
+        file_name = "cumulusci_tasks.rst"
+        title = "Tasks Reference"
+
+    result = ["=" * len(title), title, "=" * len(title), ""]
```
davisagli, line 1374: Since the title is now calculated above in the project/non-project conditional, it's now easy to output the correct number of = signs matching the length of the title (which is what's preferred for reStructuredText)

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
+    for name, task_config_dict in full_tasks.items():
```
davisagli, line 1375: I moved `.items()` here instead of where the tasks dict is defined above, because that's the usual idiom in Python. You can only iterate over the result of .items() once, so it makes sense to call it on the line that's doing the iteration.

davisagli: I also renamed `options` to `task_config_dict`, because a task config includes more than just options (such as class_path and description)

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
+        if name not in selected_tasks:
+            continue
```
davisagli, line 1377: Now exiting early if it's not one of the tasks we want to output. This avoids doing extra work for those tasks, and avoids repeating the line adding to the result.

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
+        task_config = TaskConfig(task_config_dict)
         doc = doc_task(name, task_config)
-        if project:
-            if name in runtime.project_config.config_project["tasks"]:
-                result += [f"{doc}", ""]
-        else:
-            result += [f"{doc}", ""]
+        result += [doc, ""]
```
davisagli, line 1380: There's no reason to use an f-string when all it does is include a variable that is already a string.

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
+
     if write:
         Path("docs").mkdir(exist_ok=True)
-        file_name = "project_tasks" if project else "cumulusci_tasks"
-        with open(f"./docs/{file_name}.rst", "w", encoding="utf-8") as f:
-            f.write(f"{result}")
-        return  # return to not echo output
-
-    click.echo(f"{result}")
+        (Path("docs") / file_name).write_text(result, encoding="utf-8")
```
davisagli, line 1385: Changed this to use the `write_text` method of pathlib.Path which opens and writes in a single operation.

cumulusci/cli/cci.py
```diff
@@ -1343,49 +1343,48 @@ def task_list(runtime, plain, print_json):
...
+    else:
+        click.echo(result)
```
davisagli, line 1387: I prefer putting this inside an `else` clause, instead of returning from the `if write` branch, to make it more obvious that we're only echoing the result some of the time.

cumulusci/cli/tests/test_cci.py
```diff
@@ -1801,31 +1801,42 @@ def test_task_doc(self, doc_task):
         run_click_command(cci.task_doc, runtime=runtime, project=False)
         doc_task.assert_called()
 
+    def test_task_doc__project__outside_project(self):
+        runtime = mock.Mock()
+        runtime.project_config = None
+        with pytest.raises(click.UsageError):
+            run_click_command(cci.task_doc, runtime=runtime, project=True)
```
davisagli, line 1808: Added this test to cover the new check for using --project outside of a project directory

cumulusci/cli/tests/test_cci.py
```diff
@@ -1801,31 +1801,42 @@ def test_task_doc(self, doc_task):
...
-            "project": {"name": "Test"},
-            "tasks": {"option": {"a": "b"}},
-        }
-        runtime.project_config.config_project = {"tasks": {"option": {"a": "b"}}}
+        runtime.universal_config = {"tasks": {}}
+        runtime.project_config = BaseProjectConfig(
+            runtime.universal_config,
+            {
+                "project": {"name": "Test"},
+                "tasks": {"task1": {"a": "b"}, "task2": {}},
```
davisagli, line 1819: Added a second task to make sure line 1377 above gets covered.

Also switched to using a real BaseProjectConfig object instead of a mock, because my code is getting `.tasks` on it, not just `.config`)

cumulusci/cli/tests/test_cci.py
```diff
@@ -1801,31 +1801,42 @@ def test_task_doc(self, doc_task):
...
         run_click_command(cci.task_doc, runtime=runtime, project=True)
         doc_task.assert_called()
         echo.assert_called()
 
-    @mock.patch("cumulusci.cli.cci.open")
+    @mock.patch("cumulusci.cli.cci.Path")
     @mock.patch("click.echo")
     @mock.patch("cumulusci.cli.cci.doc_task")
-    def test_task_doc_project_write(self, doc_task, echo, open):
+    def test_task_doc_project_write(self, doc_task, echo, Path):
```
davisagli, line 1830: Updated the mock since writing the file is now done using the pathlib object.

cumulusci/cli/cci.py
davisagli: I also had to update the doc_task mock in the tests to make sure it returns a string rather than a mock object.

cumulusci/cli/cci.py
jofsky: good idea!

cumulusci/cli/cci.py
jofsky: I think this makes more visual sense, I just did this to save lines.

cumulusci/cli/cci.py
jofsky: I like this a lot

PR #2132, opened 2020-10-28
handle merge conflict when create merge conflict PR

Br4nd0R:
# Changes
* Handle exception when creating a merge conflict pull request.

### DIFF
--- a/cumulusci/tasks/github/merge.py
+++ b/cumulusci/tasks/github/merge.py
@@ -460,9 +460,9 @@ class MergeBranch(BaseGithubTask):
                 self.logger.info(
                     f"Merge conflict on branch {branch_name}: created pull request #{pull.number}"
                 )
-            except github3.exceptions.Conflict as e:
+            except github3.exceptions.UnprocessableEntity as e:
                 self.logger.error(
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/github/merge.py
```diff
@@ -450,17 +449,21 @@ def _merge(self, branch_name, source, commit):
...
+                    title=f"Merge {source} into {branch_name}",
+                    base=branch_name,
+                    head=source,
+                    body="This pull request was automatically generated because "
+                    "an automated merge hit a merge conflict",
+                )
+                self.logger.info(
+                    f"Merge conflict on branch {branch_name}: created pull request #{pull.number}"
+                )
+            except github3.exceptions.Conflict as e:
```
davisagli, line 463: From the build we were looking at, we know that creating the pull request can raise a github3.exceptions.UnprocessableEntity representing an HTTP 422 response (not Conflict, which is 409). So that's the error we want to catch here.

And then, it would be more helpful if we can log the body of the response (e.response.text) rather than just the name of the error, which we already know.

davisagli (changes requested):

davisagli (approved):

PR #2139, opened 2020-11-03
Minor 2GP build fixes

davidmreed:
# Critical Changes

# Changes

- The `create_package_version` task can accept a `static_resource_path` option.

# Issues Closed

- An error that occurred when building a second-generation package using a cross-project task has been fixed.
- The `github_package_data` task will now work for projects using API versions prior to 44.0.


Br4nd0R (dismissed):

cumulusci/tasks/github/commit_status.py
```diff
@@ -19,6 +19,12 @@ class GetPackageDataFromCommitStatus(BaseGithubTask, BaseSalesforceApiTask):
         "version_id": {"description": "Package version id"},
     }
 
+    def _init_api(self, base_url=None):
+        if float(self.project_config.project__package__api_version) < 44.0:
+            # We need at least API 44.0 to query for the Dependencies field.
+            self.api_version = "44.0"
+        return super()._init_api(base_url=base_url)
```
davisagli, line 26: This task only does the one query against the org, so I think we can simply set `api_version` to a specific version that we know works, and not worry about overriding the project api_version.

cumulusci/utils/git.py
```diff
@@ -6,14 +6,17 @@ def git_path(repo_root, tail=None):
     with tail appended (if present) or None if repo_root is not set.
     """
     path = None
-    if repo_root:
+    if repo_root and (pathlib.Path(repo_root) / ".git").exists():
+        # When we're a source tarball, we'll have a repo_root but no .git/HEAD.
```
davisagli, line 10: What was the scenario where we were ending up with a source tarball? CumulusCI currently generally requires a project to be a git repository, and I think changing that would be desirable but involve more than just this.

davidmreed: @davisagli I was observing this behavior when running `create_package_version` as a cross-project task - it failed in the `source`'s working dir while trying to get the current branch. (This seemed like an expedient way to fix it but happy for any better recommendations!)

davisagli: For cross-project tasks, the github source attempts to construct a project config which has repository details preconfigured, so that looking for .git isn't needed: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/core/source/github.py#L117

It's not setting the branch though. Can you give me steps to reproduce this, or a traceback? I'd like to see what is looking for the branch.

Also, what's the use case for running create_package_version as a cross-project task?

davidmreed: The use case and the repro - Elevate. https://github.com/SalesforceFoundation/spc-integration/pull/291/files

This PR builds the Transport Layer from a feature branch as a 2GP, and also builds and installs the NPSP Connector as a 2GP (cross-project) to facilitate the team doing end-to-end testing of feature changes, which is currently very difficult. NPSP Connector has a dependency on TL.

davisagli: @davidmreed I think I'm going to fix this by making create_package_version not set the branch when creating the Package2CreateVersionRequest, because I don't think it actually does anything for us. I think changing `git_path` to return None instead of raising when there's no .git folder will require a more careful analysis to make sure we aren't introducing NPE-style errors elsewhere.

davisagli:
@davidmreed any concerns about my adjustments in https://github.com/SFDO-Tooling/CumulusCI/pull/2139/commits/7cf2411f20bdd9f6e87b2cb714324722d36dcd59?

davidmreed:
@davisagli  Totally fine by me, thanks for following up

davisagli (approved):

PR #2146, opened 2020-11-05
[CCI] Modified FindReplace Method to allow environment variables to b…

jofsky:
…e passed in

# Critical Changes

# Changes

-  The FindReplace task now has a `replace_env` option which, if true, will interpret the `replace` option as the name of an environment variable whose value should be used for the replacement."
# Issues Closed


davisagli:
@jofsky what's the use case?

jofsky:
@davisagli Jenkins.

cumulusci/tasks/util.py
```diff
@@ -129,6 +130,10 @@ class FindReplace(BaseTask):
             "description": "The string to replace matches with. Defaults to an empty string",
             "required": True,
         },
+        "env_replace": {
+            "description": "The boolean value to treat the replace option as systems environment variable for replace. Ensure it's set in your system for best results. Defaults to False",
```
davidmreed, line 134:
```suggestion
            "description": "If True, treat the value of the replace option as the name of an environment variable, and use the value of that variable as the replacement string. Defaults to False",
```

cumulusci/tasks/util.py
```diff
@@ -140,11 +145,22 @@ class FindReplace(BaseTask):
...
         )
+        self.options["env_replace"] = process_bool_arg(
+            self.options.get("env_replace") or False
+        )
+        if self.options["env_replace"]:
+            if self.options["replace"] in os.environ.keys():
+                self.options["replace"] = os.environ[self.options["replace"]]
+            else:
+                raise TaskOptionsError(
+                    "Please declare the replace variable in  your local environment, not found at runtime. To turn this off, set env_replace option to False."
```
davidmreed, line 162:
```suggestion
                    f"The environment variable {self.options['replace']} was not found. Ensure that this value is populated or set env_replace to False."
```

cumulusci/tasks/tests/test_util.py
```diff
@@ -9,9 +10,18 @@
 from cumulusci.core.config import TaskConfig
 from cumulusci.utils import temporary_dir
 from cumulusci.tests.util import DummyLogger
+from cumulusci.core.exceptions import TaskOptionsError
 
 
 class TestUtilTasks(unittest.TestCase):
+    # @pytest.fixture
```
davidmreed, line 17: Remove this if we're not going to use it.

davidmreed (changes requested):

cumulusci/tasks/tests/test_util.py
```diff
@@ -111,6 +121,45 @@ def test_FindReplace(self, find_replace):
...
+                    }
+                }
+            )
+            task = util.FindReplace(self.project_config, task_config, self.org_config)
+            task()
+            find_replace.assert_called_once()
+
+    @mock.patch("cumulusci.tasks.util.find_replace")
+    def test_FindReplace_env_replace(self, find_replace):
+        os.environ["bar"] = "bars"
```
davisagli, line 145: Please use mock.patch.dict to make sure we only add this to the environment temporarily. We don't want to let it leak into other tests.

I see you're cleaning it up on line 161, but that won't run if the test fails before then. Using mock.patch.dict is the best way to make sure that happens. If you search for mock.patch.dict I think you'll find some other examples where we do this.

cumulusci/tasks/util.py
```diff
@@ -140,11 +145,22 @@ class FindReplace(BaseTask):
...
             self.options["replace"] = ""
         self.options["file_pattern"] = process_list_arg(
             self.options.get("file_pattern") or "*"
         )
+        self.options["env_replace"] = process_bool_arg(
+            self.options.get("env_replace") or False
+        )
+        if self.options["env_replace"]:
+            if self.options["replace"] in os.environ.keys():
+                self.options["replace"] = os.environ[self.options["replace"]]
```
davisagli, line 159: Actually getting the value from the environment should be done in _run_task, not _init_options. At least with MetaDeploy, it's possible for the task to run on a completely different system than where it was initialized, and we want to use the runtime environment variables.

davisagli (changes requested):

davisagli (approved):
Looks good to me. Before merging, can you make sure the "Changes" section of the pull request description includes a bullet point that is suitable for the release notes?

jofsky:
@davisagli Done, feel free to make comments on my note.

davisagli:
@jofsky I think I would put it like this: "The FindReplace task now has a `replace_env` option which, if true, will interpret the `replace` option as the name of an environment variable whose value should be used for the replacement." 

Whether it is secure depends on whether you use it appropriately, so let's not make that claim.

jofsky:
@davisagli Updated.

davidmreed (approved):

PR #2149, opened 2020-11-09
flatten structure of UpdateDependencies._install_dependency

### LINKED ISSUE #2148: Suspicious line: "zip_url" or "repo_name" in dependency
https://github.com/SFDO-Tooling/CumulusCI/blob/1d9a8623244fa79e491ebd0e74a0b0a776e83840/cumulusci/tasks/salesforce/update_dependencies.py#L248

I think this branch will always evaluate True. Should probably be:

         if "zip_url" in dependency or "repo_name" in dependency:

I don't know a lot about this code path though, so I don't want to fiddle with unit tests to make the change pass.

davisagli:
(to remove a tautological condition, fixes #2148)

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/update_dependencies.py
+++ b/cumulusci/tasks/salesforce/update_dependencies.py
@@ -245,51 +245,52 @@ class UpdateDependencies(BaseSalesforceMetadataApiTask):
 
     def _install_dependency(self, dependency):
         package_zip = None
-        if "zip_url" or "repo_name" in dependency:
-            zip_src = None
-            if "zip_url" in dependency:
-                self.logger.info(
-                    "Deploying unmanaged metadata from /{} of {}".format(
-                        dependency["subfolder"], dependency["zip_url"]
-                    )
+
+        zip_src = None
+        if "zip_url" in dependency:
+            self.logger.info(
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -245,51 +245,52 @@ def _install_dependencies(self):
...
-                    )
-                )
-                gh_for_repo = self.project_config.get_github_api(
-                    dependency["repo_owner"], dependency["repo_name"]
+
+        zip_src = None
+        if "zip_url" in dependency:
+            self.logger.info(
+                "Deploying unmanaged metadata from /{} of {}".format(
+                    dependency["subfolder"], dependency["zip_url"]
```
prescod, line 253: Line 257 hints that dependency["subfolder"] can be missing sometimes but this log message assumes that it is here.

davisagli: https://github.com/SFDO-Tooling/CumulusCI/pull/2149/commits/e79179e0976d706fd8544a1735b2739ea64e9314

prescod (approved):

PR #2151, opened 2020-11-10
Automatically detect context for namespace injection

davisagli:
# Critical Changes

# Changes
- Tasks which perform namespace injection can now automatically determine whether they are running in the context of a managed installation or a namespaced scratch org. This means that in many cases it is no longer necessary to explicitly specify options like `managed`/`unmanaged`/`namespaced`/`namespaced_org`/`namespace_inject`, or to use a separate flow for namespaced scratch orgs.
- The `deploy_unmanaged` flow now deploys sfdx-formatted metadata using the Metadata API rather than the `sfdx force:source:push` command. This avoids an issue where sfdx could show an error about the pushed components conflicting with other changes that already happened in the org. It also improves consistency between how metadata is deployed to a scratch org and how it is deployed to a packaging org.
- Removed the `namespaced_org` option for the `update_dependencies` task, which was not functional.

# Issues Closed
- Fixed an bug where namespace injection of the `%%%NAMESPACED_ORG%%%` token with the `namespaced_org` option enabled did not actually add the namespace prefix unless the `managed` option was also enabled.


Br4nd0R (approved):
LGTM!

cumulusci/tasks/apex/testrunner.py
```diff
@@ -493,6 +491,16 @@ def _enqueue_test_run(self, class_ids):
             method="POST", url=self.tooling.base_url + "runTestsAsynchronous", json=body
         ).json()
 
+    def _init_task(self):
```
davidmreed, line 494: Clarification: this positioning is so that the value does not get frozen for MetaDeploy, correct?

davidmreed (approved):
This is awesome. Code-review approval -  I'll do a few hands-on tests as well but everything makes sense.

PR #2160, opened 2020-11-12
Update standard flow library to use beta_dependencies everywhere prior to ci_master

davidmreed:
See internal Quip doc "RFC: Beta Dependencies Everywhere".

# Critical Changes

- The flows `dev_org`, `dev_org_namespaced`, `qa_org`, `ci_feature`, and `install_beta` now use beta dependencies by default.
- The flows `ci_feature_beta_deps` and `dev_org_beta_deps` are now deprecated and should be replaced by their default equivalents above.

# Changes

# Issues Closed


davisagli (commented):
Instead of using the `beta_dependencies` flow, I wonder if we should use the `dependencies` flow with an option override:

    flow: dependencies
    options:
        update_dependencies:
            include_beta: True

That way projects which customize the dependencies flow don't have to do so twice.

davidmreed:
> Instead of using the `beta_dependencies` flow, I wonder if we should use the `dependencies` flow with an option override:

@davisagli  I like this idea. Since the options override is so straightforward in project YAML, that makes me think we could go ahead and deprecate the `beta_dependencies` flow as well.

davisagli (approved):

PR #2168, opened 2020-11-16
Better JSON checks

prescod:
# Changes

Clean up some instances where we assume that the platform will 100% of the time return JSON and never HTML error text or a proxy error or something like that.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -34,11 +34,14 @@ from cumulusci.core.config import ServiceConfig
 from cumulusci.core.config import TaskConfig
 from cumulusci.core.config import UniversalConfig
 from cumulusci.core.github import create_gist, get_github_api
-from cumulusci.core.exceptions import OrgNotFound
-from cumulusci.core.exceptions import CumulusCIException
-from cumulusci.core.exceptions import CumulusCIUsageError
-from cumulusci.core.exceptions import ServiceNotConfigured
-from cumulusci.core.exceptions import FlowNotFoundError
+from cumulusci.core.exceptions import (
+    OrgNotFound,
+    CumulusCIException,
+    CumulusCIUsageError,
+    ServiceNotConfigured,
+    FlowNotFoundError,
+)
+from cumulusci.utils.http.requests_utils import safe_json_from_response
 
 
 from cumulusci.core.utils import import_global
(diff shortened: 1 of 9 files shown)


cumulusci/core/config/OrgConfig.py
```diff
@@ -366,3 +374,10 @@ def resolve_04t_dependencies(self, dependencies):
 
             new_dependencies.append(dependency)
         return new_dependencies
+
+
+def _safe_json_from_response(response):
+    try:
+        return response.json()
+    except JSONDecodeError:
+        raise CumulusCIException(f"Cannot decode as JSON:  {response.text}")
```
davisagli, line 383: Do you think this is something we would want to use elsewhere? It seems like something we could put in cumulusci.utils...I'm fine waiting to do that until it's actually used in more than one place, as long as we don't forget it's here.

cumulusci/core/tests/test_config.py
```diff
@@ -1162,6 +1162,39 @@ def test_refresh_oauth_token(self, SalesforceOAuth2):
...
+    def test_load_user_info__bad_json(self):
+        config = OrgConfig(
+            {
+                "refresh_token": mock.sentinel.refresh_token,
+                "instance_url": "http://instance_url_111.com",
+            },
+            "test",
+        )
+        keychain = mock.Mock()
+        # resp.json.side_effect = json.JSONDecodeError("blah", "Blah", 0)
```
davisagli, line 1189:
```suggestion
```

davisagli (commented):
Thanks.

cumulusci/cli/cci.py
```diff
@@ -34,11 +34,14 @@
...
-from cumulusci.core.exceptions import CumulusCIUsageError
-from cumulusci.core.exceptions import ServiceNotConfigured
-from cumulusci.core.exceptions import FlowNotFoundError
+from cumulusci.core.exceptions import (
+    OrgNotFound,
+    CumulusCIException,
+    CumulusCIUsageError,
+    ServiceNotConfigured,
+    FlowNotFoundError,
+)
```
davisagli, line 43: Please don't rewrite existing imports to group them like this. I'm fine with adding new imports in either style, but we don't have a standard preference for one over the other so I object to including it in the diff.

(Personally, I prefer the style you removed because it is easier to copy and paste between modules and helps produce diffs that only include the imports that changed.)

davisagli (approved):

cumulusci/cli/cci.py
prescod: Okay, I thought that the grouped style was the preference.

PR #2170, opened 2020-11-17
Clean up DB connections carefully for Windows.

prescod:
Previously: During the load and extract tasks, on Windows 10 if there is an error triggered by a load or extract task, CCI masks the error message by throwing a second error related to cleaning up the temp data that was built for the process.

This should fix it. I'll do tests to checks whether it also fixes `generate_and_load_from_yaml`. Maybe, maybe not

For release notes:

During the load and extract tasks, on Windows 10 if there is an error triggered by a load or extract task, CCI would mask the real error message by throwing a second error related to cleaning up the temp data that was built for the process. Now the real error will be reported correctly.

### DIFF
--- a/cumulusci/tasks/bulkdata/extract.py
+++ b/cumulusci/tasks/bulkdata/extract.py
@@ -99,8 +99,6 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
             # initialize the DB engine
             parent_engine = create_engine(database_url)
             with parent_engine.connect() as connection:
-                self.engine = connection
-
                 # initialize DB metadata
                 self.metadata = MetaData()
                 self.metadata.bind = connection
--- a/cumulusci/tasks/bulkdata/extract.py
+++ b/cumulusci/tasks/bulkdata/extract.py
@@ -97,8 +97,9 @@ class ExtractData(SqlAlchemyMixin, BaseSalesforceApiTask):
         with self._database_url() as database_url:
 
             # initialize the DB engine
-            self.engine = create_engine(database_url)
-            with self.engine.connect() as connection:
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/bulkdata/extract.py
```diff
@@ -97,8 +97,9 @@ def _init_db(self):
         with self._database_url() as database_url:
 
             # initialize the DB engine
-            self.engine = create_engine(database_url)
-            with self.engine.connect() as connection:
+            parent_engine = create_engine(database_url)
+            with parent_engine.connect() as connection:
+                self.engine = connection
```
davisagli, line 102: This is now storing a connection, not an engine. Can we update the code to use self.connection instead of self.engine, for clarity?

(If that is a tricky change to make for some reason, I think we could keep this the way it was, but be sure to explicitly close the session and dispose the engine in a finally block, like some of the tests were doing.)

cumulusci/tasks/bulkdata/load.py
```diff
@@ -411,37 +411,39 @@ def _init_db(self):
...
-                    self._create_record_type_table(
-                        mapping.get_destination_record_type_table()
-                    )
-            self.metadata.create_all()
-
-            self._validate_org_has_person_accounts_enabled_if_person_account_data_exists()
-            yield
+            parent_engine = create_engine(database_url)
+            with parent_engine.connect() as connection:
+                self.engine = connection
```
davisagli, line 416: ditto

prescod: Okay it turns out that we didn't need this self.engine except for some unit tests which I updated.

davisagli (approved):
Please add a draft release note to the PR description.

PR #2171, opened 2020-11-17
Feature/robot second user

boakley:
This modifies `Open Test Browser` such that you can specify an alternate user. This is an attempt to let tests log in as another user without having browser session issues since this method involves opening a new browser window. 

The core change to enable this is a **new method to OrgConfig** for getting the access token for a user. This method is then used in the `login url` keyword of the CumulusCI library, which in turn is used by `Open Test Browser` keyword. 

Tests will be forthcoming.

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/config/OrgConfig.py
```diff
@@ -314,6 +318,62 @@ def get_orginfo_cache_dir(self, cachename):
...
+                result = self.salesforce_client.query_all(query).get("records", [])
+                if len(result) == 0:
+                    raise Exception("Couldn't find a username for the specified user.")
+                elif len(result) > 1:
+                    raise Exception("More than one user matched the search critiera.")
+                else:
+                    username = result[0]["Username"]
+
+            p = sfdx(f"force:org:display --targetusername={username} --json")
+            info = json.loads(p.stdout_text.read())
```
davisagli, line 368: This should probably go below the returncode check. If the command was unsuccessful it's more likely that the output is not valid json

davisagli (commented):
This looks reasonable to me as long as the method moves to SfdxOrgConfig

cumulusci/robotframework/CumulusCI.py
```diff
@@ -40,6 +40,7 @@ def __init__(self, org_name=None):
         self.org_name = org_name
         self._project_config = None
         self._org = None
+        self._access_token_cache = {}
```
davisagli, line 43:
```suggestion
```

(appears to be unused)

cumulusci/robotframework/Salesforce.robot
```diff
@@ -88,9 +88,14 @@ Open Test Browser
     ...  The keyword `Log Browser Capabilities` will automatically be called.
     ...  The keyword will also call `Wait Until Salesforce is Ready` unless
     ...  the `wait` parameter is set to False.
+    ...
+    ...  The arguments 'username' or 'useralias' may be used to specify
+    ...  a specific user; if not specified then the org default will be used.
+
+    [Arguments]  ${size}=${DEFAULT BROWSER SIZE}  ${alias}=${NONE}  ${wait}=True  ${useralias}=${NONE}
```
davisagli, line 95: I'm a little confused between the `alias` and `useralias` arguments here. `useralias` is passed to `Login URL`, and `alias` is passed to `Open Browser`. Do we need both?

cumulusci/core/config/SfdxOrgConfig.py
```diff
@@ -74,6 +76,11 @@ def sfdx_info(self):
         )
         return sfdx_info
 
+    @property
+    def sfdx(self):
+        """This exists primarily so that I can mock it out for testing"""
+        return sfdx
```
davisagli, line 82: I'd rather rename the module to sfdx_org_config.py so that we can patch it without needing this. This module was only created one or two releases ago so I'm not worried about backwards-compatibility.

cumulusci/core/config/SfdxOrgConfig.py
```diff
@@ -59,6 +60,7 @@ def sfdx_info(self):
             "access_token": org_info["result"]["accessToken"],
             "org_id": org_id,
             "username": org_info["result"]["username"],
+            "access_token_cache": {},
```
davisagli, line 63: Can we put this on a separate attribute (self._access_token_cache) please? Two reasons:
1. sfdx_info currently only contains information from the `force:org:display` command for the admin user, and I'd rather keep info related to other users separate.
2. sfdx_info is added to self.config below on line 67, which means it will get stored on the filesystem if the org config is saved to the keychain. I don't think we want to do this. Access tokens expire after a period of time, but you haven't implemented anything to invalidate the cache. So it'll probably work best if we get a new access token each time cci is run.

davisagli: I'm almost inclined to say that we shouldn't cache it at all. It's not something that we're likely to need to get multiple times for the same user very often, and when we do, it's safest to call out to sfdx each time to make sure that we have a token that hasn't expired.

cumulusci/core/config/OrgConfig.py
```diff
@@ -70,6 +70,7 @@ def refresh_oauth_token(self, keychain, connected_app=None):
                 info = self._refresh_token(keychain, connected_app)
             if info != self.config:
                 self.config.update(info)
+        self._access_token_cache = {}
```
davisagli, line 73:
```suggestion
```

davisagli (changes requested):

cumulusci/robotframework/Salesforce.robot
boakley: The `alias` argument  has been there for a while, and is an alias for the browser instance. I would prefer that `alias` was `browseralias` but I didn't want to change it for backward-compatibility reasons.

cumulusci/core/config/SfdxOrgConfig.py
boakley: How does renaming the module make patching easier? I feel like I'm missing something that should be obvious.

cumulusci/core/config/SfdxOrgConfig.py
boakley: If you want it in a separate attribute, I can do that. I originally had it that way, but figured combining with the others made more sense. 

As for the need for a cache at all, I was concerned that a test might need to open a browser for a second user multiple times, and I didn't want to add what could be minutes to an already long test run. I see your point about an expired token, though we cache the token for the default user and don't seem to have a problem with it expiring.

cumulusci/core/config/SfdxOrgConfig.py
davisagli: Right now we can't do `mock.patch("cumulusci.core.config.SfdxOrgConfig.sfdx")` because cumulusci/core/config/__init__.py imports the SfdxOrgConfig class, so cumulusci.core.config.SfdxOrgConfig ends up referring to the class and there is no way to target the module. If we rename the module then we can patch `cumulusci.core.config.sfdx_org_config.sfdx` and be sure we're targeting the module rather than the class.

cumulusci/core/config/SfdxOrgConfig.py
boakley: I guess I can remove the cache and see if anyone complains about performance. We can always add it back in later.

cumulusci/robotframework/Salesforce.robot
davisagli: Okay, that makes sense. Can you add `alias` to the docstring here to clarify the difference?

cumulusci/core/config/SfdxOrgConfig.py
davisagli: Yeah, I'd prefer to wait and add the cache only if it's needed.

davisagli (approved):
I did a bit more cleanup and think this is ready.

PR #2172, opened 2020-11-17
OrgConfig returns if multi-currency is enabled

spelak-salesforce:
Both [Multiple Currencies](https://help.salesforce.com/articleView?id=admin_enable_multicurrency.htm) and [Advanced Currency Management (ACM)](https://help.salesforce.com/articleView?id=administration_enable_advanced_currency_management.htm) can programmatically be enabled via Metadata API through the [CurrencySettings](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_currencysettings.htm).

```xml
<?xml version="1.0" encoding="UTF-8"?>
<CurrencySettings xmlns="http://soap.sforce.com/2006/04/metadata">
    <!-- Enables Multiple Currencies -->
    <enableMultiCurrency>true</enableMultiCurrency>

    <!-- Enables Advanced Currency Management (ACM) -->
    <enableCurrencyEffectiveDates>true</enableCurrencyEffectiveDates>
</CurrencySettings>
```

Notes
- Multiple Currencies cannot be disabled once enabled.
- ACM requires Multiple Currencies to be enabled.

Via REST API, we can detect if:
- Multiple Currencies is enabled by checking if the `CurrencyType` describe returns a 200.
    - Enabling Multiple Currencies exposes both the `CurrencyType` and `DatedConversionRate` Sobjects.
- ACM is enabled by checking if the `DatedConversionRate` describe both returns a 200 + is `createable`.

Implementation:
- Since Multiple Currencies and ACM can be programmatically enabled, we cannot generally cache the result.
- Since Multiple Currencies cannot be disabled once enabled, we can cache the result when we detect Multiple Currencies is enabled to save CumulusCI from doing future describe calls.
- ACM can be programmatically disabled, so we need to always check the describe call to verify if ACM is enabled or not.

# Critical Changes

# Changes
Dynamically choose to run tasks in a flow whether or not [Multiple Currencies](https://help.salesforce.com/articleView?id=admin_enable_multicurrency.htm) and/or [Advanced Currency Management (ACM)](https://help.salesforce.com/articleView?id=administration_enable_advanced_currency_management.htm) is enabled or not in the org.

Example:

```yaml
flows:
    load_storytelling_data:
        steps:
            1:
                task: load_dataset
                options:
                    mapping: datasets/without_multiple_currencies/mapping.yml
                    sql_path: datasets/without_multiple_currencies/data.sql
                when: not org_config.is_multiple_currencies_enabled
             2:
                task: load_dataset
                options:
                    mapping: datasets/with_multiple_currencies/mapping.yml
                    sql_path: datasets/with_multiple_currencies/data.sql
                when: org_config.is_multiple_currencies_enabled and not org_config.is_advanced_currency_management_enabled
             3:
                task: load_dataset
                options:
                    mapping: datasets/with_acm/mapping.yml
                    sql_path: datasets/with_acm/data.sql
                when: org_config.is_advanced_currency_management_enabled

```

# Issues Closed


spelak-salesforce:
Note: I ran `make docs`  just in case it was needed to capture the doc string for `OrgConfig.is_multi_currency_enabled`.   I saw there were a lot of changes to `flows.rst` and `tasks.rst` that were unrelated to my changes, so I did not commit those changes.

davisagli:
@spelak-salesforce Is it possible for a task to turn on multi-currency? If so, it's theoretically possible for the cached value to be out of date.

I am thinking that we should give CumulusCI the ability to run the same preflight checks that are used for MetaDeploy install plans (which can be used to skip a step, among other things). Then they would run before the main flow starts, rather than prior to a particular step. It would also mean we could do things like `cci flow info --org [org]` where we run just the preflight checks on the org and then show the adjusted flow steps that are applicable to that org.

spelak-salesforce:
> @spelak-salesforce Is it possible for a task to turn on multi-currency? If so, it's theoretically possible for the cached value to be out of date.

@davisagli , I don't know but will look into if it's possible to programmatically enable multi-currency.  This approach at least checks the state of the org when the first `org_config.is_multi_currency_enabled` is called.

spelak-salesforce:
> > @spelak-salesforce Is it possible for a task to turn on multi-currency? If so, it's theoretically possible for the cached value to be out of date.
> 
> @davisagli , I don't know but will look into if it's possible to programmatically enable multi-currency. This approach at least checks the state of the org when the first `org_config.is_multi_currency_enabled` is called.

@davisagli , I can't find definitive proof that mutli-currency cannot be enabled via API, but [MultiCurrency](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_scratch_orgs_def_file_config_values.htm#so_multicurrency) is a scratch org definition **feature** which suggests enabling this feature cannot be done via API.

davisagli:
@spelak-salesforce ok, it's probably not worth worrying about then. I'll try to give this a closer review soon.

davidmreed:
A task can't turn on multi-currency, as far as I know, because it is a scratch org feature (not a Settings deployment).

@spelak-salesforce  Can you form your `when` clauses by looking at the org shape instead? That should be no code required.

spelak-salesforce:
> A task can't turn on multi-currency, as far as I know, because it is a scratch org feature (not a Settings deployment).
> 
> @spelak-salesforce Can you form your `when` clauses by looking at the org shape instead? That should be no code required.

@davidmreed , what do you mean by org shape?   Like checking the name of the scratch org def file?

The intent of this PR is to support running tasks only if an optional Salesforce feature is enabled while being easy to interpret its usage in `cumulusci.yml`.  In this case, support checking if multi-currency is enabled.   

This method is a similar to `OrgConfig.is_person_accounts_enabled`, and it's been super useful to have when clauses like `not org_config.is_person_accounts_enabled`.

davisagli:
Checking the scratch org definition file doesn't sound like a great option because then it wouldn't work for persistent orgs.

cumulusci/core/tests/test_config.py
```diff
@@ -1702,6 +1702,100 @@ def test_is_person_accounts_enabled__is_enabled(self):
...
+                "instance_url": "https://example.com",
+                "access_token": "TOKEN",
+                "id": "OODxxxxxxxxxxxx/user",
+            },
+            "test",
+        )
+        self.assertIsNone(
+            config._is_multi_currency_enabled,
+            "_is_multi_currency_enabled should be initialized as None",
+        )
```
davisagli, line 1718: Can we use pytest-style assertions please? The codebase is inconsistent, but we are trying to be consistent about using them for new tests.

    assert config._is_multi_currency_enabled is None

davisagli (changes requested):

davidmreed:
I don't think this is a concern for the purpose here, but wanted to call out that Advanced Currency Management is a separate layer from the basic multicurrency enablement and doesn't respond to the same checks (as I learned recently in collaboration with Michael Smith).

spelak-salesforce:
> ncern for the purpose here, but wanted to call out that Advanced Currency Management is a separate layer from the basic multicurrency enablement and doesn't respond to the same checks (as I learned recently in collaboration with Michael Smith).

I'll dig into this scenario and make sure this solution supports both Currency Management + Advanced Currency Management.

spelak-salesforce:
@davisagli  @davidmreed 

I dug into both [Multiple Currencies](https://help.salesforce.com/articleView?id=admin_enable_multicurrency.htm) and [Advanced Currency Management (ACM)](https://help.salesforce.com/articleView?id=administration_enable_advanced_currency_management.htm) and found both can programmatically be enabled via Metadata API through the [CurrencySettings](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_currencysettings.htm).

```xml
<?xml version="1.0" encoding="UTF-8"?>
<CurrencySettings xmlns="http://soap.sforce.com/2006/04/metadata">
    <!-- Enables Multiple Currencies -->
    <enableMultiCurrency>true</enableMultiCurrency>

    <!-- Enables Advanced Currency Management (ACM) -->
    <enableCurrencyEffectiveDates>true</enableCurrencyEffectiveDates>
</CurrencySettings>
```

Notes
- Multiple Currencies cannot be disabled once enabled.
- ACM requires Multiple Currencies to be enabled.

Via REST API, we can detect if:
- Multiple Currencies is enabled by checking if the `CurrencyType` describe returns a 200.
    - Enabling Multiple Currencies exposes both the `CurrencyType` and `DatedConversionRate` Sobjects.
- ACM is enabled by checking if the `DatedConversionRate` describe both returns a 200 + is `createable`.

Implementation:
- Since Multiple Currencies and ACM can be programmatically enabled, we cannot generally cache the result.
- Since Multiple Currencies cannot be disabled once enabled, we can cache the result when we detect Multiple Currencies is enabled to save CumulusCI from doing future describe calls.
- ACM can be programmatically disabled, so we need to always check the describe call to verify if ACM is enabled or not.

spelak-salesforce:
@davisagli , this is ready for re-review.

cumulusci/core/tests/test_config.py
```diff
@@ -1702,6 +1702,247 @@ def test_is_person_accounts_enabled__is_enabled(self):
...
+        # Since Multiple Currencies is enabled, so the describe call returns a 200.
+        responses.add(
+            "GET",
+            "https://example.com/services/data/v48.0/sobjects/CurrencyType/describe",
+            json={
+                # The actual payload doesn't matter; only matters is we get a 200.
+            },
+        )
+
+        # Add a second describe call to demonstrate we always check the describe until we detect Multiple Currencies is enabled.  From then on, we cache the fact that Multiple Currencies is enabled knowing Multiple Currencies cannot be disabled.
```
davidmreed, line 1799: This is inert since the call isn't being made.

cumulusci/core/tests/test_config.py
```diff
@@ -1702,6 +1702,247 @@ def test_is_person_accounts_enabled__is_enabled(self):
...
+        )
+
+        # is_advanced_currency_management_enabled should be False since:
+        # - DatedConversionRate describe gives a 404 implying the Sobject is not exposed becuase Multiple Currencies is not enabled.
+        actual = config.is_advanced_currency_management_enabled
+        assert (
+            actual is False
+        ), "config.is_advanced_currency_management_enabled should be False since the describe gives a 404."
+
+        # We should have made 3 calls: 1 token call + 1 describe call
```
davidmreed, line 1867:
```suggestion
        # We should have made 2 calls: 1 token call + 1 describe call
```

davisagli (commented):
Looks good to me once @davidmreed's comments are addressed.

cumulusci/core/tests/test_config.py
spelak-salesforce: @davidmreed , fixed.

davidmreed (approved):

PR #2174, opened 2020-11-18
Include a link for each task

Br4nd0R:
# Changes
* Include a link for each task so other sections can reference them.

### DIFF
--- a/cumulusci/utils/__init__.py
+++ b/cumulusci/utils/__init__.py
@@ -417,6 +417,7 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
     from cumulusci.core.utils import import_global
 
     doc = []
+    doc.append(f"\n\n.. _{task_name}:\n")
     doc.append(f"**{task_name}**\n==========================================\n")
     doc.append(f"**Description:** {task_config.description}\n")
     doc.append(f"**Class:** {task_config.class_path}\n")
--- a/cumulusci/utils/__init__.py
+++ b/cumulusci/utils/__init__.py
@@ -417,8 +417,7 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
     from cumulusci.core.utils import import_global
 
     doc = []
-    doc.append(f"\n\n.. _{task_name}:\n")
-    doc.append(f"**{task_name}**\n==========================================\n")
+    doc.append(f"{task_name}\n==========================================\n")
     doc.append(f"**Description:** {task_config.description}\n")
     doc.append(f"**Class:** {task_config.class_path}\n")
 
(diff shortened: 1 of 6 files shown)


davisagli:
It looks like there's a way to automatically create an anchor from each heading: https://sublime-and-sphinx-guide.readthedocs.io/en/latest/references.html#links-to-sections-in-the-same-document ...should we try that?

Br4nd0R:
@davisagli - Adding `sphinx.ext.autosectionlabel` to the current extensions in `conf.py` and then attempting to link to the name of a task with :ref:`task_name` doesn't work. Also noticing this causes `WARNING: duplicate label command syntax` to be output for each task while building the docs. I also checked to see that the difference between:
```
task_name
=============

task_name
========
```
was the issue, but it didn't have an effect.

davisagli:
@Br4nd0R I looked into it a bit. :ref:`update_admin_profile` is not finding the label because the label is actually `**update_admin_profile**` with asterisks, based on the section header. I think we would want to remove the bold emphasis on the task names and let them use the heading style as configured by the sphinx theme's css.

The warnings are a little annoying, but not an actual problem as long as we don't need to link to one of the headings that is duplicated. I think using autosectionlabels is probably worth that annoyance since it'll save us a lot of work in cross-linking.

cumulusci/utils/__init__.py
```diff
@@ -417,8 +417,7 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
     from cumulusci.core.utils import import_global
 
     doc = []
-    doc.append(f"\n\n.. _{task_name}:\n")
-    doc.append(f"**{task_name}**\n==========================================\n")
+    doc.append(f"{task_name}\n==========================================\n")
```
davisagli, line 420:
```suggestion
    doc.append(f"{task_name}\n{"=" * len(task_name)}\n")
```

just to tidy up the length of the underline so sphinx doesn't complain about it.

davisagli: Also it looks like there's a test adjustment needed

cumulusci/utils/__init__.py
```diff
@@ -417,7 +417,7 @@ def doc_task(task_name, task_config, project_config=None, org_config=None):
     from cumulusci.core.utils import import_global
 
     doc = []
-    doc.append(f"**{task_name}**\n==========================================\n")
+    doc.append(f"{task_name}\n{"=" * len(task_name)}\n")
```
davisagli, line 420:
```suggestion
    doc.append(f"{task_name}\n{'=' * len(task_name)}\n")
```

Oops, sorry.

Issue #2027, opened 2020-09-16
Cannot abort attempts to create connections


acrosman:
**Describe the bug**
I cannot cancel an attempt to create a connection in Windows 10/PowerShell using Ctrl-c.

**Reproduction steps**
1. Run `cci org connect org_name`
2. Web browser launches, and expected messages appear in PowerShell.
3. In powershell hit ctrl-c
4. No response.  Eventually the attempt will time out.

**Context**
Please include the following information at minimum.

- Your CumulusCI version (`cci version`). 
- Your Python version (`python --version`)
```
> cci version
CumulusCI version: 3.18.0 (C:\Users\ahcrosman\.local\bin\cci)
Python version: 3.8.5 (c:\users\ahcrosman\.local\pipx\venvs\cumulusci\scripts\python.exe)
```

- Your operating system and version.
Windows 10

- If using Windows, whether you are using Command Prompt, PowerShell, Bash on Windows, or WSL.
PowerShell

- If possible, whether you installed CumulusCI using `brew`, `pip`, or `pipx`.
pipx

davisagli:
@acrosman We'll try to look into this, but don't have easy access to a Windows environment. Can you check if Ctrl+Break works as an alternative?

acrosman:
As my keyboard lacks a break key, no I cannot.

acrosman:
During the SFDO sprint this week @prescod and I observed the same behavior on another PC's machine.  She also lacked a break key, looks like those are either uncommon or so badly labelled on laptops that no can find them these days.

prescod:
@acrosman Are you able to test the PR #2185  and see if it fixes your problem?

prescod:
@acrosman : If you get a chance, please test these scenarios:

1. Does normal cci org connect work?
2. Does Ctrl-C work?
3. Does the 5 minute timeout work?

acrosman:
Upgraded package cumulusci from 3.20.1 to 3.23.0, then swapped in saleforce.py from the PR.  That led to the new file failing.  Is there another version I need to be working from?  That line isn't from this PR,

```
> cci org connect org_name
Traceback (most recent call last):
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.8_3.8.1776.0_x64__qbz5n2kfra8p0\lib\runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
[...]

PR #2185, opened 2020-11-19
Feature/windows shutdown nicely

prescod:
# Changes

When connecting to an org with `cci org connect`, Ctrl-C often could not kill the server on Windows in case the user wanted to exit without finishing. Now Ctrl-C works as expected.

# Issues Closed

https://github.com/SFDO-Tooling/CumulusCI/issues/2027

### DIFF
--- a/cumulusci/oauth/salesforce.py
+++ b/cumulusci/oauth/salesforce.py
@@ -14,8 +14,10 @@ from urllib.parse import urlparse
 import webbrowser
 import threading
 import random
+import time
 
 from cumulusci.oauth.exceptions import SalesforceOAuthError
+from cumulusci.core.exceptions import CumulusCIUsageError
 
 HTTP_HEADERS = {"Content-Type": "application/x-www-form-urlencoded"}
 SANDBOX_DOMAIN_RE = re.compile(
@@ -123,6 +125,28 @@ class SalesforceOAuth2(object):
         return response
 
 
+class HTTPDTimeout(threading.Thread):
+    daemon = True
+
+    def __init__(self, httpd, timeout):
+        self.httpd = httpd
+        self.timeout = timeout
+        super().__init__()
+
+    def run(self):
+        for i in range(self.timeout):
+            time.sleep(1)
+            if not self.httpd:
+                break
+
+        if self.httpd:  # extremely minor race condition
+            self.httpd.shutdown()
+
+    def quit(self):
+        print("Quittiing")
+        self.httpd = None
+
+
 class OAuthCallbackHandler(BaseHTTPRequestHandler):
     parent = None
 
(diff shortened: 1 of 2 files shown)


cumulusci/oauth/tests/test_salesforce.py
```diff
@@ -163,6 +165,7 @@ def test_oauth_flow_error_from_auth(self):
         # wait for thread to complete
         t.join()
 
+    @mock.patch("time.sleep", time.sleep)
```
davisagli, line 168: These patches shouldn't be needed. There is a fixture which automatically patches time.sleep for all tests: https://github.com/SFDO-Tooling/CumulusCI/blob/master/cumulusci/conftest.py#L16

cumulusci/oauth/salesforce.py
```diff
@@ -131,7 +157,16 @@ def do_GET(self):
...
+            gif = random.choice(
+                [
+                    "tMUMi4Ylnzk88",
+                    "Cj3Ce7e8h2EKY",
+                    "3MibWRcT0AOYG7HUeT",
+                    "9uoYC7cjcU6w8",
+                ]
+            )
+            http_body = f"""<html><p>Congratulations! Your authentication succeeded.</p>
+            <iframe src="https://giphy.com/embed/{gif}" width="480" height="360" frameBorder="0"></iframe></html>"""
```
davisagli, line 169: This is maybe a bit too cute. The 2nd gif is already a 404.

davisagli (commented):
I see what you're doing here, but could you add a few more comments to help remind us when we come back to it?

cumulusci/oauth/tests/test_salesforce.py
prescod: This is an UN-patch.

Your fixture confused the heck out of me and I don't think it makes sense in these cases. There is an explicit time.sleep() in the test cases in this file and turning time.sleep into a no-op doesn't make sense if the tests themselves are relying on the real time.sleep. I think my algorithm is probably resistant to the patching now (it wasn't before) but still, the patching causes more problems than it solves in this case, IMO.

cumulusci/oauth/salesforce.py
prescod: The last one is just a cut and paste error in the PR (fixed above) and I don't think Giphy deletions are common.

https://medium.com/@makeagif/http-help-makeagif-com-common-makeagif-q-s-will-my-gifs-be-deleted-2a0de1e75a57

The 99% of users who are "free users" have to ask to get files deleted through a "support request".

https://giphy.com/embed/Cj3Ce7e8h2EKY

cumulusci/oauth/tests/test_salesforce.py
davisagli: Oh I see, hmm. The point of my fixture was to avoid slowing down the tests when they test code that includes a sleep just to avoid polling too frequently, and it did have a real impact on test performance. But I hadn't noticed that there are also sleeps within test code. Would it help at all if the fixture did a real sleep with a shortened interval instead of a no-op?

cumulusci/oauth/salesforce.py
davisagli: Still not a fan of embedding something from a domain we don't control.

prescod: 4a0b926

prescod:
> I see what you're doing here, but could you add a few more comments to help remind us when we come back to it?

d69682d

cumulusci/oauth/tests/test_salesforce.py
prescod: I think it would be worth experimenting with a time.sleep that sleeps for 1/10 or 1/20 of the time but it should be a separate PR. I tweaked my algorithm so it doesn't presume that time.sleep(1) really sleeps for 1 second so it doesn't matter so much to me anymore, but the (pre-existing) tests themselves do try to use sleep to reduce their busy-waitiing.

davisagli (approved):

PR #2189, opened 2020-11-19
Support deploying lwc subfolders

davisagli:
# Critical Changes

# Changes

# Issues Closed
- Fixed a bug where deploying an LWC component bundle using the `deploy` task did not include files in subfolders.


davisagli:
@spelak-salesforce Does this make sense to you? Asking because this logic for filtering what we deploy from lwc folders originally came from you. I think the point was to avoid including __tests__ as well as files like eslint configuration or readmes, which I've attempted to preserve while adding support for actual component files in subfolders.

spelak-salesforce:
@davisagli , [__mocks__](https://jestjs.io/docs/en/manual-mocks) is a valid subdirectory that's used by Jest where Jest automatically imports the mock in tests vs the production code.   

Maybe:
- Ignore all LWC subdirectories starting with/wrapped in `__`?
- The cci task has a configurable blocklist of LWC subdirectories to ignore?

```
lwc
|-- __mocks__ # lwc-jest automatically overrides
|             # imports matching paths in __mocks__.
|             # Useful for overriding @salesforce imports.
| 
|-- myComponent
|   |-- __tests__ # the usual suspect
|   |-- __mocks__ # technically could be here.
|
|-- ...   
```

davisagli:
@spelak-salesforce Thanks, I like the idea of ignoring folders with `__`. I'll also check salesforce-alm to see what the CLI does.

spelak-salesforce:
> @spelak-salesforce Thanks, I like the idea of ignoring folders with `__`. I'll also check salesforce-alm to see what the CLI does.

I don't think salesforce-alm does anything.  Devs always have to put these specific items in `.forceignore`.

davisagli:
Oh wow, I'm surprised that has to be done manually. Does including the tests in the deployment lead to an error or is it just to avoid including unnecessary data?

Paying attention to .forceignore does sound like something that we should try to do.

davisagli:
@spelak-salesforce actually I think items in .forceignore are probably already filtered out when converting sfdx format to mdapi format prior to deploying.

spelak-salesforce:
> Oh wow, I'm surprised that has to be done manually. Does including the tests in the deployment lead to an error or is it just to avoid including unnecessary data?
> 
> Paying attention to .forceignore does sound like something that we should try to do.

Yes, we get sfdx deploy/push errors if we forget to add the items in the `.forceignore`.   Something like unexpected directories for `__tests__`/"can't find component..."; different error for `.eslintrc.json`.

davisagli:
I think where this leaves us is:
1. This filtering is unnecessary for sfdx format because it gets filtered using .forceignore while converting
2. I'll make it skip any folder starting with `__` so that we handle `__mocks__` in mdapi format automatically.
3. We can look into supporting .forceignore for mdapi format but let's keep that as a separate user story.

davidmreed (approved):

PR #2190, opened 2020-11-19
[CCI] Ensure record types are included in retrieve profile task

jofsky:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
+++ b/cumulusci/tasks/salesforce/tests/test_ProfileGrantAllAccess.py
@@ -549,3 +549,34 @@ def test_generate_package_xml__retrieve():
         task._expand_package_xml = mock.Mock()
         task._generate_package_xml(MetadataOperation.RETRIEVE)
         task._expand_package_xml.assert_not_called()
+        task = create_task(
+            ProfileGrantAllAccess,
+            {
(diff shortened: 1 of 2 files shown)


davisagli (commented):
@jofsky Please write a pull request description before submitting for review. Is the point here to make sure that the record types listed in the record_types option are included in the retrieved profile?

davisagli:
@jstvz Josh says this is from you...could you explain the context?

jstvz:
@davisagli From d7ff367:

    Tasks can call ProfileGrantAllAccess with:
    
    ```yaml
    update_admin_profile:
        options:
            record_types:
                    - record_type: "{managed}Foo__c.{namespaced_org}Baz"
                      default: true
    ```
    
    If the referenced record type is not included in the default package.xml
    the task will throw an exception unless a custom package.xml is
    provided.
    
    Instead, this commit updates the class such that any record types are
    appended to the package.xml before retrieval.
    
    FIXME: This is tagged as a `[wip]` commit because this may not be the
    right way to handle namespace injection for these.

This was probably a little premature. I'll take a second pass and resubmit.

davisagli:
@jstvz Thanks, I had not found that commit message. It seems pretty reasonable to me, though I am wondering what happens if there is an existing default record type that is not included in the retrieval.

PR #2191, opened 2020-11-20
[CCI] Fixing github_release_notes trial_info option for MetaCI compatibility

jofsky:
# Critical Changes

# Changes

- Change `github_release_notes`'s trial_info option logic to accept strings not booleans for MetaCI compatibility.  

# Issues Closed


davisagli (commented):
We won't be able to set this from MetaCI (since it's not responsible for creating the trial template), but I don't see a problem with the change.

davisagli (approved):

PR #2192, opened 2020-11-23
Fix silly bugs in custom_settings_value_wait

force2b:
# Critical Changes

# Changes

- Fix the `custom_settings_value_wait` task to not throw an exception if a custom settings record doesn't yet exist. Instead it'll continue to query the object until the record exists and the value matches
- Fix the `custom_settings_value_wait` task to not throw an exception if the field api name passed as an option doesn't exactly match the case of the field api name in the object.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/custom_settings_wait.py
+++ b/cumulusci/tasks/salesforce/custom_settings_wait.py
@@ -73,21 +73,24 @@ class CustomSettingValueWait(BaseSalesforceApiTask):
         except SalesforceError as e:
             message = e.content[0]["message"]
             if "No such column 'SetupOwnerId'" in message:
-                message = "Only Hierarchical Custom Settings objects are supported."
+                message = "Only Custom Settings objects are supported."
             raise TaskOptionsError(f"Query Error: {message}")
 
         self.record = None
         for row in query_results["records"]:
-            setupOwnerId = str(row["SetupOwnerId"])
-            if setupOwnerId.startswith("00D"):
-                self.record = row
+            if "SetupOwnerId" in row:
+                setupOwnerId = str(row["SetupOwnerId"])
+                if setupOwnerId.startswith("00D"):
+                    self.record = row
+            else:
+                raise TaskOptionsError("Only Custom Settings objects are supported.")
 
-        if not self.record:
-            raise SalesforceException(
-                "Hierarchical Custom Settings Org Default record not found"
+        if self.record:
+            self.poll_complete = not self._poll_again()
+        else:
+            self.logger.info(
+                f"{self.field_name}: Looking for {self.check_value} and found no custom settings record"
             )
-
-        self.poll_complete = not self._poll_again()
 
     def _poll_again(self):
         return not self.success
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/custom_settings_wait.py
```diff
@@ -106,7 +109,8 @@ def _apply_namespace(self):
 
     @property
     def success(self):
-        self.field_value = self.record[self.field_name]
+        lowerCaseRecord = {k.lower(): v for k, v in self.record.items()}
+        self.field_value = lowerCaseRecord[self.field_name.lower()]
```
davisagli, line 113: In Python we typically use snake case (underscores) for variables and reserve camel case for class names.

```suggestion
        lower_case_record = {k.lower(): v for k, v in self.record.items()}
        self.field_value = lower_case_record[self.field_name.lower()]
```

cumulusci/tasks/salesforce/custom_settings_wait.py
```diff
@@ -73,22 +73,25 @@ def _poll_action(self):
         except SalesforceError as e:
             message = e.content[0]["message"]
             if "No such column 'SetupOwnerId'" in message:
-                message = "Only Hierarchical Custom Settings objects are supported."
+                message = "Only Custom Settings objects are supported."
```
davisagli, line 76: @force2b So is this also aiming to support non-hierarchical custom settings? If so the description for the `object` option needs to be tweaked above too.

davisagli (changes requested):

cumulusci/tasks/salesforce/custom_settings_wait.py
force2b: Good point. I think I accidentally removed that line. Will put it back.

force2b:
Thanks for the review @davisagli. I made a couple of updates per your review.

davidmreed:
I pushed a change to simplify error handling, removing a condition that could not occur.

I'll approve this too but I think the existing task (_not_ these changes; thank you, @force2b) needs more work - it has no timeout so can spin a worker until it dies looking for a custom setting, and it does not currently have support for automatic injection. I'll add a WI for those changes.

davisagli (dismissed):

davidmreed (dismissed):

force2b:
Thanks @davidmreed. I wrote the original task, so I'll be keeping an eye out for changes to learn how to do a better job building these in the future.

cumulusci/tasks/salesforce/custom_settings_wait.py
```diff
@@ -78,17 +77,16 @@ def _poll_action(self):
...
+            if row["SetupOwnerId"].startswith("00D"):
                 self.record = row
 
-        if not self.record:
-            raise SalesforceException(
-                "Hierarchical Custom Settings Org Default record not found"
+        if self.record:
+            self.poll_complete = not self._poll_again()
+        else:
+            self.logger.info(
```
davisagli, line 86: This line doesn't have test coverage, so I suspect we could use a test for the case where the settings object exists but no record was found.

davidmreed:
@force2b We changed CumulusCI underneath the task. We just need to update it to keep up with the times! The move to automatic inference for namespace injection was required to make 2GP work.

davisagli:
@davidmreed I added automatic namespace injection and filled in test coverage; could you give this one more look?

davidmreed (approved):

PR #2193, opened 2020-11-24
Support assigning Permission Set Groups and Licenses

spelak-salesforce:
- Abstracted `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSets` to be reusable to assign both Permission Set Groups and Permission Set Licenses.
- Added `cumulisci.tasks.salesforce.users.permsets.AssignPermissionSetGroups` to assign Permission Set Groups if not already assigned.
    - Added default cci task `assign_permission_set_groups` calling `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSetLicenses`.
- Added `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSetLicenses`to assign Permission Set Licenses if not already assigned.
    - Added default cci task `assign_permission_set_licenses` calling `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSetLicenses`.
    - NOTE: The User sub-query for PermissionSetLicenseAssignments reture `None` instead of an empty list if no PermissionSetLicenseAsssignments are found.  This seems like a bug, and there is an explicit unit test for this scenario.

# Critical Changes

# Changes
- Added `cumulisci.tasks.salesforce.users.permsets.AssignPermissionSetGroups` to assign Permission Set Groups if not already assigned.
    - Added default cci task `assign_permission_set_groups` calling `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSetLicenses`.
- Added `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSetLicenses`to assign Permission Set Licenses if not already assigned.
    - Added default cci task `assign_permission_set_licenses` calling `cumulusci.tasks.salesforce.users.permsets.AssignPermissionSetLicenses`.
# Issues Closed


cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -19,18 +23,42 @@ def _init_options(self, kwargs):
...
+    def assignment_name(self):
+        return "PermissionSetAssignment"
+
+    @property
+    def assignment_lookup(self):
+        return "PermissionSetId"
+
+    @property
+    def assignment_child_relationship(self):
+        return "PermissionSetAssignments"
```
davisagli, line 48: There is no reason for these to be properties. You can just set a class attribute, e.g. 

    assignment_child_relationship = "PermissionSetAssignments"

cumulusci/tasks/salesforce/users/tests/test_permsets.py
```diff
@@ -199,3 +203,456 @@ def test_create_permset_raises(self):
...
+        task = create_task(
+            AssignPermissionSetLicenses,
+            {
+                "api_names": "PermSetLicense1,PermSetLicense2",
+            },
+        )
+
+        responses.add(
+            method="GET",
+            url=f"{task.org_config.instance_url}/services/data/v50.0/query/?q=SELECT+Id%2C%0A++++++++++++++++++++++++++++%28SELECT+PermissionSetLicenseId%0A+++++++++++++++++++++++++++++FROM+PermissionSetLicenseAssignments%29%0A++++++++++++++++++++++++FROM+User%0A++++++++++++++++++++++++WHERE+Username+%3D+%27test-cci%40example.com%27",
```
davisagli, line 220: This is just cosmetic but it might be good to write the query in a way that avoids including all these extra spaces. In Python one way to do this is:

    sf.query(
        "SELECT something "
        "from object"
    )

and those strings will be automatically concatenated into a single literal by the compiler.

cumulusci/tasks/salesforce/users/tests/test_permsets.py
```diff
@@ -199,3 +203,456 @@ def test_create_permset_raises(self):
...
+                    {
+                        "Id": "0PG000000000001",
+                        "DeveloperName": "PermSetGroup2",
+                    },
+                ],
+            },
+        )
+
+        with pytest.raises(CumulusCIException):
+            task()
```
davisagli, line 658: I appreciate the verbose tests that don't reuse code. Makes it easy to confirm that the intended behavior makes sense.

davisagli (changes requested):
Thanks -- this looks quite good. I just made a couple suggestions re idiomatic Python.

Tagging @davidmreed in case he wants to take a look.

cumulusci/tasks/salesforce/users/permsets.py
spelak-salesforce: Done.  Change is in upcoming commit.

davidmreed:
This looks good to me post-David Glick's comments. I did not know that wrinkle about the specific PSL assignment behavior in developer orgs; that's potentially useful for testing some of our 2GP challenges.

spelak-salesforce:
> This looks good to me post-David Glick's comments. I did not know that wrinkle about the specific PSL assignment behavior in developer orgs; that's potentially useful for testing some of our 2GP challenges.

@davidmreed , the odd behavior is because in non-namespaced dev orgs, the associated Permission Set to the PSL doesn't have a namespace.  Yet, the PSL is hard-coded to only associate to the Permission Set with both the name + namespace.  In a way, this is expected behavior.

cumulusci/tasks/salesforce/users/tests/test_permsets.py
spelak-salesforce: This is fixed.   The tests are much easier to digest.

cumulusci/tasks/salesforce/users/permsets.py
spelak-salesforce: Change is commited.

davisagli (approved):

PR #2194, opened 2020-11-24
2GP Standard Library

davidmreed:
# Critical Changes

# Changes

- An experimental `qa_org_2gp` flow is provided in the standard library. This flow uses `config_qa`, meaning that `config_qa` must be tokenized and not include explicit `managed` or `unmanaged` options so that CumulusCI can manage performing the appropriate injection. This flow may change at any time.
- The `ci_feature_2gp` flow has been changed to use `config_apextest` instead of `config_managed`. This means that `config_apextest` must be tokenized and not include explicit `managed` or `unmanaged` options so that CumulusCI can manage performing the appropriate injection.

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -904,7 +904,7 @@ flows:
                 options:
                     context: "Build Feature Test Package"
             2:
-                flow: beta_dependencies
+                flow: dependencies
                 options:
                     update_dependencies:
                         dependencies: ^^github_package_data.dependencies
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -615,21 +615,10 @@ flows:
         description: Install as a managed 2gp package and run Apex tests. Intended for use after build_feature_test_package.
         steps:
             1:
-                task: github_package_data
-                options:
-                    context: "Build Feature Test Package"
+                flow: install_2gp_commit
             2:
-                flow: beta_dependencies
-                options:
-                    update_dependencies:
-                        dependencies: ^^github_package_data.dependencies
+                flow: config_apextest
             3:
-                task: install_managed
-                options:
-                    version: ^^github_package_data.version_id
-            4:
-                flow: config_managed
-            5:
                 task: run_tests
 
     ci_master:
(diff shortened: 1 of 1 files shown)


cumulusci/cumulusci.yml
```diff
@@ -895,6 +895,24 @@ flows:
             2:
                 task: uninstall_managed
 
+    install_2gp_commit:
+        group: Install / Uninstall
+        description: Install the 2GP package for the current commit
```
davisagli, line 900: We could expand this description to explain that it's expecting a commit status with the version_id to be set for the "Build Feature Test Package" context.

cumulusci/cumulusci.yml
```diff
@@ -895,6 +895,24 @@ flows:
...
+        steps:
+            1:
+                task: github_package_data
+                options:
+                    context: "Build Feature Test Package"
+            2:
+                flow: beta_dependencies
+                options:
+                    update_dependencies:
+                        dependencies: ^^github_package_data.dependencies
```
davisagli, line 910: Btw, there's an important subtlety in how this works that I want to point out. Our feature test packages are built without specifying any dependencies, which means that the `github_package_data` task will return an empty list of dependencies. Because that's falsy, `update_dependencies` will fall back to using the dependencies defined in project_config.project__dependencies, which is how the 1gp dependencies end up getting installed.

What this means is that specifying the option here is not really doing anything -- but it might in the rare case that we need to customize a project to build a feature test package without skipping validation (because then the 2gp package has to list dependencies, and `github_package_data` will find them, and this option will be truthy).

davisagli (approved):
This makes sense to me, as long as we are okay with 2gp feature test builds failing for projects that haven't updated config_apextests to handle the managed context yet.

davidmreed:
> This makes sense to me, as long as we are okay with 2gp feature test builds failing for projects that haven't updated config_apextests to handle the managed context yet.

I think this is an acceptable risk in the cause of not using `config_managed`, which is widely and heavily customized and will dramatically slow down these test builds. It's relatively unusual to customize `config_apextest`, so I don't anticipate a large scale breakage, but we'll have to fix it where it crops up.

cc @codexmarcel @jstvz @jofsky

davisagli (approved):

PR #2195, opened 2020-11-25
Feature/robot cross project tests

boakley:
Robotframework task Adds repo root to sys.path if it's not already on it.

This is necessary for a task which runs in another repo to be able to find project-relative files.

# Critical Changes

# Changes

# Issues Closed


davisagli (changes requested):
I think we need to be sure to remove the paths that were added after calling robot_run.

Imagine that project A runs the robot task from project B as a cross-project task. So this code adds project B's repo_root to sys.path.

Then if there are subsequent tasks from project A that do imports, they might end up resolving modules from project B.

There will still be the potential of leaving modules from project B in sys.modules for any module that was imported during the robot task's execution. (We discussed that, and decided it's unlikely to be a problem since robot libraries tend to be in a path that includes their project name.) But by cleaning up sys.path we can at least avoid causing problems for later imports of modules that weren't imported during the robot task.

davisagli (approved):

PR #2204, opened 2020-11-27
Clarify what to do if scratch org can't be deleted

prescod:
# Critical Changes

Improve the error messaging around scratch_delete of an org that throws an exception from sfdx.

Force cci_org_remove to remove the org whether SFDX is able to delete it or not.

### DIFF
--- a/cumulusci/core/config/ScratchOrgConfig.py
+++ b/cumulusci/core/config/ScratchOrgConfig.py
@@ -177,7 +177,7 @@ class ScratchOrgConfig(SfdxOrgConfig):
     def can_delete(self):
         return bool(self.date_created)
 
-    def delete_org(self):
+    def delete_org(self, force=False):
         """ Uses sfdx force:org:delete to delete the org """
         if not self.created:
             self.logger.info(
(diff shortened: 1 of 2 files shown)


cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -197,7 +197,15 @@ def delete_org(self):
 
         if p.returncode:
             message = f"Failed to delete scratch org: \n{''.join(stdout)}"
-            raise ScratchOrgException(message)
+            message += "\nPerhaps it was already deleted?"
+            if force:
+                message += "\nRemoving org regardless."
+                self.logger.info(message)
```
davisagli, line 203: I think that suppressing the exception and showing this message belongs in `org_remove`, not here. It would be possible to call `org_config.delete_org(force=True)` in a context that does not proceed to remove the org.

davisagli (changes requested):

cumulusci/core/config/ScratchOrgConfig.py
prescod: I'm confused. I intended the force=True as "remove the org even if you can't delete it". So why would you want to use it in a context where you did not want to remove the org?

davisagli: I agree with that -- which is why I don't think it makes sense to make it possible to use it without removing the org. Either the actual removal of the org from the keychain should also become the responsibility of this function (when `force` is true) or the log message about removing it should move to where it is removed.

davisagli: Also this method could be called from a web app and should not be emitting a message (below) that assumes it's running in the CLI context.

PR #2206, opened 2020-11-30
Add renames option to github_copy_subtree

jstvz:
# Critical Changes

# Changes
- We added the `renames` option to `github_copy_subtree`. This allows mapping between local and target path names when publishing to support renaming a file or directory from the source repository in the target repository.

# Issues Closed


cumulusci/tasks/github/publish.py
```diff
@@ -45,21 +49,60 @@ def _init_options(self, kwargs):
...
         self.options["dry_run"] = process_bool_arg(self.options.get("dry_run", False))
 
+    def _process_renames(self, renamed_paths):
+        """
+        For each entry in renames, any renames and store them
+        in self.local_to_target_paths.
+        """
+        is_list_of_dicts = all(
+            [isinstance(pair, dict) for pair in renamed_paths if renamed_paths]
+        )
```
davisagli, line 78: Can't this simply be:
```suggestion
        is_list_of_dicts = all(isinstance(pair, dict) for pair in renamed_paths)
```
?

cumulusci/tasks/github/publish.py
```diff
@@ -45,21 +49,60 @@ def _init_options(self, kwargs):
...
+        dicts_have_correct_keys = all(
+            [
+                {"local", "target"} == pair.keys()
+                for pair in renamed_paths
+                if is_list_of_dicts
+            ]
+        )
```
davisagli, line 85: You don't need the list here. `all` can iterate over a generator expression. And I don't understand the `if is_list_of_dicts` check.

cumulusci/tasks/github/publish.py
```diff
@@ -45,21 +49,60 @@ def _init_options(self, kwargs):
...
+            "Renamed paths must be a list of dicts with `local:` and `target:` keys."
+        )
+        if not is_list_of_dicts and dicts_have_correct_keys:
+            raise TaskOptionsError(ERROR_MSG)
+
+        local_to_target_paths = {}
+
+        for rename in renamed_paths:
+            local_path = rename.get("local", None)
+            target_path = rename.get("target", None)
```
davisagli, line 97: You can assume these keys are present since you checked above.

cumulusci/tasks/github/publish.py
jstvz: I always forget about generator expressions, thanks.

davisagli (approved):

PR #2208, opened 2020-11-30
Clarify what to do if scratch org can't be deleted and why it can't be deleted

prescod:
Replaces #2204 

Improve the error messaging around scratch_delete of an org that throws an exception from sfdx.

Also a minor cleanup of a double-import of reuqests.exceptions.


cumulusci/cli/cci.py
```diff
@@ -1240,7 +1240,11 @@ def org_scratch_delete(runtime, org_name):
     if not org_config.scratch:
         raise click.UsageError(f"Org {org_name} is not a scratch org")
 
-    org_config.delete_org()
+    try:
+        org_config.delete_org()
+    except Exception as e:
+        click.echo(e)
+        click.echo(f"Use `cci org remove {org_name}` to remove it from your keychain.")
```
davisagli, line 1247: We should probably return here to avoid saving org_config in an unknown state.

cumulusci/cli/tests/test_cci.py
```diff
@@ -1671,14 +1671,14 @@ def test_org_scratch_delete_not_scratch(self):
...
+    def test_org_scratch_delete_error(self, echo):
         org_config = mock.Mock()
         org_config.delete_org.side_effect = ScratchOrgException
         runtime = mock.Mock()
         runtime.keychain.get_org.return_value = org_config
-
-        with self.assertRaises(ScratchOrgException):
-            run_click_command(cci.org_scratch_delete, runtime=runtime, org_name="test")
+        run_click_command(cci.org_scratch_delete, runtime=runtime, org_name="test")
+        assert "org remove" in str(echo.mock_calls)
```
davisagli, line 1681: This is fine, but I think it's also possible to use this pytest fixture to capture output: https://docs.pytest.org/en/stable/capture.html#accessing-captured-output-from-a-test-function

cumulusci/core/config/ScratchOrgConfig.py
```diff
@@ -188,15 +188,15 @@ def delete_org(self):
         p = sfdx("force:org:delete -p", self.username, "Deleting scratch org")
 
         stdout = []
-        for line in p.stdout_text:
+        for line in p.stderr_text:
             stdout.append(line)
```
davisagli, line 192: Let's please not take stderr and call it stdout

I want to make sure we're not suppressing useful information from sfdx about why the deletion failed. How sure are you that everything relevant is in stderr?

davisagli (changes requested):

davisagli (approved):

Issue #2093, opened 2020-10-14
Incorrect deprecation warning or incorrect docs for record_type fields during generate_and_load_from_yaml 


acrosman:
When setting record types by name during `generate_and_load_from_yaml ` the [docs currently suggest](https://cumulusci.readthedocs.io/en/latest/bulk_data.html#record-types) using `RecordTypeId`, and throws a deprecation warning if you use RecordType. However if you run with RecordTypeId you get error messages when attempting to extract record types: 

```
2020-10-14 12:34:18: Running step: Insert Account
2020-10-14 12:34:18: Extracting Record Types for Account
Error: 'Account_rt_mapping'
```

Running `generate_dataset_mapping ` does not resolve this, and I don't see directions on how it should be generated.

davisagli:
@acrosman The docs you're looking at are talking about a scenario where data is extracting using the extract_dataset task, and then loaded into a different org with the load_dataset task. In that scenario, the extract task will store a numeric id for the record type in the RecordTypeId field, and create a separate Account_rt_mapping table in the extracted dataset to store the relationship between record type id and name. This can then be used by the load task to convert to the different ids used for the same record type names in a new org.

But it sounds like you are trying to make Snowfakery generate record type names in the RecordTypeId field. The load task is assuming these are ids which need to be converted, and it is failing to find the Account_rt_mapping table to do so.

For loading record types from Snowfakery, it looks like you'll have to use RecordType, for now.

@prescod @davidmreed we should probably update snowfakery to use the new approach for storing record types, or update the load task to support specifying a record type name in a field.

davisagli:
I think this was resolved

PR #2224, opened 2020-12-05
Issue #2093: Fix confusing warning message about record_types

prescod:
Currently Snowfakery uses a feature that is technically deprecated in mapping files. CumulusCI "warns" the user not to use a feature that they didn't mean to use.

    https://github.com/SFDO-Tooling/CumulusCI/issues/2093

This is a bit of a hack but doing it "right" is fairly hard with Pydantic and all of the layers of CumulusCI and the hack seems pretty low risk. The correct option would need to be threaded through about 6-8 layers of CumulusCI/Pydantic stack frames.

Also the whole thing should be fixed in a better way some time soon: CumulusCI should support record names in a column rather than just indirected record IDs. Then Snowfakery will just output the record names in an intuitive fashion.

### DIFF
--- a/cumulusci/tasks/bulkdata/generate_and_load_data_from_yaml.py
+++ b/cumulusci/tasks/bulkdata/generate_and_load_data_from_yaml.py
@@ -1,5 +1,8 @@
+from unittest import mock
+
 from cumulusci.tasks.bulkdata import GenerateAndLoadData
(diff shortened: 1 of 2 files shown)


davisagli (approved):
I might have made this a flag on the MappingStep class instead of a global, or simply checked the CURRENT_TASK threadlocal from cumulusci.core.tasks, but this is fine.

PR #2225, opened 2020-12-06
Fix docs and requirements-specification for add_standard_value_set

prescod:
I don't think the task actually functions properly when either of the main parameters is missing.

### DIFF
--- a/cumulusci/tasks/metadata_etl/value_sets.py
+++ b/cumulusci/tasks/metadata_etl/value_sets.py
@@ -16,7 +16,7 @@ class AddValueSetEntries(MetadataSingleEntityTransformTask):
             "required": True,
         },
         "api_names": {
-            "description": "List of API names of StandardValuesets to affect, "
+            "description": "List of API names of StandardValueSets to affect, "
             "such as 'OpportunityStage', 'AccountType', 'CaseStatus'",
             "required": True,
         },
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/metadata_etl/value_sets.py
```diff
@@ -6,15 +6,20 @@
...
             "description": "Array of standardValues to insert. "
             "Each standardValue should contain the keys 'fullName', the API name of the entry, "
             "and 'label', the user-facing label. OpportunityStage entries require the additional "
             "keys 'closed', 'won', 'forecastCategory', and 'probability'; CaseStatus entries "
             "require 'closed'.",
-            "required": False,
+            "required": True,
+        },
+        "api_names": {
+            "description": "List of API names of StandardValuesets to affect, "
```
davidmreed, line 19:
```suggestion
            "description": "List of API names of StandardValueSets to affect, "
```

cumulusci/tasks/metadata_etl/value_sets.py
```diff
@@ -6,15 +6,20 @@
...
+            "such as 'OpportunityStage', 'AccountType', 'CaseStatus'",
+            "required": True,
```
davidmreed, line 21: n.b. the original reason `api_names` was not required at the top level is that leaving it blank is equivalent to wildcard (`*`). Wildcard is not meaningful for this task. Thanks for fixing.

davidmreed (changes requested):

davisagli (approved):
Makes sense to me.

davidmreed (approved):

PR #2226, opened 2020-12-07
WIP: Improve error handling, make some errors into warnings

prescod:
A column or table declared in the mapping but missing in the SQL is now a warning. This is in keeping with our decision that one mapping file should work for virtually any dataset related to a project and not just a single dataset.

A missing rt_mapping table is still an error but the error explains the context more.

Also: increase unit testiing of the module to 100% (including one pragma).


davidmreed:
> A column or table declared in the mapping but missing in the SQL is now a warning. This is in keeping with our decision that one mapping file should work for virtually any dataset related to a project and not just a single dataset.

I'm not sure I agree with this suggestion. A mapping and a dataset that are out of sync usually means to me that a user has failed to recapture after altering their mapping, which I see as an error. Can you say more about the situation where this is a desirable workflow?

prescod:
> I'm not sure I agree with this suggestion. A mapping and a dataset that are out of sync usually means to me that a user has failed to recapture after altering their mapping, which I see as an error. Can you say more about the situation where this is a desirable workflow?

NPSP ships with 4 datasets. I don't think it should need to ship with 4 different (mostly hand-edited) mapping files as well. If I add a column or table to mapping.yml for dataset A, I shouldn't need to re-export datasets B, C and D.

The idea of a "schema" which is specific to a single dataset is also kind of weird in terms of industry patterns/standards.

prescod:
@davidmreed Fundamentally we need to come to a consensus on whether our intent is that the relationship between datasets and mapping files should tend towards 1-1 or whether the relationship between mapping files and PROJECTS should tend towards 1-1.

jstvz:

PR #2227, opened 2020-12-07
Fix adding the username to the connected app cmd

atrancandoris:
# Critical Changes

# Changes
Prior to this change, the username is being added to the option, but not the command. I believe _init_options runs before _get_command. The result is that the username is omitted.

# Issues Closed


davisagli:
@atrancandoris Thanks for your pull request. I'm going to copy this to a branch in our github organization so the automated checks can run, and I will also update the tests.

PR #2228, opened 2020-12-07
Fix SOQLQuery task

davisagli:
It was using a method from salesforce-bulk which does not exist.

This task could use a bigger reimagining, but I just wanted to fix the error without spending a lot of time.

# Critical Changes

# Changes

# Issues Closed
- Fixed the `query` task, which was broken.


davidmreed (approved):
(not tested hands on)

prescod:
I tested it and the error handling in particular. It's "good enough" I guess. Spits out raw XML errors but I guess in the context this is probably "good enough."

PR #2236, opened 2020-12-08
EnsureRecordTypes supports Record Type Description

spelak-salesforce:
- Added optional `record_type_description` option that sets the new Record Type's Description.
- If no option is provided, the Description is blank which is the previous behavior.

# Critical Changes

# Changes
`ensure_record_types` supports an optional `record_type_description` option that sets the new Record Type's Description if a Record Type is created.    If no option is provided, the Description is blank.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/EnsureRecordTypes.py
+++ b/cumulusci/tasks/salesforce/EnsureRecordTypes.py
@@ -25,6 +25,7 @@ SOBJECT_METADATA = """<?xml version="1.0" encoding="utf-8"?>
         <active>true</active>
         {business_process_link}
         <label>{record_type_label}</label>
+        <description>{record_type_description}</description>
     </recordTypes>
 </CustomObject>
 """
@@ -56,6 +57,10 @@ class EnsureRecordTypes(BaseSalesforceApiTask):
             "description": "The Label of the Record Type.",
             "required": True,
         },
+        "record_type_description": {
+            "description": "The Description of the Record Type.  Only uses the first 255 characters.",
+            "required": False,
+        },
         "sobject": {
             "description": "The sObject on which to deploy the Record Type and optional Business Process.",
             "required": True,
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/EnsureRecordTypes.py
```diff
@@ -74,6 +79,11 @@ def _init_options(self, kwargs):
                 "Record Type Developer Name value must contain only alphanumeric or underscore characters"
             )
 
+        # Validate Description has 255 characters
+        self.options["record_type_description"] = (
+            str(self.options.get("record_type_description") or "")
+        )[:255]
```
davisagli, line 85: It might be better to raise a TaskOptionsError if the value is too long, but it's probably fine either way.

cumulusci/tasks/salesforce/tests/test_EnsureRecordTypes.py
```diff
@@ -151,6 +154,9 @@ def test_no_business_process_where_unneeded(self):
         self.assertNotIn("stage_name", task.options)
 
     def test_generates_record_type_and_business_process(self):
+        """
+        Asserts Record Type Description is optional.
+        """
```
davisagli, line 159: Please don't add docstrings to test methods -- IIRC the test runner is a bit weird and will show them instead of the name of the method. Use a comment starting with `#` instead.

davisagli (changes requested):
Good idea. One small request about docstrings to avoid a weird quirk of the test runner, but otherwise looks good.

cumulusci/tasks/salesforce/tests/test_EnsureRecordTypes.py
spelak-salesforce: Will convert.  Thank you!

spelak-salesforce: Done.

cumulusci/tasks/salesforce/EnsureRecordTypes.py
spelak-salesforce: I went the simpler-to-test route.   Should I change this?

davisagli: Nah, it's fine. I assume this is more to provide some context and it's not critical to reproduce a long description faithfully.

davisagli (approved):

PR #2239, opened 2020-12-09
Cci org name  and cci org default improvements

prescod:
One can now use --org for commands like `cci org info`, `cci org default`, `cci org remove` etc. This is a usability improvement requested by users on Slack.

One can now pass no org_name to cci org default to print the default.

One can now pass arguments too cci.main and have those arguments go to Click, which is useful in unit testing and generally a good design.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -869,7 +869,7 @@ def set_org_name(required):
             )
         ctx.params.setdefault("org_name", value)
         if required and not ctx.params.get("org_name"):
-            raise click.UsageError("Missing argument 'ORGNAME'")
+            raise click.UsageError("Please specify ORGNAME or --org ORGNAME")
 
     return callback
 
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -856,7 +856,7 @@ def service_info(runtime, service_name, plain):
 
 def set_org_name(required):
     """Generate a callback for processing the `org_name` option or argument
-    
+
     `required` is a boolean for whether org_name is required
     """
     # could be generalized to work for any mutex pair (or list) but no obvious need
@@ -867,7 +867,6 @@ def set_org_name(required):
             raise click.UsageError(
                 f"Either ORGNAME or --org ORGNAME should be supplied, not both ({value}, {prev_value})"
             )
-        ctx.params.setdefault("org_name", value)
         ctx.params["org_name"] = value or prev_value
         if required and not ctx.params.get("org_name"):
             raise click.UsageError("Please specify ORGNAME or --org ORGNAME")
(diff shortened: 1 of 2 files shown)


cumulusci/cli/cci.py
```diff
@@ -854,11 +854,58 @@ def service_info(runtime, service_name, plain):
 # Commands for group: org
 
 
+def set_org_name(required):
+    """Generate a callback that enforces the 'required' rule as required"""
```
davisagli, line 858:
```suggestion
    """Generate a callback for processing the `org_name` option or argument
    
    `required` is a boolean for whether org_name is required
    """
```

cumulusci/cli/cci.py
```diff
@@ -854,11 +854,58 @@ def service_info(runtime, service_name, plain):
...
+    # could be generalized to work for any mutex pair (or list) but no obvious need
+    def callback(ctx, param, value):
+        """Callback which enforces mutex and 'required' behaviour (if required)."""
+        prev_value = ctx.params.get("org_name")
+        if value and prev_value and prev_value != value:
+            raise click.UsageError(
+                f"Either ORGNAME or --org ORGNAME should be supplied, not both ({value}, {prev_value})"
+            )
+        ctx.params.setdefault("org_name", value)
```
davisagli, line 867:
```suggestion
        else:
            ctx.params["org_name"] = value
```

(Correct me if I'm overlooking some semantic difference, but if not I'd rather write it this way so the reader doesn't need to be familiar with setdefault)

cumulusci/cli/cci.py
```diff
@@ -854,11 +854,58 @@ def service_info(runtime, service_name, plain):
...
+        if required and not ctx.params.get("org_name"):
+            raise click.UsageError("Missing argument 'ORGNAME'")
```
davisagli, line 869:
```suggestion
            raise click.UsageError("Please specify ORGNAME or --org ORGNAME")
```

(Where possible exceptions should suggest what to do, not just say what was wrong.)

cumulusci/cli/cci.py
```diff
@@ -854,11 +854,58 @@ def service_info(runtime, service_name, plain):
...
+
+    return callback
+
+
+def orgname_option_or_argument(*, required):
+    "Create decorator that allows org_name to be an option or an argument"
```
davisagli, line 875: Please follow the normal convention for formatting Python docstrings.

```suggestion
    """Create decorator that allows org_name to be an option or an argument"""
```

davisagli (changes requested):
Thanks, these will be nice improvements to have and the implementation looks good. I have a few small tweaks to suggest.

I would ask for adjustments to the docs but I don't want to conflict with the big doc project, so it's probably best if we do that separately.

davisagli (approved):

PR #2240, opened 2020-12-09
execute_anon supports minimizing Apex

spelak-salesforce:
The intent of this feature is to reduce the URI length for the GET Tooling API execute anonymous request by giving an option to `minimize_apex`.   This option may have unintended consequences since it alters the Apex before the GET request, and therefore, this option is off/False by default.

- `minimize_apex` option  defaults to `False` to not interfere with existing implementations.
- If `minimize_apex` option is `True`, `_prepare_apex` does the following:
    - Removes block comments
    - Removes inline comments
    - Removes whitespace at the start of lines

# Critical Changes

# Changes
`execute_anon` task supports a `minimize_apex` option that removes block comments, inline comments, and whitespace at the start of lines.

# Issues Closed


docs/flows.rst
```diff
@@ -170,6 +170,26 @@ qa_org
 	    2) task: update_admin_profile
 	4) task: snapshot_changes
 
+qa_org_2gp
```
spelak-salesforce, line 173: @davisagli , I called `make docs`, and this changed appeared.  I can remove this file change.

docs/tasks.rst
```diff
@@ -1755,6 +1760,11 @@ Options
 
 	 Parameter to pass to the Apex. Use as %%%PARAM_2%%% in the Apex code.Defaults to an empty value.
 
+``--minimize_apex MINIMIZEAPEX``
+	 *Optional*
+
+	 Removes block comments, inline comments, and whitespace at the start of lines.  Default: False
+
```
spelak-salesforce, line 1767: @davisagli , this is the only change I want.

I called make docs, and all of the other changes appeared.   Any suggestions how to proceed?

davisagli: @spelak-salesforce sorry, this is because I made a change to the generated task doc format but forgot to actually regenerate it. You may keep these changes or remove them.

Long term we are going to stop storing this file in the repository, and only generate it while building the docs.

spelak-salesforce: I'll keep the changes then if they're ok to merge.

davisagli:
@spelak-salesforce Before reviewing this I want to make sure I understand what prompted the change. Did you run into an error due to the query being too long?

spelak-salesforce:
> @spelak-salesforce Before reviewing this I want to make sure I understand what prompted the change. Did you run into 
an error due to the query being too long?

Yes.

We have a large Execute Anonymous script to create Action Plan Template storytelling data in Grants-Management because of quirks how Action Plan Templates are created + cci bulk data doesn't support these quirks yet.  I was getting a `URI too long` error message trying to execute this script.

By using this commit, I reduced a 14,000+ character GET request to 10,000+ characters.  Because we use prettier, our Apex Anonymous files includes a lot of whitespace at the start of lines.  Then it was easy to also remove comments.

davisagli:
@spelak-salesforce I wonder if there is a different way we can run the anonymous Apex that would not be subject to this limit. Could we do it via the SOAP API instead of Tooling API? I'd rather not make users need to think about this option if we can avoid it, especially since it only moves where the limit will be hit.

davidmreed:
The Anon Apex call is available in the SOAP API. I've never used it myself and don't see any documentation of a limit there.

spelak-salesforce:
> @spelak-salesforce I wonder if there is a different way we can run the anonymous Apex that would not be subject to this limit. Could we do it via the SOAP API instead of Tooling API? I'd rather not make users need to think about this option if we can avoid it, especially since it only moves where the limit will be hit.

I scratched the surface researching SOAP API.  Will look into it further.   The intent PR was more of a quick win vs permanent solution.

davidmreed:
Another option that feels more stable to me is to actually deploy the Anonymous Apex as an unmanaged class, and simply _invoke_ it via Anonymous Apex. That should be achievable with current functionality.

spelak-salesforce:
> Another option that feels more stable to me is to actually deploy the Anonymous Apex as an unmanaged class, and simply _invoke_ it via Anonymous Apex. That should be achievable with current functionality.

@davidmreed , if using sfdx, one can add the directory containing the unpackaged Apex class to the `sfdx-project.json` `packageDirectories` and and push/pulls/deploys/retrieves will be routed to the correct, non-packaged directory.

jstvz:

PR #2241, opened 2020-12-09
tostring() works for individual elements and parent namespaces can be printed

prescod:
@bethbrains found a bug whereby the tostring() method would crash due to trying to look up namespace prefixes for a namespace.

This raised the question of whether elements like that should be printed with or without a namespace. I decided to make it a user-option.

### DIFF
--- a/cumulusci/utils/xml/test/test_metadata_tree.py
+++ b/cumulusci/utils/xml/test/test_metadata_tree.py
@@ -211,3 +211,30 @@ class TestMetadataTree:
 </CustomMetadata>""".strip()
         CustomMetadata = fromstring(xml)
         assert xml.strip() == CustomMetadata.tostring().strip()
+
+    def test_namespaced_to_string(self):
+        xml = """
+<CustomMetadata xmlns="http://soap.sforce.com/2006/04/metadata" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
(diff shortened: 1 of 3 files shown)


cumulusci/utils/xml/test/test_metadata_tree.py
```diff
@@ -211,3 +211,30 @@ def test_multiple_namespaces(self):
...
+        xml_out = (
+            CustomMetadata.find("values")
+            .find("field")
+            .tostring(xml_declaration=True, include_parent_namespaces=True)
+        )
+        expected_out = """<?xml version="1.0" encoding="UTF-8"?>
+             <field xmlns="http://soap.sforce.com/2006/04/metadata" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">dlrs__Active__c</field>"""
+        assert (
+            " ".join(xml_out.split()).strip() == " ".join(expected_out.split()).strip()
+        ), xml_out.strip()
```
davisagli, line 240: I'll leave this up to you, but I tend to prefer writing separate test cases rather than a single one that tests multiple scenarios. It makes sure that a failure on the first scenario doesn't prevent running the second one. And it's easy enough to move shared setup to a global or fixture.

davisagli (approved):

davisagli (approved):

PR #2243, opened 2020-12-10
Feature/robot spring 21

boakley:
Tweaks to make our keywords and tests pass on Spring 21.

# Critical Changes

# Changes
* new keyword 'Field value should be'
* new keyword 'Modal should show edit error for fields' for Spring 21 form field error notification
* fixed bug preventing 'Get field value' from working in some cases
* fixed bug preventing 'Select dropdown value' from working in Spring 21

# Issues Closed


davisagli (approved):
@boakley This all looks good to me! Is there a reason it's still in draft? I'd like to get it into today's release if possible.

boakley:
I had left it in draft status because there's still some tests that are failing on a prerelease org (in CI, I'm having trouble reproducing them locally). But it looks like all the checks are green so I've removed the draft status.

PR #2246, opened 2020-12-11
Trying to improve task_doc

Br4nd0R:
Attempting to improve the task_doc for `create_managed_src`. I'd like to include an example of how the `//cumulusc-managed` string is used.

Updated module to use `pathlib` and fstrings.


cumulusci/tasks/metadata/managed_src.py
```diff
@@ -7,6 +7,9 @@
 
 
 class CreateManagedSrc(BaseTask):
+    task_docs = """The //cumulusci-managed string is to guard the @deprecated
+annotation which is not legal to deploy in an unmanaged context."""
```
davisagli, line 11: How about:

    task_docs = """Apex classes which use the @deprecated annotation can comment it out using //cumulusci-managed so that it can be deployed as part of unmanaged metadata, where this annotation is not allowed. This task is for use when deploying to a packaging org to remove the comment so that the annotation takes effect."""

cumulusci/tasks/metadata/managed_src.py
```diff
@@ -81,22 +78,18 @@ class RevertManagedSrc(BaseTask):
...
-        self.logger.info("{} is now reverted".format(self.options["path"]))
+        self.logger.info(f"Reverting {path} from {revert_path}")
+        copy_tree(str(revert_path), str(path), update=1)
+        self.logger.info(f"{path} is now reverted")
 
         # Delete the revert_path
-        self.logger.info("Deleting {}".format(self.options["revert_path"]))
-        remove_tree(self.options["revert_path"])
+        self.logger.info(f"Deleting {revert_path}")
+        remove_tree(revert_path)
```
davisagli, line 95: Probably need str() here?

davisagli (changes requested):
Need to update tasks.rst

Br4nd0R:
done

davisagli (approved):

PR #2249, opened 2020-12-13
WIP: CLI improvements for Robot

prescod:
1. Allow passing Robot --options X:Y,A:B on the CLI.
2. Allow passing --outputdir on the CLI. Takes precedence over --options outputdir and YAML.
3. Allow using $CUMULUSCI_ROBOT_OUTPUT_DIR for MetaCI and other CI contexts
4. Default robot/{self.project_config.repo_name}/results

Before I write tests I would like feedback on the semantics.

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -15,9 +15,13 @@ from cumulusci.core.exceptions import (
 from cumulusci.core.tasks import BaseTask
 from cumulusci.core.utils import process_bool_arg
 from cumulusci.core.utils import process_list_arg
+from cumulusci.core.utils import process_list_of_pairs_dict_arg
 from cumulusci.robotframework.utils import set_pdb_trace
 from cumulusci.tasks.salesforce import BaseSalesforceTask
 from cumulusci.tasks.robotframework.debugger import DebugListener
+
+
+# WIP: No tests for new code.
 
 
 class Robot(BaseSalesforceTask):
@@ -40,7 +44,14 @@ class Robot(BaseSalesforceTask):
             "required": False,
         },
         "options": {
-            "description": "A dictionary of options to robot.run method.  See docs here for format.  NOTE: There is no cci CLI support for this option since it requires a dictionary.  Use this option in the cumulusci.yml when defining custom tasks where you can easily create a dictionary in yaml."
+            "description": "In simple cases this can be specified on the command line using "
+            "name:value,name:value syntax. More commplex cases can be specified "
+            "in cumuuluci.yml using YAML dictionary syntax.",
+            "required": False,
+        },
+        "outputdir": {
+            "description": "The directory in which to place HTML and XML output",
+            "required": False,
         },
         "name": {"description": "Sets the name of the top level test suite"},
         "pdb": {"description": "If true, run the Python debugger when tests fail."},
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -40,7 +44,14 @@ class Robot(BaseSalesforceTask):
             "required": False,
         },
         "options": {
-            "description": "A dictionary of options to robot.run method.  See docs here for format.  NOTE: There is no cci CLI support for this option since it requires a dictionary.  Use this option in the cumulusci.yml when defining custom tasks where you can easily create a dictionary in yaml."
+            "description": "In simple cases this can be specified on the command line using "
+            "name:value,name:value syntax. More commplex cases can be specified "
+            "in cumuuluci.yml using YAML dictionary syntax.",
```
davisagli, line 49:
```suggestion
            "in cumulusci.yml using YAML dictionary syntax.",
```

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -98,15 +110,38 @@ def _init_options(self, kwargs):
...
+        # This code is not clear to me so I'm a bit afraid to change it
+        def relativize_path(output_dir):
+            return os.path.relpath(
+                os.path.join(self.working_path, output_dir), os.getcwd()
+            )
+
+        output_dir = None
+
+        if os.environ.get("CUMULUSCI_ROBOT_OUTPUT_DIR"):
+            output_dir = relativize_path(os.environ.get("CUMULUSCI_ROBOT_OUTPUT_DIR"))
```
davisagli, line 124: Brandon is currently working on adjusting the way that MetaCI imports robot results so that it will be aware of the task's outputdir option (https://github.com/SFDO-Tooling/MetaCI/pull/1786). So it won't be necessary to change it for MetaCI, and won't be necessary to have this way to override the default.

prescod: Does the rest make sense to you?

davisagli: Yes, the rest does, at least at first glance.

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -40,7 +44,14 @@ class Robot(BaseSalesforceTask):
...
         },
         "options": {
-            "description": "A dictionary of options to robot.run method.  See docs here for format.  NOTE: There is no cci CLI support for this option since it requires a dictionary.  Use this option in the cumulusci.yml when defining custom tasks where you can easily create a dictionary in yaml."
+            "description": "In simple cases this can be specified on the command line using "
+            "name:value,name:value syntax. More commplex cases can be specified "
+            "in cumuuluci.yml using YAML dictionary syntax.",
+            "required": False,
+        },
+        "outputdir": {
+            "description": "The directory in which to place HTML and XML output",
```
boakley, line 53: This is slightly inaccurate, there could be other types of files there, such as screenshots. How about "the directory where robotframework will  write all output"?

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -98,15 +110,38 @@ def _init_options(self, kwargs):
...
+            )
+
+        output_dir = None
+
+        if os.environ.get("CUMULUSCI_ROBOT_OUTPUT_DIR"):
+            output_dir = relativize_path(os.environ.get("CUMULUSCI_ROBOT_OUTPUT_DIR"))
+
+        fallback_paths = [f"robot/{self.project_config.repo_name}/results", "."]
+
+        while not output_dir:
```
boakley, line 128: I don't like this behavior. If I set the outputdir to robot/output but robot/output doesn't exist, this will end up dumping everything in the current directory, right? This is different than robot's own default which will create the path if it doesn't exist. I think we should accept whatever is given, and let robot create the folder if it doesn't exist.

jstvz:

Issue #2250, opened 2020-12-14
UTF Decode Error on Windows in SFDX exception handling


prescod:
A user generated this gist which indicated bugs in our decoding/exception handling:

```
CumulusCI version: 3.25.0
Python version: 3.8.2 (c:\users\username\.local\pipx\venvs\cumulusci\scripts\python.exe)
Environment Info: Windows / AMD64


Last Command Run
================================
C:\Users\username\Documents\SFRepositories\omega>cci org scratch release OmegaMain --default --days 7
OmegaMain is now the default org

C:\Users\username\Documents\SFRepositories\omega>cci flow run regression_org
2020-12-14 11:54:27: ============================================================
2020-12-14 11:54:27: Initializing flow: FlowCoordinator (regression_org)
2020-12-14 11:54:27: Simulates an org that has been upgraded from the latest relea
se of to the current beta and its dependencies, but deploys any unmanaged metadata from the current beta.
2020-12-14 11:54:27: ============================================================ 
2020-12-14 11:54:27:
2020-12-14 11:54:27: Verifying and refreshing credentials for the specified org: O
megaMain.
2020-12-14 11:54:27: Creating scratch org with command: sfdx force:org:create -f o
rgs/release.json -n --durationdays 7 -a "Omega__OmegaMain" -s -w 120 adminEmail="username@salesforce.com"

Error: Failed to create scratch org: 

WARNING: Warning: If you're using Contacts to Multiple Accounts, beginning with the Spring 
’21 release, the feature isn't automatically enabled. You must manually enable the setting 
for new scratch orgs. Enable the setting in one of these ways: update the Account Settings 
page in Setup, add the enableRelateContactsToMultipleAccounts setting to your scratch org definition file, or enable it via Metadata API.
ERROR running force:org:create:  An unknown server error occurred. Please try again. If you still see this error, contact Salesforce support for assistance. Include the information from "sfdx force:data:record:get -s ScratchOrgInfo -i 2SR4p000000QYb9GAG -u username@sfdo.devhub".
Run this command for more information about debugging errors: cci error --help
Traceback (most recent call last):
  File "c:\users\username\.local\pipx\venvs\cumulusci\lib\site-packages\cumulusci\cli\cci.py", line 229, in main
    cli(args[1:], standalone_mode=False)
  File "c:\users\username\.local\pipx\venvs\cumulusci\lib\site-packages\click\core.py", line 829, in __call__
    return self.main(*args, **kwargs)
[...]

PR #2251, opened 2020-12-14
Improve Handling Windows Decode Error

Br4nd0R:
# Changes
* Replace backslashes if we hit an error decoding the temporary logfile

### DIFF
--- a/cumulusci/utils/logging.py
+++ b/cumulusci/utils/logging.py
@@ -41,8 +41,9 @@ def tee_stdout_stderr(args, logger, tempfile):
         logger.handlers[0].close()
 
         # log contents of tempfile to rotating log files
-        with open(tempfile, "r", encoding="utf-8") as f:
-            contents = f.read()
+        with open(tempfile, "rb") as f:
+            binary_data = f.read()
+            contents = binary_data.decode("utf-8", errors="backslashreplace")
 
         logger = get_gist_logger()
         logger.debug(contents)


Br4nd0R:
I still need to see if I can write a Windows only test for this.

prescod:
The test probably doesn't need to be Windows-specific. If the tempfile is corrupt/non-utf-8 for any reason, on any platform, the same error will be triggered.

I'm quite confused about how the tempfile ends up non-utf-8 in the first place.

`get_tempfile_logger` seems fine....

prescod:
David Glick prefers that we get to the root cause: https://github.com/SFDO-Tooling/CumulusCI/issues/2250#issuecomment-744612290

davisagli:
I'd prefer to figure out why the tempfile isn't getting written as utf-8, but it's not obvious to me either. I'd suggest a time-boxed debugging session in Windows, and then go with this workaround if it's still unclear.

prescod:
@Br4nd0R If you can reproduce the problem at all, one way of figuring out where exactly the bad data is inserted might be to insert some kind of mock file object that inspects every write. But that's assuming that you have a test case.

davisagli:
I wonder if this is somehow related to the capturing of output from sfdx, and its output being cp1252 (the Windows default) rather than utf-8. But if that's the case I would expect it to blow up as soon as the sfdx output is processed, not at the end of the command while processing the tempfile log.

PR #2252, opened 2020-12-14
Make robot set outpudir as return value

Br4nd0R:
# Changes
* The `Robot` task now sets a key in the tasks `return_values` for the `robot_outputdir`. This holds the file path to where the robot test results were written.

This change is so that MetaCI can easily get the output directory of the robot test results file (`output.xml`).
See: https://github.com/SFDO-Tooling/MetaCI/pull/1786


davisagli (commented):
I think this should be pretty uncontroversial, but FYI @boakley in case you want to weigh in.

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -108,6 +108,9 @@ def _run_task(self):
         options["outputdir"] = os.path.relpath(
             os.path.join(self.working_path, options.get("outputdir", ".")), os.getcwd()
         )
+        # Set as a return value so other things that want to use
+        # this file (e.g. MetaCI) know where it is
+        self.return_values["robot_outputdir"] = options["outputdir"]
```
boakley, line 113: This looks like a great enhancement. I didn't know `self.return_values` was a thing.  I guess I should find time over the holidays to RTFM.

cumulusci/tasks/robotframework/tests/test_robotframework.py
```diff
@@ -252,6 +252,7 @@ def test_sources(self, mock_add_path, mock_robot_run):
         )
         self.assertNotIn("dummy1", sys.path)
         self.assertNotIn("dummy2", sys.path)
+        self.assertEquals(".", task.return_values["robot_outputdir"])
```
boakley, line 255: It might be good to have another test to verify that if we set the outputdir option, it is reflected in the return values to make sure we aren't just always returning the default rather than actual.

boakley (dismissed):
👍 I think the test could maybe be improved slightly, but otherwise this looks like a good enhancement.

cumulusci/tasks/robotframework/robotframework.py
davisagli: @boakley I think you'll probably find the FM sadly lacking... Return values are not displayed anywhere, but are available for later flow steps to reference in their options (i.e. there could be a "process_robot_output" step which uses this return value:
```
steps:
    1:
        task: robot
    2:
        task: process_robot_output
        options:
            path: ^^robot.robot_outputdir
```

But the thing that's prompting the addition of this particular return value is wanting to use it in the system that is running the flow (MetaCI), rather than by another task.

cumulusci/tasks/robotframework/tests/test_robotframework.py
```diff
@@ -292,6 +292,44 @@ def test_sources_not_found(self, mock_robot_run):
...
+        },
+    )
+    # get_namespace returns a config. The only part of the config
+    # that the code uses is the repo_root property, so we don't need
+    # a full blown config.
+    project_config.get_namespace = mock.Mock(
+        side_effect=lambda source: mock.Mock(
+            repo_root=project_config.sources[source]["path"]
+        )
+    )
```
davisagli, line 315: Why not just leave `sources` out of the task options and then you don't need to mock this?

Br4nd0R: This is definitely not needed. I'll remove it.

davisagli (approved):

PR #2255, opened 2020-12-15
Justin's Comments on Key Concepts Section of the Docs

Br4nd0R:

### DIFF
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -65,7 +65,7 @@ Each project can have one or more **datasets**: on-disk representations of recor
 
 Robot Framework provides browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the project's Robot Framework test suites. New projects start with a simple Robot test case that creates a Contact record.
 
-While Robot Framework is used primarily for browser automation testing, it can also be harnessed to help configure orgs where other strategies and APIs are insufficient. [TODO LINK TO EXPLAIN THIS FURTHER?]
+While Robot Framework is used primarily for browser automation testing, it can also be harnessed to help configure orgs where other strategies and APIs are insufficient [TODO: link Robot Framework].
 
 ``unpackaged`` metadata
 ^^^^^^^^^^^^^^^^^^^^^^^
@@ -110,7 +110,7 @@ Popular task commands include:
 * ``cci task info <name>``: Learn more about a task and how to configure its options. ``<name>`` is the name of the task.
 * ``cci task run <name> --org <org>``: Run a task. ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against.
 
-        Example: ``run_tests`` executes Apex unit tests. If you have an org called ``dev``, you can run this task against it with the command ``cci task run run_tests --org dev``
+        Example: ``run_tests`` executes Apex unit tests. If you have an org called ``dev``, you can run this task against it with the command ``cci task run run_tests --org dev``.
 
 Many operations in CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and even other flows!) that produce a cohesive outcome, such as an org that's configured to suit a workflow like development, QA, or product demonstration.
 
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -1,119 +1,108 @@
 Key Concepts
 ============
 
-The Product Delivery Model
------------------------------
-CumulusCI extends the concepts of the `Org Development Model <https://trailhead.salesforce.com/en/content/learn/modules/org-development-model>`_ and the `Package Development Model <https://trailhead.salesforce.com/en/content/learn/modules/sfdx_dev_model>`_ to form a new model, which we call the Product Delivery Model.
+Key concept. Developer jargon. 10-cent words that sound like a technical writer patting himself on the back. (Hi there!) We imagine your cursor subconsciously drifting its way to the hyperlink for the next chapter already. BUT in an effort to give you the tools to build and test customized features with CumulusCI, it's only right that we demystify the process by explaining the *how* and the *why* in order to help you achieve your goal of seamless automation.
 
-The Product Delivery Model focuses on the customer experience, not on the technical artifacts you're delivering, and focuses on making it possible to deliver a first-class, fully configured customer experience – no matter how complex the product might be. When building a product, there can be detailed technical considerations for whether a specific component is best distributed within a package, as additional unpackaged metadata, or as sophisticated setup automation that runs before or after a package is installed. And, of course, a product may include not one, but several packages, which must be installed in sequence. CumulusCI makes it possible to deliver a cohesive product that makes use of these different techniques.
+Don't worry, we'll make this as painless as possible.
 
-CumulusCI provides the core framework of operations for automating the Product Delivery Model throughout the development lifecycle, and a command line interface to those operations. Once configured, the same automation can also be used by other applications in the CumulusCI Suite, such as MetaCI for automated testing and MetaDeploy for delivering products to customers.
+Packages
+--------
 
-The Product Delivery Model aims to represent a holistic view of delivery of a product instead of simply releasing a package. Here's a real-world example drawn from Salesforce.org's product portfolio.
+If you've come this far in your search of customization, we're guessing you're familiar with packages as well as their role in customizing a Salesforce org. However, because most everything that is built in CumulusCI is deployed via packages, it's essential to clarify once more:
 
-The Nonprofit Success Pack and the Product Delivery Model
-++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+A **package** is a container for something as small as an individual component or as large as a set of related apps. After creating a package, you can distribute it to other Salesforce users and organizations, including those outside your company.
 
-The Nonprofit Success Pack, or NPSP, is one of Salesforce.org's flagship products. The NPSP consists of six managed packages, with complex dependency relationships between them, alongside unpackaged metadata and setup automation that helps create a usable NPSP org.
+**Unmanaged packages** are typically used to distribute open-source (non-proprietary) features or application templates to provide developers with the basic building blocks for an application. Once the components are installed from an unmanaged package into a specific org, it's what's known as an org implementation. These freshly installed components can be edited as seen fit by the owners of this implementation. The developer who created and uploaded the unmanaged package has no control over the installed components, and can't change or upgrade them.
+ 
+**Managed packages** are typically used by Salesforce partners to distribute and sell applications to customers. They are proprietary code that can only be upgraded and deployed by the developer that built them. To ensure seamless upgrades, certain destructive changes, like removing objects or fields, cannot be performed.
 
-Salesforce.org delivers NPSP using CumulusCI and the Product Delivery Model because those tools are what make it possible for us to ship this large, complex, heterogeneous application efficiently to tens of thousands of customers — without a lengthy setup guide to be completed by the end user. CumulusCI automation seamlessly
-
-* deploys NPSP's five dependent managed packages in the right sequence;
-* delivers unpackaged Record Types for the Account and Opportunity objects;
-* installs NPSP;
-* and performs final configuration to make the customer's experience better, like setting up Global Actions and delivering translations.
-
-CumulusCI runs this automation throughout our development lifecycle, starting from feature branches in the hands of our developers and culminating in the delivery of new versions of the application to our users through MetaDeploy.
-
-Throughout the CumulusCI documentation, we'll have the Product Delivery Model in mind. 
+Packages are also built, stored and deployed via **projects**.
 
 Projects
 --------
 
-When you work with CumulusCI, you do so inside a project. A project is a version control repository that contains Salesforce metadata as well as the CumulusCI automation that builds and releases the project. A project usually has a one-to-one relationship with a managed package, if building a package, or with an org implementation. A project may constitute the entirety of a product in the Product Delivery Model, or may be one of multiple projects making up a product.
+When you work with CumulusCI, you do so inside a project. A project is an individual git repository (think GitHub) that contains Salesforce metadata as well as the CumulusCI automation (e.g., tasks and flows) that builds and releases the project. Most importantly, a project can build one - and only one(!) - package in its repository. This one-to-one relationship between project and package is what allows version control to track any changes your team made to the repository with minimal collision.
 
-CumulusCI scopes many of its activities to the project, so you'll always run your CumulusCI commands inside your repository directory. Each project gets its own keychain, which we'll talk about in detail below. Each project has its own set of scratch orgs, which CumulusCI namespaces in the Salesforce CLI keychain to prevent collisions. And each project is linked to a single GitHub repository, so you can always be confident of the scope of the actions you're taking.
-
-CumulusCI knows which project you are working on by the current working directory of your shell. Make sure to change directories inside your project before you run project-specific commands.
-
-In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We'll show you how to get set up with a new or existing CumulusCI project in the (TODO link) Get Started section.
+It's important to note that a project may not even contain a package; for example, it can be built to deliver functionality to records, or test data for QA. A project may constitute the entirety of a product made for release, or may be one of multiple projects banded together to make a product (think NPSP). Basically, whatever feature in your product that needs specialized Salesforce metadata and CumulusCI automation in order to be built and deployed, *especially* packages (we cannot stress this enough!), requires its own unique project. 
 
 Project Structure
 -----------------
 
-A CumulusCI project represents an application or a major component of an application on the Salesforce platform. Each project corresponds with a single managed package, or with an org-based implementation. In this documentation, we'll often speak about a package; if you're building a project that isn't packaged, please take this to refer to an unmanaged package within which your implementation is deployed.
-
-Together with the core package, a CumulusCI project encompasses a variety of unpackaged application components, automation, tests, and other support material. We'll quickly survey the primary components of a CumulusCI project here. Each component is covered in more depth later in this guide.
-
-Automation Configuration (``cumulusci.yml``)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-``cumulusci.yml`` is the definition of the project's automation. It includes information about the project itself as well as customizations that define how scratch orgs are built and provide other important functionality to be used as part of the project's lifecycle.
-
-When you work with CumulusCI in a project, the contents of your ``cumulusci.yml`` are merged with the ``cumulusci.yml`` that comes with CumulusCI and defines out-of-the-box functionality. From time to time we'll refer to that as the standard library. Learn more about customizing CumulusCI automation in TODO: link Customization section.
-
-Project Source Code
-^^^^^^^^^^^^^^^^^^^
-
-The main body of the project's code and metadata lives in the ``force-app`` directory for Salesforce DX-format projects and ``src`` for Metadata API-format projects. For managed package projects, only this metadata is part of the package. Most projects have the vast majority of their components stored here.
-
-Orgs
-^^^^
-
-The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in TODO: link Scratch Org Management. 
-
-Datasets
+Keychain
 ^^^^^^^^
 
-Each project may have one or more *datasets*: on-disk representations of record data that can be inserted into Salesforce orgs, and which can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in TODO: link Automating Data Operations.
+Each project comes with a Salesforce CLI keychain, a unique identifier of characters to separate a project and its files from any other. This deters collisions with another org. (Example: Most projects contain a "dev" org. While the name is ubiquitous across Salesforce orgs, the keychain attached is what makes it unique.) 
 
-Robot Framework Browser Automation
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Project Directory
+^^^^^^^^^^^^^^^^^
 
-CumulusCI integrates Robot Framework browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the resources and test suites for the project's Robot Framework test suites. New CumulusCI projects start with a simple Robot test case that creates a Contact record. 
+The root directory of your CumulusCI project. Because each project is linked to a single GitHub repository, you can always be confident of the scope of the actions you're taking. And thanks to the keychain, CumulusCI knows which project you are working on by the current working directory of your shell. 
 
-While Robot Framework is used primarily for browser automation testing, in a small number of situations, Robot Framework may also be harnessed to help configure orgs where other strategies and APIs are insufficient.
+[NOTE: Save yourself a headache by making sure you're in the correct repository for your project before running project-specific commands; otherwise, your project will produce an error. (This tends to be one of the biggest troubleshooting issues that developers run into with CumulusCI. Check for this first and you might save yourself an extra trip to this guide.)]
 
-Unpackaged Metadata
+In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We'll show you how to get set up with a new or existing CumulusCI project in the [TODO link Get Started section].
+
+cumulusci.yml
+^^^^^^^^^^^^^
+
+The ``cumulusci.yml`` file defines a project's automation in CumulusCI. It contains all the customizations and configurations that pertain to your project's lifecycle. It can encompass everything from customizing the shapes of scratch orgs to configuring tasks and flows.
+
+Learn more about customizing CumulusCI automation in the [TODO: link Customization section].
+
+force-app (or src)
+^^^^^^^^^^^^^^^^^^
+
+The main body of the project's code and metadata lives in the ``force-app`` directory for Salesforce DX-format projects and ``src`` for Metadata API-format projects. For managed package projects, only this metadata is part of the package. Projects have the vast majority of their components stored here.
+
+orgs Directory
+^^^^^^^^^^^^^^
+
+The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in [TODO: link Scratch Org Management].
+
+datasets
+^^^^^^^^
+
+Each project may have one or more **datasets**: on-disk representations of record data that can be inserted into Salesforce orgs, and which can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
+
+Robot
+^^^^^
+
+Robot Framework provides browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the resources and test suites for the project's Robot Framework test suites. New CumulusCI projects start with a simple Robot test case that creates a Contact record.
+
+While Robot Framework is used primarily for browser automation testing, it can also be harnessed to help configure orgs where other strategies and APIs are insufficient. [TODO LINK TO EXPLAIN THIS FURTHER?]
+
+unpackaged metadata
 ^^^^^^^^^^^^^^^^^^^
 
-In the Product Delivery Model, a product doesn't just encompass the contents of a managed package or a single deployment. It also includes *unpackaged metadata*: extra bundles of Salesforce metadata that further tailor an org or complete the product.
+A product doesn't just encompass the contents of a managed package or a single deployment. It also includes **unpackaged metadata**: extra bundles of Salesforce metadata that further tailor an org or complete the product.
 
-In a CumulusCI project, all unpackaged metadata is stored in subdirectories within the ``unpackaged`` directory. Unpackaged metadata plays multiple roles within the Product Delivery Model, including preparing an org for installing packages, adding more customization after the package or application is deployed, and customizing specific orgs that are used in the product's development process.
+In a CumulusCI project, all unpackaged metadata is stored in subdirectories within the ``unpackaged`` directory. Unpackaged metadata plays multiple roles, including preparing an org for installing packages, adding more customization after the package or application is deployed, and customizing specific orgs that are used in the product's development process.
 
-Learn more about managing unpackaged metadata in TODO: link Managing unpackaged configuration.
+Learn more about managing unpackaged metadata in [TODO: link Managing unpackaged configuration].
 
 Project Orgs & Services
 -----------------------
 
-Orgs and services are external, authenticated resources that each project uses. CumulusCI makes it easy to connect orgs, as well as services like a GitHub account or a MetaDeploy instance, to a single project or to use them across projects. 
+Orgs and services are external, authenticated resources that each project uses. CumulusCI makes it easy to connect orgs, as well as services like GitHub or MetaDeploy, to a single project, or to use them across many projects.
 
-Orgs
-^^^^
+Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain, making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be shared across multiple projects.
 
-Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain, making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be `shared across multiple projects <TODO: link to persistent orgs section>`_.
+Configuring orgs in CumulusCI is powerful, but comes with some complexity. To review the details, read the sections [TODO: link Scratch org environments] and [TODO: link persistent org section].
 
-Configuring orgs in CumulusCI is powerful, but comes with some complexity. To review all of the details, read the sections TODO: link "Scratch org environments" and TODO: link persistent org section.
+Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. The ``command cci service`` list shows you which services are connected in the context of the current project.
 
-Services
-^^^^^^^^
-
-Services represent external resources used by CumulusCI automation, such as access to a GitHub account or a MetaDeploy instance. Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. 
-
-The command ``cci service list`` shows you which services are connected in the context of the current project.
-
-Some services can be connected at the project level, which means that they're scoped to a single project and aren't shared. We recommend primarily using global services because they're easier to use, but you may encounter scenarios where, for example, you need to use a specific Dev Hub for one and only one project. Connecting that service with ``cci service connect devhub --project`` supports that use case.
+Services can be connected at the project level, which means that they're scoped to a single project and cannot be shared. Global services are easy to use and share and, therefore, we recommend you rely on them for the most part. However, when you encounter a scenario where, for example, you need to use a specific Dev Hub for one - and only one(!) - project, you can simply connect to that service by way of the ``cci service connect devhub --project`` command.
 
 Tasks and Flows
 ---------------
 
-CumulusCI uses a framework of *tasks* and *flows* to organize the automation that is available to each project.
+CumulusCI uses a framework of **tasks** and **flows** to organize the automation that is available to each project.
 
-Tasks are units of automation. A task could perform a deployment, load a dataset, retrieve material from an org, install a managed package, or undertake many other activities. CumulusCI ships with scores of tasks out of the box. 
+Tasks are units of automation. A task could perform a deployment, load a dataset, retrieve material from an org, install a managed package, or undertake many other activities. CumulusCI ships with scores of tasks out of the box.
 
-You can review the tasks available in a project by running ``cci task list``, learn more about a task and how to configure its options with ``cci task info <name>``, where ``<name>`` is the name of the task, and run a task with ``cci task run <name> --org <org>``, where ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against. For example, the ``run_tests`` task executes Apex unit tests. If you have an org called ``dev``, you can run this task against this org with the command ``cci task run run_tests --org dev``.
+You can review the tasks available in a project by running ``cci task list``; learn more about a task and how to configure its options with ``cci task info <name>``, where ``<name>`` is the name of the task; and run a task with ``cci task run <name> --org <org>``, where ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against. For example, the ``run_tests`` task executes Apex unit tests. If you have an org called ``dev``, you can run this task against this org with the command ``cci task run run_tests --org dev``.
 
-Many operations that you'll undertake with CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and other flows!) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow. 
+Many operations that you'll undertake with CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and other flows!) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow.
 
 Find the list of flows available in a project by running ``cci flow list``. Learn more about a flow and the tasks it contains with ``cci flow info <name>``, where ``<name>`` is the name of the flow, and run a flow with ``cci flow run <name> --org <org>``, where ``<name>`` is the name of the flow and ``<org>`` is the org you'd like to run it against.
 
@@ -122,7 +111,7 @@ Many of the most common flows you'll work with in CumulusCI are designed to buil
 * ``dev_org``: This is an unmanaged org that is designed for development use. This flow is typically used with an org whose configuration is ``dev`` or ``dev_namespaced``.
 * ``qa_org``: This is an unmanaged org that is designed for testing. This flow is typically used with an org whose configuration is ``qa``.
 * ``install_beta``: This is a managed org with the latest beta release installed, for projects that build managed packages. This flow is typically used with an org whose configuration is ``beta``.
-* ``install_prod``: This is a managed org with the latest release installed, for projects that build managed packages. 
-* ``regression_org``: This is a managed org that starts with the latest release installed and is then upgraded to the latest beta to simulate a subscriber upgrade, for projects that build managed packages. This flow is typically used with an org whose configuration is ``release``.
+* ``install_prod``: This is a managed org with the latest release installed, for projects that build managed packages.
+* ``regression_org``: This is a managed org that starts with the latest release installed and is then upgraded to the latest beta to simulate a subscriber upgrade for projects that build managed packages. This flow is typically used with an org whose configuration is ``release``.
 
-CumulusCI derives the library of flows and tasks available for any project by combining its internal, out-of-the-box library with the customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way out-of-the-box tasks behave, and extend, combine, and modify out-of-the-box flows to suit the specific needs of the project. We cover customization in depth in TODO: reference "Customizing CumulusCI"
+CumulusCI derives the library of flows and tasks available for any project by combining its internal standard library with your customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way featured tasks behave, and extend, combine, and modify featured flows to suit the specific needs of the project. We cover customization in depth in [TODO: reference Customizing CumulusCI].
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -6,22 +6,22 @@ Let's review some important concepts when building and testing features using Cu
 Packages
 --------
 
-If you've come this far in your search for customization tools, we're guessing you're familiar with packages as well as their role in customizing a Salesforce org. However, because most everything that you build in CumulusCI is deployed via packages, it's essential to review the basics of packages.
+If you've come this far in your search for customization tools, we're guessing you're familiar with packages as well as their role in customizing a Salesforce org. However, because most everything that you build in CumulusCI is deployed via packages, it's essential to review the basics once more.
 
-A **package** is a container for something as small as an individual component or as large as a sophisticated application. After creating a package, you can distribute it to other Salesforce users and organizations, including those outside your company.
+A *package* is a container for something as small as an individual component or as large as a sophisticated application. After creating a package, you can distribute it to other Salesforce users and organizations, including those outside your company.
 
-**Unmanaged packages** are typically used to distribute open-source (non-proprietary) features or application templates to provide developers with the basic building blocks for an application. After the components are installed from an unmanaged package in a specific org, it's what's known as an **org implementation**. These freshly installed components can be edited by the owners of the implementation. The developer who created and uploaded the unmanaged package has no control over the installed components, and can't change or upgrade them.
+*Unmanaged packages* are typically used to distribute open-source (non-proprietary) features or application templates to provide developers with the basic building blocks for an application. After the components are installed from an unmanaged package in a specific org, it's what's known as an *org implementation*. These freshly installed components can be edited by the owners of the implementation. The developer who created and uploaded the unmanaged package has no control over the installed components, and can't change or upgrade them.
  
-**Managed packages** are typically used by Salesforce partners to distribute and sell applications to customers. They are proprietary code that can be upgraded and deployed only by the developer that built them. To ensure seamless upgrades, we prevent certain destructive changes, such as deleting objects or fields.
+*Managed packages* are typically used by Salesforce partners to distribute and sell applications to customers. They are proprietary code that can be upgraded and deployed only by the developer that built them. To ensure seamless upgrades, we prevent certain destructive changes, such as deleting objects or fields.
 
-In CumulusCI, packages are built and deployed via **projects**.
+In CumulusCI, packages are built and deployed via projects.
 
 Projects
 --------
 
-When you work with CumulusCI, you do so inside a project. A project is an individual Git repository that contains both Salesforce metadata and CumulusCI automation (such as tasks and flows) that builds and releases the project. Most importantly, a project can build one--and only one(!)--package in its repository. This one-to-one relationship between project and package is what allows version control to track any changes your team makes to the repository.
+When you work with CumulusCI, you do so inside a *project*. A project is an individual Git repository that contains both Salesforce metadata and CumulusCI automation (such as tasks and flows) that builds and releases the project. Most importantly, a project can build one--and only one(!)--package in its repository. This one-to-one relationship between project and package is what allows version control to track any changes your team makes to the repository.
 
-It's important to note that a project doesn't have to contain a package. For example, a project can deliver unpackaged metadata, deliver automation but no metadata at all, or provide test data for QA. A project can constitute the entirety of a product offered to customers, or be just one of multiple projects that combine to form a complete product.
+It's important to note that a project doesn't have to contain a package. For example, a project can deliver unpackaged metadata, deliver automation but no metadata at all, or provide test data for QA. A project can contain the entirety of a product offered to customers, or be just one of multiple projects that combine to form a complete product.
 
 To sum up, although a project doesn't require a package, a package requires a project to be built and deployed.
 
@@ -48,7 +48,7 @@ Learn more about customizing CumulusCI automation in the [TODO: link Customizati
 ``force-app`` (or ``src``)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The main body of the project's code and metadata lives in the default package directory, which is the ``force-app`` directory for Salesforce DX-format projects and the ``src`` directory for Metadata API-format projects. **force-app** defines what's included when you release a managed package from your CumulusCI project. (Or when you release an unlocked package, or if you're not releasing a package at all but running the ``deploy`` task to get the metadata into an org in unmanaged form.)
+The main body of the project's code and metadata lives in the default package directory, which is the ``force-app`` directory for Salesforce DX-format projects and the ``src`` directory for Metadata API-format projects. ``force-app`` defines what's included when you release a managed package from your CumulusCI project. (Or when you release an unlocked package; or if you're not releasing a package at all but running the ``deploy`` task to get the metadata into an org in unmanaged form.)
 
 ``orgs`` directory
 ^^^^^^^^^^^^^^^^^^
@@ -58,7 +58,7 @@ The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org
 ``datasets``
 ^^^^^^^^^^^^
 
-Each project can have one or more **datasets**: on-disk representations of record data that can be inserted into Salesforce orgs, and that can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
+Each project can have one or more ``datasets``: on-disk representations of record data that can be inserted into Salesforce orgs, and that can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
 
 ``robot``
 ^^^^^^^^^
@@ -70,7 +70,7 @@ While Robot Framework is used primarily for browser automation testing, it can a
 ``unpackaged`` metadata
 ^^^^^^^^^^^^^^^^^^^^^^^
 
-As we touched upon earlier, project doesn't just encompass the contents of a managed package or a single deployment. It also includes **unpackaged metadata**: extra bundles of Salesforce metadata that further tailor an org or complete the product.
+As we touched upon earlier, project doesn't just encompass the contents of a managed package or a single deployment. It also includes *unpackaged metadata*: extra bundles of Salesforce metadata that further tailor an org or complete the product.
 
 In a CumulusCI project, all unpackaged metadata is stored in subdirectories within the ``unpackaged`` directory. Unpackaged metadata plays multiple roles, including preparing an org for installing packages, adding more customization after the package or application is deployed, and customizing specific orgs that are used in the product's development process.
 
@@ -100,7 +100,7 @@ Global services are easy to use and share. We recommend that you use them as muc
 Tasks and Flows
 ---------------
 
-CumulusCI uses a framework of **tasks** and **flows** to organize the automation that is available to each project.
+CumulusCI uses a framework of *tasks* and *flows* to organize the automation that is available to each project.
 
 Tasks are units of automation. A task can perform a deployment, load a dataset, retrieve data from an org, install a managed package, or do many other things. CumulusCI ships with scores of tasks out of the box.
 
--- a/docs/concepts.rst
+++ b/docs/concepts.rst
@@ -3,6 +3,8 @@ Key Concepts
 
 Let's review some important concepts when building and testing features using CumulusCI.
 
+
+
 Packages
 --------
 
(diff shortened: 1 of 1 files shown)


docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
-Datasets
+force-app (or src)
+^^^^^^^^^^^^^^^^^^
+
+The main body of the project's code and metadata lives in the ``force-app`` directory for Salesforce DX-format projects and ``src`` for Metadata API-format projects. For managed package projects, only this metadata is part of the package. Projects have the vast majority of their components stored here.
+
+orgs Directory
+^^^^^^^^^^^^^^
+
+The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in [TODO: link Scratch Org Management].
```
justinvholt, line 60: "We cover scratch org management in depth"

Haven't mentioned scratch orgs at this point, and there's not enough detail here (yet) that discusses the relationship between orgs directory and scratch orgs. Should we elaborate?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
+The ``cumulusci.yml`` file defines a project's automation in CumulusCI. It contains all the customizations and configurations that pertain to your project's lifecycle. It can encompass everything from customizing the shapes of scratch orgs to configuring tasks and flows.
 
-The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in TODO: link Scratch Org Management. 
+Learn more about customizing CumulusCI automation in the [TODO: link Customization section].
 
-Datasets
+force-app (or src)
+^^^^^^^^^^^^^^^^^^
+
+The main body of the project's code and metadata lives in the ``force-app`` directory for Salesforce DX-format projects and ``src`` for Metadata API-format projects. For managed package projects, only this metadata is part of the package. Projects have the vast majority of their components stored here.
```
justinvholt, line 55: "For managed package projects, only this metadata is part of the package."

A bit vague. Should we clarify further?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
+Each project may have one or more **datasets**: on-disk representations of record data that can be inserted into Salesforce orgs, and which can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
 
-Robot Framework Browser Automation
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Robot
+^^^^^
 
-CumulusCI integrates Robot Framework browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the resources and test suites for the project's Robot Framework test suites. New CumulusCI projects start with a simple Robot test case that creates a Contact record. 
+Robot Framework provides browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the resources and test suites for the project's Robot Framework test suites. New CumulusCI projects start with a simple Robot test case that creates a Contact record.
```
justinvholt, line 70: "browser automation for end-to-end testing"

Does this need further clarification? For instance, what is browser automation? What is its relationship to CumulusCI?

justinvholt, line 70: "stores the resources and test suites for the project's Robot Framework test suites"

Redundant to use "test suites" twice? Can we drop the second one and just end with "Robot Framework"?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
 
-While Robot Framework is used primarily for browser automation testing, in a small number of situations, Robot Framework may also be harnessed to help configure orgs where other strategies and APIs are insufficient.
+While Robot Framework is used primarily for browser automation testing, it can also be harnessed to help configure orgs where other strategies and APIs are insufficient. [TODO LINK TO EXPLAIN THIS FURTHER?]
```
justinvholt, line 72: TODO Link to explain this further?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
+In a CumulusCI project, all unpackaged metadata is stored in subdirectories within the ``unpackaged`` directory. Unpackaged metadata plays multiple roles, including preparing an org for installing packages, adding more customization after the package or application is deployed, and customizing specific orgs that are used in the product's development process.
 
-Learn more about managing unpackaged metadata in TODO: link Managing unpackaged configuration.
+Learn more about managing unpackaged metadata in [TODO: link Managing unpackaged configuration].
 
 Project Orgs & Services
 -----------------------
 
-Orgs and services are external, authenticated resources that each project uses. CumulusCI makes it easy to connect orgs, as well as services like a GitHub account or a MetaDeploy instance, to a single project or to use them across projects. 
+Orgs and services are external, authenticated resources that each project uses. CumulusCI makes it easy to connect orgs, as well as services like GitHub or MetaDeploy, to a single project, or to use them across many projects.
```
justinvholt, line 86: "external, authenticated resources"

Is an org an external resource? Usually? Always? Sometimes?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
-Orgs
-^^^^
-
-Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain, making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be `shared across multiple projects <TODO: link to persistent orgs section>`_.
-
-Configuring orgs in CumulusCI is powerful, but comes with some complexity. To review all of the details, read the sections TODO: link "Scratch org environments" and TODO: link persistent org section.
-
-Services
-^^^^^^^^
+Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain, making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be shared across multiple projects.
```
justinvholt, line 88: "predefined scratch org configurations"

Does reader automatically understand what this means? Further clarification?

justinvholt, line 88: "Orgs can also be shared across multiple projects."

Why is this useful? How does CumulusCI assist in making this happen?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
+Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. The ``command cci service`` list shows you which services are connected in the context of the current project.
 
-Some services can be connected at the project level, which means that they're scoped to a single project and aren't shared. We recommend primarily using global services because they're easier to use, but you may encounter scenarios where, for example, you need to use a specific Dev Hub for one and only one project. Connecting that service with ``cci service connect devhub --project`` supports that use case.
+Services can be connected at the project level, which means that they're scoped to a single project and cannot be shared. Global services are easy to use and share and, therefore, we recommend you rely on them for the most part. However, when you encounter a scenario where, for example, you need to use a specific Dev Hub for one - and only one(!) - project, you can simply connect to that service by way of the ``cci service connect devhub --project`` command.
```
justinvholt, line 94: "cannot be shared"

It was originally "aren't shared". Just wanted to confirm that my wording is OK.

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
+You can review the tasks available in a project by running ``cci task list``; learn more about a task and how to configure its options with ``cci task info <name>``, where ``<name>`` is the name of the task; and run a task with ``cci task run <name> --org <org>``, where ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against. For example, the ``run_tests`` task executes Apex unit tests. If you have an org called ``dev``, you can run this task against this org with the command ``cci task run run_tests --org dev``.
 
-Many operations that you'll undertake with CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and other flows!) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow. 
+Many operations that you'll undertake with CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and other flows!) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow.
```
justinvholt, line 105: "specific workflow"

We might need to give an example of what this means, or turn the next paragraph into a more streamlined example(?).

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
 
-You can review the tasks available in a project by running ``cci task list``, learn more about a task and how to configure its options with ``cci task info <name>``, where ``<name>`` is the name of the task, and run a task with ``cci task run <name> --org <org>``, where ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against. For example, the ``run_tests`` task executes Apex unit tests. If you have an org called ``dev``, you can run this task against this org with the command ``cci task run run_tests --org dev``.
+You can review the tasks available in a project by running ``cci task list``; learn more about a task and how to configure its options with ``cci task info <name>``, where ``<name>`` is the name of the task; and run a task with ``cci task run <name> --org <org>``, where ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against. For example, the ``run_tests`` task executes Apex unit tests. If you have an org called ``dev``, you can run this task against this org with the command ``cci task run run_tests --org dev``.
```
justinvholt, line 103: The paragraph is fine as is, but it also starts getting a little dense as one block of text. Would it be better to separate each task mentioned as its own line?

docs/concepts.rst
```diff
@@ -122,7 +111,7 @@ Many of the most common flows you'll work with in CumulusCI are designed to buil
 * ``dev_org``: This is an unmanaged org that is designed for development use. This flow is typically used with an org whose configuration is ``dev`` or ``dev_namespaced``.
```
justinvholt, line 111: "Here’s a few of the most common flows that build orgs:"

Just noticed that this is a bit confusing when I read it aloud. So to confirm: this is the name of a flow itself, not just the org? Both? Should we say "this flow creates a(n) _______ org"?

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
+The root directory of your CumulusCI project. Because each project is linked to a single GitHub repository, you can always be confident of the scope of the actions you're taking. And thanks to the keychain, CumulusCI knows which project you are working on by the current working directory of your shell. 
 
-When you work with CumulusCI in a project, the contents of your ``cumulusci.yml`` are merged with the ``cumulusci.yml`` that comes with CumulusCI and defines out-of-the-box functionality. From time to time we'll refer to that as the standard library. Learn more about customizing CumulusCI automation in TODO: link Customization section.
+[NOTE: Save yourself a headache by making sure you're in the correct repository for your project before running project-specific commands; otherwise, your project will produce an error. (This tends to be one of the biggest troubleshooting issues that developers run into with CumulusCI. Check for this first and you might save yourself an extra trip to this guide.)]
```
justinvholt, line 41: Helpful hint? Annotation?

docs/concepts.rst
davisagli: What do you think needs clarification? I think the point of the sentence is to contrast it with other sets of metadata in the unpackaged folder.

docs/concepts.rst
davisagli: I wonder if we should move the info on project structure below the Project Orgs and Services and Tasks and Flows sections, since the project structure includes references to concepts introduced in those sections.

docs/concepts.rst
davisagli: I don't think we need much detail here. The important thing is to point out that CumulusCI includes a thing called Robot Framework which can be used to create automated browser tests, and link to the later section of the docs which covers it in detail.

davisagli: I would edit it down to simply "which stores the project's Robot Framework test suites"

docs/concepts.rst
davisagli: We're going to fill in the links once the PRs for individual sections are merged so that there's a place to link to.

docs/concepts.rst
davisagli: Yes, it's always on a Salesforce server, external to the machine where you are running CumulusCI.

docs/concepts.rst
davisagli: Probably not, but we explain it more in the section on scratch orgs.

davisagli: Maybe I'm working on several different projects and I want to deploy all of them into the same persistent org in the end. I can connect that persistent org as a global org, instead of the default of being specific to one project. Then I can deploy things to it from any cci project.

It's more of an advanced use case and I don't think we need to do more than mention it.

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
-^^^^^^^^
+Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain, making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be shared across multiple projects.
 
-Services represent external resources used by CumulusCI automation, such as access to a GitHub account or a MetaDeploy instance. Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. 
+Configuring orgs in CumulusCI is powerful, but comes with some complexity. To review the details, read the sections [TODO: link Scratch org environments] and [TODO: link persistent org section].
 
-The command ``cci service list`` shows you which services are connected in the context of the current project.
+Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. The ``command cci service`` list shows you which services are connected in the context of the current project.
```
davisagli, line 92: I'm not sure why you removed the Services subheading and definition of what a service is. I know it sort of repeats the first paragraph under Orgs and Services, but I think it's important to make it clear that Services are a separate concept from Orgs.

docs/concepts.rst
davisagli: sure, I think they're equivalent

docs/concepts.rst
davisagli: These are the names of flows. When running the flow you also have to specify which org config to run it against.

docs/concepts.rst
```diff
@@ -122,7 +111,7 @@ Many of the most common flows you'll work with in CumulusCI are designed to buil
...
+* ``regression_org``: This is a managed org that starts with the latest release installed and is then upgraded to the latest beta to simulate a subscriber upgrade for projects that build managed packages. This flow is typically used with an org whose configuration is ``release``.
 
-CumulusCI derives the library of flows and tasks available for any project by combining its internal, out-of-the-box library with the customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way out-of-the-box tasks behave, and extend, combine, and modify out-of-the-box flows to suit the specific needs of the project. We cover customization in depth in TODO: reference "Customizing CumulusCI"
+CumulusCI derives the library of flows and tasks available for any project by combining its internal standard library with your customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way featured tasks behave, and extend, combine, and modify featured flows to suit the specific needs of the project. We cover customization in depth in [TODO: reference Customizing CumulusCI].
```
davisagli, line 117: I'm not understanding your use of "featured" here.

docs/concepts.rst
Br4nd0R: Dang, I didn't know we had all of those options :)

docs/concepts.rst
```diff
@@ -1,119 +1,108 @@
...
-
-CumulusCI knows which project you are working on by the current working directory of your shell. Make sure to change directories inside your project before you run project-specific commands.
-
-In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We'll show you how to get set up with a new or existing CumulusCI project in the (TODO link) Get Started section.
+It's important to note that a project may not even contain a package; for example, it can be built to deliver functionality to records, or test data for QA. A project may constitute the entirety of a product made for release, or may be one of multiple projects banded together to make a product (think NPSP). Basically, whatever feature in your product that needs specialized Salesforce metadata and CumulusCI automation in order to be built and deployed, *especially* packages (we cannot stress this enough!), requires its own unique project.
```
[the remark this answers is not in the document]
justinvholt: @davisagli I'll keep that in mind for a second run-through. There was the idea of introducing Product Delivery Model with NPSP as an example in the Introduction, but this aside can be deleted.

docs/concepts.rst
justinvholt: @davisagli Ah. See, I was interpreting external as meaning outside a Salesforce project, not the machine itself.

docs/concepts.rst
justinvholt: I figured, but I added this particular TODO, so just clarifying.

docs/concepts.rst
justinvholt: @davisagli I think I need a better understanding then of where metadata is stored. To start: A managed package also has an unpackaged folder?

docs/concepts.rst
justinvholt: I believe I deleted it because it was already mentioned in the first paragraph. Will fix. (There was never a separate subheading for Org and Services but might be prudent to add one for each.)

docs/concepts.rst
davisagli: @justinvholt It can. There's the primary package of metadata in force-app which is going to be released as a managed package, and then there can be other sets of metadata in various folders under `/unpackaged`. These can serve a variety of different purposes such as:
- occasionally it's necessary to deploy some metadata in order to enable a feature that is a prerequisite for installing a managed package
- there are some types of metadata that aren't supported in managed packages, but can still be deployed as unmanaged metadata
- there can be metadata which only needs to be deployed for particular configurations, such as some additional fields that are used for QA purposes but not wanted in other types of orgs

docs/concepts.rst
justinvholt: @davisagli I don't like _out-of-the-box_, thought _featured_ would be a better substitute. I can change it back, or is there a better word?

davisagli: out-of-the-box and featured don't seem like synonyms to me. The point we're trying to make is that a project can either use the standard tasks and flows as is, or customize them, or add new ones. Any of the standard tasks and flows can be customized, not just "featured" ones.

justinvholt: Then we won't use either, simply go with tasks.

docs/concepts.rst
justinvholt: So is it that the unmanaged metadata that's part of a package is only used as a means to deploy/install the managed package but not "officially" part of (supported by?) the managed package once it's installed? (Does this question make sense?)

Is there a better way to convey how integral force-app folder/this metadata is to a managed package?

docs/concepts.rst
justinvholt: So would it be better to say "This flow builds/creates/heroically unleashes a ______ org blah blah blah"?

docs/concepts.rst
davisagli: Unmanaged metadata is, by definition, not part of the managed package, but it can still be an important part of delivering a full product to customers. The end goal is not to install a package; the end goal is to deliver a fully configured product. Sometimes doing that requires deploying unmanaged metadata before or after the managed package is installed.

> Is there a better way to convey how integral force-app folder/this metadata is to a managed package?

The default package directory (usually called force-app) defines what will be included when you release a managed package from your cci project (or when you release an unlocked package, if you're doing that, or if you are not releasing the package at all but running the `deploy` task to get the metadata into an org in unmanaged form). It's the right place to put metadata except for when there's a reason to put it somewhere else (like one of the reasons I listed above).

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
 Key Concepts
 ============
 
-The Product Delivery Model
------------------------------
-CumulusCI extends the concepts of the `Org Development Model <https://trailhead.salesforce.com/en/content/learn/modules/org-development-model>`_ and the `Package Development Model <https://trailhead.salesforce.com/en/content/learn/modules/sfdx_dev_model>`_ to form a new model, which we call the Product Delivery Model.
+Key concept. Developer jargon. 10-cent words that sound like a technical writer patting himself on the back. (Hi there!) We imagine your cursor subconsciously drifting its way to the hyperlink for the next chapter already. BUT in an effort to give you the tools to build and test customized features with CumulusCI, it's only right that we demystify the process by explaining the *how* and the *why* in order to help you achieve your goal of seamless automation.
```
lianeyee, line 4: I see that you're trying to ease into this topic but this approach draws too much attention to the writer's voice. There's a time and a place for that; when content is dense, it's reassuring to hear the voice of someone knowledgeable right by your side. Seeing as this content is pretty high level and general, consider a more straightforward approach. "Let's review some important concepts when building and testing features using CumulusCI." Also be vigilant for creeping jargon like "seamless." =)

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-The Product Delivery Model focuses on the customer experience, not on the technical artifacts you're delivering, and focuses on making it possible to deliver a first-class, fully configured customer experience – no matter how complex the product might be. When building a product, there can be detailed technical considerations for whether a specific component is best distributed within a package, as additional unpackaged metadata, or as sophisticated setup automation that runs before or after a package is installed. And, of course, a product may include not one, but several packages, which must be installed in sequence. CumulusCI makes it possible to deliver a cohesive product that makes use of these different techniques.
+Don't worry, we'll make this as painless as possible.
```
lianeyee, line 6: If you go with the suggestion up top, this line isn't needed.

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
-
-The Nonprofit Success Pack and the Product Delivery Model
-++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-
-The Nonprofit Success Pack, or NPSP, is one of Salesforce.org's flagship products. The NPSP consists of six managed packages, with complex dependency relationships between them, alongside unpackaged metadata and setup automation that helps create a usable NPSP org.
+Packages
+--------
 
-Salesforce.org delivers NPSP using CumulusCI and the Product Delivery Model because those tools are what make it possible for us to ship this large, complex, heterogeneous application efficiently to tens of thousands of customers — without a lengthy setup guide to be completed by the end user. CumulusCI automation seamlessly
+If you've come this far in your search of customization, we're guessing you're familiar with packages as well as their role in customizing a Salesforce org. However, because most everything that is built in CumulusCI is deployed via packages, it's essential to clarify once more:
```
lianeyee, line 11:
```suggestion
If you've come this far in your search for customization tools, we're guessing you're familiar with packages as well as their role in customizing a Salesforce org. However, because most everything that you build in CumulusCI is deployed via packages, it's essential to review the basics of packages.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
-* installs NPSP;
-* and performs final configuration to make the customer's experience better, like setting up Global Actions and delivering translations.
+A **package** is a container for something as small as an individual component or as large as a set of related apps. After creating a package, you can distribute it to other Salesforce users and organizations, including those outside your company.
 
-CumulusCI runs this automation throughout our development lifecycle, starting from feature branches in the hands of our developers and culminating in the delivery of new versions of the application to our users through MetaDeploy.
+**Unmanaged packages** are typically used to distribute open-source (non-proprietary) features or application templates to provide developers with the basic building blocks for an application. Once the components are installed from an unmanaged package into a specific org, it's what's known as an org implementation. These freshly installed components can be edited as seen fit by the owners of this implementation. The developer who created and uploaded the unmanaged package has no control over the installed components, and can't change or upgrade them.
```
lianeyee, line 15:
```suggestion
**Unmanaged packages** are typically used to distribute open-source (non-proprietary) features or application templates to provide developers with the basic building blocks for an application. After the components are installed from an unmanaged package in a specific org, it's what's known as an _org implementation_. These freshly installed components can be edited by the owners of the implementation. The developer who created and uploaded the unmanaged package has no control over the installed components, and can't change or upgrade them.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+ 
+**Managed packages** are typically used by Salesforce partners to distribute and sell applications to customers. They are proprietary code that can only be upgraded and deployed by the developer that built them. To ensure seamless upgrades, certain destructive changes, like removing objects or fields, cannot be performed.
```
lianeyee, line 17:
```suggestion
**Managed packages** are typically used by Salesforce partners to distribute and sell applications to customers. They are proprietary code that can be upgraded and deployed only by the developer that built them. To ensure seamless upgrades, we prevent certain destructive changes, such as deleting objects or fields.
```
Style point on use of "only": https://confluence.internal.salesforce.com/display/DOCTEAM/only

I'm also noticing a lot of passive voice. Edited the last sentence to make it more active. There might be a few other places to try to do the same (doesn't have to be in this round of edits).

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-Throughout the CumulusCI documentation, we'll have the Product Delivery Model in mind. 
+In CumulusCI, packages are also built, stored and deployed via **projects**.
 
 Projects
 --------
 
-When you work with CumulusCI, you do so inside a project. A project is a version control repository that contains Salesforce metadata as well as the CumulusCI automation that builds and releases the project. A project usually has a one-to-one relationship with a managed package, if building a package, or with an org implementation. A project may constitute the entirety of a product in the Product Delivery Model, or may be one of multiple projects making up a product.
+When you work with CumulusCI, you do so inside a project. A project is an individual git repository (think GitHub) that contains Salesforce metadata as well as the CumulusCI automation (e.g., tasks and flows) that builds and releases the project. Most importantly, a project can build one - and only one(!) - package in its repository. This one-to-one relationship between project and package is what allows version control to track any changes your team made to the repository with minimal collision.
```
lianeyee, line 24:
```suggestion
When you work with CumulusCI, you do so inside a project. A project is an individual Git repository (think GitHub) that contains both Salesforce metadata and CumulusCI automation (such as tasks and flows) that builds and releases the project. Most importantly, a project can build one—and only one(!)—package in its repository. This one-to-one relationship between project and package is what allows version control to track any changes your team makes to the repository.
```
I like this explanation! Consider whether you really need the GitHub parenthetical? It seems more distracting than illuminating.

docs/concepts.rst
davidmreed: In this case I am the perpetrator but my zeal for sesquipedalianism remains unbowed!

davidmreed: Current thinking is that this entire section moves into the Introduction, right?

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+--------
 
-Salesforce.org delivers NPSP using CumulusCI and the Product Delivery Model because those tools are what make it possible for us to ship this large, complex, heterogeneous application efficiently to tens of thousands of customers — without a lengthy setup guide to be completed by the end user. CumulusCI automation seamlessly
+If you've come this far in your search of customization, we're guessing you're familiar with packages as well as their role in customizing a Salesforce org. However, because most everything that is built in CumulusCI is deployed via packages, it's essential to clarify once more:
 
-* deploys NPSP's five dependent managed packages in the right sequence;
-* delivers unpackaged Record Types for the Account and Opportunity objects;
-* installs NPSP;
-* and performs final configuration to make the customer's experience better, like setting up Global Actions and delivering translations.
+A **package** is a container for something as small as an individual component or as large as a set of related apps. After creating a package, you can distribute it to other Salesforce users and organizations, including those outside your company.
```
davidmreed, line 13: I don't think this is wrong, but the ambitious phrasing may give the wrong impression. It would be unusual for a single managed package to include a set of related apps.

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
-
-CumulusCI knows which project you are working on by the current working directory of your shell. Make sure to change directories inside your project before you run project-specific commands.
-
-In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We'll show you how to get set up with a new or existing CumulusCI project in the (TODO link) Get Started section.
+It's important to note that a project may not even contain a package; for example, it can be built to deliver functionality to records, or test data for QA. A project may constitute the entirety of a product made for release, or may be one of multiple projects banded together to make a product. Basically, whatever feature in your product that needs specialized Salesforce metadata and CumulusCI automation in order to be built and deployed, *especially* packages (we cannot stress this enough!), requires its own unique project.
```
lianeyee, line 26:
```suggestion
A project doesn't have to contain a package. For example, it can deliver functionality to records or provide test data for QA. A project can constitute an entire product for release, or be just one of multiple projects banded together to make a product. Here's a rule of thumb for when a project _is_ required: whenever the thing that you're building and deploying requires specialized Salesforce metadata and CumulusCI automation, and *especially* packages. We can't stress this enough! 
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+Project Directory
+^^^^^^^^^^^^^^^^^
 
-Together with the core package, a CumulusCI project encompasses a variety of unpackaged application components, automation, tests, and other support material. We'll quickly survey the primary components of a CumulusCI project here. Each component is covered in more depth later in this guide.
+The root directory of your CumulusCI project. Because each project is linked to a single GitHub repository, CumulusCI knows which project you are working on by the current working directory of your shell. 
 
-Automation Configuration (``cumulusci.yml``)
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+.. tip:: Avoid headaches by making sure you're in the correct repository for your project before running project-specific commands; otherwise, your project will produce an error. (This tends to be one of the biggest troubleshooting issues that developers run into with CumulusCI. Check for this first and you might save yourself an extra trip to this guide.).
```
lianeyee, line 36:
```suggestion
.. tip:: Avoid headaches by making sure you're in the correct repository for your project before running project-specific commands. Otherwise, errors can happen in your project. This issue tends to be one of the biggest CumulusCI troubleshooting issues. Check your repo first and you might save yourself an extra trip to this guide.
```
Try to avoid future tense and avoid omitting a subject, like "This tends to . . ." Specify the subject, as suggested. I like the parenthetical thought but I like it better as a non-parenthetical. ;-)

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-``cumulusci.yml`` is the definition of the project's automation. It includes information about the project itself as well as customizations that define how scratch orgs are built and provide other important functionality to be used as part of the project's lifecycle.
+In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We'll show you how to get set up with a new or existing CumulusCI project in the [TODO link Get Started section].
```
lianeyee, line 38:
```suggestion
In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We cover how to get set up with a new or existing CumulusCI project in the [TODO link Get Started section].
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-When you work with CumulusCI in a project, the contents of your ``cumulusci.yml`` are merged with the ``cumulusci.yml`` that comes with CumulusCI and defines out-of-the-box functionality. From time to time we'll refer to that as the standard library. Learn more about customizing CumulusCI automation in TODO: link Customization section.
+cumulusci.yml
+^^^^^^^^^^^^^
 
-Project Source Code
-^^^^^^^^^^^^^^^^^^^
+The ``cumulusci.yml`` file defines a project's automation in CumulusCI. It contains all the customizations and configurations that pertain to your project's lifecycle. It can encompass everything from customizing the shapes of scratch orgs to configuring tasks and flows.
```
lianeyee, line 43:
```suggestion
The ``cumulusci.yml`` file defines a project's automation. It contains all the customizations and configurations that pertain to your project's lifecycle. It can encompass everything from customizing the shapes of scratch orgs to configuring tasks and flows.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-The main body of the project's code and metadata lives in the ``force-app`` directory for Salesforce DX-format projects and ``src`` for Metadata API-format projects. For managed package projects, only this metadata is part of the package. Most projects have the vast majority of their components stored here.
+Learn more about customizing CumulusCI automation in the [TODO: link Customization section].
 
-Orgs
-^^^^
+force-app (or src)
+^^^^^^^^^^^^^^^^^^
+
+The main body of the project's code and metadata lives in the default package directory, commonly shown as the ``force-app`` folder for Salesforce DX-format projects and the ``src`` folder for Metadata API-format projects. **force-app** defines what will be included when you release a managed package from your CumulusCI project (or when you release an unlocked package, or if you're not releasing the package at all but running the ``deploy`` task to get the metadata into an org in unmanaged form).
```
lianeyee, line 50:
```suggestion
The main body of the project's code and metadata lives in the default package directory, commonly shown as the ``force-app`` folder for Salesforce DX-format projects and the ``src`` folder for Metadata API-format projects. **force-app** defines what's included when you release a managed package from your CumulusCI project. Or, when you release an unlocked package, or if you're not releasing the package at all but running the ``deploy`` task to get the metadata into an org in unmanaged form.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in TODO: link Scratch Org Management. 
+The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in [TODO: link Scratch Org Management].
 
-Datasets
+datasets
 ^^^^^^^^
 
-Each project may have one or more *datasets*: on-disk representations of record data that can be inserted into Salesforce orgs, and which can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in TODO: link Automating Data Operations.
+Each project may have one or more **datasets**: on-disk representations of record data that can be inserted into Salesforce orgs, and which can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
```
lianeyee, line 60:
```suggestion
Each project can have one or more **datasets**: on-disk representations of record data that can be inserted into Salesforce orgs, and that can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+force-app (or src)
+^^^^^^^^^^^^^^^^^^
+
+The main body of the project's code and metadata lives in the default package directory, commonly shown as the ``force-app`` folder for Salesforce DX-format projects and the ``src`` folder for Metadata API-format projects. **force-app** defines what will be included when you release a managed package from your CumulusCI project (or when you release an unlocked package, or if you're not releasing the package at all but running the ``deploy`` task to get the metadata into an org in unmanaged form).
+
+orgs directory
+^^^^^^^^^^^^^^
 
-The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in TODO: link Scratch Org Management. 
+The ``.json`` files found in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in [TODO: link Scratch Org Management].
```
lianeyee, line 55:
```suggestion
The ``.json`` files in the ``orgs`` directory define the Salesforce DX org configurations that are available to the project. We cover scratch org management in depth in [TODO: link Scratch Org Management].
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+Each project may have one or more **datasets**: on-disk representations of record data that can be inserted into Salesforce orgs, and which can also be modified and re-captured during the evolution of the project. Datasets are stored in the ``datasets`` directory. Learn more about datasets in [TODO: link Automating Data Operations].
 
-Robot Framework Browser Automation
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+robot
+^^^^^
 
-CumulusCI integrates Robot Framework browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the resources and test suites for the project's Robot Framework test suites. New CumulusCI projects start with a simple Robot test case that creates a Contact record. 
+Robot Framework provides browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the project's Robot Framework test suites. New CumulusCI projects start with a simple Robot test case that creates a Contact record.
```
lianeyee, line 65:
```suggestion
Robot Framework provides browser automation for end-to-end testing. Each project contains a ``robot`` directory, which stores the project's Robot Framework test suites. New projects start with a simple Robot test case that creates a Contact record.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-While Robot Framework is used primarily for browser automation testing, in a small number of situations, Robot Framework may also be harnessed to help configure orgs where other strategies and APIs are insufficient.
+While Robot Framework is used primarily for browser automation testing, it can also be harnessed to help configure orgs where other strategies and APIs are insufficient. [TODO LINK TO EXPLAIN THIS FURTHER?]
```
lianeyee, line 67:
```suggestion
While Robot Framework is primarily for browser automation testing, it's also useful for configuring orgs when other strategies and APIs are insufficient. [TODO LINK TO EXPLAIN THIS FURTHER?]
```
Is it more common to say "automated browser testing" than "browser automation testing"? I didn't suggest the change because I'm not sure.

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-Unpackaged Metadata
+unpackaged metadata
 ^^^^^^^^^^^^^^^^^^^
 
-In the Product Delivery Model, a product doesn't just encompass the contents of a managed package or a single deployment. It also includes *unpackaged metadata*: extra bundles of Salesforce metadata that further tailor an org or complete the product.
+A product doesn't just encompass the contents of a managed package or a single deployment. It also includes **unpackaged metadata**: extra bundles of Salesforce metadata that further tailor an org or complete the product.
```
lianeyee, line 72: Should this be moved up higher? We talk about metadata pretty early on in Key Concepts, and this tidbit seems helpful to know before it's first mentioned.

lianeyee, line 72:
```suggestion
A product doesn't just include the contents of a managed package or a single deployment. It also includes **unpackaged metadata**: extra bundles of Salesforce metadata that further tailor an org or complete the product.
```
Slightly simpler.

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+Orgs and services are external, authenticated resources that each project uses. CumulusCI makes it easy to connect orgs and services to a single project, or to use them across many projects.
 
 Orgs
 ^^^^
 
-Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain, making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be `shared across multiple projects <TODO: link to persistent orgs section>`_.
+Each project has its own set of orgs, including active scratch orgs, persistent orgs like a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain (via a unique identifier of characters), making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can also be shared across multiple projects.
```
lianeyee, line 86:
```suggestion
Each project has its own set of orgs, including active scratch orgs, persistent orgs such as a production or packaging org, and predefined scratch org configurations. CumulusCI securely stores org authentication information in its keychain (via a unique ID), making it easy to access connected orgs at any time. The ``cci org list`` command shows all of the orgs connected to a project. Orgs can be shared across multiple projects.
```
"unique identifier of characters" sounds a bit awkward. Trim to "unique ID" unless it's important to specify that the ID is character-based?

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-Configuring orgs in CumulusCI is powerful, but comes with some complexity. To review all of the details, read the sections TODO: link "Scratch org environments" and TODO: link persistent org section.
+Configuring orgs in CumulusCI is powerful, but comes with some complexity. To review the details, read the sections [TODO: link Scratch org environments] and [TODO: link persistent org section].
```
lianeyee, line 88:
```suggestion
Configuring orgs in CumulusCI is powerful, but comes with some complexity. For details, see [TODO: link Scratch org environments] and [TODO: link persistent org section].
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
 Services
 ^^^^^^^^
 
-Services represent external resources used by CumulusCI automation, such as access to a GitHub account or a MetaDeploy instance. Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. 
+Services represent external resources used by CumulusCI automation, such as access to a GitHub account or a MetaDeploy instance. Services are usually, but not always, connected to CumulusCI across projects as part of the global keychain. The command ``cci service list`` shows you which services are connected in the context of the current project.
 
-The command ``cci service list`` shows you which services are connected in the context of the current project.
+Global services are easy to use and share and, therefore, we recommend you rely on them for the most part. However, services can also be connected at the project level, which means that they're scoped to a single project and cannot be shared.
```
lianeyee, line 95:
```suggestion
Global services are easy to use and share, and we recommend that you use them when possible. However, services can also be connected at the project level, which means that they're scoped to a single project and can't be shared.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-Some services can be connected at the project level, which means that they're scoped to a single project and aren't shared. We recommend primarily using global services because they're easier to use, but you may encounter scenarios where, for example, you need to use a specific Dev Hub for one and only one project. Connecting that service with ``cci service connect devhub --project`` supports that use case.
+        Example: If you need to use a specific Dev Hub for one - and only one(!) - project, you can connect to that service with the command ``cci service connect devhub --project``.
```
lianeyee, line 97: Replace with em dashes.

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
 Tasks and Flows
 ---------------
 
-CumulusCI uses a framework of *tasks* and *flows* to organize the automation that is available to each project.
+CumulusCI uses a framework of **tasks** and **flows** to organize the automation that is available to each project.
+
+Tasks are units of automation. A task could perform a deployment, load a dataset, retrieve material from an org, install a managed package, or undertake many other activities. CumulusCI ships with scores of tasks out of the box.
```
lianeyee, line 104:
```suggestion
Tasks are units of automation. A task can perform a deployment, load a dataset, retrieve data from an org, install a managed package, or do many other things. CumulusCI ships with scores of tasks out of the box.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+
+Popular tasks:
```
lianeyee, line 106:
```suggestion
Popular task commands include:
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+
+* ``cci task list``: Review the tasks available in a project.
+* ``cci task info <name>``: Learn more about a task and how to configure its options; ``<name>`` is the name of the task.
+* ``cci task run <name> --org <org>``: Run a task; ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against.
```
lianeyee, line 110:
```suggestion
* ``cci task run <name> --org <org>``: Run a task; ``<name>`` is the name of the task and ``<org>`` is the org to run it against.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+        Example: ``run_tests`` executes Apex unit tests. If you have an org called ``dev``, you can run this task against it with the command ``cci task run run_tests --org dev``
 
-You can review the tasks available in a project by running ``cci task list``, learn more about a task and how to configure its options with ``cci task info <name>``, where ``<name>`` is the name of the task, and run a task with ``cci task run <name> --org <org>``, where ``<name>`` is the name of the task and ``<org>`` is the org you'd like to run it against. For example, the ``run_tests`` task executes Apex unit tests. If you have an org called ``dev``, you can run this task against this org with the command ``cci task run run_tests --org dev``.
+Many operations that you'll undertake with CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and other flows!) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow.
```
lianeyee, line 114:
```suggestion
Many operations in CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and even other flows) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-Many operations that you'll undertake with CumulusCI, including creating new orgs, use flows. Flows are ordered sequences of tasks (and other flows!) that produce a cohesive outcome, such as an org that's configured to suit a specific workflow. 
+Popular flows:
```
lianeyee, line 116:
```suggestion
Popular flow commands include:
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-Find the list of flows available in a project by running ``cci flow list``. Learn more about a flow and the tasks it contains with ``cci flow info <name>``, where ``<name>`` is the name of the flow, and run a flow with ``cci flow run <name> --org <org>``, where ``<name>`` is the name of the flow and ``<org>`` is the org you'd like to run it against.
+* ``cci flow list``: Review the flows available in a project.
+* ``cci flow info <name>``: Learn more about a flow and the tasks it contains; ``<name>`` is the name of the flow
```
lianeyee, line 119:
```suggestion
* ``cci flow info <name>``: Learn more about a flow and the tasks it contains; ``<name>`` is the name of the flow.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+* ``cci flow run <name> --org <org>``: Run a flow; ``<name>`` is the name of the flow and ``<org>`` is the org you'd like to run it against.
```
lianeyee, line 120:
```suggestion
* ``cci flow run <name> --org <org>``: Run a flow; ``<name>`` is the name of the flow and ``<org>`` is the org you'd like to run it against.
```
```suggestion
* ``cci flow run <name> --org <org>``: Run a flow; ``<name>`` is the name of the flow and ``<org>`` is the org to run it against.
```

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
 Many of the most common flows you'll work with in CumulusCI are designed to build and configure specific orgs for you. Here's a few of the most common flows that build orgs:
```
lianeyee, line 122:
```suggestion
Many of the most common flows in CumulusCI are designed to build and configure specific orgs for you. Here's a few of the most common flows that build orgs.
```

docs/concepts.rst
lianeyee: If we go with "This flow builds/creates/etc." (which sounds good to me), consider starting the second sentences with "It's typically used..."

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+* ``regression_org``: This is a managed org that starts with the latest release installed and is then upgraded to the latest beta to simulate a subscriber upgrade for projects that build managed packages. This flow is typically used with an org whose configuration is ``release``.
 
-CumulusCI derives the library of flows and tasks available for any project by combining its internal, out-of-the-box library with the customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way out-of-the-box tasks behave, and extend, combine, and modify out-of-the-box flows to suit the specific needs of the project. We cover customization in depth in TODO: reference "Customizing CumulusCI"
+CumulusCI derives the library of tasks and flows available for any project by combining its internal standard library with your customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way tasks behave, and extend, combine, and modify flows to suit the specific needs of the project. We cover customization in depth in [TODO: reference Customizing CumulusCI].
```
lianeyee, line 130:
```suggestion
CumulusCI derives the library of tasks and flows available for any project by combining its internal standard library with your customizations in ``cumulusci.yml``. Customizations can add new tasks and flows, customize the way tasks behave, and extend, combine, and modify flows to better suit the project's needs. We cover customization in depth in [TODO: reference Customizing CumulusCI].
```

docs/concepts.rst
lianeyee, line 13: The typical Salesforce standard for indicating emphasis and indicating a newly introduced term is italics. We do sometimes use bold for emphasis, though not with any consistent pattern that I can detect. Flagging it for now as a formatting question for this whole doc set. I see David Reed also called out a formatting question about formatting code. We should discuss.

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
 Project Structure
 -----------------
 
-A CumulusCI project represents an application or a major component of an application on the Salesforce platform. Each project corresponds with a single managed package, or with an org-based implementation. In this documentation, we'll often speak about a package; if you're building a project that isn't packaged, please take this to refer to an unmanaged package within which your implementation is deployed.
+Project Directory
+^^^^^^^^^^^^^^^^^
 
-Together with the core package, a CumulusCI project encompasses a variety of unpackaged application components, automation, tests, and other support material. We'll quickly survey the primary components of a CumulusCI project here. Each component is covered in more depth later in this guide.
+The root directory of your CumulusCI project. Because each project is linked to a single GitHub repository, CumulusCI knows which project you are working on by the current working directory of your shell.
```
lianeyee, line 34: I'm not sure how to read "The root directory of your CumulusCI project." Either words are missing or it should instead be a heading like "Root Directory"?

docs/concepts.rst
justinvholt: Soooooooooooooo satisfying to fix this paragraph.

docs/concepts.rst
justinvholt: Let's definitely set a time to talk about this next week. There's enough of a mix of language in both sections that we might need to sort out what takes priority in getting mentioned first.

docs/concepts.rst
justinvholt: @davidmreed Do you mean _all_ of Key Concepts moving into Introduction? As it stands, Intro is gonna be DENSE what with the explanation of CumulusCI and the Product Delivery Model.

justinvholt: @lianeyee Fiiiiiiiiiine. Writer's voice quelled... for now.

For the record, I'm gonna blame @Br4nd0R for the phrase "seamless automation". (I personally prefer my automation to be rife with seams.)

justinvholt: Entire paragraph sent out to pasture; adios, darlings. (Sir Arthur Quiller-Couch would be proud.)

docs/concepts.rst
justinvholt: @lianeyee I'll remove the "Think GitHub" bit.

@davidmreed IIRC, "one-to-one relationship" came from the original text here. I discussed that in greater detail with @Br4nd0R, and that's where I came up with this description. If I'm not still not describing it accurately, what can we say that stresses the importance of the one-to-one relationship? (@davisagli, you might want to chime in here, too.)

docs/concepts.rst
justinvholt: @davidmreed What's the word for "a set of a related apps" anyway? A zynga? A benioff? 

Seriously, though, what could we use as an example for package size? "As large as a fully functional app"?

docs/concepts.rst
davidmreed: I mean the Product Delivery Model section moving.

justinvholt: @davidmreed Yes, PDM will go into Intro. (Part of the reason why it's gonna be a bit of a beast to overhaul.)

docs/concepts.rst
justinvholt: Will definitely strip out those mays/mights, but I need to advocate here for my original text because it adds more emphasis where needed. Saying "a project doesn't have to contain a package" is succinct, but it _is_ important to note because it's the reason for this paragraph's existence.  Because projects are not just vehicles for packages. (I made this assumption several times; @Br4nd0R corrected me each time.) Also, what's being a stressed enough (at the end) isn't the fact that a project only requires SF metadata and cci automation, which had just been clarified, but that **packages** definitely require a project because they're the bread and butter of cci.

My proposed edit:

"It's important to note that a project doesn't have to contain a package. For example, it can deliver functionality to records or provide test data for QA. A project can constitute an entire product for release, or be just one of multiple projects that together constitute a product. Essentially, if whatever you're building and deploying requires specialized Salesforce metadata and CumulusCI automation, *especially* packages (we cannot stress this enough!), it requires its own unique project."

(I also prefer "cannot" over "can't" in this instance because read aloud it emphasizes the stress.)

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
+**Unmanaged packages** are typically used to distribute open-source (non-proprietary) features or application templates to provide developers with the basic building blocks for an application. Once the components are installed from an unmanaged package into a specific org, it's what's known as an org implementation. These freshly installed components can be edited as seen fit by the owners of this implementation. The developer who created and uploaded the unmanaged package has no control over the installed components, and can't change or upgrade them.
+ 
+**Managed packages** are typically used by Salesforce partners to distribute and sell applications to customers. They are proprietary code that can only be upgraded and deployed by the developer that built them. To ensure seamless upgrades, certain destructive changes, like removing objects or fields, cannot be performed.
 
-Throughout the CumulusCI documentation, we'll have the Product Delivery Model in mind. 
+In CumulusCI, packages are also built, stored and deployed via **projects**.
```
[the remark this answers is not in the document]
justinvholt: @davidmreed Edit noted . . . but to clarify: packages are **not** stored via projects? Because I've been bragging about that to all my friends.

docs/concepts.rst
davidmreed: How about "... as small as an individual component or as large as a sophisticated application"?

docs/concepts.rst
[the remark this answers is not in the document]
davidmreed: As is so often the case... it's complicated.

Projects store the source code that becomes a package. But the package itself is an artifact that is only ever stored on a Salesforce server. It's as if you had Java code but the compiled JAR file was never accessible to you.

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: @lianeyee submitted her edit on this, too, so I'm tagging her here so she can have eyes on it. 

@davidmreed I really like your version because it's even more succinct. (I'll swap out the "may".) 

That said, "builds either one or zero packages" is true, but it's not a great selling point for cci. My reason for restructuring the entirety of Key Concepts is to center everything around packages, which is the primary motivator for someone to use CCI in the first place, correct? So I think it's prudent to emphasize that while a project's existence doesn't hinge on whether or not it contains a package, it _is_ the vehicle for a package.

I suggest we end this section with this:

"To sum up, although a project doesn't require a package, a package requires a project to be built and deployed."

docs/concepts.rst
```diff
@@ -1,128 +1,130 @@
...
 
-``cumulusci.yml`` is the definition of the project's automation. It includes information about the project itself as well as customizations that define how scratch orgs are built and provide other important functionality to be used as part of the project's lifecycle.
+In order to be used as a CumulusCI project, a directory must both be a Git repository and contain a ``cumulusci.yml`` configuration file. We'll show you how to get set up with a new or existing CumulusCI project in the [TODO link Get Started section].
 
-When you work with CumulusCI in a project, the contents of your ``cumulusci.yml`` are merged with the ``cumulusci.yml`` that comes with CumulusCI and defines out-of-the-box functionality. From time to time we'll refer to that as the standard library. Learn more about customizing CumulusCI automation in TODO: link Customization section.
+cumulusci.yml
```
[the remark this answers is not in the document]
justinvholt: @davidmreed Anything that is its own file name or command gets ``code style`` formatting. Are you talking about the phrase "standard library"? We can make that bold text.

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: You folks really hate describing what a keychain is. (Glick suggested I erase the entire paragraph for it in the Project Structure section.)

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: I am severely resistant to having it all lowercase. This is technically its own sentence, not an extension of one. 

Does it help to note that the reason why I put in colons was to save writing "allows user to" each time, which looks a bit goofy when repeated in a list. We can default to "reviews/learns/runs", but "learns more about a flow" or "runs a flow" doesn't hit my ear right when read aloud.

@lianeyee, thoughts?

docs/concepts.rst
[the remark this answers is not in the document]
davidmreed: No, I'm pointing at some of the headings, which are specific references to files or directories. They can be both headings and code style, I believe.

docs/concepts.rst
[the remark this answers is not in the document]
davidmreed: How about 

```suggestion
* ``cci flow run <name> --org <org>``: Run the flow ``<name>`` against the org ``<org>``.
```

docs/concepts.rst
justinvholt: Clean and succinct. Love it. 

That said, I prefer "project will produce an error" because that's _literally_ what will happen. The system will kick back an error, not just that an error can occur because you've saved this or that to the wrong folder.

docs/concepts.rst
justinvholt: I often use parentheticals as asides, meaning not vital but still letting the reader in the know. Again, I've written this entire page as if I'm giving a presentation aloud. So any emphasis, whether bold, italics, em-dashes or parentheses, gives a rhythm to the writing. I'm not saying this because the text is precious. But if we remove all of the parentheses, we're completely stripping away the "confidential tone" that I'm using to engage the reader.

Perhaps we can discuss with JV after the holidays and find a happy medium?

docs/concepts.rst
justinvholt: @davidmreed, @davisagli, @Br4nd0R: Thoughts?

docs/concepts.rst
justinvholt: Good point. We could place it between project directory and cumulusci.yml.

@davidmreed, @davisagli, @Br4nd0R: Thoughts?

justinvholt: Yeah, but then we're using "include" as the verb twice in a row. I think I chose encompass as shorthand for "that's not all that the project contains".

docs/concepts.rst
justinvholt: Everyone hates describing a keychain. (Although to be fair, it's pretty easy to intuit.) 

Glick and Reed both suggested deleting mention of this. It's gone. :)

docs/concepts.rst
justinvholt: "Global services are easy to use and share; we recommend that you use them as much as possible."

More conversational in tone. And I'm still gonna advocate for the "cannot" here for emphasis.

docs/concepts.rst
justinvholt: So it's looking like that's not feasible with Sphinx. And one small dash is too short. Can we use two small dashes instead? (Any command with two dashes is in that special ``code`` formatting.)

docs/concepts.rst
justinvholt: @davidmreed: Would "The project directory is the root of your CumulusCI project." be accurate? Would I say root directory even if we're using it twice in one sentence (yuck)? Root folder? Source?

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: Hrm. Fair point. I defaulted on using headers without code style. But I'm not beholden to that.

@lianeyee: Thoughts on whether we should make any giving any word in a header that has its own file/folder/command code style?

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: I left this in because it was in the original text, and an example never hurts (even if, yes, it's a bit detailed).

@Br4nd0R & @davisagli: Thoughts on excising this?

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: @davidmreed Was going to change it to "popular flow commands" per Yee's suggestion. Are we being _too_ particular with flow-related? If it's important to discern, I have no problem putting it in. 

And to confirm: If we go with flow-related, do we also say "task-related commands" with the suggestions in the previous section? Just want to be accurate.

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: It's more succinct, but now it might not also be as clear to the reader. (This is where we start getting into the debate of who are we writing this for, and how much do they know or can intuit.)

Can we get more of the team to chime in on this one? @davisagli? @Br4nd0R?

docs/concepts.rst
justinvholt: Yeah, wasn't sure if bold or italics was the way to go. I figured bold would stand out better because these are VIP key concepts.

docs/concepts.rst
davidmreed: Your phrase reads fine to me. I'd rather not use both the words "directory" and "folder" because they mean the same thing but are differently idiomatic depending on the user's background and operating system.

docs/concepts.rst
lianeyee: @justinvholt Looks like we've landed in a great succinct place with David Reed's edits below and your concluding "To sum up." Yay!

docs/concepts.rst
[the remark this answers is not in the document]
lianeyee: @justinvholt If we would format a word or phrase in body text with code style, we should do the same when it appears in heading text.

docs/concepts.rst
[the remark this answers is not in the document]
lianeyee: @justinvholt I agree with using the colon to economize on words. (Another option we sometimes use for such lists is an em dash, but I wouldn't here because a dash seems slightly less clear where the command stops and the separator starts.) 

For me, it is clearer to see the info spelled out in two separate parts. "Here's what the flow does. Here are the variables you specify." How about replacing the semicolons with periods? Seems simpler. And use ending periods on all. 

For extra succinctness, omit "you'd like."

docs/concepts.rst
lianeyee: Find a way to say that without using future tense and I'm sold. :)

docs/concepts.rst
lianeyee: I fear that I'm giving off "I don't like parentheticals" vibes when actually, I do like them! In this case, it was more about sentence length getting long, making for more work to parse. Breaking up the sentence with a harder stop feels like it gives me permission to set down that first chunk and then consider the next chunk. The parenthetical seems to ask for more cognitive holding of that first chunk. Anyway, this is getting really esoteric and nitpicky. Just note a potential readability issue that may or may not be a real issue. (And happy to get JV's take on it.)

docs/concepts.rst
lianeyee: Sometimes, echoing the verb can be a desirable, helpful device that ties the second idea closely to the first. When I first read it, "include" seemed like a simpler word choice than "encompass," as well as less cognitive load--I've already processed that verb the first time. But I respect your point that "encompass" might more precisely convey what you mean. In which case, never mind my edit! Just note that repetition isn't always a bad thing.

docs/concepts.rst
lianeyee: Sounds good! That semicolon seems fussy for such a short sentence. Replace with a period?

Roger on "cannot."

docs/concepts.rst
lianeyee: Yep, let's do two small en dashes, then. If we ever convert this content to easyDITA, we can do a quick search and replace.

docs/concepts.rst
lianeyee: Let's see if @jvarese has an opinion.

docs/concepts.rst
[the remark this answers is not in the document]
justinvholt: Changed it, although the formatting looks a bit odd. Might need to revisit with next commit.

docs/concepts.rst
justinvholt: I turned the parenthetical into its own sentence. It looks a bit odd, but I still don't think it has enough weight to drop the parentheses. Additionally, it's something Parker mentioned specifically, so I'm hesitant to drop it completely because it _is_ helpful to the reader with a certain skill level.

Tagging @jvarese so he knows what we're talking about next week.

docs/concepts.rst
jvarese: @justinvholt - as written, the above is definitely an italics situation (per house style). We use italics in the flow of a sentence. Another way to extra emphasize something is with a definition list, which bolds the first word/definition automatically. If this is part of a whole list of things, you could consider doing it that way too.

docs/concepts.rst
jvarese: @justinvholt - I like parenthetical, though Liane's right about us needing a more definite break to the flow. Sounds like making it its own sentence is a great solution. If the semi-colon comes back after that first clause, that might help too. (And I do admire your circumspection around the parenthetical -- a difficult thing to master.)

jvarese: @justinvholt - a PS: if you don't want the **forceapp** that starts the sentence to look weird, you could simply redundantly start the sentence with "The forceapp folder/directory defines . . . " I know we are always trying to write less words, not more, but this is one of those cases where need for clarity trumps that ambition.

docs/concepts.rst
justinvholt: Swapped 'em out!

docs/concepts.rst
davisagli: I agree that it's inaccurate to say that the 1:1 package-repository relationship is what makes version control possible.  One package per repository is more of a convention that CumulusCI assumes in a lot of places -- it's possible to imagine storing multiple packages in the same repository and setting up custom CumulusCI flows to work with them, but you would be "working against the grain" and not able to take advantage of the default flows that CumulusCI provides. So the main point is that you'll have a better experience with CumulusCI if you accept its premise that each package should go in its own repository.

justinvholt: @davisagli Would you advise we just leave the language as is in order to "push" people to stick with the 1:1 relationship?

docs/concepts.rst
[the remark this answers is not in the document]
davisagli: No strong opinion either way, but I'll note that we also cover this in a subsection of the topic on scratch orgs: https://cumulusci.readthedocs.io/en/feature-doc-reorg__scratch-org/scratch_orgs.html#use-a-non-default-dev-hub

docs/concepts.rst
davisagli: I think "automated browser testing" is clearer.

docs/concepts.rst
davisagli: @justinvholt how about:

When you work with CumulusCI, you do so inside a project. A project is an individual Git repository (think GitHub) that contains both Salesforce metadata and CumulusCI automation (such as tasks and flows) that builds and releases the project. If you are building multiple packages, we strongly recommend organizing each package as a separate project in its own repository. CumulusCI's standard automation assumes that there is one package per repository, so it will work the best if you follow this convention.

docs/concepts.rst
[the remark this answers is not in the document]
Br4nd0R: I like the more suggestion by @davidmreed. True, it may not be as clear, but including an example (as is present just after the 'popular tasks' list) should clear that up.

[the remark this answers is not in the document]
justinvholt: @Br4nd0R: What example would you like to insert?

PR #2264, opened 2020-12-16
2GP Dependency Resolution

davidmreed:
# Critical Changes

- CumulusCI now can resolve dependencies using second-generation packages (2GPs) for upstream projects. When a 2GP build runs on a release branch (starting with `prefix/NNN`, where `prefix` is the feature branch prefix and `NNN` is an integer), CumulusCI will look for a matching release branch in each upstream dependency and use a 2GP package build on that release branch if present, falling back to a 1GP beta release if not present.

# Changes

# Issues Closed


davidmreed:
Note: tests not yet complete.

davidmreed:
I diverged somewhat from the original design, where I was pushing additional complexity to allow for matching against any branch, because the semantics for how to do matching based on branch name across repos were more complicated than I anticipated. If time permits, I will revisit.

cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -50,6 +50,15 @@ class UpdateDependencies(BaseSalesforceMetadataApiTask):
         "security_type": {
             "description": "Which users to install packages for (FULL = all users, NONE = admins only)"
         },
+        "prefer_2gp_from_release_branch": {
+            "description": "If True and this build is on a release branch (feature/NNN, where NNN is an integer), "
+            "or a child branch of a release branch, resolve GitHub managed package dependencies to 2GP builds present on "
+            "a matching release branch on the dependency."
+        },
+        "2gp_context": {
```
davisagli, line 58: Should we make this a property of the project__git section of the project config?

cumulusci/core/config/project_config.py
```diff
@@ -498,8 +507,62 @@ def get_repo_from_url(self, url):
...
+        if not remote_branch_prefix:
+            return None
+
+        remote_matching_branch = construct_release_branch(
+            remote_branch_prefix, release_id
+        )
+
+        # Check the most recent commit on this release branch looking for a 2GP package to use
+        # TODO: Should we attempt to scan backwards to address race conditions?
+        # Or look for a commit status for an in-progress package upload, and poll?
```
davisagli, line 540: I think we probably want the former, so that a new commit on the remote release branch doesn't make it impossible to deploy downstream packages using this option until the package has been created. But we can do it without first as an MVP.

I think the value of polling is probably low since these skip-validation uploads are quite quick.

cumulusci/core/config/project_config.py
```diff
@@ -498,8 +507,62 @@ def get_repo_from_url(self, url):
...
+        release_branch = remote_repo.branch(remote_matching_branch)
+
+        version_id = get_version_id_from_commit(
+            remote_repo, release_branch.commit.sha, context_2gp
+        )
+        if version_id:
+            self.logger.info(
+                "Located 2GP package version {version_id} for release {release_id} on {remote_repo.clone_url}"
```
davisagli, line 548: Should we include the commit in the log here?

davisagli (commented):
Overall shape of this looks like I was expecting.

cumulusci/tasks/salesforce/update_dependencies.py
davidmreed: I like this idea quite a bit so long as we also default it to "Build Feature Test Package" - I don't want to have to merge a PR in every single project that uses 2GP to make that change.

cumulusci/core/config/project_config.py
davidmreed: 👍

cumulusci/tasks/salesforce/update_dependencies.py
davisagli: Yeah, that's what I had in mind. Since we're inventing this convention for locating corresponding package versions, we can decide what to use as the default context for the commit status.

Possibly we should consider renaming it to something like "2gp Managed Beta Package Version" since it's no longer just for the feature test use case.

cumulusci/core/tests/test_config.py
```diff
@@ -166,6 +167,10 @@ def ref(self, s):
         ref.object.sha = "ref_sha"
         return ref
 
+    @property
+    def default_branch(self):
+        return "main"
```
davisagli, line 172:
```suggestion
    default_branch = "main"
```

cumulusci/core/config/project_config.py
```diff
@@ -498,8 +509,76 @@ def get_repo_from_url(self, url):
...
+        except NotFoundError:
+            self.logger.info(
+                f"Release branch {remote_matching_branch} not found on {remote_repo.clone_url}. Falling back to 1GP."
+            )
+            return (None, None)
+
+        version_id = None
+        count = 0
+        commit = release_branch.commit
+        while version_id is None and count < 5:
```
davisagli, line 551: You might be able to use fewer github API requests by using https://github3py.readthedocs.io/en/master/api-reference/repos.html#github3.repos.repo.Repository.commits rather than getting one commit at a time and following the parent references yourself.

davisagli (changes requested):

cumulusci/core/config/project_config.py
```diff
@@ -498,8 +509,76 @@ def get_repo_from_url(self, url):
 
         return repo
 
-    def get_ref_for_dependency(self, repo, dependency, include_beta=None):
-        release = None
+    def find_matching_2gp_release(
+        self, remote_repo: Union[BaseTaskFlowConfig, Mock]
+    ) -> Tuple[str, str]:
```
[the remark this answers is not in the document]
davidmreed: 🤦  I didn't know Pylance had a type checking mode - thanks!

cumulusci/core/config/project_config.py
```diff
@@ -24,13 +26,22 @@
...
 )
-from cumulusci.core.github import get_github_api_for_repo
-from cumulusci.core.github import find_latest_release
-from cumulusci.core.github import find_previous_release
+from cumulusci.core.github import (
+    get_github_api_for_repo,
+    find_latest_release,
+    find_previous_release,
+    get_version_id_from_commit,
+)
```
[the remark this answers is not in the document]
davidmreed: That was me - will keep in mind!

cumulusci/core/config/project_config.py
```diff
@@ -498,8 +509,76 @@ def get_repo_from_url(self, url):
...
+        head_cumulusci_yml = cci_safe_load(
+            io.StringIO(contents.decoded.decode("utf-8"))
+        )
+        remote_branch_prefix = (
+            head_cumulusci_yml.get("project", {})
+            .get("git", {})
+            .get("prefix_feature", "feature/")
+        )
+
+        remote_matching_branch = construct_release_branch(
```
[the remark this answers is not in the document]
davidmreed: Renamed to `construct_release_branch_name`; that was unclear.

davisagli (approved):

PR #2265, opened 2020-12-16
only automerge if the branch is a feature branch

Br4nd0R:
# Changes
* Flows `ci_feature` and `ci_feature_beta_deps` now only run `github_automerge_feature` if the branch begins with the configured feature branch prefix.

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -624,6 +624,7 @@ flows:
                 task: run_tests
             5:
                 task: github_automerge_feature
+                when: project_config.repo_branch.startswith(project_config.project__git__prefix_feature)
 
     ci_feature_beta_deps:
         group: Continuous Integration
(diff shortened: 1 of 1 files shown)


cumulusci/cumulusci.yml
```diff
@@ -624,6 +624,7 @@ flows:
                 task: run_tests
             5:
                 task: github_automerge_feature
+                when: project_config.repo_branch.startswith(project_config.project__git__prefix_feature)
```
davisagli, line 627:
```suggestion
                when: (project_config.repo_branch or "").startswith(project_config.project__git__prefix_feature)
```

to handle the case where repo_branch is None because the user has checked out a specific commit rather than a branch

cumulusci/cumulusci.yml
```diff
@@ -647,6 +648,7 @@ flows:
...
```
davisagli, line 651:
```suggestion
                when: (project_config.repo_branch or "").startswith(project_config.project__git__prefix_feature)
```

davisagli (changes requested):
Should we also add a condition for github_automerge_main in the release_beta flow, to make sure the flow is being run against the default branch?

davisagli (approved):

PR #2269, opened 2020-12-22
Use betas when building 2GP orgs

davidmreed:
# Critical Changes

# Changes

- CumulusCI installs beta dependencies when building 2GP feature test and QA orgs.

# Issues Closed


davisagli (approved):
Tried it on CaseMan, seems fine.

PR #2270, opened 2020-12-22
Fix issue with LWC mocks in update_package_xml

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue that caused CumulusCI to generate invalid `package.xml` entries for Metadata API-format projects that include `__mocks__` or `__tests__` LWC directories.


davisagli (approved):
I think the ideal solution would probably be to make the generic BundleParser honor entries in .forceignore the same way that sfdx does, but this looks fine as a quicker fix. (We are assuming that no one actually intends to create a LightningComponentBundle with a name that starts with double underscore.)

PR #2288, opened 2021-01-04
Re-enable deprecated record type feature.

prescod:
The deprecated syntax for record types was accidentally broken when I was trying to make the deprecation warning conditional. This fixes the feature and adds a test so we don't break it again.


davidmreed (approved):

davisagli:
@prescod please update the pull request description to include a release note in the usual format

davisagli:
@prescod never mind, I see this is a fix for a change that wasn't released yet, so doesn't need its own release note

prescod:
The change WAS released in 3.25

davisagli:
@prescod oops, I was confused. Thanks for adding the note.

PR #2299, opened 2021-01-06
handle gack while collecting installed packages

davisagli:
It's possible to encounter a gack while retrieving details about a specific installed package that was found by querying InstalledSubscriberPackage. Let's handle that so that it doesn't get in the way of detecting packages that are not broken.

# Critical Changes

# Changes

# Issues Closed
- Handle a possible gack while collecting info about installed packages

### DIFF
--- a/cumulusci/core/config/OrgConfig.py
+++ b/cumulusci/core/config/OrgConfig.py
@@ -273,7 +273,10 @@ class OrgConfig(BaseConfig):
                         "tooling/query/?q=SELECT Id, MajorVersion, MinorVersion, PatchVersion, BuildNumber, "
                         f"IsBeta FROM SubscriberPackageVersion WHERE Id='{isp['SubscriberPackageVersionId']}'"
                     )
-                except SalesforceError:
+                except SalesforceError as err:
+                    self.logger.warning(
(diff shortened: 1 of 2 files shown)


cumulusci/core/config/OrgConfig.py
```diff
@@ -268,10 +268,13 @@ def installed_packages(self):
...
-                    "tooling/query/?q=SELECT Id, MajorVersion, MinorVersion, PatchVersion, BuildNumber, "
-                    f"IsBeta FROM SubscriberPackageVersion WHERE Id='{isp['SubscriberPackageVersionId']}'"
-                )
+                try:
+                    spv_result = self.salesforce_client.restful(
+                        "tooling/query/?q=SELECT Id, MajorVersion, MinorVersion, PatchVersion, BuildNumber, "
+                        f"IsBeta FROM SubscriberPackageVersion WHERE Id='{isp['SubscriberPackageVersionId']}'"
+                    )
+                except SalesforceError:
+                    continue
```
davidmreed, line 277: Shall we add logging here? I'm afraid we'll end up with confusing preflight failures in the future if we don't somehow record that we failed to find details of some package or other.

davisagli: Good call. Added.

davidmreed (approved):

PR #2300, opened 2021-01-06
Correct off-by-one error in extracting data

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue causing the `extract_dataset` task to fail in some circumstances when both an anchor date and Record Types were used.


davidmreed:
I reproduced the error reported by Michael Kolodner and confirmed this change fixes it (successful extract and load). All existing integration tests also pass.

davisagli (approved):

PR #2301, opened 2021-01-07
Support inserting Content Documents

spelak-salesforce:
## Idea
We want to insert files as part of our storytelling data.  This task supports this by:
- Specifying a `path` to the file.   This is inserted as a new ContentVersion, and we query the ContentDocumentId.
- To link records to the new ContentDocument, specify a list of `queries` where are records have a corresponding ContentDocumentLink record created with `share_type` and `visibility`.
    - `share_type` and `visibility` have default values.
    - Creating ContentDocumentLink records is skipped if no `queries` are given or no records returned.
- This task was architected to be extended.  e.g. `upload_user_profile_photo` also has to insert a ContentDocument but has a different linking method.  `upload_user_profile_photo` could be refactored to extend `InsertContentDocument`.

# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+
+Examples
+--------
+
+Upload a profile photo for the default user.
+
+.. code-block:: yaml
+
+    tasks:
+        upload_profile_photo_default:
```
spelak-salesforce, line 27: This needs to be updated.  It was copied from `cumulusci.tasks.salesforce.user.photos`.

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+                photo: storytelling/photos/grace.png
+                where: (Alias = 'grace' OR Alias = 'walker') AND IsActive = true AND CreatedDate = TODAY
+    """
+
+    task_options = {
+        "path": {
+            "description": "Path to the file to upload as a Content Document.",
+            "required": True,
+        },
+        "queries": {
```
spelak-salesforce, line 53: We should probably (auto) namespace inject the queries

spelak-salesforce: @davisagli , any suggestions on how to get this task to have the namespace auto injected?   Are there magic option names to use to plug into the framework?

davisagli: @spelak-salesforce The anonymous Apex task serves as an example:
1. The task has options named `managed` and `namespaced` (the latter should be called `namespaced_org` for best consistency with most tasks; it's called `namespaced` for the anonymous Apex task for backwards compatibility)
2. Those options are processed here: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/tasks/apex/anon.py#L99-L110 -- If the options haven't been specified, determine the defaults using org_config.installed_packages and org_config.namespace
3. Call the inject_namespace utility with these options to replace the standard tokens in whatever string needs it: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/tasks/apex/anon.py#L112-L118

davisagli (changes requested):
I've done some code review; @davidmreed and @jstvz could you review the overall approach?

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+            self.logger.info(
+                "😴 No records IDs queried. Skipping linking the Content Document to related records."
+            )
+
+    def _run_task(self):
+        # Insert the ContentDocument.
+        content_document_id = self._insert_content_document()
+
+        # Query records to link to the new ContentDocument.
+        # "Rolls back" the ContentDocument insert if an Exception is raised.
```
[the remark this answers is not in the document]
spelak-salesforce: Maybe someone wants a non-linked File?

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+            link_records_to_content_document_exception = e
+        finally:
+            if link_records_to_content_document_exception:
+                self.logger.error(
+                    "An error occurred linking queried records to the ContentDocument."
+                )
+                self.logger.error(
+                    f'Deleting ContentDocument "{content_document_id}" to roll back the transaction.'
+                )
+                self._delete_content_document(content_document_id)
```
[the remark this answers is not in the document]
spelak-salesforce: I thought I was being helpful rolling back the entire transaction.  For scratch org development, it's not necessary.   Maybe it's useful in sandbox/production environments?   However, there's not really a "rollback" feature in any other cci task, so it could be best just to remove this functionality.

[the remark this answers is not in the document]
davisagli: I'd say take it out if there isn't a clear need for it. It adds complexity, and probably makes it harder to debug an error when it happens.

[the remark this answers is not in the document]
spelak-salesforce: My intent is `InsertContentDocument` can be extended for other uses cases.   For example, `UploadProfilePhoto` has reuses the same `_insert_content_document` method but has a different way to "link" the file to something (in this case  using the Connect API @ `"connect/user-profiles/{user_id}/photo"`.   I deduced that inserting a file always involves:

1. Inserting the Content Document
2. Determining which record(s) to "link" the file to (if any)
3. "Link" the records to the file.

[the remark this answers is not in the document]
davisagli: Sure, I think catching those exceptions and reporting them in a more helpful way makes sense -- just not removing the ContentDocument when that happens.

cumulusci/tasks/salesforce/__init__.py
```diff
@@ -22,6 +22,7 @@
     "EnsureRecordTypes": "cumulusci.tasks.salesforce.EnsureRecordTypes",
     "GetInstalledPackages": "cumulusci.tasks.preflight.packages",
     "InstallPackageVersion": "cumulusci.tasks.salesforce.install_package_version",
+    "InsertContentDocument": "cumulusci.tasks.salesforce.content_documents",
```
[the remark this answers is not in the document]
spelak-salesforce: Done

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+
+        self.options["path"] = Path(self.options.get("path"))
+
+        if not self.options["path"].exists() or not self.options["path"].is_file():
+            raise CumulusCIException(
+                f'Invalid "path". No file found at {self.options["path"]}'
+            )
```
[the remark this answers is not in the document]
spelak-salesforce: Done in upcoming push.

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+        "visibility": {
+            "description": 'ContentDocumentLink.Visibility for all Content Document Links related to the new ContentDocument. Default: "AllUsers"',
+            "required": False,
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+
+        self.options["path"] = Path(self.options.get("path"))
```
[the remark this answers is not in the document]
spelak-salesforce: Done in upcoming push.

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,209 @@
...
+        if not self.options["path"].exists() or not self.options["path"].is_file():
+            raise CumulusCIException(
+                f'Invalid "path". No file found at {self.options["path"]}'
+            )
+
+        self.options["queries"] = process_list_arg(self.options.get("queries"))
+
+        # Set defaults.
+        if not self.options.get("share_type"):
+            self.options["share_type"] = "I"
```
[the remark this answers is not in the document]
spelak-salesforce: Done in upcoming push.

cumulusci/tasks/salesforce/content_documents.py
spelak-salesforce: @davisagli  The following is totally unrelated to changes in this pull request and only for discussion humbly sharing an opinion.
> Since namespace injection is a very common feature in cci tasks + (it seems that) generally tasks only want to namespace inject using the "defaults", could `OrgConfig` support a `_inject_namespace` method that uses the defaults to call the `_inject_namespace` function?
> - It'd be super easy for tasks to use not having to do a manual configuration + manual call.
> - This shouldn't have any backwards compatibility issues since this doesn't alter task (instance) methods.
> - New tasks like `InsertContentDocument` can opt to inject namespaces only using this default config values.   We can wait for feedback that users really want to have options to override the namespace injection behavior to do the extra work of a manual calculation + call.

davisagli: @spelak-salesforce I like the idea of an `_inject_namespace` method to make this more reusable. There's already one on the base Metadata ETL task, actually, but putting it on OrgConfig makes sense, at least at first glance. I think the project_config can be accessed by way of the OrgConfig's `keychain`.

Context here is that I realized while working on automatic namespace injection that this logic was a bit repetitive and would be nice to centralize, but I had trouble figuring out a way to do that that would actually make things simpler, especially while still supporting the explicit options and 100% backwards compatibility. But I think your idea has potential, especially for new tasks.

FWIW I _mostly_ agree that the explicit options are mostly for backwards compatibility and don't need to be added for new tasks. One possible reason to continue providing them is that in the context of a metadeploy plan where it's known that the package will always be managed, it might be nice to avoid making unnecessary API calls to the org to check the installed packages.

spelak-salesforce: For an alternative, while writing the above, my initial was:
- The base task or base salesforce task always includes "base" task options of the options needed to inject namespaces: `managed`, `namespaced_org`, and `namespace_[to_]inject`.
    - This would change how task options are compiled.   Task implementations can override the base task options, but at least the options would still exist.  Something like:
       ```python
        task_options = self.base_task_options
        task_options.update(self.task_options)
        ```
- The base task then has an instance method called `_inject_namespace` that does all the calculation.
    - At first, I was worried about breaking existing implementations, but then I remember it's python: any implementation using the same method name will just override the base method.
    - Some extra care may need to be taking since we are "overriding" task options.  I think the trick is never to have the base task reset base task options.  `_inject_namespace` always reads the task options for its calculations.  
    -  If the base task options are overridden in a way that breaks `_inject_namespace`, oops!  Don't use `_inject_namespace`.
- Tasks like `Deploy` that use non-standard task options for namespace injection can do the conversion in `_init_options`.

davisagli: @spelak-salesforce It's already pretty easy to copy particular options from another source (whether it's a base task's options, or a standalone dict), so I don't see a compelling reason to add `base_task_options` as a formal part of the framework. This could be something like this:

    namespace_options = {
        "namespace": { ... },
        "managed": { ... },
        "namespaced_org": { ... },
    }

    class MyTask(BaseTask):
        task_options = {
            **namespace_options,
            "another_option": { ... },
        }

Using ** in this way also makes sure that we're constructing a new dict and not mutating the one we're copying options from.

cumulusci/tasks/salesforce/content_documents.py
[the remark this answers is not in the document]
spelak-salesforce: Done.   I still have to review the tests, but the production code is changed.

cumulusci/tasks/salesforce/content_documents.py
spelak-salesforce: Could we store the `namespace_options` + the following `BaseTask` mixin?

```python
from cumulusci.utils import inject_namespace

namespace_options = {
    "managed": {
        "description": "If False, changes namespace_inject to replace tokens with a blank string",
        "required": False,
    },
    "namespaced_org": {
        "description": "If True, the tokens %%%NAMESPACED_ORG%%% and ___NAMESPACED_ORG___ will get replaced with the namespace.  The default is false causing those tokens to get stripped and replaced with an empty string.  Set this if deploying to a namespaced scratch org or packaging org.",
        "required": False,
    },
    "namespace_inject": {
        "description": "If set, the namespace tokens in files and filenames are replaced with the namespace's prefix",
        "required": False,
    },
}

class NamespaceInjectionMixin:

    def inject_namespace(self, name, content):
        namespace = (
            self.options.get("namespace_inject")
            or self.project_config.project__package__namespace
        )
        if "managed" in self.options:
            managed = process_bool_arg(self.options["managed"])
        else:
            managed = (
                bool(namespace) and namespace in self.org_config.installed_packages
            )
        if "namespaced_org" in self.options:
            namespaced_org = process_bool_arg(self.options["namespaced_org"])
        else:
            namespaced_org = bool(namespace) and namespace == self.org_config.namespace

        return inject_namespace(
            name_content
            namespace=namespace,
            managed=managed,
            namespaced_org=namespaced_org,
        )
```

spelak-salesforce: @davisagli , I added `cumulusci.tasks.salesforce.namespaces` containing `NamespaceInjectionMixin` + `namespace_injection_options`.   Thoughts?  Thank you.

spelak-salesforce: `NamespaceInjectionMixin` has been moved to #2313

cumulusci/cumulusci.yml
```diff
@@ -286,6 +286,12 @@ tasks:
         description: Parses GitHub release notes to report various info
         class_path: cumulusci.tasks.github.ReleaseReport
         group: GitHub
+    insert_content_document:
+        description: Inserts a Content Document and links records that are returned by queries.   Queries are namespace injected before execution.
+        class_path: cumulusci.tasks.salesforce.content_documents.InsertContentDocument
+        group: Salesforce
+        options:
+            namespace_inject: $project_config.project__package__namespace
```
davidmreed, line 294: We should handle this with the new "magic" pattern rather than specifying in YAML (already implemented below)

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,237 @@
...
+        # Query the ContentDocumentId for our created record.
+        content_document_id = self.sf.query(
+            f"SELECT Id, ContentDocumentId FROM ContentVersion WHERE Id = '{content_version_id}'"
+        )["records"][0]["ContentDocumentId"]
+
+        self.logger.info(f'Success!  Inserted ContentDocument "{content_document_id}".')
+
+        return content_document_id
+
+    def _get_record_ids_to_link(self) -> list[str]:
```
davidmreed, line 147: This type hinting syntax (using the collection directly instead of `typing.List`) is Python 3.9-only.

cumulusci/tasks/salesforce/content_documents.py
```diff
@@ -0,0 +1,237 @@
...
+        finally:
+            if get_record_ids_to_link_exception:
+                self.logger.error(
+                    "An error occurred querying records to link to the ContentDocument."
+                )
+                # Reraise the Exception
+                raise get_record_ids_to_link_exception
+
+        # Links queried records to link to the new ContentDocument.
+        # "Rolls back" the ContentDocument insert if an Exception is raised.
```
davidmreed, line 222: We are not currently performing a rollback, correct?

cumulusci/tasks/salesforce/tests/mocks/InsertContentDocument.file.txt
```diff
@@ -0,0 +1 @@
+Mock of a file to be inserted as a Content Document.
```
jstvz, line 1: I'd suggest using the pytest `tmpdir` fixture instead.

cumulusci/tasks/salesforce/content_documents.py
spelak-salesforce: Correct.  We removed that feature, but I forgot to update the comments.  Thank you!

cumulusci/tasks/salesforce/content_documents.py
spelak-salesforce: Will update.

jstvz:

PR #2306, opened 2021-01-08
Prepare 3.26.0

Br4nd0R:
# Critical Changes

# Changes

# Issues Closed


HISTORY.rst
```diff
@@ -2,6 +2,40 @@
 History
 =======
 
+3.26.0 (2021-01-08)
+-------------------
+
+Critical Changes:
+
+- CumulusCI can now resolve dependencies using second-generation packages (2GPs) for upstream projects. When a 2GP build runs on a release branch (starting with ``prefix/NNN``, where ``prefix`` is the feature branch prefix and ``NNN`` is an integer), CumulusCI will look for a matching release branch in each upstream dependency and use a 2GP package build on that release branch if present, falling back to a 1GP beta release if not present.
```
davisagli, line 10: This is just a change, not a critical change. Nothing is different unless you turn on the new option.

HISTORY.rst
```diff
@@ -2,6 +2,40 @@
...
+
+Changes:
+
+- CumulusCI now reports how long it took for flows to run.
+
+- Flows ``ci_feature`` and ``ci_feature_beta_deps`` now only run the ``github_automerge_feature`` task if the branch begins with the configured feature branch prefix.
+
+- Running the ``deploy`` task with the ``path`` option set to a path that doesn't exist will log a warning instead of raising an error.
+
+- CumulusCI now installs beta dependencies when building 2GP feature test and QA orgs.
```
davisagli, line 20: This might be clearer: "When the ci_feature_2gp and qa_org_2gp flows install dependencies, the latest beta version will be used when available."

davisagli (changes requested):

HISTORY.rst
davidmreed: As implemented it is on by default, which would make it a critical change to me. 

Should we quick-change that to be off by default? That's probably a lot safer. @davisagli @Br4nd0R

davidmreed (changes requested):
Holding for discussing on 2GP behavior

HISTORY.rst
davisagli: @davidmreed It appears to already be off by default: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/tasks/salesforce/update_dependencies.py#L84

davidmreed: It's turned on in `cumulusci.yml` for 2GP builds: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/cumulusci.yml#L933

davisagli: Ah ok, that's the disconnect. Can we be more specific and name specific flows rather than vaguely mentioning "2gp builds" in this note?

HISTORY.rst
```diff
@@ -2,6 +2,41 @@
...
+
+- CumulusCI now reports how long it took for flows to run.
+
+- Flows ``ci_feature`` and ``ci_feature_beta_deps`` now only run the ``github_automerge_feature`` task if the branch begins with the configured feature branch prefix.
+
+- Running the ``deploy`` task with the ``path`` option set to a path that doesn't exist will log a warning instead of raising an error.
+
+- When the ``ci_feature_2gp`` and ``qa_org_2gp`` flows install dependencies, the latest beta version will be used when available.
+
+- CumulusCI can now resolve dependencies using second-generation packages (2GPs) for upstream projects. When a 2GP build runs on a release branch (starting with ``prefix/NNN``, where ``prefix`` is the feature branch prefix and ``NNN`` is an integer), CumulusCI will look for a matching release branch in each upstream dependency and use a 2GP package build on that release branch if present, falling back to a 1GP beta release if not present.
```
davidmreed, line 19:
```suggestion
- CumulusCI can now resolve dependencies using second-generation packages (2GPs) for upstream projects. When a `ci_feature_2gp` or `qa_org_2gp` flow runs on a release branch (starting with ``prefix/NNN``, where ``prefix`` is the feature branch prefix and ``NNN`` is an integer), CumulusCI will look for a matching release branch in each upstream dependency and use a 2GP package build on that release branch, if present, falling back to a 1GP beta release if not present.
```

davidmreed (changes requested):

davisagli (approved):

PR #2307, opened 2021-01-08
Fix query caching

prescod:
# Changes

Push upgrade caching had some small intermittent bugs which have never been observed in production but were caught by automated-tests. The cache has been replaced with a more reliable one.

# Context

I removed a bunch of calls to @memoize and created a query cache instead. This broke a bunch of test cases but my hypothesis is that the tests were broken and the brokenness was hidden by the memoization.

I'm far from an expert in the domain of push upgrades so I'd appreciate a detailed review either a meeting or in github.

And we also need to figure out #2302 before we can merge this.

And we should probably try some frm of performance testing to see if the previous caching and the new one are equivalent or at least if the new one is acceptable.

# Issues Closed


davisagli:
So first some context on the steps of the push tasks:
1. query the org for a package id and package version id based on the namespace and version in the task options
2. get the list of orgs to push to (the _get_orgs method is implemented differently for different tasks, so it could come from a file, or from a query of the packaging org's subscribers)
3. create a PackagePushRequest, with a related PackagePushJob for each org
4. start the push by setting the status of the push request to Pending
5. if the push is being scheduled immediately, poll until it completes and report the results

I don't think that doing a blanket cache of all these queries makes sense. In particular, I see that the task does multiple calls to `get_subscribers`, `get_push_requests`, and `get_push_request_objs`, and the current implementation explicitly clears the memoize cache before them. Probably makes more sense to avoid caching them in the first place. Some of the other queries to get info about packages and package versions should be safer to cache, but I'm still looking at whether we really need to.

In any case I think we probably want to cache more selectively, and not across the board for all queries.

prescod:
Yeah, I dd worry that I didn't have enough context about what the code was doing and it turned out I was right.

PR #2311, opened 2021-01-08
Improve yaml parse error handling

Br4nd0R:
I was unable to find an attribute on `ScannerError` for the name of the file being parsed.  I included "likely cumulusci.yml" in the error messaging, because we can't be sure which call to `handle_exception()` is being invoked.

# Changes
* CumulusCI displays more user friendly error message when encountering parsing errors in `cumulusci.yml`.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -25,6 +25,7 @@ import requests
 from rst2ansi import rst2ansi
 from jinja2 import Environment
 from jinja2 import PackageLoader
+from yaml.scanner import ScannerError
 
 import cumulusci
 from cumulusci.core.config import OrgConfig
@@ -204,24 +205,28 @@ def main(args=None):
         if "--json" not in args and not is_version_command:
             check_latest_version()
 
-        # Load CCI config
-        global RUNTIME
-        RUNTIME = CliRuntime(load_keychain=False)
-        RUNTIME.check_cumulusci_version()
-
-        # Configure logging
-        debug = "--debug" in args
-        if debug:
-            args.remove("--debug")
-        should_show_stacktraces = RUNTIME.universal_config.cli__show_stacktraces
-
-        # Only create logfiles for commands
-        # that are not `cci error`
+        # Only create logfiles for commands that are not `cci error`
         is_error_command = len(args) > 2 and args[1] == "error"
         tempfile_path = None
         if not is_error_command:
             logger, tempfile_path = get_tempfile_logger()
             stack.enter_context(tee_stdout_stderr(args, logger, tempfile_path))
+
+        debug = "--debug" in args
+        if debug:
+            args.remove("--debug")
+
+        # Load CCI config
+        global RUNTIME
+        try:
+            RUNTIME = CliRuntime(load_keychain=False)
+        except Exception as e:
+            handle_exception(e, is_error_command, tempfile_path, debug)
+            sys.exit(1)
+
+        RUNTIME.check_cumulusci_version()
+
+        should_show_stacktraces = RUNTIME.universal_config.cli__show_stacktraces
 
         init_logger(log_requests=debug)
         # Hand CLI processing over to click, but handle exceptions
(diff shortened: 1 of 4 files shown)


cumulusci/cli/cci.py
```diff
@@ -272,6 +279,17 @@ def connection_error_message():
     click.echo(click.style(message, fg="red"), err=True)
 
 
+def yaml_parse_error_message(error):
+    message = f"An error occurred while parsing a yaml file (likely cumulusci.yml). Check around line {error.problem_mark.line} for the issue."
```
davisagli, line 283: I think the cci_safe_load function in cumulusci.utils.yaml.cumulusci_yml would be a better place to catch parse errors and reraise them with a better message -- then we should be able to get the filename. We should also include the column if it's available, and include the original error message for those who can interpret it.

davisagli (changes requested):

cumulusci/utils/yaml/tests/test_cumulusci_yml.py
```diff
@@ -37,3 +42,39 @@ def test_converter_is_selective(self):
...
+        type(f_config).name = PropertyMock(return_value="cumulusci.yml")
+
+        with pytest.raises(CumulusCIException) as error:
+            cci_safe_load(f_config)
+
+        assert error.typename == "CumulusCIException"
+        assert (
+            "An error occurred parsing cumulusci.yml at line 12345, column 54321.\nError message: None"
+            == error.value.args[0]
+        )
```
davisagli, line 63: Could you rewrite this test to use an actual poorly formatted yaml file in a tempdir instead of mocks? The mocks make it hard to know if we're setting up the mocks realistically.

Also do a test to make sure cci_safe_load still works if we pass in a StringIO() instance instead of a file, because I think they probably don't have a `.name`

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -35,7 +39,18 @@ def _replacer_func(matchobj):
...
 def cci_safe_load(f_config: IO[Text]):
-    "Load a file, convert NBSP->space and parse it in YAML."
+    """Load a file, convert NBSP->space and parse it in YAML.
+    Raises CumulusCIException if an error occurs while parsing.
+    """
     data = _replace_nbsp(f_config.read())
-    rc = yaml.safe_load(StringIO(data))
+    try:
+        rc = yaml.safe_load(StringIO(data))
+    except ScannerError as e:
```
davisagli, line 48: Looking at the PyYAML source, there are a few classes of exceptions which extend MarkedYAMLError. I think we probably want to catch that one here instead of only ScannerError.

davisagli (changes requested):

prescod (commented):
I seem to usually fall on the side of "mocks if necessary but only if necessary" so I like the recent changes to the PR. Libraries can change under us, so it is nice to test that they still behave the way we expect (in this case, that PyYaml still responds as we expect to typos).

I try to use shorter snippets of error messages so that minor rewordings don't break unit tests but that's hardly urgent.

Worth considering if there are other weird cases worth testing. In particular 1. What if it isn't really Unicode and 2. What if the file is totally empty.

cumulusci/utils/yaml/tests/test_cumulusci_yml.py
```diff
@@ -37,3 +43,59 @@ def test_converter_is_selective(self):
...
+
+            with open(cumulusci_yml_filepath, "w+") as cumulusci_yml:
+                invalid_yml = """xyz: abc   \n>>>\nefg: lmn\n"""
+                cumulusci_yml.write(invalid_yml)
+                cumulusci_yml.seek(0)
+
+                with pytest.raises(CumulusCIException) as error:
+                    cci_safe_load(cumulusci_yml)
+
+        assert error.typename == "CumulusCIException"
```
davisagli, line 59: This is unnecessary -- the `pytest.raises` context manager already asserts that the expected exception was raised. You can also pass `matches` to `pytest.raises` to assert that a string is included in the except message, instead of a separate assertion, if you want.

davisagli (approved):
Would be good to also add tests for the other cases Paul came up with, but this is already a good improvement.

Br4nd0R:
@davisagli @prescod - I removed the class to make things more pytest-like (also added a fixture to clean up cumulusci.yml file creation). Odd though, ass the `test_converter()` method now fails. The `outp` has 12 spaces between `\n` and `y:` where as the value for `rc` has 10 spaces between `\n` and `y:`, but no code changed there.

davisagli:
@Br4nd0R fwiw, I don't see organizing tests into classes as antithetical to pytest's approach -- but I don't feel strongly about whether we do so or not.

The failing test is because you changed the indentation inside the triple-quoted string.

Br4nd0R:
> The failing test is because you changed the indentation inside the triple-quoted string.

davisagli (approved):

PR #2313, opened 2021-01-11
Support mixin for namespace injection

spelak-salesforce:
`NamespaceInjectionMixin` from `cumulusci.tasks.salesforce.namespaces` supports a `_inject_namespace` method that calls `inject_namespace` from `cumulusci.utils` while automatically detecting the org context for default `namespace`, `managed`, and `namespaced_org` arguments.

Implementations can reuse the `namespace_injection_options` from `cumulusci.tasks.salesforce.namespaces` in `task_options` to support `_inject_namespace`.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/namespaces.py
+++ b/cumulusci/tasks/salesforce/namespaces.py
@@ -1,0 +1,61 @@
+from cumulusci.core.utils import process_bool_arg
+from cumulusci.utils import inject_namespace
+from typing import Optional, Tuple
+
+namespace_injection_options = {
+    "managed": {
+        "description": "If False, changes namespace_inject to replace tokens with a blank string",
+        "required": False,
+    },
+    "namespaced_org": {
+        "description": "If True, the tokens %%%NAMESPACED_ORG%%% and ___NAMESPACED_ORG___ will get replaced with the namespace.  The default is false causing those tokens to get stripped and replaced with an empty string.  Set this if deploying to a namespaced scratch org or packaging org.",
+        "required": False,
+    },
+    "namespace_inject": {
+        "description": "If set, the namespace tokens in files and filenames are replaced with the namespace's prefix",
+        "required": False,
+    },
+}
+
+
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/namespaces.py
```diff
@@ -0,0 +1,61 @@
...
+namespace_injection_options = {
+    "managed": {
+        "description": "If False, changes namespace_inject to replace tokens with a blank string",
+        "required": False,
+    },
+    "namespaced_org": {
+        "description": "If True, the tokens %%%NAMESPACED_ORG%%% and ___NAMESPACED_ORG___ will get replaced with the namespace.  The default is false causing those tokens to get stripped and replaced with an empty string.  Set this if deploying to a namespaced scratch org or packaging org.",
+        "required": False,
+    },
+    "namespace_inject": {
```
spelak-salesforce, line 14: I copied the `namespace_inject` option from [Deploy](https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/tasks/salesforce/Deploy.py) for consistency.   IMHO, the task name is weird.  Could we instead just call it `namespace` or `namespace_to_inject`?

cumulusci/tasks/salesforce/namespaces.py
```diff
@@ -0,0 +1,61 @@
...
+    ) -> Tuple[Optional[str], Optional[str]]:
+        _namespace = (
+            namespace
+            or self.options.get("namespace_inject")
+            or self.project_config.project__package__namespace
+        )
+
+        if managed is not None:
+            _managed = managed
+        elif "managed" in self.options:
```
davidmreed, line 38: I'd suggest also looking for `"unmanaged"`, as many tasks use that option.

(Such tasks wouldn't be able to use this mixin's task options dict by reference, but I think newer tasks should be using the clearer `"managed"` anyway).

jstvz:

PR #2314, opened 2021-01-11
NamespaceInjectionMixin supports _get_namespaced_filename and _get_namespaced_content

spelak-salesforce:
> NOTE: Child branch of #2313 

`_inject_namespace` method of `NamespaceInjectionMixin` from `cumulusci.tasks.salesforce.namespaces` returns multiple values for namespace injecting both a file's name and the file's content simultaneously.   However, there are many use cases where a task only wants to namespace inject a string "content".   `NamespaceInjectionMixin` now supports individual methods to namespace inject a filename and content separately using the same functionality from `_inject_namespace`:
- `_get_namespaced_filename`
- `_get_namespaced_content`

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/namespaces.py
+++ b/cumulusci/tasks/salesforce/namespaces.py
@@ -59,3 +59,35 @@ class NamespaceInjectionMixin:
             namespaced_org=_namespaced_org,
             logger=self.logger,
         )
+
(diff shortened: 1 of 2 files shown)


jstvz:

PR #2315, opened 2021-01-11
Fix flaky push API memoization

### LINKED ISSUE #2303: Incorrect tests?
We have a mock for push.create_push_request

https://github.com/SFDO-Tooling/CumulusCI/blob/48db7daf2743ae0b518a6af2027c7e899ded4ecb/cumulusci/tasks/push/tests/test_push_tasks.py#L499

The type of the mock is Tuple[dict, int]

The return value for the real function seems to be Tuple[str,int]

https://github.com/SFDO-Tooling/CumulusCI/blob/44b0d6a3e44427652f018e23684e37d232e7eaf0/cumulusci/tasks/push/push_api.py#L584

Perhaps this was hidden by the memoization code?

davisagli:
This is an alternative to #2307 which attempts to simply fix the existing memoization rather than switch to a query cache. I switched from our own `memoize` decorator to `functools.lru_cache`. `lru_cache` uses a cache key with an actual reference to the function arguments rather than stringifying them, so I believe it should avoid the problem with cache collisions due to object ids being reused -- both by no longer using the stringified id, and by preventing the object from getting garbage collected, so its id won't be reused.

I think we're still doing more memoization than really makes sense and probably want to revisit that the next time we're doing bigger work on the push tasks, but for now I'm just looking for a quick fix to avoid the flaky tests.

I also removed the `get_where_last_version` method which was not used anywhere other than tests.

Also fixes #2303 

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/tasks/push/push_api.py
+++ b/cumulusci/tasks/push/push_api.py
@@ -291,14 +291,14 @@ class SalesforcePushApi(object):
 
         return "%s LIMIT %s" % (query, limit)
 
-    @lru_cache
+    @lru_cache(32)
     def get_packages(self, where=None, limit=None):
         where = self.format_where_clause(where)
         query = f"SELECT id, name, namespaceprefix FROM MetadataPackage{where}"
         query = self.add_query_limit(query, limit)
         return self.return_query_records(query)
 
-    @lru_cache
+    @lru_cache(32)
     def get_package_objs(self, where=None, limit=None):
         package_objs = []
         for package in self.get_packages(where, limit):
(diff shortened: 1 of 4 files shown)


jofsky (approved):
@davisagli you made the comment that 

`I think we're still doing more memoization than really makes sense and probably want to revisit that the next time we're doing bigger work on the push tasks, but for now I'm just looking for a quick fix to avoid the flaky tests.`

What part of this code is still using memoized functions? Or are you referring to other components of the push api?

davisagli:
@jofsky I'm still counting lru_cache as a form of memoization, even though it's not literally called memoize

PR #2321, opened 2021-01-16
Rename example file and make minor unit test changes

prescod:
I renamed a file and I was seeing intermittent errors in a test case so I added additional diagnostics in the test to make it easier to diagnose if it shows up again.


cumulusci/core/tests/test_pageobjects.py
```diff
@@ -87,8 +87,15 @@ class TestPageObjects(unittest.TestCase):
...
-        self.assertEqual(po.registry, BASE_REGISTRY)
+        assert po.get_keyword_names() == CORE_KEYWORDS, (
+            po.get_keyword_names(),
+            CORE_KEYWORDS,
+        )
+        assert len(po.registry) == len(BASE_REGISTRY), (
+            len(po.registry),
+            len(BASE_REGISTRY),
+        )
+        assert po.registry == BASE_REGISTRY, (po.registry, BASE_REGISTRY)
```
davisagli, line 98: It's frustrating that we keep making one-off changes to make assertions actually output something useful, since that's a feature that is built in to pytest -- but that isn't working for us because typeguard's import hook prevents pytest's from taking effect. Is there any way we can make pytest and typeguard play nicer together?

prescod: Actually TypeGuard doesn't take the blame in this case. Previously it was using self.assertEqual which TypeGuard doesn't touch. I think that it was doing too aggressive truncation. But the next time I notice the TypeGuard/Pytest conflict I can look into it.

davisagli (approved):

PR #2322, opened 2021-01-16
Allow num_records to be omitted in Snowfakery invocations.

prescod:
# Changes

When invoking Snowfakery from CumulusCI, you no longer have to supply a number of records or a controlling table name. Your recipe will be invoked exactly once and generate as many records as it generates.


docs/cookbook.rst
```diff
@@ -551,10 +551,15 @@ writing them in the `Snowfakery docs
...
+If you would like to execute the recipe multiple times to generate
+more data, you do so like this:
 
 ``$ cci task run generate_and_load_from_yaml -o generator_yaml
-datasets/some_snowfakery_yaml -o num_records 1000 -o num_records_tablename
-Account —org dev``
+datasets/some_snowfakery_recipe.yml -o num_records 1000 -o num_records_tablename
+Account —-org dev``
 
 ``generator_yaml`` is a reference to your Snowkfakery recipe.
```
davisagli, line 564:
```suggestion
``generator_yaml`` is a reference to your Snowfakery recipe.
```

davisagli (changes requested):

davisagli (approved):
Approving this, but it made me wonder -- is there anything to avoid an infinite loop if the recipe doesn't actually generate any records?

prescod:
@davisagli If you just asked to run the script once then it shouldn't matter whether it generated records or not (although there might be downstream bugs for this weird case).

If you asked to run the script with an SObject name specified (as was previously mandatory) then it checks to ensure that it succeeded in generating that SObject and throws an exception if not.

It's actually hard to make a no-op script by accident, by the way.

PR #2323, opened 2021-01-17
New option for task ensure_record_type to force create

bethbrains:
# Critical Changes

# Changes
- The `ensure_record_types` task has a new option, `force_create`, which will create the Record Type even if other Record Types already exist on the object.

# Issues Closed


cumulusci/tasks/salesforce/tests/test_EnsureRecordTypes.py
```diff
@@ -253,6 +253,33 @@ def test_no_action_if_existing_record_types(self):
...
+
+        task.sf = mock.Mock()
+        task.sf.Account = mock.Mock()
+        task.sf.Account.describe = mock.Mock(return_value=OPPORTUNITY_DESCRIBE_WITH_RTS)
+        task._infer_requirements()
+
+        with temporary_dir():
+            task._build_package()
+            with open(os.path.join("objects", "Account.object"), "r") as f:
+                opp_contents = f.read()
```
davidmreed, line 277: This variable name may be confusing. `obj_contents`?

cumulusci/tasks/salesforce/tests/test_EnsureRecordTypes.py
```diff
@@ -253,6 +253,33 @@ def test_no_action_if_existing_record_types(self):
...
+                "record_type_label": "NPSP Default",
+                "record_type_description": "Default Account Record Type created by NPSP.",
+                "sobject": "Account",
+                "force_create": True,
+            },
+        )
+
+        task.sf = mock.Mock()
+        task.sf.Account = mock.Mock()
+        task.sf.Account.describe = mock.Mock(return_value=OPPORTUNITY_DESCRIBE_WITH_RTS)
```
davidmreed, line 271: I'd love a comment here to call out that there's no impact from using the "wrong" describe (all it's doing is representing Record Types as present).

davidmreed (commented):
This looks great; I just have two nit-picks on the tests because reading through it initially confused me.

cumulusci/tasks/salesforce/tests/test_EnsureRecordTypes.py
davisagli: This way of reading the file is fine, but it's also possible to do it in a single line using pathlib:

    obj_contents = pathlib.Path("objects", "Account.object").read_text()

Also, _probably_ doesn't matter here, but it's good to get in the habit of always specifying `encoding="utf-8"` when opening a file or using read_text. There is often a different default encoding on Windows.

davisagli (changes requested):
Agree with @davidmreed's feedback: looks good; a few minor suggestions.

cumulusci/tasks/salesforce/tests/test_EnsureRecordTypes.py
bethbrains: discussed with DR... we will put off a pathlib refactor until we do the entire file.

davisagli (approved):

PR #2337, opened 2021-01-20
Add dict option for org_settings task

bethbrains:
# Critical Changes

# Changes
- `org_settings` accepts a dict option called `settings` in addition to (or instead of) the existing `definition_file` option.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/org_settings.py
+++ b/cumulusci/tasks/salesforce/org_settings.py
@@ -30,6 +30,7 @@ class DeployOrgSettings(Deploy):
 
     task_options = {
         "definition_file": {"description": "sfdx scratch org definition file"},
+        "settings": {"description": "A dict of settings to apply"},
         "api_version": {"description": "API version used to deploy the settings"},
     }
 
@@ -42,10 +43,16 @@ class DeployOrgSettings(Deploy):
         self.options["namespaced_org"] = False
 
     def _run_task(self):
-        with open(self.options["definition_file"], "r") as f:
-            scratch_org_definition = json.load(f)
+        settings = {}
 
-        settings = scratch_org_definition.get("settings", {})
+        if self.options.get("definition_file"):
+            with open(self.options["definition_file"], "r") as f:
+                scratch_org_definition = json.load(f)
+                settings.update(scratch_org_definition.get("settings", {}))
+
+        if self.options.get("settings"):
+            settings.update(self.options["settings"])
+
         if not settings:
             return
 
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/salesforce/org_settings.py
```diff
@@ -42,10 +43,16 @@ def _init_options(self, kwargs):
...
-            scratch_org_definition = json.load(f)
+        settings = {}
+
+        if self.options.get("definition_file"):
+            with open(self.options["definition_file"], "r") as f:
+                scratch_org_definition = json.load(f)
+                settings.update(scratch_org_definition.get("settings", {}))
+
+        if self.options.get("settings"):
+            settings.update(self.options["settings"])
```
davidmreed, line 54: There's a risk of confusion here because `update` does not merge sub-dicts. So if the user, say, wanted to exploit this facility to combine a settings file with a YAML tweak... they'll only get the tweak for whatever entity they were trying to override. `cumulusci.core.utils.dictmerge()` would avoid that edge case, or we could prohibit using both sources.

cumulusci/tasks/salesforce/org_settings.py
```diff
@@ -42,10 +43,16 @@ def _init_options(self, kwargs):
...
 
-        settings = scratch_org_definition.get("settings", {})
         if not settings:
             return
```
davidmreed, line 57: If we take no action, let's log that that's the case.

davidmreed (changes requested):

davisagli (commented):
@bethbrains can you tell me about the use case for this? The main use I'm aware of for this task is in Metecho, where the settings are in a scratch org file.

prescod:
Yesterday I was working with a team on a persistent org. They were trying to do a hard delete. I instructed them to run create_bulk_data_permission_set and assign_permission_sets as described in the docs.

But they failed, because as the docs also say: NOTE: the org setting to allow Set Audit Fields must be turned on.

Can Beth's work be used to do that and if so, can we document how?

davisagli:
@prescod I don't know, but based on a quick search for `salesforce "set audit fields"`, it looks like it is a permission rather than an org setting.

davidmreed:
Beth's use case is in Platform Encryption, where org settings need to be manipulated at a specific point in the automation _after_ the initial creation of the scratch org.

Set Audit Fields requires an org setting _and_ a user-level permission. The org setting is `enableAuditFieldsInactiveOwner` on `SecuritySettings`.

davidmreed (approved):

PR #2338, opened 2021-01-21
[CCI] Push API library refactored to support Bulk Data API Queries as part of Push Upgrade Automation Epic

jofsky:
# Critical Changes

# Changes
- Adjustments to push upgrade tasks:
  - The `push_sandbox` and `push_all` tasks now use the Bulk API to query for subscriber orgs.
  - The `push_sandbox` and `push_all` tasks now default to including all orgs whose status is not Inactive, rather than only orgs with a status of Active. This means that sandboxes, scratch orgs, and Developer Edition orgs are included.

# Issues Closed
- Fixed the `min_version` option for the `push_sandbox` and `push_all` tasks to include the `min_version` and not only versions greater than it.


davidmreed:
@jofsky Are you looking for a general once-over on this draft PR or something more specific?

jofsky:
@davidmreed I'm looking more for a once over, however I can't for the life of me figure out why this one test is failing. Given what I'm returning, I'm not sure If I need to mock a response or there's something I'm not including.

jofsky:
I tested it against CumulusCI-Test and received expected behavior.

cumulusci/tasks/push/push_api.py
```diff
@@ -265,12 +268,24 @@ def __init__(self, sf, logger, lazy=None, default_where=None, batch_size=None):
             batch_size = 200
         self.batch_size = batch_size
 
-    def return_query_records(self, query):
-        res = self.sf.query_all(query)
-        if res["totalSize"] > 0:
-            return res["records"]
+        self.bulk = bulk
+
+    def return_query_records(self, query, field_names=None, sObject=None):
```
davisagli, line 273: sObject is the correct capitalization in prose, but I would stick with all lowercase `sobject` for a Python variable name. (Same thing below)

cumulusci/tasks/push/push_api.py
```diff
@@ -358,13 +373,22 @@ def get_package_versions_by_id(self, where=None, limit=None):
...
-        return self.return_query_records(query)
+        field_names = [
+            "Id",
+            "MetadataPackageVersionId",
+            "InstalledStatus",
+            "OrgName",
+            "OrgKey",
+            "OrgStatus",
+            "OrgType",
+        ]
```
davisagli, line 388: It would probably be a good idea to define this list once, then use it both for generating `query` and passing in as `field_names` -- to make sure that they don't accidentally get out of sync if we change one but forget to change the other.

cumulusci/tasks/push/tasks.py
```diff
@@ -237,12 +237,12 @@ def _init_task(self):
...
-        if neither_file_option or both_file_options:
-            raise TaskOptionsError(
-                "Please call this task with either the `orgs` or `csv` option."
-            )
+        # neither_file_option = "orgs" not in self.options and "csv" not in self.options
+        # both_file_options = "orgs" in self.options and "csv" in self.options
+        # if neither_file_option or both_file_options:
+        #     raise TaskOptionsError(
+        #         "Please call this task with either the `orgs` or `csv` option."
+        #     )
```
davisagli, line 245: I see this is related to what you were asking about in standup. I think you should go ahead and fix _init_options for SchedulePushOrgQuery as part of this PR.

davidmreed:
The test that is failing is failing because `_get_orgs()` initializes its local `push_api`, causing an actual HTTP request to get through (tasks.py lines 362-364). Your mocks at test_push_tasks.py lines 483-484 aren't taking effect because `push_api` is a local, not an instance variable.

To do that mock, you'll probably want to use something like `@mock.patch("cumulusci.tasks.push.tasks.SalesforcePushApi")` on the method (which then takes an argument, representing that new mock) so that constructing that local variable returns a mock.

cumulusci/tasks/push/tasks.py
jofsky: So initially I was going to remove the additional option restriction from `SchedulePushOrgList`, but added it to `SchedulePushOrgQuery` instead let me know if you agree with this approach.

cumulusci/tasks/push/push_api.py
jofsky: Personal reminder @jofsky to do this.

cumulusci/tasks/push/push_api.py
```diff
@@ -265,12 +268,25 @@ def __init__(self, sf, logger, lazy=None, default_where=None, batch_size=None):
...
+    def return_query_records(self, query, field_names=None, sobject=None):
+        res = []
+        if self.bulk and field_names and sobject:
+            step = BulkApiQueryOperation(
+                sobject=sobject, api_options={}, context=self, query=query
+            )
+            step.query()
+            if len(list(step.get_results())) > 0:
+                for query_result in list(step.get_results()):
+                    res += [dict(zip(query_result.keys(), query_result.values()))]
```
davisagli, line 282: Keep in mind that the way this is currently implemented, we are loading all the results into memory at once rather than processing them in batches. I guess we'll see if that works, but at some point we may need to rewrite it.

davisagli (commented):
PackagePushApi.create_push_request currently adds orgs to the job by creating PackagePushJobs in batches of 200. This could probably also be switched to use the Bulk API.

Ideally, we would process all the subscriber orgs in batches (i.e. get one batch of results from the bulk query, add the PackagePushJobs for that batch using a bulk insert, repeat...) to minimize memory use but that will take more significant refactoring.

cumulusci/tasks/push/tasks.py
```diff
@@ -236,7 +240,7 @@ def _init_task(self):
 
     def _init_options(self, kwargs):
         super(SchedulePushOrgList, self)._init_options(kwargs)
-
+        self.options["dry_run"] = process_bool_arg(self.options.get("dry_run", False))
```
davisagli, line 243: I think the dry_run option is valuable for SchedulePushOrgQuery, in order to have a way to check how many orgs were selected. But I don't see the value for SchedulePushOrgList, where the user is providing the list. I would just set it to False here and not expose the option.

cumulusci/tasks/push/tasks.py
```diff
@@ -283,40 +287,43 @@ def _run_task(self):
                 "Scheduling push for %d minutes from now", delay_minutes
             )
             start_time = datetime.utcnow() + timedelta(minutes=delay_minutes)
+        if not self.options["dry_run"]:
+            self.logger.warning(
+                "Skipping push job creation due to dry_run flag enabled."
+            )
```
davisagli, line 293: This log line looks like it's in the wrong place -- it says dry_run is enabled in the case where it's not.

Also let's log at the info level, not warning, and be sure to output how many orgs are selected.

cumulusci/tasks/push/tasks.py
```diff
@@ -349,8 +356,21 @@ class SchedulePushOrgQuery(SchedulePushOrgList):
                 + " Ex: 2016-10-19T10:00"
             )
         },
+        "dry_run": {
+            "description": "If True, skip creating Github data.  Defaults to False"
```
davisagli, line 360:
```suggestion
            "description": "If True, output how many orgs were selected but do not actually create a push request. Defaults to False"
```

davisagli (changes requested):

cumulusci/tasks/push/push_api.py
```diff
@@ -249,7 +250,9 @@ def get_push_jobs_by_id(self, where=None, limit=None):
 class SalesforcePushApi(object):
     """ API Wrapper for the Salesforce Push API """
 
-    def __init__(self, sf, logger, lazy=None, default_where=None, batch_size=None):
+    def __init__(
+        self, sf, logger, lazy=None, default_where=None, batch_size=None, bulk=None
```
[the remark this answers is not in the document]
jofsky: @davisagli the way I've implemented it if an sobject and field_name parameters are passed into the return_query_records the bulk api will be used. If they're not set the rest api will be used. This allowed minimal code change and backwards compatibility. Also this way we will only use the bulk api on operations we expect to return large amounts of data such as subscribers.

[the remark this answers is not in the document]
davisagli: Okay, I see now. I think it's a little confusing and hard to tell when the bulk API is going to be used, but it's okay for now.

davisagli (approved):

PR #2334, opened 2021-01-20
New task "AddFieldsToFieldSet"

bethbrains:
# Critical Changes

# Changes
- A new Metadata ETL task, `add_fields_to_field_set`, allows adding fields to existing field sets.

# Issues Closed

PR #2340, opened 2021-01-21
Update History.rst and version.txt to release V3.27.0

prescod:
Update History.rst and version.txt to release V3.27.0


HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
 =======
 
-3.26.0 (2021-01-08)
+3.27.0 (2021-01-21)
 -------------------
 
+Changes:
+
+- Upgrades including Snowfakery 1.4 with 
+  min, max, round functions. PR #2335
```
davisagli, line 11: We haven't typically included PR links. I suppose I don't have a big objection to that, but you'll need to do a bit more reformatting to remove them when you post to the Hub and Trailblazer community. For the same reason, I usually turn on the word-wrap in my editor and avoid line breaks within paragraphs, because then they copy over better.

We don't typically mention routine upgrades of dependencies, with the exception of Snowfakery, so I would probably change this to "Updated Snowfakery to version 1.4 with ..."

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+- The ensure_record_types task has a new option, force_create,
+  which will create the Record Type even if other Record Types
+  already exist on the object. PR #2323
```
davisagli, line 15:
```suggestion
- The ``ensure_record_types`` task has a new option, ``force_create``,
  which will create the Record Type even if other Record Types
  already exist on the object. PR #2323
```

Task and option names should be double-backticked to render in a fixed-width font. There are others to adjust below...

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+- Allow num_records and num_records_tablename
+  to be omitted when using the task 
+  generate_and_load_from_yaml. PR #2322
+
+- CumulusCI displays more user friendly error message 
+  when encountering parsing errors in cumulusci.yml. #2311
```
davisagli, line 22: I would put this one under issues closed

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+- New Robot Keywords for Performance Testing: #2291
+
+    * Set Test Elapsed Time: This keyword captures a computed rather than measured elapsed time for performance-tests.
```
davisagli, line 26:
```suggestion
    * Set Test Elapsed Time: This keyword captures a computed rather than measured elapsed time for performance tests.
```

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+    * Start Perf Time, End Perf Time: start a timer and then store the result.
+
+    * Set Test Metric: store any test metric, not just elapsed time.
+
+  - CumulusCI now reports how long it took for flows to run. #2249
```
davisagli, line 32: Have you double-checked the reST rendering? This and the next two bullets look like they're indented differently than the previous top-level items.

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+  - Handle a possible gack while collecting info about installed packages #2299
```
davisagli, line 34: Should go under Issues closed

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+  - We fixed an issue causing the extract_dataset task to fail in some 
+    circumstances when both an anchor date and Record Types were used. #2300
```
davisagli, line 37: Should go under issues closed

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
```
davisagli, line 38: Don't forget to add https://github.com/SFDO-Tooling/CumulusCI/pull/2334 which I merged this morning.

HISTORY.rst
```diff
@@ -2,9 +2,50 @@
...
+
+Issues Closed:
+
+ - Fixed an error that could occur while cleaning cache directories.
+
+ - Potential bugs in the Push API support have been fixed.
```
davisagli, line 44:
```suggestion
 - Fixed potential bugs in the Push Upgrade tasks.
```

davisagli (changes requested):

HISTORY.rst
prescod: I wouldn't think of them as PR links but as PR numbers and it seems helpful to me across all channels? It allows someone (the technical subset) to investigate further what changed and why?

davisagli: I guess I'm okay with trying this. The audience for CumulusCI is less likely to be confused by it than our product customers.

davisagli (approved):

PR #2344, opened 2021-01-25
Stricter homebrew script.

prescod:
I think that the script should stop if something fails rather than making a corrupt PR which needs to be deleted.

I admittedly do not know how to test this change and its possible that this script is supposed to tolerate some forms of harmless "failures", so please take the PR as suggestive. If it doesn't make sense we can reject it.

### DIFF
--- a/utility/build-homebrew.sh
+++ b/utility/build-homebrew.sh
@@ -1,5 +1,5 @@
 #!/bin/sh
-
+set -e
 OUT_FILE="$1" || exit 1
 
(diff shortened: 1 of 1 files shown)


davisagli (changes requested):
Good call -- but perhaps we should go a step further and `set -euo pipefail`? https://wizardzines.com/comics/bash-errors/

Also, with this, we can remove all the `|| exit 1` below

davisagli (approved):

jstvz (approved):

PR #2345, opened 2021-01-25
Update AUTHORS.rst

prescod:
Update AUTHORS.rst to be more accurate.

Order is chronological by date of first contribution.


davisagli (commented):
In the new docset that's in progress, we are removing this entirely so that we don't have to maintain it. It can still be found in git commit history if someone is curious. Any objection?

prescod:
I don't feel strongly one way or the other. Mildly prefer to leave it in, I guess. But not a big deal either way.

davisagli:
If we do keep it, then:
- I doubt we want to encourage reaching out to individuals, so might want to remove the emails
- There are other people who have made contributions and it feels misleading to leave them out

prescod:
I agree. 

Okay I'll just leave the decision about what to do with this PR to you, and you and Brandon, Jason or whoever can make the final decision about the new docset.

Br4nd0R:
I don't mind keeping it :) It is one more thing that will need periodic updating, but seems like a low lift.

jstvz:
If we keep it, we can update CONTRIBUTING to tell new contributors to add themselves if they want to be included. This would shift the maintenance burden to contributors and allow them to decide how they'd like to be included (e.g., full name vs github username).

codexmarcel (approved):

davisagli (changes requested):
I would like to remove the emails. We want to point people to our official channels for support, not encourage reaching out to individuals.

Br4nd0R (dismissed):

AUTHORS.rst
```diff
@@ -7,17 +7,28 @@ Development Team
...
-* Bryan Oakley <boakley@salesforce.com>
+* Jason Lantz
+* David Glick
+* James Estevez
+* Paul Prescod
+* Bryan Oakley
+* David Reed
+* Brandon Parker
+* Josh Kofsky
+* Marcel Pufal
```
davisagli, line 18:
```suggestion
* Marcel Pufal
* David Ray
```

davisagli (approved):

PR #2346, opened 2021-01-26
Feature/robot locator manager

boakley:
Prior to this change, using a custom locator that is not on a page would always throw an error even if you didn't expect it to be on a page (eg: `Page should not contain     custom:locator`). This changes the locator manager so that it lets the keyword using the locator be responsible for throwing the error.

As an unfortunate side effect, the translated locator isn't included as part of the error message. Therefore, every time a custom locator is used, the translated locator will be logged to aid in debugging.

# Critical Changes

# Changes

# Issues Closed


davisagli (approved):
I looked at seleniumlibrary's ElementFinder.find and agree with your analysis that a location strategy should return something falsy rather than raising an exception if the element wasn't found.

PR #2356, opened 2021-02-02
A couple small tweaks to locators for Spring '21.

boakley:
Small spring '21 locator changes. Thanks to Sravani for bringing these to our attention and offering suggested solutions.  Even without these changes our own test suites were passing, but NPSP in spring '21 is using some new markup that was causing a couple of keywords to fail. 

I was unable to find a scenario that would cause our own tests to fail without these fixes, so I haven't added any new test cases. However, our own tests still pass with these changes, and I manually verified that the failing NPSP tests were fixed with these changes. 

# Critical Changes

# Changes

* `Click related list button` keyword has been modified to be more liberal in the types of DOM elements it will click on. Prior to this change it only clicked on anchor elements, but now also works for related list buttons that use an actual button element.
* `Click modal button` keyword now attempts to find the given button anywhere on the modal rather than only inside a `force-form-footer` element.

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/locators_51.py
+++ b/cumulusci/robotframework/locators_51.py
@@ -3,9 +3,7 @@ import copy
 
 lex_locators = copy.deepcopy(locators_50.lex_locators)
 
-lex_locators["modal"][
-    "button"
-] = "//div[contains(@class,'uiModal')]//force-form-footer//button[.='{}']"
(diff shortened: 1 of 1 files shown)


jofsky (approved):
Is the tuple starting on line 19 in cumulusci/robotframework/locators_51.py supposed to be comma separated, or is this intentional?

boakley:
It's intentional. It's not a tuple, it's a string expression. Makes it easy
to split long strings across multiple lines.

PR #2357, opened 2021-02-02
Add version_base option for the 2gp create_package_version task

davisagli:
W-8839087

In addition to the change noted below, this refactors some of the package version number operations into a PackageVersionNumber class instead of methods on the task.

# Critical Changes

# Changes
- The `create_package_version` has a new option, `version_base`, which can be used to increment the package version from a different base version instead of the highest existing version of the 2gp package. The `build_feature_test_package` flow now uses this option to create a package version with the minor version incremented from the most recent 1gp release published to github.

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -1056,6 +1056,8 @@ flows:
                 options:
                     package_type: Managed
                     package_name: $project_config.project__package__name Managed Feature Test
+                    version_base: latest_github_release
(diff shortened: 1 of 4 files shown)


davidmreed (approved):
I tested with several variations (packages building multiple 2GPs against the same released version, packages with no releases) and this looks great.

Br4nd0R (approved):

PR #2366, opened 2021-02-04
Set post-install and uninstall script for 2gp managed packages

davisagli:
W-8847899

# Critical Changes

# Changes
- The `create_package_version` task now supports setting a post-install script and uninstall script when creating a managed package version, by setting the `post_install_script` and `uninstall_script` options. By default, these options will use the values of `install_class` and `uninstall_class` from the package section of `cumulusci.yml`.

# Issues Closed


Br4nd0R (approved):
Nice fix! All looks good.

PR #2371, opened 2021-02-05
New preflight check task - recordtypes.GetHasRecordTypes for Sobject

ClayTomerlin:
Proposed feature will provide a preflight check task that returns an org's existing RecordTypes.
This will allow for preflight checks to confirm if any RecordTypes currently exist for an SobjectType. Will also also verification of the existence of a specific RecordType by 'DeveloperName'. 

# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/preflight/recordtypes.py
```diff
@@ -0,0 +1,21 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+
+class GetRecordTypesForSobject(BaseSalesforceApiTask):
+    task_options = {
+        "sobject": {
+            "description": "The sObject for which we want check if any recordtypes exist",
```
davidmreed, line 6:
```suggestion
            "description": "The API name of the sObject to check for Record Types.",
```

cumulusci/tasks/preflight/recordtypes.py
```diff
@@ -0,0 +1,21 @@
...
+            "required": True,
+        }
+    }
+
+    def _run_task(self):
+        sobject = self.options["sobject"]
+        results = self.tooling.query(
+            f"Select  Name FROM RecordType WHERE SobjectType = '{sobject}'"
```
davidmreed, line 14: Question: I know that many Tooling API queries cannot be bulkified. Does this one successfully fetch, say, all Record Types? If it does, we could potentially increase performance and flexibility of this task by digesting all Record Types into a dict, and then let the `when` expression specify the object rather than a task option (e.g., `when: "Opportunity" not in tasks.check_sobject_record_types()`).

```suggestion
            f"Select Name FROM RecordType WHERE SobjectType = '{sobject}'"
```

cumulusci/tasks/preflight/recordtypes.py
```diff
@@ -0,0 +1,21 @@
...
+        )["records"]
+        self.return_values = bool(results)
+        self.logger.info(
+            f"Found existing recordtype for {sobject}: {self.return_values}"
+        )
+        if self.return_values:
```
davidmreed, line 20: I suggest just keeping the log above and dropping this one as redundant.

davidmreed:
@ClayTomerlin  Thanks for submitting a first contribution! I added some suggestions and one question about a possible tweak to the design.

To make this task fully functional, we need to add an entry in `cumulusci/cumulusci.yml`. Could you add one? You can copy the format of an existing preflight check task there, using the class path of this class instead. Then, run `make docs` to update the task documentation, and commit the results.

cumulusci/tasks/preflight/recordtypes.py
ClayTomerlin: This is a great idea and my original thought - but wasnt sure how we measure expense. This query will fetch all existing RecordTypes. From what I could tell that probably wont ever be a 'huge' number, but at the same time I dont see many hard limits to how many recordtypes per object one could have. 
Will make this suggested change!

davidmreed: @ClayTomerlin  I think the pattern you can best imitate, if you'd like a pattern, is in the CheckSobjectsAvailable task, where `return_values` is a dict. You can make the keys sObject names and the values a list of Record Types.

ClayTomerlin: @davidmreed 
That makes sense, is there any way to return more than the sObjectType names (basically more than the one field in the query)? This would allow expansion to also confirm a 'specific' recordtype exists. 
For now Ill make it simple

davidmreed: Yeah, the route noted above can do that. If the return value has a structure like this:

`{ 'Opportunity': ['MyRTDeveloperName']}`

you can write `when` clauses like

`when: 'Account' not in tasks.check_sobject_record_types()`

and

`when: 'Opportunity' in tasks.check_sobject_record_types() and 'MyRTDeveloperName' in tasks.check_sobject_record_types()['Opportunity']`

So you cover both bases - checking for _any_ record type, and checking for a _specific_ record type.

cumulusci/cumulusci.yml
```diff
@@ -252,6 +252,10 @@ tasks:
         description: Retrieves a list of the currently available Permission Set License definition keys
         class_path: cumulusci.tasks.preflight.licenses.GetAvailablePermissionSetLicenses
         group: Salesforce Preflight Checks
+    get_existing_recordtypes:
```
davidmreed, line 255:
```suggestion
    get_existing_record_types:
```

Suggesting a task name that I think fits `snake_case` a little better but I don't feel strongly about it.

cumulusci/cumulusci.yml
```diff
@@ -252,6 +252,10 @@ tasks:
...
+        description: "Retrieves a list of all RecordTypes that exist in the format SobjectType: [DeveloperName, DeveloperName2]"
```
davidmreed, line 256:
```suggestion
        description: "Retrieves all Record Types in the org as a dict, with sObject names as keys and lists of Developer Names as values."
```

cumulusci/tasks/preflight/recordtypes.py
```diff
@@ -0,0 +1,15 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from collections import defaultdict
+
+
+class CheckSObjectRecordTypes(BaseSalesforceApiTask):
+    def _run_task(self):
+        rts = defaultdict(list)
+        records = self.tooling.query("Select SObjectType,FullName FROM RecordType")[
```
davidmreed, line 8:
```suggestion
        records = self.tooling.query("Select SobjectType, FullName FROM RecordType")[
```

Suggesting that we use the capitalization from the API docs (as below) because otherwise we have to remember that the schema's case-insensitive in the API and case-sensitive when represented in Python.

cumulusci/tasks/preflight/recordtypes.py
```diff
@@ -0,0 +1,15 @@
...
+            "records"
+        ]
+        for r in records:
+            rts[r["SobjectType"]].append(r["FullName"].split(".")[1])
+
+        self.return_values = rts
+        self.logger.info(f"Found existing recordtypes: {self.return_values}")
```
davidmreed, line 15:
```suggestion
        self.logger.info(f"Found existing Record Types: {self.return_values}")
```

cumulusci/tasks/preflight/tests/test_recordtypes.py
```diff
@@ -0,0 +1,31 @@
+from unittest.mock import Mock
+import unittest
+
+from cumulusci.tasks.preflight.recordtypes import CheckSObjectRecordTypes
+from cumulusci.tasks.salesforce.tests.util import create_task
+
+
+class TestRecordTypePreflights(unittest.TestCase):
+    def test_recordType_preflight(self):
```
davidmreed, line 9:
```suggestion
    def test_record_type_preflight(self):
```

cumulusci/tasks/preflight/tests/test_recordtypes.py
```diff
@@ -0,0 +1,31 @@
...
+                {
+                    "SobjectType": "ActionPlanTemplate",
+                    "FullName": "ActionPlanTemplate.Default",
+                },
+            ],
+        }
+        task._run_task()
+
+        task.tooling.query.assert_called_once_with(
+            "Select SObjectType,FullName FROM RecordType"
```
davidmreed, line 26:
```suggestion
            "Select SobjectType, FullName FROM RecordType"
```

docs/tasks.rst
```diff
@@ -2139,6 +2139,20 @@ Command Syntax
 
 
 
+**get_existing_recordtypes**
```
davidmreed, line 2142: You'll have a merge conflict here because we're actually removing this file from the source tree to smooth documentation generation.

davidmreed (changes requested):

davidmreed (dismissed):

cumulusci/tasks/preflight/recordtypes.py
ClayTomerlin: Great reminder and important detail

davidmreed (approved):

cumulusci/tasks/preflight/recordtypes.py
```diff
@@ -0,0 +1,15 @@
+from cumulusci.tasks.salesforce import BaseSalesforceApiTask
+from collections import defaultdict
+
+
+class CheckSObjectRecordTypes(BaseSalesforceApiTask):
+    def _run_task(self):
+        rts = defaultdict(list)
+        records = self.tooling.query("Select SobjectType, FullName FROM RecordType")[
```
davisagli, line 8: This should use query_all instead of query, or it won't include all results when there are many.

davidmreed: Oof, big code review miss on my part. I'll throw up a quick PR to fix this.

PR #2372, opened 2021-02-06
remove "force" from "force:auth:sfdxurl:store

Julian88Tex:
auth was moved from force to it's own namespace. Even though there is an alias, I was having trouble getting this to run in Github Actions:
Screen Shot 2021-02-05 at 7 46 44 PM

SFDX docs: https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_auth_sfdxurl.htm

# Critical Changes

# Changes
- remove "force" from "force:auth:sfdxurl:store

# Issues Closed

### DIFF
--- a/docs/cookbook.rst
+++ b/docs/cookbook.rst
@@ -351,7 +351,7 @@ Tests workflow, use your editor to create a file named
            wget -qO- https://developer.salesforce.com/media/salesforce-cli/sfdx-linux-amd64.tar.xz | tar xJ -C sfdx --strip-components 1
            ./sfdx/install
            echo ${{ secrets.SFDX_AUTH_URL }} > sfdx_auth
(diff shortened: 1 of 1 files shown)


davisagli:
@Julian88Tex Thanks -- this may indicate a bug in sfdx's backwards compatibility aliases, but either way it makes sense to update this to the new recommended commands.

Can you create a new PR using a branch within the SFDO-Tooling organization instead of your fork? We don't run builds for PRs from forks, so they can't meet the criteria for merging.

Julian88Tex:
@davisagli Yes would be happy to however as best I can tell I do not have the privileges to do so. I'm unable to create a branch on this repo. Please let me know if you had something else in mind. Thanks!

PR #2373, opened 2021-02-06
Fix  "CCI org connect takes a long time on Safari"

prescod:
We can see the timeout being respected here (client is Safari)

```
10:59:29.872775 call       694     def readinto(self, b):
10:59:29.873142 line       702         self._checkClosed()
10:59:29.873305 line       703         self._checkReadable()
10:59:29.873435 line       704         if self._timeout_occurred:
10:59:29.873587 line       707             try:
10:59:29.873708 line       708                 return self._sock.recv_into(b)
10:59:32.876480 exception  708                 return self._sock.recv_into(b)
Exception:..... socket.timeout: timed out
10:59:32.877447 line       709             except timeout:
10:59:32.877754 line       710                 self._timeout_occurred = True
10:59:32.878007 line       711                 raise
```

Without a timeout set, the wait time seems to be around 52 seconds (?) which must be based on some timeout in Safari's networking layer or the operating system or elsewhere.

Why the timeout is relevant seems (based on my limited TCP knowledge) to be caused by the fact that Chrome (and hopefully other browsers) send a TCP PSH after sending the request and Safari does not. I don't know why it does not.

Chrome:

image

Safari:

image

Note the long gap between line 20 and 33.


davisagli:
Wish I understood what was happening here, but probably makes sense to go with this pragmatic solution rather than spend more time on it. To get green builds we'll need the CI fixes that I already made in #2374 

I guess Safari is trying to use a persistent TCP connection and our server is not handling that in some way? I'm guessing that it wants to make another request on the same connection to get the favicon while other browsers open multiple connections.

davisagli (approved):

PR #2374, opened 2021-02-07
Improve docs build

davisagli:
- Remove autogenerated API docs which weren't working
- Remove docs/tasks.rst and docs/flows.rst from the repository; they no longer need to be updated manually, but will be generated at the beginning of the Sphinx build.
- Fixed `cci flow doc` to work even when not run in the root folder of the CumulusCI repository
- This is a readthedocs setting change rather than part of this PR, but I'm going to switch its "Default version" to `stable` so that changes to the docs aren't visible until they are part of a release tag.

# Critical Changes

# Changes

# Issues Closed


davisagli:
Sigh, looks like there's some issue with the latest sfdx

davidmreed (approved):

PR #2381, opened 2021-02-10
Feature/remove force namespace sfdxurl auth

Julian88Tex:
auth was moved from force to it's own namespace. Even though there is an alias, I was having trouble getting this to run in Github Actions:
Screen Shot 2021-02-05 at 7 46 44 PM

SFDX docs: https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_reference.meta/sfdx_cli_reference/cli_reference_auth_sfdxurl.htm

# Critical Changes

# Changes
- remove "force" from "force:auth:sfdxurl:store"

# Issues Closed


davisagli (approved):
Thanks for updating this @Julian88Tex

PR #2384, opened 2021-02-11
Display Response Body on 400 errors

Br4nd0R:
# Changes
* Enhanced error messaging on 400 errors encountered when acquiring a JWT session.

Thoughts on mocking `encode()`? I know, I know; bad practice to mock something that really doesn't need to be mocked. But, without an actual dummy key value (i.e. the contents of a genuine server.key file) I wasn't able to get past that method, and it didn't feel right needing to commit private key value somewhere, even if it's just a throwaway.

### DIFF
--- a/cumulusci/oauth/salesforce.py
+++ b/cumulusci/oauth/salesforce.py
@@ -18,7 +18,7 @@ import time
 import socket
 
 from cumulusci.oauth.exceptions import SalesforceOAuthError
-from cumulusci.core.exceptions import CumulusCIUsageError
+from cumulusci.core.exceptions import CumulusCIException, CumulusCIUsageError
 from cumulusci.utils.http.requests_utils import safe_json_from_response
 
 HTTP_HEADERS = {"Content-Type": "application/x-www-form-urlencoded"}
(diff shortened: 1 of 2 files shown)


cumulusci/oauth/salesforce.py
```diff
@@ -74,6 +78,9 @@ def jwt_session(client_id, private_key, username, url=None, auth_url=None):
     headers = {"Content-Type": "application/x-www-form-urlencoded"}
     token_url = urljoin(url, "services/oauth2/token")
     response = requests.post(url=token_url, data=data, headers=headers)
+    if response.status_code == 400:
+        raise CumulusCIException(ERROR_MESSAGE_400.format(response.text))
```
davisagli, line 82: I think it would make sense to use the same logic that we do for checking the response when we do the refresh token flow, and raise the same exception class: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/config/OrgConfig.py#L106-L109

davisagli (changes requested):
I'm fine with mocking jwt.encode like that; there are some existing tests in test_config.py that do the same thing.

davisagli (approved):

Issue #2389, opened 2021-02-12
Add org name to flow success message


Julian88Tex:
**Please describe the nature of this feature request**
- Add org name to flow success message `Completed flow 'regression_org' successfully!`

I often have to scroll up to verify which org I ran a flow on so it would be helpful to have it in the success message.

davisagli:
@Julian88Tex good idea. Here's where that line is produced if you want to try to fix it yourself: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/flowrunner.py#L418 ... should be a matter of including `org_config.name`

Julian88Tex:
@davisagli thanks! just created the PR. Any chance there's a good doc on setting up CumulusCI locally for development so I can test it out?

davisagli:
@Julian88Tex You can tell us whether it's good or not, but that's what https://github.com/SFDO-Tooling/CumulusCI/blob/main/CONTRIBUTING.rst is for

davisagli:
Fixed in #2390

PR #2390, opened 2021-02-12
#2389 Add org name to flow success

Julian88Tex:
# Critical Changes

# Changes
- add org's name to flow success message

# Issues Closed

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -415,7 +415,7 @@ class FlowCoordinator(object):
             for step in self.steps:
                 self._run_step(step)
             flow_name = f"'{self.name}' " if self.name else ""
-            self.logger.info(f"Completed flow {flow_name} on org {org_config.name}successfully!")
+            self.logger.info(f"Completed flow {flow_name}on org {org_config.name} successfully!")
         finally:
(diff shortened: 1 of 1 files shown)


cumulusci/core/flowrunner.py
```diff
@@ -415,7 +415,7 @@ def run(self, org_config):
             for step in self.steps:
                 self._run_step(step)
             flow_name = f"'{self.name}' " if self.name else ""
-            self.logger.info(f"Completed flow {flow_name}successfully!")
+            self.logger.info(f"Completed flow {flow_name} on org {org_config.name}successfully!")
```
davisagli, line 418: Notice from the line above that flow_name may either be the name with a space afterward, or an empty string if this flow has no name. So for the spacing to work out right I think we need:

```suggestion
            self.logger.info(f"Completed flow {flow_name}on org {org_config.name} successfully!")
```

davisagli (changes requested):

cumulusci/core/flowrunner.py
Julian88Tex: @davisagli ah thanks updated

davisagli (approved):

PR #2391, opened 2021-02-16
Feature/robot run task cross project

boakley:
This fixes the `Run Task` keyword to initialize the task with the proper project config (the project config of the task if it has one, otherwise the current project config)

This includes some pytest tests. However, I'm not super confident they are the best way to write these tests. I could definitely use some advice on how to write pytests that need to set up a repository or define a task.

I put this new pytest under the robotframework folder, which is _not_ where a similar pytest for the salesforce library lives. The pytest for the salesforce library is under the core folder but I don't think that's the right place. I wanted to keep this PR focused on the problem at hand. If you, the kind reviewer, agrees that this is the right place for these new tests, I'll move the salesforce tests out of the core folder to live alongside this new test file for consistency.

# Critical Changes

# Changes
* Run Tasks robot keyword now uses the task-specific project config when running a cross-project task
* SalesforceLibrary has a new keyword, `Scroll Element Into View`, which is more reliable on Firefox than the keyword of the same name in SeleniumLibrary
* ObjectManagerPageObject has been made more stable

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/tests/test_cumulusci_library.py
+++ b/cumulusci/robotframework/tests/test_cumulusci_library.py
@@ -65,11 +65,13 @@ class TestCumulusCILibrary(MockLoggerMixin, unittest.TestCase):
 
             with mock.patch.object(self.cumulusci, "_run_task"):
                 self.cumulusci.run_task("example:whatever")
-                task = self.cumulusci._run_task.mock_calls[0].args[0]
+
+                args, kwargs = self.cumulusci._run_task.call_args
+                self.assertEqual(len(args), 1)
 
                 # make sure it's not using the current project config
                 # for the task, and that the config it _is_ using is
                 # rooted in the directory we created
+                task = args[0]
                 self.assertNotEqual(task.project_config, self.cumulusci.project_config)
                 self.assertEqual(tmpdir, Path(task.project_config.repo_root))
-                print("bruh")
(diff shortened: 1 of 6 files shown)


cumulusci/robotframework/tests/test_cumulusci_library.py
```diff
@@ -0,0 +1,77 @@
...
+                        "options": {
+                            "command": "pwd",
+                        },
+                    },
+                },
+                "sources": {
+                    "example": {"path": "/tmp"},
+                },
+            },
+        )
```
davisagli, line 30: This looks fine as a way of setting up the project config. We have some utils around that provide a more shorthand way to do it for some tests (search for create_project_config), but I don't that would work here since the test covers getting the task from the project_config, so it needs to be set up including a task config.

cumulusci/robotframework/tests/test_cumulusci_library.py
```diff
@@ -0,0 +1,77 @@
...
+                    tasks:
+                        whatever:
+                            class_path: cumulusci.tasks.command.Command
+                            options:
+                                command: pwd
+                    """
+                )
+
+            with mock.patch.object(self.cumulusci, "_run_task"):
+                self.cumulusci.run_task("example:whatever")
```
davisagli, line 67: If the library used a separate (private) method for constructing the task, you could test that here instead of run_task, and then you wouldn't need to mock anything.

davisagli (dismissed):
I agree this is a better place for unit tests of the robot libraries. We try to keep unit tests close to the module that they are testing.

davisagli (approved):
_Thank you_! Two green checks in a row looks like a good sign.

PR #2393, opened 2021-02-16
Add dry_run option to uninstall_packaged_incremental

davidmreed:
# Critical Changes

# Changes

- The `uninstall_packaged_incremental` and `uninstall_packaged` tasks now support a `dry_run` option, which outputs the destructiveChanges package manifest to the log instead of executing it.

# Issues Closed


davisagli (approved):
Good idea.

PR #2405, opened 2021-02-19
Centralize YAML parsing and error reporting

prescod:
# Changes

Unify error handling for mapping.yml files and cumulusci.yml.

Change the error messaging to be more IDE friendly.


cumulusci/core/exceptions.py
```diff
@@ -267,3 +267,7 @@ class PackageUploadFailure(CumulusCIFailure):
 
 class PackageInstallError(Exception):
     """Error installing a 2GP package version"""
+
+
+class YAMLParseException(CumulusCIException):
+    """Error parsing a YAML File"""
```
davisagli, line 273: Should this be a subclass of ConfigError?

prescod: 80b659be8b8da28a3984437abbb2e3217acb5ac2

davisagli: @prescod on second thought, probably not. ConfigError assumes you're parsing something that has a config_name, but not all YAML is a cci config file.

prescod: I reverted but I note that a) ConfigError seems to never be instantiated with two arguments and b) its output doesn't seem to be intelligent about what to do if the config_name is missing. So there's a broader cleanup that should be done and it could include unifying YAMLParseException and ConfigError and making the config_name optional in a useful way.

Do you want me to go ahead and do that or just approve this and fix it someday?

davisagli: I'll approve and leave that up to you whether to do now or later.

davisagli (approved):

PR #2406, opened 2021-02-19
Add pytest-random-order.

prescod:
Make random-order testing easy.

Do we need a 3PP for dev requirements?

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -178,3 +178,19 @@ still be quite slow compared to normal unit tests. Nevertheless, if you are chan
 these tests, you should run them periodically.
 
 Do not commit the files ("large_cassettes/\*.yml") to the repository.
+
+Randomized tests
+~~~~~~~~~~~~~~~~
+
+Tests should be executable in any order. You can run this command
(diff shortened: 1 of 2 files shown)


Br4nd0R (approved):

CONTRIBUTING.rst
```diff
@@ -179,3 +179,19 @@ still be quite slow compared to normal unit tests. Nevertheless, if you are chan
...
 
 Do not commit the files ("large_cassettes/\*.yml") to the repository.
+
+Randomized tests
+~~~~~~~~~~~~~~~~
+
+Tests should be executable in any order. You can run this command
+a few times to verify if they are:
+
+    pytest --random-order
```
davisagli, line 189: Should we add this to the default options, or is it not stable enough yet?

davisagli (commented):
I thought there was an exception to 3pp approvals for test-only dependencies, but I'm not finding it now. Probably safest to submit the request.

CONTRIBUTING.rst
prescod: Not stable enough.

PR #2407, opened 2021-02-19
Add add_page_layout_fields Metadata ETL task

BrianKSFDO:
# Critical Changes

# Changes
Added new metadata etl feature to layouts.py that enables user to add or move fields to a specified Page Layout section.  Note - if a field is labeled as Required or Read-Only in Section A, and user runs task to move to Section B, it will lose those field properties.

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -11,6 +11,12 @@ tasks:
         class_path: cumulusci.tasks.metadata_etl.AddRelatedLists
         options:
             namespace_inject: $project_config.project__package__namespace
+    add_page_layout_fields:
+        group: Metadata Transformations
+        description: Adds specified Fields to one Page Layout.
+        class_path: cumulusci.tasks.metadata_etl.AddFields
+        options:
+            namespace_inject: $project_config.project__package__namespace
     add_standard_value_set_entries:
         group: Metadata Transformations
         description: Adds specified picklist entries to a Standard Value Set.
--- a/cumulusci/tasks/metadata_etl/layouts.py
+++ b/cumulusci/tasks/metadata_etl/layouts.py
@@ -68,3 +68,70 @@ class AddRelatedLists(MetadataSingleEntityTransformTask):
             elem.append("customButtons", button)
 
         elem.append("relatedList", text=related_list)
+
+
+class AddFields(MetadataSingleEntityTransformTask):
+    entity = "Layout"
+    task_options = {
+        "layout_section": {
+            "description": "Which layout section the fields should be added to, default is Information",
+            "required": False,
+        },
+        "fields": {
+            "description": "Array of field API names to append to specified layout section",
+            "required": True,
+        },
+        **MetadataSingleEntityTransformTask.task_options,
+    }
+
+    def _transform_entity(
+        self, metadata: MetadataElement, api_name: str
+    ) -> Optional[MetadataElement]:
+
+        # Set layout section and default to "Information" section if not passed in
+        layout_section = self._inject_namespace(self.options.get("layout_section") or "Information")
+
+        # Set fields to be added to layout section
+        adding_fields = [
+            self._inject_namespace(field)
+            for field in process_list_arg(self.options.get("fields", []))
+        ]
+
+        self._remove_existing_fields(metadata, api_name, adding_fields)
+
+        self._set_adding_fields(metadata, api_name, layout_section, adding_fields)
+
+        return metadata
+
+    ## If fields passed in that already exist in the Page Layout, remove from current sections
+    def _remove_existing_fields(self, metadata, api_name, adding_fields):
+
+        ## Need to traverse all layoutSections->layoutColumns->layoutItems->fields to find if item/field already exists
+        layout_sections = metadata.findall("layoutSections")
+        for section in layout_sections:
+            layout_columns = section.findall("layoutColumns")
+            for col in layout_columns:
+                layout_items = col.findall("layoutItems")
+                for item in layout_items:
+                    if item.field.text in adding_fields:
+                        # remove that item from existing layout to be re-added
+                        col.remove(item)
+
+    def _set_adding_fields(self, metadata, api_name, layout_section_name, adding_fields):
+        ## If no fields are passed in, do nothing
+        if len(adding_fields) == 0:
+            return
+
+        # Find specified layout sections to add fields to
+        layout_section_element = metadata.find("layoutSections", label=layout_section_name)
+        
+        # If layout section doesn't exist, exit gracefully with appropriate message
+        if layout_section_element is None:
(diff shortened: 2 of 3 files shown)


BrianKSFDO:
Here is how I tested this:
### Test 1 - add new and existing fields to an existing Layout Section

### Test 2 - if user passes in non-existent Layout Section, error with appropriate message

### Test 3 - When 'layout_sections' value is not passed in, default to Information section

### Test 4 - If specified 'field' is existing in any of the page layout sections, remove from existing layout and add to section identified by User

### Test 5 - send in multiple Fields

### Regresision Test 6 - Section order must remain:
# Information
# Relationships
# Wellness Status
# Employment Status
# Contact Information
# Employee Profile
# System Information
# Custom Links

BrianKSFDO:
My only concern is trying to move FirstName field on Employee object gave the following error "Could not process MDAPI response: Update of Layout Employee-Employee Layout: Error: Field:FirstName must be Required".

I wonder if system required fields on Standard objects will always create this error.

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -68,3 +68,70 @@ def _create_new_related_list(self, metadata, api_name, related_list):
...
+        
+        # If layout section doesn't exist, exit gracefully with appropriate message
+        if layout_section_element is None:
+            raise CumulusCIException("Layout section doesn't exist, task failed")
+
+        layout_columns_element = layout_section_element.find("layoutColumns")
+
+        for f in adding_fields:
+            layout_items_element = layout_columns_element.append("layoutItems")
+            layout_items_element.append("behavior", "Edit")
```
BrianKSFDO, line 136: If I change `Edit` to `Required` then system required fields like LastName will pass.  I wonder why EmployeeNumber didn't error out.

cumulusci/cumulusci.yml
```diff
@@ -11,6 +11,12 @@ tasks:
         class_path: cumulusci.tasks.metadata_etl.AddRelatedLists
         options:
             namespace_inject: $project_config.project__package__namespace
+    add_page_layout_fields:
+        group: Metadata Transformations
+        description: Adds specified Fields to one Page Layout.
+        class_path: cumulusci.tasks.metadata_etl.AddFields
+        options:
+            namespace_inject: $project_config.project__package__namespace
```
davidmreed, line 19: @BrianKSFDO We have a pattern for inferring this option now, so it's okay to leave it out of the task definition (more below).

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -68,3 +68,70 @@ def _create_new_related_list(self, metadata, api_name, related_list):
             elem.append("customButtons", button)
 
         elem.append("relatedList", text=related_list)
+
+
+class AddFields(MetadataSingleEntityTransformTask):
+    entity = "Layout"
+    task_options = {
+        "layout_section": {
+            "description": "Which layout section the fields should be added to, default is Information",
```
davidmreed, line 77: Is this sensitive to localization? I.e., can this default fail in an org whose default language is Spanish?

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -68,3 +68,70 @@ def _create_new_related_list(self, metadata, api_name, related_list):
...
+            "required": False,
+        },
+        "fields": {
+            "description": "Array of field API names to append to specified layout section",
```
davidmreed, line 81:
```suggestion
            "description": "List of field API names to append to specified layout section",
```

I suggest "list" because it's more non-developer friendly.

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -68,3 +68,70 @@ def _create_new_related_list(self, metadata, api_name, related_list):
...
+            for field in process_list_arg(self.options.get("fields", []))
+        ]
+
+        self._remove_existing_fields(metadata, api_name, adding_fields)
+
+        self._set_adding_fields(metadata, api_name, layout_section, adding_fields)
+
+        return metadata
+
+    ## If fields passed in that already exist in the Page Layout, remove from current sections
```
davidmreed, line 106: I'd be interested to hear more about this piece. In many Metadata ETL situations, we deliberately do nothing if the configuration already exists. For example, if the field is already on the page layout, moving it to the end of a section _seems_ relatively unlikely to me to be the right outcome. Is there a specific use case that needs this mechanic, as opposed to simply leaving it where it is?

cumulusci/tasks/metadata_etl/layouts.py
davidmreed: What do you think about asking the Describe API if the field is required, so that you can supply a smarter default here?

cumulusci/tasks/metadata_etl/layouts.py
davidmreed: Even if user-specified, do we have the same language dependency? What would you think about an index-based approach - specify say the first or last section, without naming it, either as the primary target or as a fallback if the specified section does not exist?

cumulusci/tasks/metadata_etl/layouts.py
```diff
@@ -68,3 +68,70 @@ def _create_new_related_list(self, metadata, api_name, related_list):
...
+    def _set_adding_fields(self, metadata, api_name, layout_section_name, adding_fields):
+        ## If no fields are passed in, do nothing
+        if len(adding_fields) == 0:
+            return
+
+        # Find specified layout sections to add fields to
+        layout_section_element = metadata.find("layoutSections", label=layout_section_name)
+        
+        # If layout section doesn't exist, exit gracefully with appropriate message
+        if layout_section_element is None:
```
davidmreed, line 129: This is highly vulnerable to end-user changes, since sections can be removed from managed page layouts. Can we define (as alluded to above) fallback mechanics, or allow the task to define a "no-action" path that is not an error?

cumulusci/tasks/metadata_etl/layouts.py
davidmreed: Is the behavior the same with custom fields that are required at the database level?

davidmreed (commented):
@BrianKSFDO My comments/questions here are primarily about design rather than code. Happy to discuss further.

We'll need full test coverage in order to merge. I'm happy to offer some examples and guidance on how we usually test these classes as well.

cumulusci/tasks/metadata_etl/layouts.py
BrianKSFDO: @davidmreed okay I see a few examples of the Describe API in the existing CumulusCI framework that I can try here.

cumulusci/cumulusci.yml
BrianKSFDO: I'll remove `namespace_inject` in my next commit 👍

Issue #2018, opened 2020-09-15
Include Login URL with "cci org info"


Julian88Tex:
**Please describe the nature of this feature request**

- Add login url (`secur/frontdoor.jsp?sid=<accesstoken>`) to `cci org info` or `cci org browser` as an option.

**Does your feature request solve a current problem? Please describe.**
It makes it quicker to get a login url. Currently the only way to access this url is through `sfdx force:org:open -r -u <username>`

**Additional context**
I think adding it to `cci org info` or at least `cci org info --json` is more useful than `cci org browser` since it's a more intuitive place even though it differs from the sfdx model of adding it to the login command.

davisagli:
I'd prefer to add it to `cci org browser` because `cci org info` is currently only a static listing of the information stored in the org config file and doesn't access the org at all. (We could discuss changing that to add other useful info, but it would make it harder to copy an org from the CLI to MetaCI using the info from `cci org info --json`.)

What's your use case for wanting the login url rather than immediately opening the browser with it?

Julian88Tex:
@davisagli Interesting you mention that since I did a `cci org info --json` and I noticed the access token used in `sfdx force:org:open -r -u <username>` is the exact same:

<img width="973" alt="Screen Shot 2020-09-16 at 5 39 48 PM" src="https://user-images.githubusercontent.com/459099/93406198-f2e7db80-f843-11ea-827d-48114cede7f7.png">

As far as I can tell, if the access token is static on `cci org info` then it seems to be static on `sfdx force:org:open -r -u <username>` too.

The use case is the same as `sfdx force:org:open -r -u <username>`. It is convenient to get a URL that can be shared with other for access to the org without giving them credentials. `cci org browser` also does not show this URL like `sfdx force:org:open` does:

<img width="744" alt="Screen Shot 2020-09-16 at 5 43 57 PM" src="https://user-images.githubusercontent.com/459099/93406319-34788680-f844-11ea-9651-fc3376cfd442.png">

davidmreed:
Tracking this feature request as W-8887568.

PR #2410, opened 2021-02-22
Add path/url options to org_browser

jstvz:
This PR adds:
- an option that mirrors SFDX's `--path` option to pass a URL path as the retURL URL parameter.
- an option, `url_only`, that will print `org_config.start_url` to stdout without opening a browser.

# Critical Changes

# Changes
- We added two new options to `cci org browser`: 
  - `--path`:  to allow navigating to a specific page after logging in. For example, passing `--path /lightning/setup/ImportedPackage/home` would navigate to the list of installed packages in Setup.
  - `--url_only`: that will print a login URL to the terminal without launching a browser.

# Issues Closed
- #2018


Br4nd0R (approved):
These options look quite handy. Code looks great!

PR #2414, opened 2021-02-23
Do not override existing project files on init.

### LINKED ISSUE #1325: Do not overwrite existing project files during init
Currently when you run `cci project init` it generates `README.md` and `.gitignore` files, which is great if you are setting CCI for the first time in a new project.  But if you are trying to add CCI as tooling for an existing project it may overwrite wanted files.
It would be better if these files were only generated when missing, or probably better still to be a question during init that checks for existing files, and when found, asks if the user wants to replace, ignore, or append the content.

davidjray:
# Critical Changes

# Changes
- Check is the project files already exist when running `cci project init`. If the files already exist, then display a message telling the user, and output the rendered file template.

# Issues Closed
- Fixes #1325


davidjray:
Example output of files already existing:
image

cumulusci/cli/cci.py
```diff
@@ -600,8 +600,20 @@ def project_init(runtime):
...
+        if not os.path.isfile(name):
+            with open(name, "w") as f:
+                f.write(template.render(**context))
+        else:
+            click.echo(
+                click.style(
+                    name
+                    + " file already exists. Please manually edit "
+                    + name
+                    + " to include:",
```
davisagli, line 612: This looks like a good opportunity to use an [f-string](https://realpython.com/python-f-strings/)

```suggestion
                    f"{name} already exists. Please manually edit {name} to include:"
```

davisagli: Also, I'd suggest this wording instead: "As a reference, here is what would be placed in the file if it didn't already exist" ... My thinking is that most of what we're generating is just reasonable defaults, and if the file already exists, there's a good chance it's been intentionally customized and our defaults wouldn't make sense.

cumulusci/cli/cci.py
```diff
@@ -600,8 +600,17 @@ def project_init(runtime):
     # Render templates
     for name in (".gitignore", "README.md", "cumulusci.yml"):
         template = env.get_template(name)
-        with open(name, "w") as f:
-            f.write(template.render(**context))
+        if not os.path.isfile(name):
```
Br4nd0R, line 603: I'd recommend using Pathlib's [is_file()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file) method instead of the `os` module for stuff like this.

cumulusci/cli/cci.py
davisagli: @davidjray Did you see my suggestion here re a different message to show the user?

davisagli (approved):
I tested and it works as expected. Nice work, thanks and congrats @davidjray

PR #2419, opened 2021-02-26
Improve `cci task run <task_name> --help`

Br4nd0R:
# Changes
* The `cci task run` command now includes the tasks configured description and any `task_docs` present on the task's class when passing the `--help` flag.


davisagli (commented):
But not `task_class.task_docs`?

Br4nd0R:
`task_docs` is welcome to come along, but some of them can look a little odd once rendered:
```
❯ cci task run command --help
Usage: cci task run command [OPTIONS]

  Run an arbitrary command

          **Example Command-line Usage:**         ``cci task run command -o
          command "echo 'Hello command task!'"``

          **Example Task to Run Command:**

          ..code-block:: yaml

              hello_world:                 description: Says hello world
              class_path: cumulusci.tasks.command.Command
              options:                 command: echo 'Hello World!'
.
.
.
```

davisagli:
@Br4nd0R for output to the terminal we would want to run it through rst2ansi the same way that `cci task info` does.

Br4nd0R:
`task_docs` are now present. I'm still noticing that `rst2ansi` doesn't play well with [the formatter](https://github.com/pallets/click/blob/c76fea1696c0ffe7edff8a36cadd4686cda8cbfb/src/click/core.py#L1188) used by click for things like code-blocks. It works fine when outputting via `click.echo`, but since we are assigning this help text via the `command.help` property the formatter is likely stripping out multiple tabs. That's my current working hypothesis...

davisagli:
@Br4nd0R The problem is that click is trying to [rewrap the text](https://click.palletsprojects.com/en/7.x/documentation/#preventing-rewrapping)

As another approach we could override `get_help` to handle the entire rendering:

```python
            cmd.get_help = lambda ctx: rst2ansi(
                doc_task(task_name, task_config).encode("utf-8")
            )
```

but that doesn't include the generic options for task run (--debug and such)

Br4nd0R:
> ...but that doesn't include the generic options for task run (--debug and such)

@davisagli - I think that `task run task_name --help` should include those global options, as `task info` doesn't currently output info on those, and I don't want to lose the only place where we tell the user they are valid options while running the tasks.

davisagli:
@Br4nd0R yes, I agree. So I think we ideally need to do more work on a solution that formats the task docs correctly _and_ includes all the generic options. But if you want to just take the task_docs back out here for now, I'm okay with that.

davisagli (approved):

PR #2422, opened 2021-02-28
Remove global RUNTIME

Br4nd0R:
# Changes
* We now have a new method for accessing a single instance of `CliRuntime` class: `CliRuntime.get_instance()`.

This is the first piece of work needed in order to refactor the commands contained in `cci.py` into their own (separate) files.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -156,21 +156,17 @@ def render_recursive(data, indent=None):
                 click.echo(f"{indent_str}{key_str} {value}")
 
 
-# global reference to active runtime
-RUNTIME = None
-
-
 def pass_runtime(func=None, require_project=True, require_keychain=False):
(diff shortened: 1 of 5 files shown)


prescod:
This is fine but David Glick had also come up with a nice pattern that does not require callers to call get_instance().

davisagli:
I agree with Paul that it would probably make sense to use the standard Python method for changing how an instance gets constructed, instead of inventing our own.

But I'm also not sure these changes gain us much. We could have accomplished the same goal (making it possible to split apart cci.py) simply by moving the RUNTIME global to a different module (to avoid import cycles).

I'd be more excited about a change so that the command functions always get the runtime object as an argument (filled in from the click context object by a decorator if it's not passed explicitly). I think this would make it possible to write tests with a lot less use of mock.patch

davisagli:
My suggestion for how to do this is in https://github.com/SFDO-Tooling/CumulusCI/pull/2533

PR #2425, opened 2021-03-02
Cleanup HTTPD Startup

prescod:
Previously, an inability to grab port 8080 would hang the tests forever.


davisagli (approved):
Would be even better to make it pick a free port to use for the test, but this is better than hanging for sure.

prescod:
I didn't know if there was some magic associated with the port 8080 somewhere in the code so I just left it.

davisagli:
Hopefully not, since cci listens to whatever port is configured in the connected app callback URL

PR #2427, opened 2021-03-03
Update homebrew formula to install using pip

davisagli:
This updates our script for generating a Homebrew formula to install using pip rather than setup.py install. This brings us more in line with how Python packages are typically installed and makes it possible to take advantage of binary wheels (which will be the easiest way to install the newest versions of cryptography).

The script uses the pip-compile tool from pip-tools to build a pip requirements file for cumulusci and its dependencies which includes the sha256 hashes from PyPI. This is embedded in a heredoc in the formula and written out to a requirements.txt file during installation before running pip install.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/utility/build-homebrew.sh
+++ b/utility/build-homebrew.sh
@@ -5,7 +5,8 @@ set -euo pipefail
 OUT_FILE="$1"
 
 ENV_DIR="$(mktemp -d)"
-RES_FILE="$(mktemp)" 
+REQS_IN_FILE="$(mktemp)"
+REQS_FILE="$(mktemp)"
 PYPI_JSON="$(mktemp)" 
 
 echo " "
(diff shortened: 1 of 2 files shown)


utility/build-homebrew.sh
```diff
@@ -44,21 +38,17 @@ class Cumulusci < Formula
...
+REQS
 
-    deps = resources.map(&:name).to_set
-    deps.each do |r|
-      resource(r).stage do
-        system "python3", *Language::Python.setup_install_args(libexec)
-      end
-    end
+    File.write("requirements.txt", reqs)
+    system "python3", "-m", "pip", "install", "-r", "requirements.txt", "--require-hashes", "--no-deps", "--prefix", libexec
```
jstvz, line 51: So [`setup_install_args`](https://github.com/Homebrew/brew/blob/b024954a0f203a4f439cc95eefb1180c96e943ce/Library/Homebrew/language/python.rb#L87) passes `--single-version-externally-managed` to `setup.py`. I don't understand what purpose it serves, so I'm unsure whether it's needed.

jstvz (approved):

utility/build-homebrew.sh
davisagli: `--single-version-externally-managed` tells setuptools to install the package in the form we're used to from pip (package folders plus a separate folder with the distribution metadata) instead of in an egg. The reason we're used to it is that pip started as a wrapper calling setuptools with this flag (see https://carljm.github.io/pipvirtualenv-preso/ for lots of details from ca. 2012), and I think it still does in certain legacy cases, but not when using wheels.

PR #2442, opened 2021-03-09
Use absolute paths for outputdir and robot_outputdir

prescod:
# Changes

robot_outputdir return value is now an absolute path.

This is implemented by making robot's outputdir option absolute in run_task. Then, when it is copied to robot_outputdir, it is still absolute.

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -441,7 +441,7 @@ tasks:
             revert_path: src.orig
         group: Salesforce Metadata
     robot:
-        description: Runs a Robot Framework test from a .robot file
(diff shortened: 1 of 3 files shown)


boakley (commented):
I'm curious why this is necessary.

prescod:
Cross-product flows change the current working directory so relative paths are ambiguous.

It broke one of my builds:

https://sfdo-metaci.herokuapp.com/builds/425987

And this fixed it:

```
class MonkeyPatchRobot(BaseTask):
    def _run_task(self, *args, **kwargs):
        def new_robot_run_task(self):
            rc = real_robot_run_task(self)
            self.return_values["robot_outputdir"] = str(Path(self.return_values["robot_outputdir"]).resolve())
        Robot._run_task = new_robot_run_task

```

davisagli (approved):

PR #2451, opened 2021-03-11
Docs about robot_outputdir

prescod:
`make docs` did not work on my computer so I hope this formats correctly.

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -23,6 +23,11 @@ from cumulusci.utils.xml.robot_xml import log_perf_summary_from_xml
 
 
 class Robot(BaseSalesforceTask):
+    task_docs = """Runs Robot test cases using a browser, if
+    necessary and stores its results in a directory. The
+    path to the directory can be retrieved from the
+    `robot_outputdir` return variable."""
+
     task_options = {
         "suites": {
             "description": 'Paths to test case files/directories to be executed similarly as when running the robot command on the command line.  Defaults to "tests" to run all tests in the tests directory',
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -23,10 +23,11 @@ from cumulusci.utils.xml.robot_xml import log_perf_summary_from_xml
 
 
 class Robot(BaseSalesforceTask):
-    task_docs = """Runs Robot test cases using a browser, if
+    task_docs = """
+    Runs Robot test cases using a browser, if
     necessary and stores its results in a directory. The
     path to the directory can be retrieved from the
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -23,6 +23,11 @@
 
 
 class Robot(BaseSalesforceTask):
+    task_docs = """Runs Robot test cases using a browser, if
```
davisagli, line 26: Needs consistent indenting in order to format correctly.

```suggestion
    task_docs = """
    Runs Robot test cases using a browser, if
```

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -23,6 +23,11 @@
...
+    necessary and stores its results in a directory. The
+    path to the directory can be retrieved from the
+    `robot_outputdir` return variable."""
```
davisagli, line 29: rST uses double backticks for literal code; single means italics.

```suggestion
    ``robot_outputdir`` return variable."""
```

davisagli (commented):
What error did you get? I had to delete the docs/api folder on mine in order to run it, since it was left over and out of date.

There are a couple formatting fixes which I'll make momentarily

prescod:
@davisagli 

$ pip install -r requirements_dev.txt && make docs

It seems like Sphinx is trying to probe some globals for docs and it triggers some unexpected behaviour.

```
  File "/Users/pprescod/.venvs/CCI-37/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py", line 584, in get_sourcename
    fullname = '%s.%s' % (self.object.__module__, self.object.__qualname__)
  File "/Users/pprescod/code/CumulusCI/cumulusci/core/template_utils.py", line 28, in __str__
    return self.func()
  File "/Users/pprescod/code/CumulusCI/cumulusci/core/template_utils.py", line 45, in <lambda>
    lambda *args, **kwargs: self.faker.format(name, *args, **kwargs)
  File "/Users/pprescod/.venvs/CCI-37/lib/python3.9/site-packages/faker/generator.py", line 76, in format
    return self.get_formatter(formatter)(*args, **kwargs)
  File "/Users/pprescod/.venvs/CCI-37/lib/python3.9/site-packages/faker/generator.py", line 86, in get_formatter
    raise AttributeError(msg)
AttributeError: Unknown formatter '__qualname__' with locale 'en_US'
```

Do we even still publish the Sphinx autodoc?

davisagli:
@prescod that's what I was getting too until I deleted docs/api

autodoc is different from the apidoc extension we were using, but I don't think we are actually using autodoc either and it would be fine to turn it off.

prescod:
Sorry I forgot that you had told me how to fix it. Yes its fixed for me now.

PR #2443, opened 2021-03-10
feat: adds the noancestors option to scratch org def

lionelarmanet:
# Critical Changes

# Changes

Adds a new option for scratch org definitions to specify the `noancestors` flag.

A note for the reviewers: I tried to find a unit test to verify that the flag is properly built and passed to dx. Ultimately, using a builder pattern could be handy to build the sfdx command string.

# Issues Closed
Closes #2441

PR #2452, opened 2021-03-12
Adds a noancestors option for scratch org configs

davisagli:
This is based on @lionelarmanet's contribution in #2443, but I also refactored how args for the sfdx force:org:create command are built to make that easier to test.

# Critical Changes

# Changes
- It is now possible to pass the `--noancestors` flag to sfdx when creating a scratch org by setting `noancestors: True` in the scratch org config in `cumulusci.yml`. Thanks @lionelarmanet

# Issues Closed

### DIFF
--- a/cumulusci/core/sfdx.py
+++ b/cumulusci/core/sfdx.py
@@ -4,6 +4,7 @@ import logging
 import platform
 import sarge
 import sys
+import typing as T
 
 from cumulusci.core.exceptions import SfdxOrgException
 
@@ -15,7 +16,7 @@ def sfdx(
     username=None,
     log_note=None,
     access_token=None,
-    args=None,
+    args: T.Optional[T.List[str]] = None,
     env=None,
     capture_output=True,
     check_return=False,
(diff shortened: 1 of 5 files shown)


cumulusci/core/sfdx.py
```diff
@@ -15,7 +16,7 @@ def sfdx(
     username=None,
     log_note=None,
     access_token=None,
-    args=None,
+    args: T.Optional[T.List[str]] = None,
```
Br4nd0R, line 19: Is there a specific reason you aren't importin `Optional` and `List` directly from `typing`? That's how I've seen it in our codebase for he most part, so just curious here.

Br4nd0R (approved):
Great refactor, and glad we have the new option to work with @lionelarmanet!

cumulusci/core/sfdx.py
davisagli: I am -- I imported typing as T so that each individual name doesn't need to be imported separately. It's an idiom @prescod has started using elsewhere.

prescod: Yeah it just gets tedious maintaining the import list. "No unions used in this file anymore, need to delete it. Now I need Optional. Need to add it."

Really, this is the kind of thing that import * was invented for but the taboo against it is so strong in the Python world that I don't dare buck the trend! (and linters would probably hate it)

PR #2453, opened 2021-03-12
configure the task to use the robot logger

boakley:
This adds a `logger` keyword argument to BaseTask, and updates the `run task` keyword to use it, causing task output to be captured in the robot logs.

Here's the robot log **with** the fix:

Screen Shot 2021-03-12 at 1 31 58 PM

Here's the robot log **without** the fix

Screen Shot 2021-03-12 at 1 33 01 PM

# Critical Changes

# Changes
* The `run task` keyword now captures all task output in the robot log instead of printing it to stdout.
* The command task now sets the `return_values` to a dictionary that contains the return code of the command that was run.
* 
# Issues Closed


davisagli:
@boakley It makes sense to try to include logs for tasks that are run using Robot Framework in the robot logs. Doing it this way will miss logs that are emitted while the task is initialized (such as from _init_options). How about if we make BaseTask.__init__ take an optional `logger` keyword argument, and if that's supplied, use it rather than setting up the default logger.

boakley:
@davisagli I've made the changes you requested.

davisagli (approved):

PR #2454, opened 2021-03-12
Promote a 2GP Package

Br4nd0R:
# Changes
* CumulusCI has a new task: `promote_2gp_package`. This task will promotes a `Package2Version` to the "IsReleased" state making it available for installation in production orgs. The task notifies users of any 1GP dependencies detected, as well as, any unpromoted 2GP dependencies (including unlocked packages). If any unpromoted 2GP dependencies are detected then the promotion operation is aborted. To promote _all_ unpromoted dependencies, in addition to the target `Package2Version`, use `--auto-promote True`.

Sample Usage:
```    $ cci task run promote_2gp_package --version_id 04t000000000000 --auto_promote True```

This should be pretty close. Let me know if we think that more unit tests would be a good thing. I covered the three main execution paths that can occur when running the task:

1. No dependencies that need promoting are found, so target `Package2Version` is promoted
2. Unpromoted dependencies with `--auto-promote False` so we abort.
3. Unpromoted dependencies detected with `--auto-promote True` so we promote them, as well as the target `Package2Version`. 

Task Output for when `--auto-promote` is `False`:
Screen Shot 2021-03-10 at 1 42 59 PM

Task Output for when `--auto-promote` is `True`:
Screen Shot 2021-03-10 at 2 26 33 PM

### CODE REVIEW

**Comment by Br4nd0R** [comment]
I did a refactor on the majority of the task after finishing, I have some tests failing still, but I'll fix those on Monday morning when I get in. This should be ready to review for the most part though.

File: cumulusci/core/config/util.py [hunk 1]
```diff
@@ -0,0 +1,18 @@
+from cumulusci.core.config import BaseProjectConfig
+from cumulusci.core.exceptions import ServiceNotConfigured
+from cumulusci.core.config.sfdx_org_config import SfdxOrgConfig
+from cumulusci.core.sfdx import get_default_devhub_username
+
+
+def get_devhub_config(project_config: BaseProjectConfig) -> SfdxOrgConfig:
+    """
+    @param project_config: project configuration that has a configured devhub service
```
davisagli, line 9: It doesn't need to have a configured devhub service; if it doesn't we get the default devhub from sfdx

cumulusci/cumulusci.yml
```diff
@@ -344,6 +344,10 @@ tasks:
         description: Apply org settings from a scratch org definition file or dict
         class_path: cumulusci.tasks.salesforce.org_settings.DeployOrgSettings
         group: Salesforce DX
+    promote_2gp_package:
```
davisagli, line 347: I'd suggest calling this promote_package_version for congruence with create_package_version. (And change the Python class to match.) Not explicitly saying 2gp is intentional since 2gp is meant to be the primary packaging method in the future.

cumulusci/tasks/package_2gp.py
```diff
@@ -445,6 +433,7 @@ def _create_version_request(
             "SkipValidation": skip_validation,
             "Tag": f"hash:{package_hash}",
             "VersionInfo": version_info,
+            "CalculateCodeCoverage": not skip_validation,
```
davisagli, line 436: I suspect this also needs to be off when creating an org-dependent unlocked package.

cumulusci/tasks/tests/test_package_2gp.py
```diff
@@ -133,8 +106,10 @@ def task(project_config, devhub_config, org_config):
         ),
         org_config,
     )
-    task._init_devhub = mock.Mock(return_value=devhub_config)
-    task._init_task()
+    with mock.patch(
+        "cumulusci.tasks.package_2gp.get_devhub_config", return_value=devhub_config
+    ):
```
davisagli, line 111: Why don't you make the devhub_config fixture take care of doing this mocking so that you don't have to repeat it for each test?

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+
+    The terms "promoted" and "released" are the same in the context of
+    a 2GP package. Once "promoted" a package is considered to be "released".
+    Hence, the "promotion" of a package is just an update to the
+    Package2Version.IsReleased field.
+
+    The abbreviation `spv` in variable names is shorthand for "SubscriberPackageVersion".
+    """
+
+    task_docs = """Promote a Second Generation managed package.
```
davisagli, line 22: Unlocked packages also need to be promoted https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_unlocked_pkg_create_pkg_ver_promote.htm

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+    Lists any 1GP dependencies that are detected, as well as,
```
davisagli, line 23:
```suggestion
    Lists any 1GP dependencies that are detected, as well as
```

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+    any dependency packages that have not benn promoted.
```
davisagli, line 24:
```suggestion
    any dependency packages that have not been promoted.
```

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+    Once promoted, 2GP package can be installed into production orgs."""
```
davisagli, line 25:
```suggestion
    Once promoted, the 2GP package can be installed into production orgs."""
```

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+
+    task_options = {
+        "version_id": {
+            "description": "The SubscriberPackageVersion (04t) Id for the target package.",
+            "required": True,
+        },
+        "auto_promote": {
```
davisagli, line 32: How about `promote_dependencies`?

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+            ),
+            "required": False,
+        },
+    }
+
+    def _init_options(self, kwargs) -> None:
+        super()._init_options(kwargs)
+
+        if "version_id" not in self.options:
+            raise TaskOptionsError("Task option `version_id` is required.")
```
davisagli, line 44: I think we probably need a way to find this based on the current commit so that it can be used from the Upload Release plan in MetaCI, where there's no place to enter a version_id.

Maybe something like: If the version_id isn't specified, look for a beta release tag that references the current commit, and use the version_id from its tag message. (Obviously in order for that to work we need to implement making the github_release task store the version_id in the message for the tag that it creates.)

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+        @param spv_id: a SubscirberPackageVersionId
+        @returns: a dict with the following shape:
+        {
+            "is_2gp": (bool) whether or not this dependency is a 2GP package
+            "is_promoted": whether or not this package is already promoted (only present if is_2gp is True)
+            "Package2VersionId": The Id from the corresponding Package2Version record
+            "name": (str) The name of the package
+            "release_state": (str) the packages release state
+            "version_id": (str) the SubscriberPackageVersionId (04t) of the dependency
+        }
```
davisagli, line 88: You don't necessarily need to change this, but this looks like a situation where it might have made sense to define a pydantic model and return an instance of it instead of a dict. That way you can get help from your IDE with typing in the correct keys and assigning the correct types to them. On the other hand it's a bit more effort so maybe not worth it if we're also unit testing the code and not anticipating it changing frequently.

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+        dictionaries with info on all dependency packages present.
+
+        @param spv_id: a SubscirberPackageVersionId
+        @returns: a dict with the following shape:
+        {
+            "is_2gp": (bool) whether or not this dependency is a 2GP package
+            "is_promoted": whether or not this package is already promoted (only present if is_2gp is True)
+            "Package2VersionId": The Id from the corresponding Package2Version record
+            "name": (str) The name of the package
+            "release_state": (str) the packages release state
```
davisagli, line 86:
```suggestion
            "release_state": (str) the package's release state
```

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+
+        @param dependencies: list of dependencies to process
+        """
+        one_gp_deps = self._filter_one_gp_deps(dependencies)
+        if one_gp_deps:
+            self.logger.warning("This package has the following 1GP dependencies:")
+            self.logger.warning("")
+            for dep in one_gp_deps:
+                self.logger.warning(
+                    f"    Package Name: {dep['name']:20} ReleaseState: {dep['release_state']}"
```
davisagli, line 142: How hard would it be to also show the package version?

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+        """
+        two_gp_deps = self._filter_two_gp_deps(dependencies)
+        unpromoted_two_gp_deps = self._filter_unpromoted_two_gp_deps(dependencies)
+
+        self.logger.info(f"Total 2GP dependencies: {len(two_gp_deps)}")
+        self.logger.info(f"Unpromoted 2GP dependencies: {len(unpromoted_two_gp_deps)}")
+
+        should_exit = False
+        if unpromoted_two_gp_deps and self.options.get("auto_promote", False):
+            [self._promote_2gp_package(d) for d in unpromoted_two_gp_deps]
```
davisagli, line 161: Using a list comprehension here caught me off guard because I think of their primary purpose as being to construct a list, not to perform a side effect. Would be a bit easier to read if you write it as a normal for loop on 2 lines, I think.

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+        elif unpromoted_two_gp_deps:
+            # we only want _run_task() to exit if unpromoted
+            # 2GP deps are presentand we aren't auto-promoting
```
davisagli, line 164:
```suggestion
            # 2GP deps are present and we aren't auto-promoting
```

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+        }
+
+    def _promote_2gp_package(self, package_info: Dict) -> None:
+        """
+        Promote the 2GP package associated with the given SubscriberPackageVersionId
+
+        @param spv_id: the SubscriberPackageVersionId to promote
+        """
+        self.logger.info("")
+        self.logger.info(f"Promoting package: {package_info['name']}")
```
davisagli, line 206: Would be nice to include the version number here

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+        @param raise_error: If True, raises an error if no records are found
+
+        @return: None if no records are found, single sObject if `return_one` is True,
+        or a list of sObject records if `return_one` is False.
+        """
+        query = f"SELECT {', '.join(fields)} FROM {obj_name}"
+        if where_clause:
+            query += f" WHERE {where_clause}"
+
+        res = self.tooling.query(query)
```
davisagli, line 289: query_all or you won't get all the results when there are more than 200

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+
+        return res["records"]
+
+    def _filter_one_gp_deps(self, dependencies: List[Dict]) -> List[Dict]:
+        """Return only 1GP dependencies"""
+        return [d for d in dependencies if not d["is_2gp"]]
+
+    def _filter_two_gp_deps(self, dependencies: List[Dict]) -> List[Dict]:
+        """Return only 2GP dependencies"""
+        return [d for d in dependencies if d["is_2gp"]]
```
davisagli, line 304: Maybe _get_dependency_info should put the dependencies into the appropriate lists as it collects the info and return the multiple lists, instead of needing to filter later? Your call.

davisagli (commented):
@Br4nd0R nice work, I've got a lot of little comments but the overall direction of this looks right.

davisagli:
We'll also need some doc updates here: https://cumulusci.readthedocs.io/en/stable/unlocked_package.html#promote-a-package-version

cumulusci/tasks/tests/test_package_2gp.py
Br4nd0R: If the path that needs mocking may change depending on which test is running, is there a way to patch this in the `devhub_config` fixture, such that it doesn't need a parameter?

For example, in `test_package_2gp.py`, "cumulusci.tasks.package_2gp.get_devhub_config" is patched, but in `test_promote_package_version.py`, "cumulusci.tasks.salesforce.promote_package_verions.get_devhub_config" is patched.

davisagli: Oh I see, that does make it more complicated and I'm not sure I have a better suggestion for how to simplify it then.

cumulusci/tasks/salesforce/promote_2gp_package.py
Br4nd0R: I'm not sure how often we need to aggregate package information from several (related) sObjects, but I'll definitely keep this in mind next time this situation comes up. Also, looking at this again, now I want to rename the keys to match the corresponding sObject field name (e.g. {'SubscriberPackage.Name': name}).

cumulusci/tasks/package_2gp.py
```diff
@@ -200,24 +199,13 @@ def _init_options(self, kwargs):
         )
 
     def _init_task(self):
-        self.devhub_config = self._init_devhub()
         self.tooling = get_simple_salesforce_connection(
             self.project_config,
-            self.devhub_config,
+            get_devhub_config(self.project_config),
```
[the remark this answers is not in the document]
davisagli: It could be, but I always feel a little bad about adding more stuff to BaseProjectConfig and OrgConfig which are already so big...

davisagli:
@Br4nd0R For logging the version I meant the version number, not the SubscriberPackageVersionId -- I'm thinking it would make it easier for someone to confirm at a glance whether the expected dependencies are there. But it requires querying more fields and can be a future enhancement.

cumulusci/tasks/salesforce/promote_2gp_package.py
```diff
@@ -0,0 +1,308 @@
...
+    def _query_SubscriberPackageVersion(self, spv_id: str) -> Optional[Dict]:
+        """Queries for a SubscriberPackageVersion record with the given SubscriberPackageVersionId"""
+        return self._query_one_tooling(
+            ["Id", "Dependencies", "ReleaseState", "SubscriberPackageId"],
+            "SubscriberPackageVersion",
+            where_clause=f"Id='{spv_id}'",
+            raise_error=True,
+        )
+
+    def _query_one_tooling(
```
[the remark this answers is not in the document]
Br4nd0R: I think we need a work item to look more at a general purpose SalesforceQuery class.

docs/unlocked_package.rst
```diff
@@ -53,10 +53,22 @@ For dependencies that are an unpackaged bundle of metadata, CumulusCI creates an
...
+    $ cci task run promote_package_version --version_id 04t000000000000
+
+If additional unlocked packages were created to hold unpackaged dependencies, they must be promoted as well.
+To promote dependencies automatically use ``--promote_dependencies True``.
+
+.. code-block:: console
+
+    $ cci task run promote_package_version --version_id 04t000000000000 --promote_dependencies True
+
+Alternatively, you can use ``sfdx force:package:version:promote`` command to promote a 2GP package.
```
davisagli, line 71:
```suggestion
Alternatively, you can use the ``sfdx force:package:version:promote`` command to promote a 2GP package.
```

Br4nd0R:
I've gone ahead and created the following work items:
* One for further discussion around a `SalesforceQuery` util class (in the tech debt epic)
* One for auto-resolving 04ts which is I'm currently working on for the current sprint.
* One for logging `version_number` along with the other information we display for dependencies in the `promote_package_version` task (in the 2GP epic).

cumulusci/tasks/package_2gp.py
davisagli: Based on https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_unlocked_pkg_code_coverage.htm it sounds like it's not supported for org-dependent unlocked packages, but also not invalid to specify it. So I guess I'll resolve this.

cumulusci/cumulusci.yml
```diff
@@ -348,6 +348,10 @@ tasks:
         description: Apply org settings from a scratch org definition file or dict
         class_path: cumulusci.tasks.salesforce.org_settings.DeployOrgSettings
         group: Salesforce DX
+    promote_package_version:
+        description: Promote a 2gp package so that it can be installed in a production org
+        class_path: cumulusci.tasks.salesforce.promote_2gp_package.PromotePackageVersion
```
davisagli, line 353: I tried running the task and got:

ModuleNotFoundError: No module named 'cumulusci.tasks.salesforce.promote_2gp_package'

davisagli (changes requested):

cumulusci/cumulusci.yml
Br4nd0R: Ya, we might want the task to actually run 🤷‍♂️

davisagli (approved):

PR #2455, opened 2021-03-15
[W-8998476] Add preflight task to retrieve available permissions sets

vmarasow:
# Critical Changes

# Changes
We added a new custom task that retrieves the list of available permission sets in the org. 

# Issues Closed
[W-8998476](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000008a58bIAA/view)

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -256,6 +256,10 @@ tasks:
         description: Retrieves a list of the names of any permission sets assigned to the running user.
         class_path: cumulusci.tasks.preflight.permsets.GetPermissionSetAssignments
         group: Salesforce Preflight Checks
+    get_available_permission_set_licenses:
+        description: Retrieves a list of the currently available Permission Set definition keys
+        class_path: cumulusci.tasks.preflight.licenses.GetAvailablePermissionSets
+        group: Salesforce Preflight Checks
     get_existing_record_types:
         description: "Retrieves all Record Types in the org as a dict, with sObject names as keys and lists of Developer Names as values."
         class_path: cumulusci.tasks.preflight.recordtypes.CheckSObjectRecordTypes
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -257,7 +257,7 @@ tasks:
         class_path: cumulusci.tasks.preflight.permsets.GetPermissionSetAssignments
         group: Salesforce Preflight Checks
     get_available_permission_sets:
-        description: Retrieves a list of the currently available Permission Set definition keys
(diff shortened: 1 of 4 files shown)


cumulusci/cumulusci.yml
```diff
@@ -256,6 +256,10 @@ tasks:
         description: Retrieves a list of the names of any permission sets assigned to the running user.
         class_path: cumulusci.tasks.preflight.permsets.GetPermissionSetAssignments
         group: Salesforce Preflight Checks
+    get_available_permission_set_licenses:
```
davisagli, line 259: @vmarasow  This should be get_available_permission_sets, right?

Also could you please adjust this to NOT rewrite most of the whitespace in cumulusci.yml? (see diff below)

vmarasow: Good catch. I got all the whitespace back to the original state.

cumulusci/cumulusci.yml
```diff
@@ -256,6 +256,10 @@ tasks:
         description: Retrieves a list of the names of any permission sets assigned to the running user.
         class_path: cumulusci.tasks.preflight.permsets.GetPermissionSetAssignments
         group: Salesforce Preflight Checks
+    get_available_permission_sets:
+        description: Retrieves a list of the currently available Permission Set definition keys
```
davisagli, line 260:
```suggestion
        description: Retrieves a list of the currently available Permission Sets
```

cumulusci/tasks/preflight/licenses.py
```diff
@@ -23,3 +23,15 @@ def _run_task(self):
         ]
         licenses = "\n".join(self.return_values)
         self.logger.info(f"Found permission set licenses:\n{licenses}")
+
+
+class GetAvailablePermissionSets(BaseSalesforceApiTask):
+    def _run_task(self):
+        self.return_values = [
+            result["Name"]
+            for result in self.sf.query("SELECT Name FROM PermissionSet")[
```
davisagli, line 32:
```suggestion
            for result in self.sf.query_all("SELECT Name FROM PermissionSet")[
```

query only returns the first page of results

davisagli (changes requested):
Thanks for the contribution -- I have a few small tweaks to suggest but it looks close.

vmarasow:
Added the suggested changes, please let me know if you can think of anything else

davisagli:
@vmarasow The test also needs a slight change to provide a mock value for query_all rather than query

davisagli (approved):
Looks good. I'm rerunning the builds because one hit an unrelated error creating a scratch org, but this should be good to merge once the builds are green.

davisagli (approved):

PR #2456, opened 2021-03-16
Refactor dependency management and package installation

davidmreed:
# Outstanding Work

- There are a few code paths not currently covered that need additional unit tests written.
- Run `isort` across the changed files to tidy them up.
- The `dependencies.dependencies` module path is not very nice and needs revision. I originally planned multiple submodules, but use of type annotations and relatively tighter binding between some of the resolvers and their dependency classes produced a lot of circular dependencies.
- The rewritten `freeze()` methods need very careful review (side-by-side with existing freeze outcomes).
- There's an inefficiency where some of the resolvers will load the remote repo's `cumulusci.yml` twice; this needs optimization.
- Manual QA with real projects. The Asteroids are willing to help test the brand-new Exact Match 2GP resolver, which they requested.

## New WIs to Create

- Refactor `sources` to use new resolution strategies
- Refactor `install_managed`'s `latest_beta` etc. options to use new resolution strategies.
- Ideally, rework the mock GitHub objects in `project_config` to be better mocks overall.

# Critical Changes

# Changes

# Issues Closed


davidmreed:
I know the issue with the Release Test flow and am working on a fix.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+            )
+        assert (
+            values.get("zip_url")
+            or (values.get("github") and values.get("ref"))
+            or (
+                values.get("repo_name")
+                and values.get("repo_owner")
+                and values.get("ref")
+            )
+        ), "Must specify `zip_url`, or `github` and `ref`"
```
davisagli, line 450: Would this get any clearer if zipurl-based and github-based unpackaged dependencies were processed by separate dependency classes?

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+        ManagedPackageDependency,
+        GitHubDynamicDependency,
+        UnmanagedDependency,
+    ]:
+        try:
+            dep = dependency_class.parse_obj(dep_dict)
+            if dep:
+                return dep
+        except pydantic.ValidationError:
+            pass
```
davisagli, line 558: This is quite a nice way to deal with the backwards compatibility while giving us more solidly typed objects to work with.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+
+
+## Resolvers
```
davisagli, line 561: Put them in their own module maybe?

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+
+RESOLVER_CLASSES = {
+    DependencyResolutionStrategy.STRATEGY_STATIC_TAG_REFERENCE: GitHubTagResolver,
+    DependencyResolutionStrategy.STRATEGY_2GP_EXACT_BRANCH: GitHubReleaseBranchExactMatch2GPResolver,
+    DependencyResolutionStrategy.STRATEGY_2GP_RELEASE_BRANCH: GitHubReleaseBranch2GPResolver,
+    DependencyResolutionStrategy.STRATEGY_2GP_PREVIOUS_RELEASE_BRANCH: GitHubPreviousReleaseBranch2GPResolver,
+    DependencyResolutionStrategy.STRATEGY_BETA_RELEASE_TAG: GitHubBetaReleaseTagResolver,
+    DependencyResolutionStrategy.STRATEGY_RELEASE_TAG: GitHubReleaseTagResolver,
+    DependencyResolutionStrategy.STRATEGY_UNMANAGED_HEAD: GitHubUnmanagedHeadResolver,
+}
```
davisagli, line 823: It's okay for now but in the future if we want to support projects adding their own custom resolvers, we'll need a different way to register them.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+            f"No 2GP package version located for release {release_id} on {repo.clone_url}."
+        )
+        return (None, None)
+
+
+RESOLVER_CLASSES = {
+    DependencyResolutionStrategy.STRATEGY_STATIC_TAG_REFERENCE: GitHubTagResolver,
+    DependencyResolutionStrategy.STRATEGY_2GP_EXACT_BRANCH: GitHubReleaseBranchExactMatch2GPResolver,
+    DependencyResolutionStrategy.STRATEGY_2GP_RELEASE_BRANCH: GitHubReleaseBranch2GPResolver,
+    DependencyResolutionStrategy.STRATEGY_2GP_PREVIOUS_RELEASE_BRANCH: GitHubPreviousReleaseBranch2GPResolver,
```
davisagli, line 819: I think we should probably not use 2gp in the name for these resolvers that look for a version_id in a particular commit status on a particular branch. 2gp describes the end result rather than the strategy. Once we have full support for 2gp as a project's primary release vehicle, we'll need both these strategies, and a strategy which looks for the version_id in an annotation on the latest release tag.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+            unique(
+                itertools.chain(
+                    *list(
+                        d.flatten(context)
+                        for d in dependencies
+                        if not _should_ignore_dependency(d, ignore_deps or [])
+                    )
+                ),
+            )
+        )
```
davisagli, line 890: I think we could probably come up with a better name than `flatten`. It's not very evocative of being the function that finds transitive dependencies.

cumulusci/core/exceptions.py
```diff
@@ -68,6 +68,12 @@ class DependencyResolutionError(CumulusCIException):
     pass
 
 
+class DependencyParseError(CumulusCIException):
+    """ Raised when CumulusCI cannot parse a dependency specification. """
+
+    pass
```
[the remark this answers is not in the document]
davidmreed: I wanted to be able to distinguish between an _invalid specification_ and a _valid specification that does not successfully resolve_.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+    repo_owner: Optional[str]  # Deprecate - use full URL
+    repo_name: Optional[str]  # Deprecate - use full URL
+
+    unmanaged: bool = False
+    subfolder: Optional[str]
+    namespace_inject: Optional[str]
+    namespace_strip: Optional[str]
+
+    tag: Optional[str]
+    ref: Optional[str]
```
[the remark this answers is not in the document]
davidmreed: Agreed. It's on my list to write a WI for that further effort.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+                "The dependency keys `repo_owner` and `repo_name` are deprecated. Use `github` instead."
+            )
+        return values
+
+    @pydantic.root_validator
+    def check_complete(cls, values):
+        assert values.get("github") or (
+            values.get("repo_owner") and values.get("repo_name")
+        ), "Must specify `github` or `repo_owner` and `repo_name`"
+        assert values["ref"] is None, "Must not specify `ref` at creation."
```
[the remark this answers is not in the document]
davidmreed: It _should_ be supported here too, it just parses to a different class - once `ref` is populated, it's not a `GitHubDynamicDependency` any more but a static `UnmanagedDependency`. To be tested.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+            install_1gp_package_version(
+                context,
+                org,
+                self.namespace,
+                self.version,
+                options,
+                retry_options=DEFAULT_PACKAGE_RETRY_OPTIONS,
+            )
+        elif self.version_id:
+            context.logger.info(f"Installing {self.version_id}")
```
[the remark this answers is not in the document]
davidmreed: In many cases this dependency should have the `package_name` field populated by the resolver, so we can often do a better job already. Will add.

cumulusci/core/dependencies/dependencies.py
davidmreed: I think it would. That's a nice clarity improvement. Will change.

cumulusci/core/dependencies/dependencies.py
davidmreed: That earned me a lot more circular dependencies in the type annotations, but I'll try revisiting it after I make the rest of the changes.

cumulusci/core/dependencies/dependencies.py
davidmreed: Great point. I'll revise the messaging for these.

cumulusci/core/exceptions.py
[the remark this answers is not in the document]
davisagli: I'm not sure it's useful to make that distinction in terms of exception classes unless there's somewhere where both could happen and we want to handle them differently. But I don't feel strongly about it.

cumulusci/core/dependencies/dependencies.py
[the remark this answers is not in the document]
davisagli: Ok yeah, I see now.

cumulusci/core/dependencies/dependencies.py
davisagli: Fair -- I feel like there's probably a way to make it work out, but I don't have a concrete suggestion yet.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+            )
+        )
+
+        return deps
+
+    def __str__(self):
+        return f"Dependency: {self.github}"
+
+
+class ManagedPackageDependency(StaticDependency):
```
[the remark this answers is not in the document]
davidmreed: Changed to `PackageDependency` - but your other comments made me realize that both `StaticDependency` subclasses can be further decomposed, into a `PackageVersionIdDependency` + `PackageNamespaceVersionDependency`, and a `UnmanagedGitHubRefDependency` + `UnmanagedZipURLDependency`. 

That removes a lot of their internal logic in favor of letting the parse operation identify the right code path.

cumulusci/salesforce_api/package_install.py
```diff
@@ -12,6 +19,42 @@
...
+    ADMIN = "NONE"  # System Administrator only
+    PUSH = "PUSH"  # No profiles
+
+
+class NameConflictResolution(str, Enum):
+    BLOCK = "Block"
+    RENAME = "RenameMetadata"
+
+
+class ManagedPackageInstallOptions(CCIModel):
```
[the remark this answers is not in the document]
davidmreed: Good call. Changed.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+                self.skip,
+                managed=managed,
+                namespace=namespace,
+            )
+        )
+
+        return deps
+
+    def __str__(self):
+        return f"Dependency: {self.github}"
```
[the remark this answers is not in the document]
davidmreed: I've added a `description` property that is more verbose, so it can be used in logging messages.

cumulusci/salesforce_api/package_install.py
```diff
@@ -42,34 +90,102 @@ def _install_package_version(project_config, org_config, options):
...
+
+
+def _install_1gp_package_version(
+    project_config: BaseProjectConfig,
+    org_config: OrgConfig,
+    namespace: str,
+    version: str,
+    install_options: ManagedPackageInstallOptions,
+    retry_options=None,
+):
```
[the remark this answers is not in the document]
davidmreed: I went with `install_package_by_namespace_version()` and `install_package_by_version_id()`. (The Tooling API does not accept a namespace + version as far as I can tell).

cumulusci/core/dependencies/dependencies.py
davidmreed: I revised these throughout to use "Commit Status" instead of "2GP". Some internal code still refers to 2GPs.

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,910 @@
...
+
+        namespace = config.project__package__namespace
+        package_name = (
+            config.project__package__name_managed
+            or config.project__package__name
+            or "Package"
+        )
+
+        return package_name, namespace
+
```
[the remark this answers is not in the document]
davidmreed: I moved these mixins to functions in a separate submodule.

cumulusci/core/dependencies/dependencies.py
davidmreed: These have now been broken out into separate classes. I similarly broke out version id-based package dependencies and version/namespace dependencies, which gets to a 1:1 relationship between dependency class and install pathway.

cumulusci/cli/cci.py
```diff
@@ -725,11 +730,24 @@ def project_info(runtime):
...
-        click.echo(line)
+def project_dependencies(runtime, resolution_strategy):
+    dependencies = get_static_dependencies(
+        parse_dependencies(runtime.project_config.project__dependencies),
+        get_resolver_stack(
+            runtime.project_config,
+            resolution_strategy,
+        ),
+        runtime.project_config,
+    )
```
davisagli, line 747: Calling get_static_dependencies looks more complicated than it needs to be -- the dependencies input and resolver stack can be derived from the project_config in cases where they aren't being overridden to come from somewhere else, no?

cumulusci/core/config/OrgConfig.py
```diff
@@ -489,27 +489,35 @@ def is_advanced_currency_management_enabled(self):
...
+        # Circular dependency.
+
         new_dependencies = []
         for dependency in dependencies:
-            dependency = {**dependency}
-
-            if "namespace" in dependency:
+            if (
+                isinstance(dependency, PackageNamespaceVersionDependency)
+                and dependency.namespace
```
davisagli, line 504: If it's a PackageNamespaceVersionDependency then we can assume it has a namespace, right?

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,978 @@
...
+    STRATEGY_STATIC_TAG_REFERENCE = "tag"
+    STRATEGY_COMMIT_STATUS_EXACT_BRANCH = "commit_status_exact_branch"
+    STRATEGY_COMMIT_STATUS_RELEASE_BRANCH = "commit_status_release_branch"
+    STRATEGY_COMMIT_STATUS_PREVIOUS_RELEASE_BRANCH = (
+        "commit_status_previous_release_branch"
+    )
+    STRATEGY_BETA_RELEASE_TAG = "latest_beta"
+    STRATEGY_RELEASE_TAG = "latest_release"
+    STRATEGY_UNMANAGED_HEAD = "unmanaged"
```
davisagli, line 72: seems redundant to prefix these with STRATEGY_; they're already identified as strategies by being accessed as members of DependencyResolutionStrategy

cumulusci/core/dependencies/dependencies.py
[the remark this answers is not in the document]
davisagli: Can we use that instead of `name` in `__str__` so that we see the additional info in the output from `cci project dependencies`?

cumulusci/core/dependencies/dependencies.py
```diff
@@ -0,0 +1,978 @@
...
+            retry_options=DEFAULT_PACKAGE_RETRY_OPTIONS,
+        )
+
+    @property
+    def name(self):
+        return f"Install {self.package} {self.version}"
+
+    @property
+    def description(self):
+        return self.name
```
davisagli, line 412: This can probably go in a base class as the default description.

cumulusci/core/dependencies/utils.py
```diff
@@ -0,0 +1,3 @@
+from collections import namedtuple
+
+TaskContext = namedtuple("TaskContext", ["org_config", "project_config", "logger"])
```
davisagli, line 3:
```suggestion
# data structure mimicking a task for use with the metadata API classes
TaskContext = namedtuple("TaskContext", ["org_config", "project_config", "logger"])
```

cumulusci/cumulusci.yml
```diff
@@ -863,7 +863,7 @@ flows:
                 flow: dependencies
                 options:
                     update_dependencies:
-                        include_beta: True
+                        resolution_strategy: preproduction
```
davisagli, line 866: Are you sure we want to use the preproduction resolution strategy by default, even for flows that are oriented toward development? It means that if I'm a partner developing an extension to NPSP, I'll get the beta versions of NPSP and its underlying packages, which I probably don't want.

Or is this just a temporary situation maintaining the status quo until you've had a chance to get the PRs in to change the default resolution strategy for our internal repositories?

davisagli: Never mind, I continued reading and realized preproduction is an alias for latest_release by default.

cumulusci/tasks/metaxml.py
```diff
@@ -57,19 +63,27 @@ class UpdateDependencies(MetaXmlBaseTask):
...
-                    (dependency["namespace"], str(dependency["version"]))
-                )
+            if (
+                isinstance(dependency, PackageNamespaceVersionDependency)
+                and dependency.namespace
+            ):
+                self.dependencies.append((dependency.namespace, dependency.version))
 
     def _process_xml(self, root):
+        print(self.dependencies)
```
davisagli, line 86:
```suggestion
```

cumulusci/tasks/metaxml.py
```diff
@@ -57,19 +63,27 @@ class UpdateDependencies(MetaXmlBaseTask):
...
+            get_resolver_stack(
+                self.project_config,
+                "production",
+            ),
+            self.project_config,
+        )
         self._process_dependencies(dependencies)
 
     def _process_dependencies(self, dependencies):
+        print(dependencies)
```
davisagli, line 77:
```suggestion
```

cumulusci/tasks/package_2gp.py
```diff
@@ -493,52 +511,35 @@ def _get_dependencies(self):
...
             new_dependency = {}
-            if dependency.get("version_id"):
-                name = (
-                    f"{dependency['namespace']}@{dependency['version']} "
-                    if "namespace" in dependency
-                    else ""
-                )
+            if (
+                isinstance(dependency, PackageVersionIdDependency)
+                and dependency.version_id
```
davisagli, line 529: We can assume a PackageVersionIdDependency has a version_id, right?

cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -93,268 +79,165 @@ def _init_options(self, kwargs):
...
-                "Updating dependencies would require uninstalling these packages "
-                "but uninstalls are not enabled: {}".format(
-                    ", ".join(dep["namespace"] for dep in self.uninstall_queue)
+    def _install_dependency(self, dependency):
+        if isinstance(
+            dependency, (PackageNamespaceVersionDependency, PackageVersionIdDependency)
+        ):
+            installed = False
+
+            if isinstance(dependency, PackageNamespaceVersionDependency):
```
davisagli, line 183: Is there a reason this check for the package already being installed shouldn't go in the dependency's `install` method? It seems like something we would want regardless of where in the code a package is being installed, not just for update_dependencies.

cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -93,268 +79,165 @@ def _init_options(self, kwargs):
...
         steps = []
-        for i, dependency in enumerate(self._flatten(dependencies), start=1):
-            name = dependency.pop("name", None)
-            if "namespace" in dependency:
+        for i, dependency in enumerate(dependencies, start=1):
+            if isinstance(
+                dependency,
+                (PackageNamespaceVersionDependency, PackageVersionIdDependency),
+            ):
                 kind = "managed"
```
davisagli, line 229: We should probably change this to "package" since a version id could be different types of package -- but that will take some coordination on the metadeploy side to change the available options. This is not a blocker.

cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -55,34 +46,29 @@ class UpdateDependencies(BaseSalesforceMetadataApiTask):
             "or a child branch of a release branch, resolve GitHub managed package dependencies to 2GP builds present on "
             "a matching release branch on the dependency."
         },
+        "resolution_strategy": {
+            "description": "The name of a sequence of resolution_strategy (from project__dependency_resolutions) to apply to dynamic dependencies."
+        },
```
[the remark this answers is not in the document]
davidmreed: I'm working on the docs now but haven't pushed yet.

cumulusci/tasks/salesforce/update_dependencies.py
davidmreed: Let's make it an enum while we're at it. I can add that WI to my list-to-create.

cumulusci/core/config/OrgConfig.py
davidmreed: Yep, good catch. Fixed.

cumulusci/core/dependencies/dependencies.py
davidmreed: Agreed. Changed.

cumulusci/core/dependencies/dependencies.py
davidmreed: Done.

cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -39,14 +37,7 @@ class UpdateDependencies(BaseSalesforceMetadataApiTask):
...
         },
-        "allow_newer": {
-            "description": "If the org already has a newer release, use it. Defaults to True."
-        },
-        "allow_uninstalls": {
-            "description": "Allow uninstalling a beta release or newer final release "
-            "in order to install the requested version. Defaults to False. "
-            "Warning: Enabling this may destroy data."
-        },
+        "allow_newer": {"description": "Deprecated. This option has no effect."},
```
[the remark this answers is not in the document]
davidmreed: I was trying to keep the number of backwards-incompatible changes minimal, and this one is fairly harmless.

cumulusci/tasks/salesforce/update_dependencies.py
davidmreed: That's an excellent idea that will make this code quite a bit cleaner, as well.

davidmreed: It could even go in the core `install_package_by_*()` methods, although I feel a little leery of adding that much behavior to those primitives. I think there is a valid case to be able to reinstall a package (with different install options), although it's pretty edge-casey.

cumulusci/tasks/salesforce/update_dependencies.py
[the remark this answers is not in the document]
davisagli: I was thinking it didn't matter if it's here or not, but I was only thinking about options passed in yaml. It does make a difference for whether the CLI will accept the option. I doubt anyone is passing this as a CLI option but I could be wrong.

cumulusci/tasks/salesforce/update_dependencies.py
davisagli: Yeah I agree with that assessment.

davidmreed:
Completed hands-on test charter: https://salesforce.quip.com/JfZwAPuPTbhE

docs/dev.rst
```diff
@@ -393,6 +399,83 @@ When tokenized, all occurrences of the namespace prefix, are replaced with ``%%%
...
+
+CumulusCI converts dynamic dependencies specified via GitHub repositories into specific package versions and commit references by applying one or more *resolvers*. You can customize the resolvers that CumulusCI applies to control how beta managed packages and second-generation feature packages, or to intervene more deeply in the dependency resolution process.
+
+CumulusCI organizes resolvers into *resolution strategies*, which are named, ordered lists of resolvers to apply. When CumulusCI applies a resolution strategy to a dependency, it applies each resolver from top to bottom until a resolver succeeds in resolving the dependency.
+
+Three resolution strategies are provided in the CumulusCI standard library:
+
+ * ``latest_release``, which will attempt to resolve to the latest managed release of a managed package project.
+ * ``include_beta``, which will attempt to resolve to the latest beta, if any, or managed release of a managed package project.
+ * ``commit_status``, which will resolve to second-generation package betas created on feature branches, if any, before falling back to managed package releases. This strategy is used only in the ``qa_org_2gp`` flow.
```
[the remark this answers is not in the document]
davidmreed: Sure is. Fixed.

davisagli (approved):

PR #2458, opened 2021-03-16
Dependency updates

Br4nd0R:
Notable Updates:
Keyring 22.3.0 --> 23.0.0


davisagli (changes requested):
Let's also try updating cryptography this time. And there is a security update for urllib3.

davisagli (approved):

PR #2481, opened 2021-03-20
Debug mode is made available to tasks

prescod:
# Changes

Tasks now get access to the --debug-mode option and can output debugging information conditional on it. 

Reviewer Note:

debug-mode will eventually be a collection of flags rather than a simple boolean.


prescod:
ContextVars has been approved for use in Salesforce products.

cumulusci/core/runtime.py
```diff
@@ -1,23 +1,29 @@
...
     project_config_class = BaseProjectConfig
     keychain_class = BaseProjectKeychain
     callback_class = FlowCallback
 
-    def __init__(self, *args, load_keychain=True, **kwargs):
+    def __init__(self, *args, load_keychain=True, debug_mode=False, **kwargs):
         self.universal_config = None
         self.project_config = None
         self.keychain = None
+        DEBUG_MODE.set(debug_mode)
```
davisagli, line 26: I really don't like using a global variable for this (even one that is context-safe).

When we added the runtime object, the thinking was that we would put any contextual information needed by tasks on the runtime object, and then add a way for tasks to access their runtime. I think we should handle this in that way.

davisagli (changes requested):

cumulusci/core/runtime.py
prescod: I did implement it that way at first but changed my mind.

There are a few features that naturally lend themselves to context vars or globals.

Logging, for example. Python logging is global-variable based because it is unreasonable to thread the log variable through to every corner of the code.  --debug is highly tied to and analogous with logging.

If it were only tasks that need access to it, I'd do it that way, but any algorithm in the whole system might need access, which means it becomes a part of the type signature of every constructor and function in all of CCI. I think that's a bigger problem than using a ContextVar.

Note also that this option is unlikely to be used in multi-RUNTIME contexts like servers.

prescod: BTW, today I wanted access to the --debug option in `SfdxOrgConfig.sfdx_info`.

I could imagine a property on classes like that self.debug which is opaque about where the data is really coming from. Then we could change it later if needed.

Passing the RUNTIME to tasks is a two-edged sword and might cause even bigger long-term problems. If custom tasks start to depend on having access to basically everything, refactoring could be tricky later.

davisagli: Okay... I concede that there are some things where it's handy to be able to assume they're present as part of a global context and not have to pass them around explicitly.

But, part of what makes me uneasy here is that we already have a few things which are global, and there's no consistency in how we're handling them. RUNTIME is just a plain module global. The CURRENT_TASK stack in cumulusci.core.tasks is a threading.local. Logging is an interesting one (because we should be able to just use global loggers from the logging module and rely on it to route log entries to the right handlers based on the logger name -- but we're actually passing it around in many places that we probably don't need to). Now you're proposing to add a ContextVar. So it seems like maybe it's the right time to try to make things more consistent (possibly consolidate into a single context object, but I do share that concern about things starting to implicitly depend on having access to everything).

Globals are also a bit of a pain to test, since you have to patch them in the correct module where they are used, not where they are defined.

I'd like to hear more about your anticipated use cases for this. If it's just about logging debug information, that's what loglevels are for. So I suspect it's more than that?

prescod: In many of my use-cases I would get rid of task-specific flags like --keep-working-directory (or whatever its called) and just use a single global debug flag. In particular for data-loading I'd like to provide the user with copious intermediate data (e.g. the raw CSVs imported and exported, the mapping file used, including runtime annotations, etc.)

I suspect that keeping around lots of working files and logs could also be helpful for debugging metadata deploy issues.

ContextVars are just the 21st century way to do Thread-Local Variables, so yeah, now that we've got the module, CURRENT_TASK should definitely be updated (or replaced).

IMO, RUNTIME should probably not be global at all, which is why I didn't just add this flag to it. Assuming that there is only one RUNTIME could cause all sorts of scalability problems in future versions of Pipeline API.

Thankfully, RUNTIME seems to be used only in the one file, so it isn't "really" used as a Global yet and we could get rid of it before someone uses it that way.

IMO, ContextVars are actually only like 10% as bad as "true" globals. The one thing you do not want is that when you refactor your app to be multi-threaded or multi-runtimed that you'll get leaking between the threads or runtimes and ContextVars *should* do a good job of preventing that.

davisagli: Thanks, that's helpful for understanding the use case.

And I agree that (separately) CURRENT_TASK should be switched to a contextvar and RUNTIME should ideally not be a global.

What I'd like to see if we're going to use a contextvar for the debug mode is:
- Move the related bits (DebugMode class, DEBUG_MODE contextvar) into their own module, cumulusci.core.debug_mode
- Provide functions in that module for getting and setting the debug mode. That way the use of a contextvar is an implementation detail, not the actual API. And the current debug mode can always be patched in the same module for tests, rather than needing to remember to patch it where it's imported.
- Set the debug mode directly from the `main` function rather than tying it to BaseCumulusCI's initializer.
- an approved 3pp record for contextvars (it's not enough that it's been approved for some other use; we need an approval for our use)

cumulusci/core/debug.py
```diff
@@ -0,0 +1,32 @@
+from contextlib import contextmanager
+from contextvars import ContextVar
+
+_DEBUG_MODE = ContextVar("debug_mode", default=False)
+
+
+# This may have levels/modes/flags someday
+# For now let's discourage anyone from using "DebugMode is True"
```
davisagli, line 8: What are we encouraging instead?

cumulusci/core/debug.py
```diff
@@ -0,0 +1,32 @@
...
+    def set(self, enable: bool):
+        token = _DEBUG_MODE.set(enable)
+        try:
+            yield
+        finally:
+            _DEBUG_MODE.reset(token)
+
+
+DebugMode = DebugMode()
+
```
davisagli, line 30: I think I would have just written `set_debug_mode` and `get_debug_mode` functions rather than a singleton class, but I guess that's more a matter of taste and don't really have a substantive critique of the class.

A couple minor things though:
- It took me a minute to figure out the intended way to set and get the debug mode. Would be helpful to have examples in the docstring.
- It irks me that the instance of the singleton is written in camel case, which usually signifies a class in Python

cumulusci/cli/tests/test_cci.py
```diff
@@ -2251,10 +2253,54 @@ def test_lines_from_traceback(self):
...
+    @mock.patch("cumulusci.tasks.robotframework.RobotLibDoc", MagicMock())
+    @mock.patch("cumulusci.cli.cci.get_tempfile_logger")
+    def test_run_flow_debug(self, get_tempfile_logger):
+        get_tempfile_logger.return_value = (mock.Mock(), "tempfile.log")
+        rtd = "cumulusci.tasks.robotframework.RobotTestDoc._run_task"
+
+        with mock.patch(rtd, mock_validate_debug(False)):
+            cci.main(["cci", "flow", "run", "robot_docs"])
+        with mock.patch(rtd, mock_validate_debug(True)):
+            cci.main(["cci", "flow", "run", "robot_docs", "--debug"])
```
davisagli, line 2291: It's not sustainable to keep adding new tests that need to mock this many things. Doesn't need to be addressed as part of this PR, but let's try to figure out a better test harness for cci.main

cumulusci/cli/cci.py
```diff
@@ -216,33 +217,39 @@ def main(args=None):
...
+                    )
+                sys.exit(1)
+
+
+def _load_runtime(is_error_command, tempfile_path, debug):
+    global RUNTIME
+    try:
+        RUNTIME = CliRuntime(load_keychain=False)
+    except Exception as e:
+        handle_exception(e, is_error_command, tempfile_path, debug)
```
davisagli, line 248: Maybe handle_exception should make use of the debug mode now so that it doesn't have to be passed around.

cumulusci/cli/tests/test_cci.py
prescod: I agree and started a cleanup on the weekend.

https://github.com/SFDO-Tooling/CumulusCI/pull/2519

cumulusci/core/debug.py
prescod: I didn't feel as strongly about it as you did, so I've changed to a function-based syntax in 28aea0b9

cumulusci/cli/cci.py
davisagli: Right, and we have a way to turn on should_show_stacktraces without turning on debug mode, okay.

cumulusci/core/debug.py
```diff
@@ -0,0 +1,33 @@
...
+    debug.save_intermediate_load_files etc."""
+
+    def __repr__(self):
+        return f"<Debug mode: {'enabled' if self else 'disabled'}>"
+
+
+_DEBUG_MODE = ContextVar("debug_mode", default=DebugMode(False))
+
+
+def get_debug_mode():
```
davisagli, line 18:
```suggestion
def get_debug_mode() -> DebugMode:
```

cumulusci/cli/cci.py
```diff
@@ -216,33 +217,39 @@ def main(args=None):
...
             args.remove("--debug")
 
-        # Load CCI config
-        global RUNTIME
-        try:
-            RUNTIME = CliRuntime(load_keychain=False)
-        except Exception as e:
-            handle_exception(e, is_error_command, tempfile_path, debug)
-            sys.exit(1)
+        with set_debug_mode(debug):
```
davisagli, line 220: No strong opinion about what's best here, but there's an ExitStack you could add it to.

cumulusci/conftest.py
```diff
@@ -108,3 +108,9 @@ def open_db():
                 yield connection, Base.metadata, session
 
         yield open_db
+
+
+@pytest.fixture(scope="session", autouse=True)
+def disable_exit(request):
+    with mock.patch("sys.exit"):
+        yield
```
davisagli, line 116: This doesn't strike me as a good idea, since it could cause us to overlook code that is exiting when it shouldn't. What's the rationale for patching it everywhere instead of a targeted patch or exception handler?

prescod: Okay, removed. I was trying to reduce the mocking burden as you described. But I accept your criticisms of this particular approach.

cumulusci/core/debug.py
prescod: Done.

cumulusci/conftest.py
davisagli: I think maybe the way to go is to create a context manager or pytest fixture (`cli_test_harness`?) that sets up all the existing mocks. That way at least they are centralized and can be requested with a single line instead of a giant stack of patches.

davisagli: Not saying that needs to be in the PR though. I'll approve.

davisagli (dismissed):

davisagli (approved):

PR #2482, opened 2021-03-21
Better error message for org import

prescod:
# Issues Closed

- The `cci org import` command now shows a clearer error message if you try to import an org that is not a locally created scratch org.

### DIFF
--- a/cumulusci/cli/cci.py
+++ b/cumulusci/cli/cci.py
@@ -1042,6 +1042,11 @@ def org_import(runtime, username_or_alias, org_name):
     scratch_org_config.config["created"] = True
 
     info = scratch_org_config.sfdx_info
+    if not info.get("created_date"):
+        raise click.UsageError(
+            "cci org import only works for locally created "
+            "scratch orgs.\nUse `cci org connect` for other orgs."
+        )
     scratch_org_config.config["days"] = calculate_org_days(info)
     scratch_org_config.config["date_created"] = parse_api_datetime(info["created_date"])
 
--- a/cumulusci/cli/tests/test_cci.py
+++ b/cumulusci/cli/tests/test_cci.py
@@ -1175,6 +1175,33 @@ Environment Info: Rossian / x68_46
             "Imported scratch org: access, username: test@test.org" in "".join(out)
         )
 
+    @mock.patch("sarge.Command")
+    def test_org_import__persistent_org(self, cmd):
+        runtime = mock.Mock()
+        result = b"""{
+            "result": {
+                "createdDate": null,
+                "instanceUrl": "url",
+                "accessToken": "access!token",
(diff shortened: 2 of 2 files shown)


davisagli (approved):
Thanks -- I think we could make `cci org import` work for non-scratch orgs, but this is helpful in the meantime.

prescod:
I thought that maybe it was impossible to import persistent orgs because the persistent org wouldn't have the connected app installed/authorized or something like that?

I thought that it only works for Scratch Orgs because of special rules wherein they inherit their authorizations from the DevHub?

The exact relationships between sfdx/cumulusci/scratch orgs/persistent orgs/devhubs/connected apps/environment variables/keychains still has many fuzzy bits for me.

davisagli:
@prescod In cci we have 2 types of org config:
1. OrgConfig authenticates against Salesforce directly. This is what you get when you connect an org using "cci org connect" Currently all of these must use the same connected app. If SFDX_CLIENT_ID and SFDX_HUB_KEY are set we do JWT auth using them; otherwise we do the web-based auth flow using the connected_app service config.
2. ScratchOrgConfig delegates authentication to sfdx (NOT the dev hub, aside from a small interaction as part of the first login when the scratch org is created). And in sfdx, each org could be using a different connected app. ScratchOrgConfig is what you get if you create a scratch org in cci org import one from sfdx using "cci org import". It also has a number of additional methods and properties that only make sense for scratch orgs, and in general cci assumes that if it has a ScratchOrgConfig, then the org is a scratch org and it can use those methods and properties. That's why we can't easily use ScratchOrgConfig to delegate authentication to sfdx for a persistent org.

So there's a need for a 3rd thing, an SfdxOrgConfig which delegates auth to sfdx but isn't a scratch org. Actually it exists -- I factored it out of ScratchOrgConfig while working on the 2gp create_package_version task, because I needed a way to get the dev hub from sfdx as an org config. The missing piece is that we don't yet have a way to store it in the keychain (for ScratchOrgConfig we set `scratch: true` in the config dict, and that's the sign to deserialize it as a ScratchOrgConfig instead of an OrgConfig when we load the keychain files).

So I think the path to making `cci org import` work for persistent orgs is something like:
1. make sure we can persist SfdxOrgConfigs in the keychain
2. make `cci org import` create either an SfdxOrgConfig or ScratchOrgConfig depending on the type of org
3. make sure that SfdxOrgConfigs are presented in a reasonable way in `cci org list` output

Separately, once we have named services we could also consider some separate changes so that not all OrgConfigs need to be using the same connected app:
1. Add a reference from OrgConfig to which connected_app service alias to use for it.
2. Add a way to indicate as part of a connected_app service that it should use JWT auth, so that we don't need to do it with environment variables that apply to everything.

PR #2483, opened 2021-03-22
Add support for creating PSAs for multiple users

jstvz:
This PR adds support for creating permset assignments for multiple users.

----

# Critical Changes

# Changes
- The `user_alias` option for the `assign_permission_sets`, `assign_permission_set_groups`, and `assign_permission_set_licenses` tasks now accept a list of user aliases, and can now create permission assignments for multiple users with a single task invocation. 
# Issues Closed


cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -29,65 +31,98 @@ def _init_options(self, kwargs):
...
                 self.logger.warning(
-                    f'{self.permission_label} "{api_name}" is already assigned.'
+                    f'{self.permission_label} "{perm_name}" is already assigned to {user_id}.'
                 )
+        return assignments
+
+    def _insert_assignments(self, records_to_insert):
+        request_body = json.dumps({"allOrNone": False, "records": records_to_insert})
+        result = self.sf.restful("composite/sobjects", method="POST", data=request_body)
+        self.logger.debug(result)
```
davidmreed, line 125: This output seems a little user-unfriendly. Could we just throw an exception with a specific message if any subrequest fails?

cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -29,65 +31,98 @@ def _init_options(self, kwargs):
...
             else:
                 self.logger.warning(
-                    f'{self.permission_label} "{api_name}" is already assigned.'
+                    f'{self.permission_label} "{perm_name}" is already assigned to {user_id}.'
                 )
+        return assignments
+
+    def _insert_assignments(self, records_to_insert):
+        request_body = json.dumps({"allOrNone": False, "records": records_to_insert})
+        result = self.sf.restful("composite/sobjects", method="POST", data=request_body)
```
davidmreed, line 124: Hitting the 200-row limit here seems unlikely but definitely possible. Could we do a check first to avoid an inscrutable exception?

cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -29,65 +31,98 @@ def _init_options(self, kwargs):
...
+        )
 
     def _run_task(self):
-        # Determine existing assignments
-        if "user_alias" not in self.options:
+        users = self._query_existing_assignments()
+        users_assigned_perms = {
+            user["Id"]: self._get_assigned_perms(user) for user in users
+        }
+        self.perms_by_id = self._get_perm_ids()
```
davidmreed, line 43: I don't feel strongly about it to argue for a change, but the mixing between local parameters and the instance var confused me on first read.

davidmreed (changes requested):

cumulusci/tasks/salesforce/users/permsets.py
```diff
@@ -120,9 +123,40 @@ def _get_assignments(self, user_id, assigned_perms):
...
-        self.logger.debug(result)
+        result_list = []
+        for i in range(0, len(records_to_insert), 200):
+            request_body = json.dumps(
+                {"allOrNone": False, "records": records_to_insert[i : i + 200]}
+            )
+            result = self.sf.restful(
+                "composite/sobjects", method="POST", data=request_body
+            )
+            self.logger.debug(result)
```
davidmreed, line 134: With the table, can we now avoid logging the raw JSON?

jstvz: Generally, I like having API responses when debugging, but you're right that it's duplicative here.

davidmreed (approved):

PR #2493, opened 2021-03-23
Add community name return vals to ListCommunities

jstvz:
-------

# Critical Changes

# Changes

- The `list_communities` task now returns a list of Experience Cloud site names in its `return_values` for use in preflight checks.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/ListCommunities.py
+++ b/cumulusci/tasks/salesforce/ListCommunities.py
@@ -36,3 +36,4 @@ class ListCommunities(BaseSalesforceApiTask):
 
         communities_output2 = "\n".join(communities_output).encode("utf-8")
         self.logger.info(rst2ansi(communities_output2))
+        self.return_values = [community["name"] for community in communities]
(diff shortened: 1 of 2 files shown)


jstvz:
There's also `get_community_info` [here](https://github.com/SFDO-Tooling/CumulusCI/blob/feature%2Flistcommunity-return-vals/cumulusci/core/config/OrgConfig.py#L210), but that throws an exception. I'm open to the idea that the preflight should use OrgConfig, but this seemed both simpler and consistent with how most preflight checks are called.

davidmreed:
I don't really remember what this task was for in the first place. I wonder if we should actually just move it to the preflight checks module. Or what do you think about adding a new named task with the same Python class to harmonize with the other preflights?

jstvz:
I see a couple of references in FC-Reviewer and FGM-Portal that say:

> Use `cci task run list_communities` for quick access to the community links.

..so I'm guessing that's it. Adding a task config to match other preflights makes sense, I think. How about `get_existing_sites` or `get_existing_communities`?

davidmreed:
That sounds good to me.

davidmreed (approved):

PR #2494, opened 2021-03-24
Feature/dray/recursion update

davidjray:
[W-8886893](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B00000096XHJIA2/view)
# Critical Changes

# Changes
- It is now possible to create a flow which runs the same sub-flow multiple times, as long as they don't create a self-referential cycle.

# Issues Closed


cumulusci/core/flowrunner.py
```diff
@@ -628,33 +630,37 @@ def _check_old_yaml_format(self):
...
 
-        :param steps: Set of step definitions to loop through
-        :param flows: Flows already visited.
+        :param flow_config: Set of step definitions to loop through
+        :param flow_stack: list of flow signatures already visited
         :return: None
         """
-        if visited_flows is None:
-            visited_flows = set()
+        if flow_stack is None:
```
Br4nd0R, line 641: `None` is also a "falsy" value so you can shorten this to:
```suggestion
        if not flow_stack:
```

cumulusci/core/flowrunner.py
```diff
@@ -628,33 +630,37 @@ def _check_old_yaml_format(self):
...
                 raise FlowConfigError("No steps found in the flow definition")
 
-    def _check_infinite_flows(self, flow_config, visited_flows=None):
+    def _check_infinite_flows(self, flow_config, flow_stack=None):
         """
         Recursively loop through the flow_config and check if there are any cycles.
 
-        :param steps: Set of step definitions to loop through
-        :param flows: Flows already visited.
+        :param flow_config: Set of step definitions to loop through
```
Br4nd0R, line 637: This isn't actually a `Set`, it's a `FlowConfig` object (which we want to ensure doesn't have cycles/infinite loops).

cumulusci/core/flowrunner.py
```diff
@@ -628,33 +630,37 @@ def _check_old_yaml_format(self):
...
         :return: None
         """
-        if visited_flows is None:
-            visited_flows = set()
+        if not flow_stack:
+            flow_stack = []
         project_config = flow_config.project_config
+
         for step in flow_config.steps.values():
             if "flow" in step:
```
Br4nd0R, line 646: I'd make this
```suggestion
            if step == "flow":
```

cumulusci/core/flowrunner.py
```diff
@@ -628,33 +630,37 @@ def _check_old_yaml_format(self):
...
                 flow_name = step["flow"]
```
Br4nd0R, line 647: It might be helpful to name this to `next_flow_name` or `sub_flow_name` to indicate this is the different than the flow we are looping through.

cumulusci/core/flowrunner.py
```diff
@@ -628,33 +630,37 @@ def _check_old_yaml_format(self):
...
                 if flow_name == "None":
                     continue
```
Br4nd0R, line 649: This is odd. Running `make coverage` shows that we have 100% code coverage on this file (nice job!), but when I step through these two tests I don't see that this case is triggered. @davisagli - I'm guessing this has to do with flows listed as `None` in the config file to skip a step. Would it be worth it for @davidjray to add another flow that would trigger this case?

Br4nd0R: It looks like this is covered by `SimplTestFlowCoordinator.test_init_ambiguous_step()`, so I'm guessing that you don't really need to explicitly test this case, but let's see what David thinks.

Br4nd0R (changes requested):
Overall, awesome work on this! 🎉  Just some minor tweaks that I mentioned.

cumulusci/core/flowrunner.py
davidjray: That's a good idea 👍🏽

cumulusci/core/flowrunner.py
davisagli: I don't think that's right. `step` is a dict here, with "flow" as one of the keys.

cumulusci/core/flowrunner.py
davisagli: that existing test looks sufficient

davisagli (dismissed):
Sort out the details from Brandon's comments as you like, but this looks good to me.

davisagli:
I'd edit the changelog entry to say something like "It is now possible to create a flow which runs the same sub-flow multiple times, as long as they don't create a self-referential cycle."

cumulusci/core/flowrunner.py
davidjray: @davisagli is correct. `(flow_config.steps.values()` looks like:
`dict_values([{'flow': 'grandchild_flow'}, {'flow': 'parent_flow'}, {'task': 'pass_name'}, {'flow': 'recursive_flow'}])`

davidjray:
> It is now possible to create a flow which runs the same sub-flow multiple times, as long as they don't create a self-referential cycle.

Thanks, I suck at writing those. :D

davisagli:
@davidjray The way I think about it is that for release notes I'm generally trying to put myself in the user's shoes and imagine what the change looks like from their perspective. It's the place to call out what functionality it breaks or makes possible, not details about what changed in the implementation (unless it's a detail that we know people are treating as a public API).

cumulusci/core/flowrunner.py
Br4nd0R: Apologies. I was reading the contents of my debugger incorrectly. 🙃

Br4nd0R (approved):

PR #2499, opened 2021-03-25
Named Services

Br4nd0R:
# Changes
* CumulusCI now supports named services! This means you can configure multiple services of the same _type_ under different names. If you run `cci service list` you will note that your existing global services will have the name `service_type__global`, and any project specific services will have the name `service_type__project_name`.
* CumulusCI has a new command: `cci service default`. This command sets the default service for a given type.
* CumulusCI has a new command: `cci service rename`. This command renames a given service.

## Internal Considerations
* `keychain.list_services()` now returns a dict with service types as keys, and each key returns a list of available aliases/names of services configured for the given type. (I believe this has potential impacts on MetaCI).
* The keychain has a new public API method `set_default_service(service_type, alias)` for setting the default service for a given type. (This will be useful for the new `cci service default` command).
* The keychain's `get_service()` method will return the the default service for the given type if no alias is specified.

## Misc
* The `make coverage` target now supports a `CLASS_PATH` argument to allow running coverage for a subset of tests.


cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -1,16 +1,40 @@
...
+"""
+EncryptedFileProjectKeychain
+----------------------------
+This class represents a project keychain that stores services and orgs in encrypted
+files. These files have the extensions: '.org' and '.service'.
+
+Service files (.service) are organized under ~/.cumulusci/services/ into sub-directories that
+pertain to a particular service type (e.g. github, connected_app). These files hold
+the encrypted 'attributes' of each service as they are listed in the universal
+cumulusci.yml file. There are multiple DEFAULT_SERVICES.json files, the map a particular
```
davisagli, line 23:
```suggestion
cumulusci.yml file. There are multiple DEFAULT_SERVICES.json files that map a particular
```
```suggestion
cumulusci.yml file. There are multiple DEFAULT_SERVICES.json files, the map a particular
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -1,16 +1,40 @@
...
+service type to the default alias for that taype.
+
+The DEFUALT_SERVICES.json file that resides at ~/.cumulusci holds the global default
```
davisagli, line 26:
```suggestion
The DEFAULT_SERVICES.json file that resides at ~/.cumulusci holds the global default
```
```suggestion
The DEFUALT_SERVICES.json file that resides at ~/.cumulusci holds the global default
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -1,16 +1,40 @@
...
+services mappings. Each local project directory (~/.cumulusci/project/) also has a
+DEFUALT_SERVICES.json file for any services that have been configured to be project specific.
```
davisagli, line 28:
```suggestion
DEFAULT_SERVICES.json file for any services that have been configured to be project specific.
```
```suggestion
DEFUALT_SERVICES.json file for any services that have been configured to be project specific.
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -1,16 +1,40 @@
...
+
+Org files (.org) live in both under ~/.cumulusci and in local project direcotries.
```
davisagli, line 30:
```suggestion
Org files (.org) live in both under ~/.cumulusci and in local project directories.
```
```suggestion
Org files (.org) live in both under ~/.cumulusci and in local project direcotries.
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -1,16 +1,40 @@
...
+The .org files store an encrypted access_token and ____________. There is a
```
davisagli, line 31:
```suggestion
The .org files store an encrypted access_token and other org attributes. There is a
```
```suggestion
The .org files store an encrypted access_token and ____________. There is a
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -26,107 +50,28 @@ def global_config_dir(self):
...
+        self._load_app_file(self.global_config_dir, "connected.app", "app")
+        self._load_app_file(self.project_local_dir, "connected.app", "app")
+
+    def _load_app_file(self, dirname: str, filename: str, key: str) -> None:
+        """This is only used for loading the legacy connected_app configurations"""
         if dirname is None:
             return
-        full_path = os.path.join(dirname, filename)
+        full_path = Path(f"{dirname}/{filename}")
         if not os.path.exists(full_path):
```
davisagli, line 66:
```suggestion
        if not full_path.exists():
```
```suggestion
        if not os.path.exists(full_path):
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+                self._default_services[s_type] = alias
+
+    def _save_default_services(self, project: bool = False) -> None:
+        """Write out the contents of self._default_services to the
+        DEFAULT_SERVICES.json file based on the provided scope"""
+        dir_path = (
+            Path(self.project_local_dir) if project else Path(self.global_config_dir)
+        )
+        with open(dir_path / DEFAULT_SERVICES_FILENAME, "w") as f:
+            f.write(json.dumps(self._default_services))
```
davisagli, line 258: Doesn't this mean that setting the default for one service type as a project-level default will turn the defaults for all other services into project-level defaults as well? I think when a default service is set at the project level we want to do a more targeted update of the project-level DEFAULT_SERVICES file to only add/update that one service in it.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+        out to a DEFAULT_SERVICES.json file in the directory. This occurs
+        once before .service files are migrated to the appropriate
+        services/ sub-directory, so we set the default to the alias
+        that will be assigned during migration.
+
+        @param dir_path: the directory to look through and write
+        the DEFAULT_SERVICES.json file in.
+        """
+        if dir_path is None:
+            return  # pragma: no cover
```
davisagli, line 288: Can this actually happen?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+
+        This also has the advantage that when a new service type
+        is added to cumulusci.yml a new directory for that service type
+        will be created.
+        """
+        services_dir_path = Path(f"{self.global_config_dir}/services")
+        if not Path.is_dir(services_dir_path):
+            Path.mkdir(services_dir_path)
+
+        configured_service_types = self.project_config.config["services"].keys()
```
davisagli, line 326: You don't actually need .keys() if you're just going to iterate over it. Iterating over a dict gives you the dict keys. Some people like to explicitly write .keys() though to make it clearer, and that's fine.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+        |   |-- ...
+        .
+
+        This also has the advantage that when a new service type
+        is added to cumulusci.yml a new directory for that service type
+        will be created.
+        """
+        services_dir_path = Path(f"{self.global_config_dir}/services")
+        if not Path.is_dir(services_dir_path):
+            Path.mkdir(services_dir_path)
```
davisagli, line 324:
```suggestion
        services_dir_path.mkdir(exist_ok=True)
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+        """Migrate all .service files from the given directory to
+        the appropriate service sub-directory and apply the default
+        alias. This is intended to be run against either the
+        global_config_dir, or a local project directory.
+
+        Default aliases are in the form `service_type__scope`.
+        Scope is either 'project' or 'global' depending
+        on where the .service file is located."""
+        if not dir_path:
+            return  # pragma: no cover
```
davisagli, line 349: Can this actually happen?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+
+    def _migrate_services_from_dir(self, dir_path: str) -> None:
+        """Migrate all .service files from the given directory to
+        the appropriate service sub-directory and apply the default
+        alias. This is intended to be run against either the
+        global_config_dir, or a local project directory.
+
+        Default aliases are in the form `service_type__scope`.
+        Scope is either 'project' or 'global' depending
+        on where the .service file is located."""
```
davisagli, line 347: I take it you decided not to try to support multiple projects with project-specific services by naming them based on the project name?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +105,300 @@ def unset_default_org(self):
...
+        """Iterate over all local project dirs in ~/.cumulusci"""
+        for item in Path(self.global_config_dir).iterdir():
+            if item.is_dir() and item.name not in ["logs", "services"]:
+                yield item
+
+    def _raise_service_not_configured(self, name):
+        raise ServiceNotConfigured(  # pragma: no cover
+            f"'{name}' service configuration could not be found. "
+            f"Maybe you need to run: cci service connect {name}"
+        )
```
davisagli, line 394: I usually figure it's better to write the simple test for a function like this than to ignore coverage.

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -53,16 +118,12 @@ def _convert_connected_app(self):
                 "client_secret": connected_app.client_secret,
             }
         )
-        self.set_service("connected_app", ca_config)
+        # We're using an obnoxious alias to see who's using this
+        self.set_service("connected_app", "please_contact_sfdo_releng", ca_config)
```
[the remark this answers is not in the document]
Br4nd0R: @davisagli - Do you want me to just rip out the `_convert_connected_app()` function (and all references to it) in BaseProjectKeychain? (This would make me quite happy :) )

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -201,64 +206,136 @@ def get_org(self, name: str):
...
+        pass
+
+    def _init_default_services(self):
+        pass
+
+    def set_service(self, service_type, alias, service_config, project=False):
+        """Store a ServiceConfig in the keychain"""
+        if (
+            "services" not in self.project_config.config
+            or service_type not in self.project_config.config["services"]
```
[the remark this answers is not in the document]
Br4nd0R: I suppose it's more readable the way you mentioned. I was debugging tests and broke it out this way because of the context I was in at the time. I'll switch it back.

cumulusci/core/keychain/base_project_keychain.py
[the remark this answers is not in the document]
davisagli: Yeah, I think we should go ahead and rip out everything that's related to connected apps as a thing that is separate from services, and just be sure to call it out in the release notes. This can be part of this PR or a new story for another sprint.

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -201,64 +206,136 @@ def get_org(self, name: str):
...
+    #              Services               #
+    #######################################
+
+    def _load_services(self):
+        pass
+
+    def _init_default_services(self):
+        pass
+
+    def set_service(self, service_type, alias, service_config, project=False):
```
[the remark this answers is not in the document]
Br4nd0R: @davisagli - Looks like this needs to be present as no other keychain class currently overrides this function. If we're going to merge EncryptedFileProjectKeychain with BaseEncryptedProjectKeychain, I could add a new `set_service()` method there that accepts the `project` kwarg.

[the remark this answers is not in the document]
davisagli: @Br4nd0R I thought we decided all services would be stored at the global level and only defaults would be overridable at the project level. With that in mind, what's the purpose of the project flag here?

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -201,64 +206,136 @@ def get_org(self, name: str):
...
+        if alias == service_type:
+            raise ServiceNotValid(
+                "Service alias cannot be the same as the service type."
+            )
+
+        default = f"{service_type}__{DEFAULT_SERVICE_ALIAS}"
+        if alias == default:
+            raise ServiceNotValid(
+                f"Service alias cannot be the default alias: {default}"
+            )
```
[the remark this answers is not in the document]
Br4nd0R: We will definitely allow people to adjust the names of services that have been migrated. This comment made me realize that this check is actually incorrect. This should be ensuring that when we set a new service that there is not another service of the same type with the same alias.

@codexmarcel will be implementing the `cci service rename <service_type> <current_alias> <new_alias>` command in this PR so renaming will be possible.

cumulusci/core/keychain/base_project_keychain.py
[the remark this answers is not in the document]
Br4nd0R: You are correct; all services are stored at the global level. Since this is the only place where the `set_service()` method is defined, then when it is invoked from an instance of `EncryptedFileProjectKeychain` we still need to be able to pass in this argument so that the keychain knows that we need to write out the new project level default service into the local project's `DEFAULT_SERVICES.json` file in `EncryptedFileProjectKeychain._save_default_services()`.

But I think that we can remove the project argument from both `BaseEncryptedProjectKeychain` and `EncryptedFileProjectKeychain`'s `_set_encrypted_service()` methods.

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: good catch

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: Just tested and looks like we can remove this.

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: I'll leave it for now. I think the variable name better conveys what the keys pertain to in this context.

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: Seems like it won't be an issue if we remove it.

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: This was something I just forgot to incorporate. I'll go back and name them based on project name.

cumulusci/core/keychain/base_encrypted_project_keychain.py
```diff
@@ -27,23 +31,37 @@ class BaseEncryptedProjectKeychain(BaseProjectKeychain):
...
-            ServiceConfig, self.services[name], context=f"service config ({name})"
-        )
+        self._default_services[service_type] = alias
+
+    def _set_service(self, service_type, alias, service_config, project):
+        if service_type not in self.services:
+            self.services[service_type] = {}
+            # TODO: Should this be moved after _set_encrypted_service?
+            self._default_services[service_type] = alias
+            self._save_default_services(project)
```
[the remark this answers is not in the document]
Br4nd0R: I've made a new work item to address this in a future sprint: W-9110395

cumulusci/cli/cci.py
```diff
@@ -825,14 +844,19 @@ def callback(*args, **kwargs):
...
                 validator(serv_conf)
 
-            runtime.keychain.set_service(name, ServiceConfig(serv_conf), project)
+            service_name = kwargs["service_name"]
+            runtime.keychain.set_service(
+                service_type, service_name, ServiceConfig(serv_conf), project
+            )
             if project:
-                click.echo(f"{name} is now configured for this project.")
+                click.echo(f"{service_type} is now configured for this project.")
```
davisagli, line 852: This is potentially misleading. We are now only using the project flag to determine where the service gets set as the default, and the newly connected service will only get set as the default if it is the first one of this service_type.

I think we might want to do this:
1. Add a new `--default` flag which determines whether to set the new service as the global default for its type.
2. Repurpose `--project` to mean that the new service should be set as the project-specific default for its type.
3. Set both of these defaults via separate calls to set_default_service, so set_service doesn't need the project flag.

cumulusci/cli/cci.py
```diff
@@ -843,32 +867,63 @@ def service_connect():
...
     except ServiceNotConfigured:
         click.echo(
-            "{0} is not configured for this project.  Use service connect {0} to configure.".format(
-                service_name
-            )
+            f"{service_name} is not configured for this project.  Use service connect {service_name} to configure."
         )
 
 
+@service.command(name="default", help="Show the details of a connected service")
```
davisagli, line 898: wrong help

cumulusci/cli/cci.py
```diff
@@ -843,32 +867,63 @@ def service_connect():
...
+@click.argument("service_name")
+@pass_runtime(require_project=False, require_keychain=True)
+def service_default(runtime, service_type, service_name):
+    try:
+        runtime.keychain.set_default_service(service_type, service_name)
+    except ServiceNotConfigured as e:
+        click.echo(f"An error occurred setting the default service: {e}")
+        return
+
+    click.echo(f"'{service_name}' set as the default for {service_type} services.")
```
davisagli, line 909: Doesn't this need a way to set a project-level override of the default?

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -24,46 +24,55 @@ class BaseProjectKeychain(BaseConfig):
...
         self._load_orgs()
         self._load_scratch_orgs()
         self._load_services()
+        self._load_default_services()
 
-    def _load_app(self):
+    def _validate_key(self):
         pass
 
+    def change_key(self, key):
```
davisagli, line 47: This isn't used anywhere other than tests; we can probably remove it.

cumulusci/core/keychain/base_project_keychain.py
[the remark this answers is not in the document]
davisagli: I think if a new service is being set as the default because it's the first service of a type, we always want to set that at the global level, so don't need the project flag here. Explicitly setting a project-level override of the default can be done by a separate call to set_default_service.

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -201,64 +154,138 @@ def get_org(self, name: str):
...
+    def _load_default_services(self):
+        pass
+
+    def set_service(self, service_type, alias, service_config, project=False):
+        """Store a ServiceConfig in the keychain"""
+        if (
+            not self.project_config.services
+            or service_type not in self.project_config.services
+        ):
+            self._raise_service_not_valid(service_type)
```
davisagli, line 204: Should this check of the service type be part of _validate_service?

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -201,64 +154,138 @@ def get_org(self, name: str):
...
-            self._raise_service_not_valid(name)
+            raise ServiceNotValid(
+                f"Missing required attribute(s) for service: {missing_required}"
+            )
+
+    def _validate_service_alias(self, service_type, alias):
+        if alias == service_type:
+            raise ServiceNotValid(
+                "Service alias cannot be the same as the service type."
+            )
```
davisagli, line 283: Is there a reason we need this restriction or is it just to avoid confusion?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +89,410 @@ def unset_default_org(self):
...
+    ) -> None:
+        """Public API for setting a default service e.g. `cci service default`"""
+        if service_type not in self.project_config.services:
+            raise ServiceNotConfigured(f"No such service type: {service_type}")
+        elif alias not in self.services[service_type]:
+            raise ServiceNotConfigured(
+                f"No service of type {service_type} configured with name: {alias}"
+            )
+        elif alias == self._default_services[service_type]:
+            return  # already the default
```
davisagli, line 173: But it might be the global default and user is trying to set it as the project-specific default (so that it won't change for this project if the global default is later changed)

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +89,410 @@ def unset_default_org(self):
...
+        or if no service of the given type has the current_alias
+        """
+
+        if service_type not in self.services:
+            raise ServiceNotConfigured(
+                f"No services of type {service_type} are currently configured"
+            )
+        elif current_alias not in self.services[service_type]:
+            raise ServiceNotConfigured(
+                f"No service of type {service_type} configrued with the name: {current_alias}"
```
davisagli, line 196:
```suggestion
                f"No service of type {service_type} configured with the name: {current_alias}"
```

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +89,410 @@ def unset_default_org(self):
...
+            )
+
+        self.services[service_type][new_alias] = self.services[service_type][
+            current_alias
+        ]
```
davisagli, line 201: Don't we need a check to make sure the new_alias doesn't already exist?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +89,410 @@ def unset_default_org(self):
...
+        current_filepath = Path(
+            f"{self.global_config_dir}/services/{service_type}/{current_alias}.service"
+        )
+        new_filepath = Path(
+            f"{self.global_config_dir}/services/{service_type}/{new_alias}.service"
+        )
+        current_filepath.replace(new_filepath)
+
+        # look through all DEFAULT_SERVICE.json files and
+        # change current_alias to new_alias (if present)
```
davisagli, line 214: Kudos for thinking of this

docs/cli.rst
```diff
@@ -263,6 +263,61 @@ and :doc:`Connect Persistent Orgs <connected_orgs>`.
...
+    $ cci service default <service_type> <service_name> 
+
+Rename a Service
+****************
+To rename a service use:
+
+.. code-block:: console
+
+    $ cci service rename <service_type> <old_name> <new_name>
+
```
davisagli, line 318: Have you checked for existing docs that mention services to see if they need to be updated?

cumulusci/cli/cci.py
```diff
@@ -825,14 +844,19 @@ def callback(*args, **kwargs):
...
             else:
-                click.echo(f"{name} is now configured for all CumulusCI projects.")
+                click.echo(
+                    f"{service_type} is now configured for all CumulusCI projects."
+                )
 
-        ret = click.Command(name, params=params, callback=callback)
-        return ret
+        params.append(click.Argument(["service_name"]))
+        return click.Command(service_type, params=params, callback=callback)
```
davisagli, line 860: github won't let me comment on the line below where the "service connect" command is registered, but let's adjust the help to say "Connect an external service to CumulusCI" instead of "Connect a CumulusCI task service"

cumulusci/cli/cci.py
```diff
@@ -843,32 +867,63 @@ def service_connect():
 
 
 @service.command(name="info", help="Show the details of a connected service")
+@click.argument("service_type")
```
davisagli, line 870: Should we make this optional so you can get the info for the default service without specifying its name?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -160,12 +89,410 @@ def unset_default_org(self):
...
+        is of an unrecognized type.
+        """
+        service_type = item.name.replace(".service", "")
+        configured_service_types = self.project_config.config["services"].keys()
+        if service_type not in configured_service_types:
+            self.logger.info(f"Skipping migration of unrecognized service: {item.name}")
+            return None
+
+        scope = "global" if item.parent.name == ".cumulusci" else "project"
+        new_filename = f"{service_type}__{scope}.service"
```
davisagli, line 442: Should this just be f"{scope}.service"? I was expecting to have a github service named `global` after the migration, but instead I have one named `github__global`:

Screen Shot 2021-04-09 at 11 02 58 AM

davisagli:
This is only tangentially connected to this PR, but we should remove the service types `apextestsdb` and `saucelabs` from the universal cumulusci.yml, because we've already removed the code for the tasks that used them.

cumulusci/cli/cci.py
Br4nd0R: > 3\. Set both of these defaults via separate calls to set_default_service, so set_service doesn't need the project flag.

set_service would still need the project flag it is going to set a default for the first service of a given type that is added. I suppose that I could take that logic out and implment it in the cci command. So if after we add a given service, and we detect that it is the first of its kind, then we call set_default_service() in that case as well (only if  --default and --project 
are not specified)

cumulusci/core/keychain/base_project_keychain.py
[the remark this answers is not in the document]
davisagli: Are you sure we don't want to allow connecting a service if the name already exists? The use case I have in mind is _updating_ a service configuration. I tried doing this for an existing service and got this error. I think if the service name already exists we should make the `cci service connect` command prompt the user for confirmation, but still provide a way to replace it.

davisagli:
It can be a separate work item but we're going to want a `cci service remove <service_type> <service_name>` command.

cumulusci/cli/cci.py
davisagli: @Br4nd0R I think it's okay to keep the "set the first service of a type as the default for that type" logic in the keychain (and that seems preferable to make sure that it always happens). This implicit default should always be a global default, not project. And then the cci command can additionally use set_default_service to make it a project-level service if that's what the user asked for.

cumulusci/core/keychain/base_project_keychain.py
Br4nd0R: This was simply to avoid confusion.

cumulusci/core/keychain/base_project_keychain.py
Br4nd0R: I like it.

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: good catch

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: We did intend to have the names like this but I think your suggestion is better.

Br4nd0R:
> This is only tangentially connected to this PR, but we should remove the service types `apextestsdb` and `saucelabs` from the universal cumulusci.yml, because we've already removed the code for the tasks that used them.

I went ahead and removed these from the standard library, and made a 1pt story for this and included it in the sprint.

cumulusci/cli/cci.py
davisagli: Confirmed that works, but can we show a message that indicates which default was set (e.g. "for all projects" vs "for this project").

cumulusci/cli/cci.py
```diff
@@ -796,25 +813,53 @@ def _build_param(self, attribute, details):
...
-                project = False
+                set_project_default = False
             else:
-                project = kwargs.pop("project", False)
+                set_project_default = kwargs.pop("project", False)
+
+            set_global_default = kwargs.pop("default", False)
+            if set_global_default and set_project_default:
+                raise click.UsageError(
+                    "Cannot specify both --default and --project. Please choose one."
```
davisagli, line 860: This is fine, but there's no technical reason they need to be mutually exclusive.

Br4nd0R: I switched it to accept both.

davisagli (approved):

PR #2501, opened 2021-03-26
Apply the port to the instance URL if the org has a port defined.

force2b:
# Critical Changes

# Changes

- Ensure that CCI is able to connect to scratch orgs running in a local build environment by not removing the instance port

# Issues Closed


cumulusci/salesforce_api/utils.py
```diff
@@ -12,12 +13,27 @@
...
+        instance_url = urlparse(org_config.instance_url).hostname
+    except:
+        pass
+
+    try:
+        port = urlparse(org_config.instance_url).port
+        if port:
+            instance_url = instance_url.replace(".com", ".com:" + str(port))
+    except:
+        pass
```
davisagli, line 33: Are there particular errors you're expecting? I think it should generally work to use urlparse to get the hostname and port as long as the input is actually a URL, and if it's not, that's probably an error we would not want to suppress.

cumulusci/salesforce_api/utils.py
```diff
@@ -12,12 +13,27 @@
...
 ):
+
     # Retry on long-running metadeploy jobs
     retries = Retry(total=5, status_forcelist=(502, 503, 504), backoff_factor=0.3)
     adapter = HTTPAdapter(max_retries=retries)
+    instance_url = org_config.instance_url
+
+    # Attept to get the host and port from the URL, but ignore any errors retrieving it.
+    try:
+        instance_url = urlparse(org_config.instance_url).hostname
```
davisagli, line 24: I would call the variable `instance` here since it's no longer a full URL, and pass that along on line 20.

davisagli (commented):
This is relying a bit on internal details of simple-salesforce. It would be good to have a test that instantiates a simple-salesforce client using this (both with and without a port in the instance_url) and then verifies that it generates the expected base_url.

cumulusci/salesforce_api/utils.py
force2b: I had added the try/except blocks only because it was failing in the tests. It looks like the mock generates an invalid url that gets kicked back by the call to urlparse().

force2b:
@davisagli I do agree that some tests for simple_salesforce would be useful here, but I admittedly could not get the existing `test_utils.py` to pass. @prescod offered to take a look once I created the draft PR.

cumulusci/salesforce_api/utils.py
davisagli: We should probably handle that by setting up more realistic instance_urls in the tests, including when they are coming from a mock.

davisagli:
@force2b yeah no problem, we can help with the tests. Thanks for getting it started.

force2b:
@davisagli and @prescod - Code changes made per David's review, and I added a test method to validate that the port is retained. Will change this to officially ready for review now.

cumulusci/salesforce_api/utils.py
```diff
@@ -12,12 +13,27 @@
...
     project_config, org_config, api_version=None, base_url: str = None
 ):
+
     # Retry on long-running metadeploy jobs
     retries = Retry(total=5, status_forcelist=(502, 503, 504), backoff_factor=0.3)
     adapter = HTTPAdapter(max_retries=retries)
+    instance = org_config.instance_url
+
+    # Attept to get the host and port from the URL, but ignore any errors retrieving it.
+    try:
```
prescod, line 23: Just a tip to help you level-up in Python.

try/except: pass will not survive a code review and most linters will also flag it. It will swallow all sorts of weird stuff like Ctrl-C signals, out-of-memory signals etc.

```
try:

except Exception:
   pass
```

is very slightly better, but not as good as good as catching just the exception that you see being raised.

But I'll take a look at what changes to the test suite might make this unnecessary altogether.

force2b: Thanks. I had to add the try/except blocks because without them the tests (almost all) were failing with the following error. I think it's because the mocked URL that gets passed to this method (in tests) is an invalid URL that causes the urlparse library to throw an exception.
`TypeError: '>' not supported between instances of 'Mock' and 'int'`

force2b:
[W-9050867](https://gus.lightning.force.com/lightning/_classic/%2Fa07AH00000093UjYAI)

prescod (approved):

PR #2504, opened 2021-03-27
Allow dashes for task options

prescod:
# Changes

Task option command line arguments can now be specified with either an underscore or a dash. So, for example, clean_meta_xml can be specified as either --clean_meta_xml or --clean-meta-xml or -o clean-meta-xml

This conforms to normal conventions for CLI tools, compensates for a predictable error and reduces the load on the pinky finger.

# Notes

10 minutes of implementation
1+ hours figuring out how to write the test and mocks


Br4nd0R (approved):

cumulusci/cli/cci.py
```diff
@@ -1673,7 +1677,10 @@ def _get_click_options_for_task(self, task_options):
             # task level via task._validate_options() or Pydantic models.
             click_options.append(
                 click.Option(
-                    param_decls=(f"--{name}",),
+                    param_decls=(
+                        f"--{name}",
+                        f"--{name.replace('_', '-')}",
```
davisagli, line 1682: This leads to showing even options that don't have an underscore twice in --help:

     --namespace, --namespace TEXT   Package namespace

Let's add the normalized form only if it's different.

davisagli (changes requested):

cumulusci/cli/tests/test_cci.py
```diff
@@ -359,6 +359,84 @@ def test_cci_org_default__no_orgname(
...
+        task_config = Deploy.mock_calls[0][1][1]
+        assert "clean_meta_xml" in task_config.options
+
+    @mock.patch("cumulusci.cli.cci.init_logger", mock.Mock())
+    @mock.patch("cumulusci.cli.cci.tee_stdout_stderr", mock.MagicMock())
+    @mock.patch("cumulusci.tasks.salesforce.Deploy.__call__", mock.Mock())
+    @mock.patch("sys.exit", mock.Mock())
+    @mock.patch("cumulusci.cli.cci.get_tempfile_logger")
+    @mock.patch("cumulusci.cli.cci.CliRuntime")
+    @mock.patch("cumulusci.tasks.salesforce.Deploy.__init__")
```
davisagli, line 402: Is there some way to change the approach to testing here so that we either require fewer mocks or centralize setting them up? This seems hard to maintain.

cumulusci/cli/cci.py
prescod: 09a046b

davisagli (approved):

PR #2514, opened 2021-03-31
Entry point for snowfakery

prescod:
# Changes

The `snowfakery` command is now available to CumulusCI users.

# Issues Closed

### DIFF
--- a/cumulusci/tests/test_entry_points.py
+++ b/cumulusci/tests/test_entry_points.py
@@ -1,0 +1,19 @@
+from subprocess import Popen, PIPE
+import pytest
+
+
+class TestEntryPoints:
+    def _popen(self, cmd):
+        process = Popen([cmd], stdout=PIPE, stderr=PIPE)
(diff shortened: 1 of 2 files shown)


davisagli:
This seems to already be the case:
- I created a new virtualenv, ran pip install cumulusci, and verified that there is a snowfakery script in the virtualenv's `bin`.
- When installing using homebrew, our formula links the snowfakery script here: https://github.com/SFDO-Tooling/homebrew-sfdo/blob/master/cumulusci.rb#L512

So, what is the scenario where it's not already working?

davisagli:
Does this specifically make a different for pipx?

prescod:
It seems that Windows users do not have access to the 'snowfakery' command so this is an attempt to fix that. I don't know for sure if it will work, to be honest, but seemed low-risk to try.

davisagli (approved):
It makes sense to me that that would be because of pipx, and that this would help.

PR #2517, opened 2021-03-31
Added "ordering" and "testlevelsplit" options to the robot task

boakley:
This adds two new robot task options: `ordering` and `testlevelsplit`. Both of these are options that are passed through to pabot when the `processes` option is set higher than 1.  Parallel test runs are still in the experimental stage, but this will give teams a bit more to experiment with.

We don't do anything with these options other than to pass them through, so the only tests I've added assert that the options are part of the pabot command but are ignored for the normal path through the code.

Jason and I had a brief discussion on whether these should be first class options or a layer deeper under a "parallel" or "pabot" option, and we agreed that first class options are the way to go.

test_robotframework.py is getting a bit large so I split out the parallel tests into their own separate test file.

# Critical Changes

# Changes
* Added two new options to the robot task: `ordering` and `testlevelsplit`. These only have an effect when combined with the `processes` option to run tests in parallel.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/robotframework/robotframework.py
+++ b/cumulusci/tasks/robotframework/robotframework.py
@@ -59,9 +59,33 @@ class Robot(BaseSalesforceTask):
         "debug": {
             "description": "If true, enable the `breakpoint` keyword to enable the robot debugger"
         },
+        "ordering": {
+            "description": (
+                "Path to a file which defines the order in which parallel tests are run. "
+                "This maps directly to the pabot option of the same name. It is ignored "
+                "unless the processes argument is set to 2 or greater."
+            ),
+        },
         "processes": {
-            "description": "*experimental* Number of processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robotframework. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
+            "description": (
+                "*experimental* Number of processes to use for running tests in parallel. "
+                "If this value is set to a number larger than 1 the tests will run using the "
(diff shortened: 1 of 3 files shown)


cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -59,10 +59,34 @@ class Robot(BaseSalesforceTask):
...
+                "unless the processes argument is set to 2 or greater."
+            ),
+        },
         "processes": {
-            "description": "*experimental* Number of processes to use for running tests in parallel. If this value is set to a number larger than 1 the tests will run using the open source tool pabot rather than robotframework. For example, -o parallel 2 will run half of the tests in one process and half in another. If not provided, all tests will run in a single process using the standard robot test runner.",
+            "description": (
+                "*experimental* Number of processes to use for running tests in parallel. "
+                "If this value is set to a number larger than 1 the tests will run using the "
+                "open source tool pabot rather than robotframework. For example, -o parallel "
+                "2 will run half of the tests in one process and half in another. If not "
```
jstvz, line 74: Quibble: Another way to word this could be something like "*n* will split tests across *n* processes".

jstvz (approved):
Looks good to me. 

I made a suggestion about the description for `processes`, but I think it's fine as-is.

cumulusci/tasks/robotframework/robotframework.py
```diff
@@ -59,10 +59,34 @@ class Robot(BaseSalesforceTask):
...
+                "provided, all tests will run in a single process using the standard robot "
+                "test runner. See https://pabot.org/ for more information on pabot."
```
davisagli, line 76: Note to self: we should textwrap.dedent the option descriptions so that it's possible to put them in a triple-quoted block rather than breaking lines like this.

PR #2533, opened 2021-04-08
Split cci.py into multiple modules, remove global RUNTIME

davisagli:
Split apart cci.py:

- The runtime object is now passed as the "obj" on the `click` context object, instead of using a global
- Moved project commands to project.py, task commands to task.py, etc
- The remaining group is the service commands, which I'm waiting on to avoid conflicts with the named services branch.

# Critical Changes

# Changes

# Issues Closed


Br4nd0R:
I could cry I'm so happy right now 🎉

Br4nd0R (approved):

PR #2540, opened 2021-04-12
Specify a 2GP Ancestor Package

Br4nd0R:
# Changes
* The `create_package_version` task now accepts an `--ancestor-id` option. This can be set to the 04t Id of the ancestor of the package version you are about to create. If it is not specified, CumulusCI will attempt to look up the `SubscriberPackageVersionId` of your project's latest production release and use that as the ancestor Id value.


cumulusci/tasks/package_2gp.py
```diff
@@ -454,6 +461,45 @@ def _create_version_request(
...
         return response["id"]
 
+    def _resolve_ancestor_id(self, spv_id: str = None) -> str:
+        """
+        If an ancestor_id (04t) is not specified, get it
+        from the latest production release.
+
+        @param spv_id The SubscriberPackageVersionId (04t) that is the ancestor
+        to the version being created.
+        """
```
davisagli, line 471: We can't set an ancestor id for an unlocked package, so we should skip the automatic lookup if package_type is Unlocked

cumulusci/tasks/package_2gp.py
```diff
@@ -454,6 +461,45 @@ def _create_version_request(
...
+        if not spv_id:
+            tag_name = self.project_config.get_latest_tag(beta=False)
```
davisagli, line 473: This will raise GithubException if no release is found. In that case I think we want to catch the exception and use an empty ancestor id (otherwise it won't be possible to create package versions before the first release)

davisagli (changes requested):

cumulusci/tasks/package_2gp.py
Br4nd0R: I went ahead and returned an empty string if we detect an unlocked package. I think I really should test this with a sample unlocked package, unless you have confirmed that setting an empty string as an ancestor Id doesn't throw an error. 

Alternatively, I could alter the code so that we only set a value for ancestor Id on packages that are not of type `Unlocked`.

davisagli: We've been setting ancestorId to an empty string for all 2gp package versions prior to this PR, so I'm not too worried about it.

cumulusci/tasks/package_2gp.py
```diff
@@ -454,6 +461,52 @@ def _create_version_request(
...
+        If an ancestor_id (04t) is not specified, get it
+        from the latest production release.
+
+        @param spv_id The SubscriberPackageVersionId (04t) that is the ancestor
+        to the version being created.
+        """
+        if self.package_config.package_type == PackageTypeEnum.unlocked:
+            return ""
+
+        if not spv_id:
```
davisagli, line 475: I just realized -- finding the ancestor id in this way is something we want to do if the user is creating a 2gp as the primary release vehicle of a project, but not if they're creating a different package like our 2gp feature test packages (in that case the 04ts located this way would be for the wrong package!)

So how about this:
- make the task default to not setting an ancestor
- ancestor can be explicitly set to a 04t using the `ancestor_id` task option
- `ancestor_id` can be set to a string like `auto` or `last_release` in order to opt in to finding the ancestor in this way. We would set the option to this in the release_2gp_beta flow.

Br4nd0R: This sounds good to me, but something must be off in my understanding of our 2gp feature test packages.

> in that case the 04ts located this way would be for the wrong package!

I'm not understanding how resolving the ancestor Id to the latest production version of the package, is the wrong package to assign as the ancestor for a 2gp feature test package. I would have thought that we would want ancestor's to be specified as this would help to better simulate manageability rules of the package at the time commits are made.

davisagli: We may well decide to do that in some cases, but in general we can't assume that every package version created using create_package_version is for the same package. The package is determined by the package_name task option. In general that defaults to `project_config.project__package__name`, but the build_feature_test_package flow sets it to `$project_config.project__package__name Managed Feature Test`. The biggest reason we made it a separate package was out of an abundance of caution as we learn about 2gp. I think in some cases we will want to use the same package for feature testing so that we can set an ancestor.

davisagli (approved):
Approving, but let's wait to merge until after today's release.

PR #2546, opened 2021-04-13
`package_type` option for `github_release` task

Br4nd0R:
# Critical Changes
* If you have custom flows that utilize the `github_release` task, they will need to be updated to include the `package_type` option (which is required).

# Changes
* The `github_release` task now has a `package_type` option which is included in the information written to GitHub release tags. The following standard library "release" flows have been updated with hardcoded values (either `1GP` or `2GP`) for this option:
    * `release_beta` (1GP)
    * `release_production` (1GP)
    * `release_2gp_beta` (2GP)
    * `release_2gp_production` (2GP)


cumulusci/tasks/github/release.py
```diff
@@ -68,6 +71,8 @@ def _run_task(self):
         message = self.options.get("message", "Release of version {}".format(version))
         if self.options.get("version_id"):
             message += f"\n\nversion_id: {self.options['version_id']}"
+        if self.options.get("package_type"):
+            message += f"\npackage_type: {self.options['package_type']}"
```
davidmreed, line 75: The other message concatenations use two newlines as a prefix to each data element. Let's stick with that to avoid any future confusion.

cumulusci/tasks/github/release.py
```diff
@@ -35,6 +35,9 @@ class CreateRelease(BaseGithubTask):
         "resolution_strategy": {
             "description": "The name of a sequence of resolution_strategy (from project__dependency_resolutions) to apply to dynamic dependencies. Defaults to 'production'."
         },
+        "package_type": {
```
davidmreed, line 38: What do you think about making this required? If it's not populated, going forward, we will assume 1GP (those releases being created by an older version of CumulusCI), so any 2GP release _must_ declare itself as a 2GP release.

davisagli (commented):
I forget why we think we need this. Can't we install any 04t using PackageInstallRequest?

cumulusci/tasks/github/release.py
Br4nd0R: I was trying to think through the implications of making it required; I suppose I didn't want to break anyone who is using this in a CI system and figured that since we will adhere to using our standard library release flows, that these would always be populated for us (so long as we're hardcoding that value in the flows). Making it required, would come with the benefit you noted though. I'll go ahead and switch it.

Br4nd0R:
> I forget why we think we need this. Can't we install any 04t using PackageInstallRequest?

@davisagli - The concern is that we want the switchover of 1GPs from installing via the current mechanism to `PackageInstallRequest` to be on our terms since there is the potential for issues to occur. This will allow us to dictate what install method we want to use for the given package type.

davisagli:
> This will allow us to dictate what install method we want to use for the given package type for the time being.

That makes sense. I guess another way might be to look it up by querying the SubscriberPackageVersion in the org that is the installation target, and check whether it has a value for the Package2ContainerOptions field. But I think this way is okay.

davidmreed:
I was eyeing SFDX's REST-deploy problems and was concerned about the possible (unknown) impact if we switched our package installs wholesale over to the Tooling API.

davisagli:
> I was eyeing SFDX's REST-deploy problems and was concerned about the possible (unknown) impact if we switched our package installs wholesale over to the Tooling API.

Yeah -- I don't think that should stop us from trying it but we should proceed cautiously and make sure we have a way to switch back if needed in the short term.

cumulusci/tasks/github/release.py
davidmreed: We should probably call that out as a Critical Change. If someone has built their own flows rather than using ours, they'll break, although they will fail safe.

Br4nd0R: I went ahead and called it out in the Critical Changes section.

cumulusci/tasks/github/tests/test_release.py
```diff
@@ -83,6 +83,7 @@ def test_run_task(self):
                         "version": "1.0",
                         "version_id": "04t000000000000",
                         "dependencies": [{"namespace": "foo", "version": "1.0"}],
+                        "package_type": "1GP",
```
davidmreed, line 86: We're getting free code coverage because of how the logic is structured; could we add an assert somewhere to validate the generated tag message?

Br4nd0R: good call. I've gone ahead and added an assert on the request body to ensure the expected package type is set.

davidmreed (approved):

PR #2547, opened 2021-04-13
Fix Package Version Incrementation

Br4nd0R:
# Changes
* The `create_package_version` task can now increment package version numbers when the package is not in a released state.

## Dev Notes
* Changed `package_2g.py` to `create_package_version.py` to align with current naming conventions (We could probably strip out `PackageTypeEnum`, `VersionTypeEnum`, `PackageVersionNumber`, and `PackageConfig` to a separate `2gp_util.py` file as well.)


cumulusci/tasks/create_package_version.py
```diff
@@ -198,7 +198,7 @@ def _init_options(self, kwargs):
             or self.project_config.project__package__namespace,
             version_name=self.options.get("version_name") or "Release",
             version_base=self.options.get("version_base"),
-            version_type=self.options.get("version_type") or "minor",
+            version_type=self.options.get("version_type") or "patch",
```
davisagli, line 201: This should default to `build` to match the previous behavior. (When building a feature test package we want to keep the same major/minor/patch numbers, and only increment the build number by setting it to NEXT)

The release_2gp_beta flow should use this task with version_type set to `minor` though.

davisagli (changes requested):

cumulusci/tasks/create_package_version.py
Br4nd0R: done

davisagli (dismissed):

davisagli (approved):

PR #2555, opened 2021-04-19
Named service fixes

davisagli:
- Fix `cci service connect` when service isn't configured yet
- Make _read_default_services return an empty dict instead of raising an exception if the default services file doesn't exist yet

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -525,7 +521,7 @@ def _write_default_services(
                 f"No {DEFAULT_SERVICES_FILENAME} file found at: {file_path}"
             )
 
-        with open(file_path, "w") as f:
+        with open(file_path, "w", encoding="utf-8") as f:
```
codexmarcel, line 524: Good catch on the encoding safety.

codexmarcel (approved):

prescod:
Works for me!

PR #2557, opened 2021-04-19
Resolve dependencies using 04t ids in release tags

davidmreed:
Requires #2546 for end-to-end testing.

# Critical Changes

# Changes

- `update_dependencies` can now consume 2GP releases in upstream repositories, provided they're stored in release tags as generated by CumulusCI.

# Issues Closed


cumulusci/core/dependencies/tests/test_resolvers.py
```diff
@@ -132,6 +157,30 @@ def test_github_release_tag_resolver(self, project_config):
             ),
         )
 
+    def test_github_release_tag_resolver__2gp(self, project_config):
```
Br4nd0R, line 160: It looks like this is a duplicate of the test above (with a different repo name).

Edit: Ah, it looks like we are testing different repos here (that are defined in `conftest.py`), but this test and the one above it need to have unique names (they currently share the same one).

Br4nd0R (commented):
Just a couple small items. Overall this looks great!

cumulusci/core/dependencies/tests/test_resolvers.py
davidmreed: Fixed!

Br4nd0R (approved):

PR #2567, opened 2021-04-20
Improvements to cumulusci.yml warnings

davisagli:
- When parsing cumulusci.yml from a github repo, log the repo it came from instead of `<stream>`
- Only show the explanation of cumulusci.yml warnings once even if multiple files have warnings.
- Cache result of get_remote_project_config in an lru_cache so we don't fetch the same file from github multiple times.

# Critical Changes

# Changes
- Improved logging of cumulusci.yml warnings from dependencies.

# Issues Closed


prescod (approved):

davisagli:
Please wait to merge.

PR #2574, opened 2021-04-21
Log errors in stderr of sfdx commands during source conversion

Br4nd0R:
# Issues Closed
* Improved logging of errors from sfdx while converting sfdx format metadata to deploy via the Metadata API, so that they are not lost when CumulusCI is embedded in another system like MetaCI or Metecho.


davisagli:
Suggested edit for release note (and I'd consider this an Issue closed): Improved logging of errors from sfdx while converting sfdx format metadata to deploy via the Metadata API, so that they are not lost when CumulusCI is embedded in another system like MetaCI or Metecho.

davisagli (approved):

PR #2576, opened 2021-04-22
Use chromedriver 89

davisagli:
Temporary workaround for apparent issue with chromedriver 90 https://bugs.chromium.org/p/chromedriver/issues/detail?id=3782

# Critical Changes

# Changes

# Issues Closed


boakley (approved):
These look reasonable to me. 

On my machine I download the various versions under their version numbers (eg: chromedriver-90) and then make a symlink to /usr/local/bin/chromedriver, though in this case that seems like overkill.

PR #2578, opened 2021-04-22
Add set records as viewed in data load task

davidjray:
Thanks @prescod  for the help with the unit tests.

# Critical Changes

# Changes
- Added an new option`set_recently_viewed` to the `load_dataset` task that sets newly inserted data as recently viewed

# Issues Closed


prescod (approved):

prescod:
I helped write this so it wouldn't hurt to have independent eyes on it. It's small.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -101,6 +104,9 @@ def _init_options(self, kwargs):
         self.options["drop_missing_schema"] = process_bool_arg(
             self.options.get("drop_missing_schema") or False
         )
+        self.options["set_recently_viewed"] = process_bool_arg(
+            self.options.get("set_recently_viewed") or False
```
davisagli, line 108: From slack it sounds like the decision was to change this default to True, right?

cumulusci/tasks/bulkdata/load.py
```diff
@@ -631,3 +639,27 @@ def _generate_contact_id_map_for_person_accounts(
...
+        if custom_objects:
+            for record in self.sf.query_all(
+                "SELECT SObjectName FROM TabDefinition WHERE IsCustom = true AND SObjectName IN ('{}')".format(
+                    "','".join(sorted(custom_objects))
+                )
+            )["records"]:
+                object_names.add(record["SobjectName"])
+
+        for mapped_item in sorted(object_names):
+            self.sf.query_all(f"SELECT Id FROM {mapped_item} LIMIT 1000 FOR VIEW")
```
davisagli, line 665: There's nothing here to make sure that we're specifically selecting records that were just inserted, is there? Should we do a reverse sort by creation date? Should we limit it to the number of items that were inserted?

cumulusci/tasks/bulkdata/load.py
davidjray: @davisagli yessir, I am looking at making the change now, it has some repercussions on some unit tests.

prescod (approved):

davisagli (approved):

PR #2582, opened 2021-04-24
Add support for branch_prefixes that do not include a slash

jstvz:
This PR restores support for branch prefixes that do not include a forward slash to `github_automerge_main`.

-------

# Critical Changes

# Changes
- The `github_automerge_main` task has a new option, `skip_future_releases`, which can be set to False to disable the default behavior of skipping branches that are numeric (and thus considered release branches) but not the lowest number.

# Issues Closed
- Fixed an error in the github_automerge_main task when using a branch prefix that doesn't contain a slash.

### DIFF
--- a/cumulusci/tasks/github/merge.py
+++ b/cumulusci/tasks/github/merge.py
@@ -28,8 +28,11 @@ class MergeBranch(BaseGithubTask):
         "branch_prefix": {
             "description": "A list of prefixes of branches that should receive the merge.  Defaults to project__git__prefix_feature"
         },
+        "skip_future_releases": {
+            "description": "If true, then exclude branches that start with the branch prefix if they are not for the lowest release number. Defaults to True."
+        },
         "update_future_releases": {
-            "description": "If source_branch is a release branch, then merge all future release branches that exist. Defaults to False."
+            "description": "If true, then include release branches that are not the lowest release number even if they are not child branches. Defaults to False."
         },
     }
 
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/github/merge.py
```diff
@@ -211,6 +212,10 @@ def _is_release_branch(self, branch_name):
         """A release branch begins with the given prefix"""
         return is_release_branch(branch_name, self.options["branch_prefix"])
 
+    def _get_base_branch(self, branch_name):
+        """Strip branch_prefix from branch name, returning the base branch"""
```
davisagli, line 216: "base branch" is the term github uses to refer to the target branch of a pull request. Can we call this something else, maybe "_get_release_number" since it's only called from places that then `int()` it

davisagli (changes requested):

Br4nd0R (dismissed):

cumulusci/tasks/github/merge.py
```diff
@@ -28,8 +28,11 @@ class MergeBranch(BaseGithubTask):
         "branch_prefix": {
             "description": "A list of prefixes of branches that should receive the merge.  Defaults to project__git__prefix_feature"
         },
+        "skip_future_releases": {
+            "description": "If true, then exclude branches that start with the branch prefix if they are not for the lowest release number. Defaults to True."
+        },
```
Br4nd0R, line 33: > If true, then exclude branches that start with the branch prefix if they...

Which prefix is being referred to: feature, release?

Br4nd0R (approved):

cumulusci/tasks/github/merge.py
davisagli: The one named in the task option called branch_prefix. It is typically set to the feature branch prefix.

PR #2588, opened 2021-04-28
Deploy unpackaged directories in deterministic (alpha) order

davidmreed:
# Critical Changes

- `update_dependencies` now guarantees to resolve unpackaged metadata directories (subdirectories of `unpackaged/pre` and `unpackaged/post`) in alphabetical order, matching the behavior of `deploy_pre` and `deploy_post`. `unpackaged/pre/bar` will deploy prior to `unpackaged/pre/foo`. The previous behavior was undefined, which caused rare problems. This change is critical only for projects that have deployment-order dependencies between unpackaged directories located in upstream dependencies and rely on the current undefined load order.

# Changes

# Issues Closed


davidmreed:
I suggest holding this until after the release of CumulusCI 3.34 to allow a "possible breaking change" announcement, although I think the likelihood of breaking is very low.

davidmreed:
W-9103644

cumulusci/core/dependencies/dependencies.py
```diff
@@ -248,7 +248,7 @@ def _flatten_unpackaged(
             contents = None
 
         if contents:
-            for dirname in list(contents.keys()):
+            for dirname in sorted(list(contents.keys())):
```
davisagli, line 251:
```suggestion
            for dirname in sorted(contents.keys()):
```

davisagli: (`.keys()` is also unnecessary, since dicts yield their keys when iterated, but helpful for clarity I think)

davisagli (approved):

davisagli (approved):

PR #2590, opened 2021-04-29
Have install_managed check for already-installed packages

davidmreed:
# Critical Changes

# Changes

- The `install_managed` and `install_managed_beta` tasks now take no action if the specified package is already installed in the target org.

# Issues Closed


davisagli (approved):
At some point for MetaDeploy I'd like to do this by generating a preflight check, so that we can show an accurate preview of what is going to be installed. But in the meantime this is a nice straightforward adjustment. Do you think there are any scenarios where we are relying on the effects of reinstalling a package even if it is already installed?

davidmreed:
There are no use cases of that kind that I'm aware of. The only situation I'm aware of where that could have an impact is if the user had unpermissioned components and then deliberately reinstalled with `security_type: FULL` to grant all access. 

I hope no one is doing that; it's an extremely bad idea.

PR #2593, opened 2021-04-29
Add --json flag to `cci org list`

Br4nd0R:
# Changes
* The `cci org list` command can now output in JSON format with the `--json` flag.

This is utilized by the VSCode extension to show info about a user's orgs.


cumulusci/cli/org.py
```diff
@@ -319,13 +323,26 @@ def org_list(runtime, plain):
                 [org_days, not org_config.active, org_config.config_name, domain]
             )
             scratch_data.append(row)
+            json_data[row[0]] = {
```
davidmreed, line 326: This feels a bit fragile to me. I'd suggest making the JSON output independent of the column content and ordering in the table. Too easy to change the latter and break the former down the line.

davidmreed (approved):

cumulusci/cli/org.py
```diff
@@ -319,13 +324,26 @@ def org_list(runtime, plain):
...
+                "isScratch": True,
+            }
         else:
             username = org_config.config.get(
                 "username", org_config.userinfo__preferred_username
             )
             row.append(username)
             row.append(org_config.expires or "Unknown")
             persistent_data.append(row)
+            json_data[org] = {"isDefault": is_org_default, "isScratch": False}
```
davisagli, line 342: Can we use is_default and is_scratch? Javascript/Typescript tend to use camel case for variable names and keys in a mapping, but Python tends to use snake case (underscores) and CumulusCI generally follows that convention.

PR #2596, opened 2021-05-03
[CCI] First pass at multi repository backwards compatible github release notes

jofsky:
# Critical Changes

# Changes

# Issues Closed


cumulusci/tasks/release_notes/task.py
```diff
@@ -62,25 +66,29 @@ def _run_task(self):
             "prefix_beta": self.project_config.project__git__prefix_beta,
             "prefix_prod": self.project_config.project__git__prefix_release,
         }
+        for repo in self.options["repos"]:
+            print(f"Repo: {repo}\n")
+            github_info["github_repo"] = repo
+            generator = GithubReleaseNotesGenerator(
+                self.github,
+                github_info,
+                self.project_config.project__git__release_notes__parsers.values(),
```
davidmreed, line 75: This is where you'd have to intervene to make this viable for projects that customize their release notes. Each loop iteration would have to source the `cumulusci.yml` for the remote repository and parse out its relevant configuration here.

cumulusci/tasks/release_notes/task.py
```diff
@@ -119,16 +127,21 @@ class ParentPullRequestNotes(BaseGithubTask):
             "description": "force rebuilding of change notes from child branches in the given branch.",
             "required": False,
         },
+        "repos": {
```
davidmreed, line 130: I'm fuzzier on the use case for this functionality in `ParentPullRequestNotes`. What does it do or is it expected to do?

cumulusci/tasks/release_notes/task.py
jofsky: @davidmreed I'm struggling to find an easy solution to this. Have we had this issue in the past?

davidmreed: Check out `cumulusci.core.dependencies.github.get_remote_project_config()`

PR #2601, opened 2021-05-05
Handle directories missing a trailing slash

jstvz:
This PR tweaks how `github_copy_subtree` filters filenames based on an an include list. Previously, this task would silently generate incorrect/truncated commits when a directory was included in `self.options.includes` without a trailing slash. Now, it checks the zipfile namelist's directories (filenames ending with a slash) for includes that are missing a slash and appends it.

-------

# Critical Changes

# Changes

# Issues Closed


davisagli (approved):
LGTM

PR #2618, opened 2021-05-13
Fix for MC service: defer getting oauth2_client until it's needed

davisagli:
This way we can avoid getting it altogether in tubular (by overriding the _refresh_token method) until we give tubular the ability to perform the refresh token flow.

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/config/marketing_cloud_service_config.py
```diff
@@ -47,8 +44,11 @@ def _refresh_token(self):
         * access_token
         * refresh_token
         """
-        oauth_client = OAuth2Client(self._client_config.config)
-        info = oauth_client.refresh_token(self.refresh_token)
+        oauth2_client_config = self._keychain.get_service(
+            "oauth2_client", self.oauth2_client
+        )
+        oauth2_client = OAuth2Client(oauth2_client_config.config)
```
Br4nd0R, line 50: @davisagli - Can we just pass the instance of `OAuth2CLientConfig` instead of it's config dict?

Br4nd0R (approved):

cumulusci/core/config/marketing_cloud_service_config.py
davisagli: @Br4nd0R great point; adjusted

Br4nd0R (approved):

PR #2622, opened 2021-05-17
Provide a way to get package install keys from environment variables

davisagli:
Possible way to support getting package install keys from environment variables (see https://trailblazers.salesforce.com/_ui/core/chatter/groups/GroupProfilePage?g=0F9300000009M9Z&fId=0D54S00000B5niN&s1oid=00D4S0000019iJX&s1nid=0DB30000000072L&emkind=chatterCommentNotification&s1uid=0053A00000CB68z&emtm=1621258075218&fromEmail=1&s1ext=0)

Things I don't love about it:
- password_env_name needs to be added to DynamicDependency even though it doesn't directly use it, then copied to other dependency types as they are resolved
- There's still no way to obtain an install_key for transitive dependencies

# Critical Changes

# Changes

# Issues Closed


davidmreed:
If we come to a consensus with input from our community stakeholder on using environment variables, I'm much happier with that solution (mostly for the reasons already articulated in the comments above).

davidmreed (approved):

PR #2628, opened 2021-05-19
Fix documentation for skipping unpackaged metadata

leboff:
# Critical Changes

# Changes
Fix documentation for skipping unpackaged metadata. CumulusCI expects a list, documentation shows a single string value

# Issues Closed


davisagli:
@leboff Thanks for submitting this but this actually looks like a regression in functionality -- we used to support it as a string or a list, and didn't intend to remove that support. Working on a fix in #2629

PR #2633, opened 2021-05-24
[CCI] Release notes preview generation

jofsky:
# Critical Changes

# Changes
Creates a task called `gather_release_notes` which gets the latest body of a given set of releases. This task generates an html file with a table of contents and writes it out to a file names `github_release_notes.html`
# Issues Closed


jofsky:
Will write tests, at later date.

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,39 @@
...
+            if project["owner"] and project["repo"]:
+                release = self.github.repository(project["owner"],project["repo"]).latest_release().body
+                table_of_contents += f"<a href=#{project['repo']}><li><slot name={project['repo']} />{project['repo']}</li></a>"
+                release_project_header = f"<h1 slot id={project['repo']} name={project['repo']}>{project['repo']}</h1><br>"
+                release_html = self.github.markdown(release,mode="gfm",context="{}/{}".format(project["owner"],project["repo"]))
+                body += f"{release_project_header}{release_html}"
+        table_of_contents += "</ul>"
+        
+        result = table_of_contents + body
+        result = result.replace("\n","<br>") # fixing newline issue
```
davidmreed, line 44: Say more - what's the newline issue? Can it be corrected closer to its source? It seems to be constraining you to construct your HTML without any line breaks to avoid extra `<br>` tags, which is a little clunky.

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,39 @@
...
+        with open("github_release_notes.html", "w") as f:
```
davidmreed, line 45: It doesn't look like we're generating a complete HTML file.

davidmreed (changes requested):

cumulusci/cumulusci.yml
```diff
@@ -314,6 +314,16 @@ tasks:
     description: Creates a Github release for a given managed package version number
     class_path: cumulusci.tasks.github.CreateRelease
     group: GitHub
+  display_release_notes:
```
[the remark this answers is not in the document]
jofsky: So I believe it's a combination of the notes to provide a html dump they can put into ezdita (I don't know the spelling). Jason had expressed the eventual desire to capture beta's however I'm unsure how to do it with the current implementation if you have any idea's I'd love to hear!

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,39 @@
...
+    }
+
+    def _run_task(self):
+        table_of_contents = "<h1>Table Of Contents</h1><ul>"
+        body = ""
+        for project in self.options["repos"]:
+            print(f"Owner: {project['owner']}")
+            print(f"Repo: {project['repo']}")
+            if project["owner"] and project["repo"]:
+                release = self.github.repository(project["owner"],project["repo"]).latest_release().body
```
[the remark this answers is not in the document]
jofsky: Do you know how I can achieve this?

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,39 @@
...
+                table_of_contents += f"<a href=#{project['repo']}><li><slot name={project['repo']} />{project['repo']}</li></a>"
+                release_project_header = f"<h1 slot id={project['repo']} name={project['repo']}>{project['repo']}</h1><br>"
```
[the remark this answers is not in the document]
jofsky: I just followed https://www.hongkiat.com/blog/create-auto-generated-table-of-contents-html-slot/ should I remove?

cumulusci/tasks/release_notes/task.py
jofsky: So I `suspect` that its reading newline raw bytes and as a result an extra \n is being displayed in the html display. I replaced it with a <br> to `html`ify it.

cumulusci/tasks/release_notes/task.py
jofsky: Do you mind expanding upon that?

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,44 @@
...
 
+class AllGithubReleaseNotes(BaseGithubTask):
+
+    task_options = {
+        "repos": {
+            "description": (
+                "The list of owner, repo key pairs for which to generate release notes."
+                + " Ex: 'owner': SalesforceFoundation 'repo': 'NPSP'"
+            ),
+            "required": False,
```
davidmreed, line 24: This needs to be required or the task throws an exception.

cumulusci/tasks/release_notes/task.py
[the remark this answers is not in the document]
davidmreed: Maybe I'm still not understanding what it's doing here, but I don't see that you really need slots since you're generating the entire document and static ToC in parallel. You define the anchor for each product as you add it.

cumulusci/tasks/release_notes/task.py
davidmreed: The output is a document fragment - it has no `<html>`, `<head>`, or `<body>` element.

It's possible that that isn't wrong for a specific use case, but it's making me a little concerned about shipping this task to the public because we will have to explain and support it. I'm wondering whether this should be kept internal to support the SFDO docs team.

cumulusci/cumulusci.yml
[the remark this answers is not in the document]
davidmreed: easyDITA. 

`gather_release_notes` perhaps could be a name, or we could go ask the professional writers who are the audience for the task :)

cumulusci/tasks/release_notes/task.py
davidmreed: The table of contents needs to go inside the `<body>` tag. I would suggest also adding a `<head>` section with the title set to something like "Release Notes".

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,42 @@
...
+        body = ""
+        for project in self.options["repos"]:
+            if project["owner"] and project["repo"]:
+                release = (
+                    self.github.repository(project["owner"], project["repo"])
+                    .latest_release()
+                    .body
+                )
+                table_of_contents += f"<a href=#{project['repo']}><li name={project['repo']}>{project['repo']}</li></a>"
+                release_project_header = f"<h1 id={project['repo']} name={project['repo']}>{project['repo']}</h1>"
```
davidmreed, line 39: HTML attributes here need to be quoted.

cumulusci/tasks/release_notes/task.py
```diff
@@ -13,6 +13,42 @@
...
+                body += f"{release_project_header}<hr>{release_html}<hr>"
+        table_of_contents += "</ul><br><hr>"
+        result = f"<html>{table_of_contents}<body>{body}</body></html>"
+        with open("github_release_notes.html", "w") as f:
+            f.write(result)
+
+
 class GithubReleaseNotes(BaseGithubTask):
 
     task_options = {
```
davidmreed, line 54: We do need tests to complete review.

I'm still interested in asking whether this should be publicly released or used internally by SFDO doc writers. Do you know where the original ask or TD is located?

davidmreed (changes requested):

cumulusci/cumulusci.yml
```diff
@@ -314,6 +314,10 @@ tasks:
     description: Creates a Github release for a given managed package version number
     class_path: cumulusci.tasks.github.CreateRelease
     group: GitHub
+  gather_release_notes:
+    description: Generates release notes by parsing pull request bodies of merged pull requests between two tags
```
Br4nd0R, line 318: This description is the same as `github_release_notes`, what is this task doing differently between the two?

cumulusci/tasks/release_notes/tests/test_task.py
```diff
@@ -1,15 +1,27 @@
...
+
+class TestAllGithubReleaseNotes:
+    def test_run_AllGithubReleaseNotes_task(
+        self,
+    ):
+        task = create_task(AllGithubReleaseNotes, options={"repos": [{"owner": "SalesforceFoundation", "repo": "NPSP"}]})
+        task.github = mock.Mock()
+        task.get_repo = mock.Mock()
+        task._run_task()
+        assert os.path.isfile("./github_release_notes.html")
```
Br4nd0R, line 24: let's stick with using pathlib for assertions like this. You can `from pathlib import Path` and then this should be `assert Path('./github_release_notes.html').is_file()`.

davidmreed: I'd also like to see some level of assertion that the task is behaving as expected, i.e., attempting to include each of the named repositories. We probably want to run this in a tempdir as well so that the CWD is not polluted.

cumulusci/tasks/push/tests/test_push_tasks.py
```diff
@@ -448,9 +448,7 @@ def test_schedule_push_org_query_get_org_error():
         options={
             "orgs": ORG,
             "version": VERSION,
-            "namespace": NAMESPACE,
```
davidmreed, line 451: Is this change intended to be in this PR?

cumulusci/tasks/release_notes/tests/test_task.py
Br4nd0R: I think we should also be validating the output of the html file to a certain extent. You should be able to mock a response from github where we are getting back the change notes from a repository, and then assert that the actual file contents are what is expected given the mocked change notes.

cumulusci/tasks/push/tests/test_push_tasks.py
jofsky: Yeah I figured I'd increase code coverage while were at it.

davidmreed (approved):

PR #2634, opened 2021-05-25
Fix 'cci org connect --sandbox'

prescod:
# Changes

Fix regression that prevented `cci org connect --sandbox` from launching test.salesforce.com.


prescod:
I don't believe that the test case in #2630 would have detected the problem.

I can reintroduce the same bug and all of the tests pass.

This PR has tests that prevent regression.

cumulusci/cli/tests/test_org.py
```diff
@@ -143,6 +145,72 @@ def test_org_connect(self, oauth2client):
...
+            "cumulusci.cli.cci.CliRuntime", mock.Mock()
+        ):
+            yield
+
+    @responses.activate
+    def test_org_connect_sandbox(self):
+        def check_and_quit_function_halfway_through(auth_uri, token_uri, **kwargs):
+            assert "test.salesforce" in auth_uri
+            assert "test.salesforce" in token_uri
+            raise StopIteration("Done!")
```
davisagli, line 167: This way of testing is, I think, unusual enough to warrant a comment explaining it. Wouldn't it be simpler to refactor org_connect to use one function for setting up the client and another one for actually making use of it, so that you can simply mock the latter?

davisagli (commented):
I'll let @Br4nd0R review since he worked on #2630. Should we also remove the test he added there which doesn't actually prove the fix?

cumulusci/cli/tests/test_org.py
Br4nd0R: > I can reintroduce the same bug and all of the tests pass.

@prescod - How are you reproducing the issue?

Br4nd0R:
> I can reintroduce the same bug and all of the tests pass.

@prescod - How are you reproducing the issue?

prescod:
@Br4nd0R : To clarify: I didn't say that I could reproduce the bug on main.

I said that if I *re-introduce* the line of code that caused the bug, the test case will not notice. So I'm suggesting these other test cases which do catch the bug.

Br4nd0R:
Thanks for clarifying. Yes, I'm realizing that it's hard to test the `--default-value` on an option because the `run_click_command` function seems to require values for _all_ options on a given command.

Br4nd0R:
@prescod @davisagli - I refactored the org_connect command into two functions and then fixed up the tests (as well as removed my original test). Confirmed that adding a default value to the `--login-url` option now causes a test failure.

davisagli (approved):

PR #2641, opened 2021-05-25
fix 2gp beta tags

Br4nd0R:
# Changes
* The `github_release` task now has an `is_beta` option which indicates whether the release is associated with a beta package version.  

# Issues Closed
* Fixed a bug where beta tags created for 2GP packages were not receiving the proper tag prefix.

Should I just make the option required outright? As is, the option is required for a production release (which I don't think is necessarily bad as a precaution). Would the option be better if it were named `is_production` instead?


Br4nd0R:
Went with an alternate solution: https://github.com/SFDO-Tooling/CumulusCI/pull/2642

PR #2642, opened 2021-05-26
fix release_2gp_beta tags

Br4nd0R:
# Changes
* The `github_release` task has a new option `tag_prefix`. This option can be set to specify what prefix you would like to use when CumulusCI creates a release tag for you in GitHub.

# Issues Closed
* Fixed a bug where beta tags created via the `release_2gp_beta` flow were not receiving the proper tag prefix.


davisagli (approved):

cumulusci/cumulusci.yml
```diff
@@ -1042,6 +1042,7 @@ flows:
           ancestor_id: latest_github_release
           version_base: latest_github_release
           version_type: minor
+          tag_prefix: $project_config.project__git__prefix_beta
```
davisagli, line 1045: @Br4nd0r The tag_prefix option is on the wrong step and I didn't notice :( It's an option for github_release.

PR #2643, opened 2021-05-26
Refactor: `cleanup_org_cache_dirs()` to `EncryptedFileProjectKeychain`

Br4nd0R:
# Implementation Notes
* The `cleanup_org_cache_dirs()` utility method has been refactored to be a part of the keychain classes. Looking at the original arguments it took, this made sense as it needed a keychain and project_config to execute (both of which the keychain is/has). It also makes sense to keep a cleanup method related to orgs with the other org related methods in the keychain.
* I ported over the existing tests (with some tweaks) and kept them as their own class inside `encrypted_file_project_keychain.py`.
* I did not add any additional tests, as the problem this fixes (this method being invoked when an `EnvironmentProjectKeychain` is utilized is no longer a possibility.


cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -160,6 +159,9 @@ def list_orgs(self):
     def _raise_org_not_found(self, name):
         raise OrgNotFound(f"Org named {name} was not found in keychain")
 
+    def cleanup_org_cache_dirs(self):
+        raise NotImplementedError("Subclasses must implement this method.")
```
davisagli, line 163: Let's make this pass rather than raising. In other words, make it so there is a default implementation which does nothing, rather than requiring subclasses to implement it as part of implementing the keychain contract. That way if there is a subclass of BaseProjectKeychain outside of cumulusci (I think there is in MetaCI), we won't break it.

prescod (approved):

prescod:
I don't see any problems but this is a complex hierarchy which is hard to keep in my head (especially when considering Tubular, MetaDeploy and MetaCI), so I'm mostly relying on unit tests more than confidence that everything is correct in my head.

PR #2644, opened 2021-05-26
Delete org when error occurs running flow

Br4nd0R:
# Issues Closed
* CumulusCI will now remove orgs when the `--delete-org` option is passed to `cci flow run`, even if an error occurs while running the flow.

## Dev Notes
I may be missing something here so let me know if I'm way off base. Not sure if there's a good reason we had just a `try/finally` block without an `except` clause in there. It looks like we would always display the "flow complete" message even when a flow encountered an error.

### DIFF
--- a/cumulusci/cli/flow.py
+++ b/cumulusci/cli/flow.py
@@ -158,20 +158,15 @@ def flow_run(runtime, flow_name, org, delete_org, debug, o, skip, no_prompt):
         coordinator.run(org_config)
         duration = datetime.now() - start_time
         click.echo(f"Ran {flow_name} in {format_duration(duration)}")
-    except Exception as e:
-        click.echo(
-            click.style(f"An error occurred while running the flow: {e}", fg="red"),
-            err=True,
-        )
-    else:
+    finally:
         runtime.alert(f"Flow Complete: {flow_name}")
 
-    # Delete the scratch org if --delete-org was set
-    if delete_org:
-        try:
-            org_config.delete_org()
-        except Exception as e:
-            click.echo(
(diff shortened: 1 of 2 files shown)


davisagli (changes requested):
@Br4nd0R The fix I had in mind for this was to indent the "if delete_org:" block so that it is inside the "finally:" block and will run even if the flow raised an exception.

I don't think we want to suppress the raising of the exception. That will prevent the stacktrace from being captured in the cci log.

And I don't think we need to explicitly log the exception here. That is already done within the flowrunner.

davisagli (approved):
LGTM.

If you also want to move `runtime.alert(f"Flow Complete: {flow_name}")` outside of `finally` so that we only log that when the flow completed without an exception, I'm okay with that too.

cumulusci/cli/flow.py
```diff
@@ -170,3 +168,5 @@ def flow_run(runtime, flow_name, org, delete_org, debug, o, skip, no_prompt):
                     "Scratch org deletion failed.  Ignoring the error below to complete the flow:"
                 )
                 click.echo(str(e))
+
+    runtime.alert(f"Flow Complete: {flow_name}")
```
davisagli, line 172: @Br4nd0R sorry for the back and forth on this but I just realized we also should call `runtime.alert` with a different message when there is an error. It's the thing that causes an OS notification to pop up and we wouldn't want to lose that when there's an error.

Br4nd0R: no problem

davisagli (approved):

PR #2651, opened 2021-06-01
Fix: `release_2gp_beta` flow

Br4nd0R:
# Changes
* The `release_2gp_beta` flow now sets the `force_upload` option to `True`. This means that a new package version will be created _even if_ there are no changes to the package contents.

# Issues Closed
* Fixed an issue where the `tag_prefix` option was being used on the wrong task in the `release_2gp_beta` flow.


davisagli:
@Br4nd0R I'd phrase the issue closed as "Fixed a bug where github tags created by the release_2gp_beta flow were not using the correct prefix." (in other words, focus on the symptom that users might have experienced, not what we did to fix it)

davisagli (approved):

PR #2653, opened 2021-06-01
Handle dependencies in generate_data_dictionary

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed a bug in how `generate_data_dictionary` handles dependencies.


cumulusci/core/dependencies/resolvers.py
```diff
@@ -402,13 +402,29 @@ def get_resolver_stack(
...
+def dependency_filter_ignore_deps(ignore_deps: List[dict]) -> Callable:
+    ignore_github = [d["github"] for d in ignore_deps if "github" in d]
+    ignore_namespace = [d["namespace"] for d in ignore_deps if "namespace" in d]
+
+    def should_include(some_dep: Dependency) -> bool:
+
+        if isinstance(some_dep, PackageNamespaceVersionDependency):
+            return some_dep.namespace not in ignore_namespace
+        if isinstance(some_dep, BaseGitHubDependency):
+            return some_dep.github not in ignore_github
```
davisagli, line 414: This looks like something that could maybe be refactored into a method on Dependency classes, but it's also simple enough as is that I wouldn't necessarily rush to do that.

cumulusci/tasks/datadictionary.py
```diff
@@ -101,72 +113,56 @@ def _init_options(self, kwargs):
...
-            if "github" in dependency:
-                repo = self.project_config.get_repo_from_url(dependency["github"])
-                if repo is None:
-                    raise DependencyResolutionError(
-                        f"Github repository {dependency['github']} not found or not authorized."
-                    )
-                _, ref = self.project_config.get_ref_for_dependency(
-                    repo, dependency, include_beta
-                )
+        github_deps = set()
```
davisagli, line 130: Unlike dicts, sets are not guaranteed to return entries in insertion order. Does that matter here?

davidmreed: I don't think so. The data dictionary generator makes no commitment about the ordering of output data, and if it did I think alphabetical by package name is more likely to be correct than dependency resolution order.

davisagli (approved):

PR #2655, opened 2021-06-02
Make `install_managed` handle 2GP releases

davidmreed:
# Critical Changes

# Changes

- The `install_managed` task now supports 2GP releases.

# Issues Closed


cumulusci/tasks/salesforce/install_package_version.py
```diff
@@ -82,18 +81,43 @@ def _init_options(self, kwargs):
...
+                    dependency.managed_dependency, PackageNamespaceVersionDependency
+                ):
+                    self.options["version"] = dependency.managed_dependency.version
+                elif isinstance(
+                    dependency.managed_dependency, PackageVersionIdDependency
+                ):
+                    self.options["version"] = dependency.managed_dependency.version_id
+            else:
+                raise CumulusCIException(
+                    f"The release for {version} does not identify a managed package."
```
davisagli, line 119: Can we not handle unlocked packages this way?

davidmreed: Good point; revising the message.

cumulusci/tasks/salesforce/install_package_version.py
```diff
@@ -82,18 +85,44 @@ def _init_options(self, kwargs):
...
+                    dependency.managed_dependency, PackageNamespaceVersionDependency
+                ):
+                    self.options["version"] = dependency.managed_dependency.version
+                elif isinstance(
+                    dependency.managed_dependency, PackageVersionIdDependency
+                ):
+                    self.options["version"] = dependency.managed_dependency.version_id
+                    self.options[
+                        "version_number"
+                    ] = dependency.managed_dependency.version_number
```
davisagli, line 121: I was a bit confused by why you're doing this rather than simply storing the dependency instance, but then I realized it needs to work like this in order to support freezing.

cumulusci/tasks/salesforce/tests/test_install_package_version.py
```diff
@@ -78,33 +84,134 @@ def test_init_options():
...
+    mock_GitHubDynamicDependency.return_value.managed_dependency = None
+
+    with pytest.raises(CumulusCIException) as e:
+        create_task(
+            InstallPackageVersion,
+            {"version": "latest_beta"},
+            project_config=project_config,
+        )
+
+        assert "does not identify a managed package" in str(e)
```
davisagli, line 213: This assertion is out of date, and not running because create_task raises the expected exception.

If you want to make an assertion about the exception message, do:

`with pytest.raises(CumulusCIException, match="some regexp"):`

davisagli (changes requested):
Mostly looks good, just one piece of test cleanup

cumulusci/tasks/salesforce/tests/test_install_package_version.py
davidmreed: Whoops! Fixed.

davisagli (approved):

PR #2656, opened 2021-06-02
fix shell quotes on windows

Br4nd0R:
# Issues Closed
* Fixed and issue where `sfdx` commands that had options with spaces failed to execute on Windows.


cumulusci/core/tests/test_sfdx.py
```diff
@@ -23,6 +25,15 @@ def test_windows_quoting(self, Command):
         cmd = Command.call_args[0][0]
         assert cmd == r'sfdx cmd "a\"b" -u token'
 
+    @mock.patch("platform.system", mock.Mock(return_value="Windows"))
+    @pytest.mark.skipif(
+        not sys.platform.startswith("win"),
+        reason="Special handling of shell quotes on windows",
+    )
```
davisagli, line 32: Since you're mocking platform.system you don't need to skip the test based on the platform, right?

davisagli (changes requested):

cumulusci/core/tests/test_sfdx.py
Br4nd0R: Ah yes. My intent was to remove the call to `mock.patch` and have this execute in the windows feature test check in GitHub. Does that sound alright to you?

davisagli: @Br4nd0R I'd rather go the other way and keep the mock but remove the skip. That way if we make a change that breaks it, we'll find out regardless of which platform we run the test on.

Br4nd0R: Sure, that works for me.

davisagli (approved):

PR #2659, opened 2021-06-04
Move context vars dependency

prescod:
Contextvars is actually a prod dependency on Python 3.6, not dev.


davisagli:
@prescod @Br4nd0R Let's do this as part of this week's dependency update because it won't actually take effect until the requirements/*.txt files are re-compiled.

We should probably use [environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers) so that it is only a requirement on the versions of Python that need it.

prescod:
@Br4nd0R It looks like contextvars is still in dev.in when it should be prod.in:

https://github.com/SFDO-Tooling/CumulusCI/blob/a5723ca173b005ded982d51382781acca047337b/requirements/dev.in#L21

davisagli (approved):
If you make changes to the `requirements/*.in` files, you must recompile to update the `requirements/*.txt` files. I took care of it.

prescod:
@davisagli I think that it is actually best to update *.txt files as part of the weekly updates. Otherwise we're setting ourselves up for tons of merge conflicts.

davisagli:
@prescod Hmm, maybe. We don't have a lot of PRs that update dependencies aside from the weekly updates, so I don't think there will be that many conflicts. And I want to make sure we're actually testing PRs with the correct dependencies before merging.

PR #2660, opened 2021-06-07
Feature/robot seleniumlibrary 5

boakley:
This updates the SeleniumLibrary dependency to be the latest version. There are a few small backward incompatibilities which don't seem to affect very many of our tests. However, the keyword 'select window' was renamed to 'switch window' which would break several of our tests, so I added a 'select window' keyword to Salesforce.py along with our own deprecation warning.

Selenium has been warning about this deprecation for a while, but some of our tests still use the keyword. Adding the keyword back to Salesforce.py gives our teams a bit longer grace period. I'll write a blog post warning people that we will be removing the keyword in a future release.

# Critical Changes

# Changes
* upgrades the SeleniumLibrary dependency to 5.x
* adds new keyword "select window" to Salesforce library, to replace the keyword of the same name which was renamed in SeleniumLibrary 5.x to 'switch window'. We will be removing this keyword in a future release, tests should use 'switch window' instead.

# Issues Closed

### DIFF
--- a/.github/workflows/release_test.yml
+++ b/.github/workflows/release_test.yml
@@ -7,7 +7,7 @@ on:
       - main
 
 env:
-  CUMULUSCI_ORG_scratch: '{"config_file": "orgs/dev.json", "scratch": true}'
+  CUMULUSCI_ORG_scratch: '{"config_file": "orgs/prerelease.json", "scratch": true}'
   CUMULUSCI_ORG_packaging: ${{ secrets.CUMULUSCI_ORG_packaging }}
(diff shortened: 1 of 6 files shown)


davisagli:
The failing Release Test build is not your fault; it's because the instance where CumulusCI-Test's packaging org lives has upgraded to 232 but the scratch org used in the test didn't yet. I'll try to fix...

boakley:
@davisagli it looks like the prerelease build is still failing. Is there something else we need to do?

davisagli:
@boakley This one is a different error, the same one that some people were seeing on Friday. Let me try something...

davisagli (approved):

PR #2664, opened 2021-06-08
Make built-in connected app a first class service

Br4nd0R:
# Critical Changes
* The `connected_app` task has been renamed to `create_connected_app` to more accurately describe what the task does. Any references to this task in scripts or CI systems will need to be adjusted accordingly.

# Changes
* The built-in connected app that ships with CumulusCI is now first-class service which is visible in the output of the `cci service list` command. This allows users to more easily view and change which connected app service is the current default if more than one is present. The built-in connected app service has the name `built-in` and cannot be renamed or removed.

# Issues Closed
* Fixed a small bug where `cci service info <service_type>` would display "None" as the name for the default service of the given type if no name was provided.

## Notes
* Let me know if we don't want to rename the task; but I'm thinking that its impact is likely pretty low.
* Thoughts on the service showing with name of `built-in`? Names that include `default` are all misleading because it is easily confused with implying that this built-in service is the default (which it might not be).


cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -223,7 +236,12 @@ def get_service(self, service_type, alias=None):
         return service
 
     def _get_service(self, service_type, alias):
-        return self.services[service_type][alias]
+        try:
+            return self.services[service_type][alias]
+        except KeyError:
+            raise ServiceNotConfigured(
+                f"No service of type {service_type} configured with name: {alias}"
+            )
```
Br4nd0R, line 244: This was needed for tests in `test_connectedapp.py` as that test class instantiates the `BaseProjectKeychain` instead of `EncryptedFileProjectKeychain`. This shouldn't affect anything as `BaseProjectKeychain` is not used on its own and `EncryptedFileProjectKeychain` implements its own `_get_service()` method.

cumulusci/cli/service.py
```diff
@@ -229,8 +229,13 @@ def service_info(runtime, service_type, service_name, plain):
             ]
         )
         wrap_cols = ["Value"] if not plain else None
+        service_name = (
+            runtime.keychain._default_services[service_type]
```
davisagli, line 233: We should avoid directly accessing a structure that is meant to be internal to the keychain implementation. Should we add a method like get_default_service_name to BaseProjectKeychain and move these lines to it? https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/keychain/base_project_keychain.py#L212-L216

That will also make sure this gives a meaningful error if there isn't a default set for some reason.

cumulusci/cumulusci.yml
```diff
@@ -117,18 +117,18 @@ tasks:
     description: Execute a series of REST API requests in a single call
     class_path: cumulusci.tasks.salesforce.composite.CompositeApi
     group: Data Operations
-  connected_app:
+  create_community:
+    description: Creates a Community in the target org using the Connect API
+    class_path: cumulusci.tasks.salesforce.CreateCommunity
+    group: Salesforce Communities
+  create_connected_app:
```
davisagli, line 124: Renaming the task config is not backwards compatible; I don't think we should do it even though the name would be clearer.

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -170,7 +176,16 @@ def _load_services(self):
...
+        self._default_services["connected_app"] = DEFAULT_CONNECTED_APP_NAME
+
+    def _load_default_connected_app(self):
+        """Load the default connected app as a first class service on the keychain.
+        Sets this as the default connected app service."""
+        if "connected_app" not in self.config["services"]:
+            self.config["services"]["connected_app"] = {}
+        self.config["services"]["connected_app"][
+            DEFAULT_CONNECTED_APP_NAME
+        ] = DEFAULT_CONNECTED_APP
```
davisagli, line 188: The docstring says this sets it as the default, but the code is doing it in a way that won't do that, isn't it?

davisagli (changes requested):
A few comments but this looks pretty solid.

cumulusci/cli/service.py
Br4nd0R: I should have realized this was a code smell when I wrote it. I like the idea of including another Public API method (get_default_service) that can be utilized. I'm doing some similar re-work in the merge-keychain PR so that as few methods as possible are directly mutating the `keychain.config` attribute.

cumulusci/core/keychain/base_project_keychain.py
Br4nd0R: You're correct. This should only load the service.

davisagli (approved):

PR #2668, opened 2021-06-08
Week 23 Dependency Updates

Br4nd0R:


requirements/prod.txt
```diff
@@ -90,21 +92,24 @@ requests==2.25.1
     #   simple-salesforce
 robotframework-lint==1.1
     # via -r requirements/prod.in
-robotframework-pabot==1.13
+robotframework-pabot==2.0.0
```
Br4nd0R, line 95: @boakley - Do you have opinions about updating to `robotframework-pabot v2.0.0`?

davisagli: It looks like the only change in v2 is that it now depends on https://pypi.org/project/robotframework-stacktrace/ and enables it as a listener.

davisagli (approved):

requirements/prod.txt
boakley: Yeah, I saw that pabot 2.0 was released a couple days ago. I haven't hand a chance to review it, but there doesn't look to be any backward-compatibile issues. robotframework-stacktrace is also very new. It's something we might want to use in our robot task in the future.

I'm cool with upgrading to 2.0.

PR #2671, opened 2021-06-09
Data Dictionary Prerelease Support

davidmreed:
# Critical Changes

# Changes

- The `generate_data_dictionary` task includes a new option, `include_prerelease`. If set to `True`, CumulusCI will include schema from the current branch on GitHub in the data dictionary with the version listed as "Prerelease".

# Issues Closed


cumulusci/tasks/datadictionary.py
```diff
@@ -122,6 +156,22 @@ def _init_options(self, kwargs):
                     "Only GitHub dependencies are currently supported."
                 )
 
+        self.options["include_prerelease"] = process_bool_arg(
+            self.options.get("include_prerelease") or False
```
Br4nd0R, line 160: Slightly more concise 🤷‍♂️ 
```suggestion
            self.options.get("include_prerelease", False)
```

cumulusci/tasks/datadictionary.py
```diff
@@ -348,10 +432,12 @@ def _process_object_element(self, sobject_name, element, version):
...
             self.sobjects[sobject_name].append(
                 SObjectDetail(
-                    version,
-                    sobject_name,
-                    element.label.text,
-                    description_elem.text if description_elem is not None else "",
+                    version=version,
+                    api_name=sobject_name,
+                    label=element.label.text,
+                    description=description_elem.text
```
Br4nd0R, line 438: I like the use of kwargs here (and elsewhere). Definitely an improvement to readability.

cumulusci/tasks/datadictionary.py
```diff
@@ -434,18 +527,24 @@ def _process_field_element(self, sobject, field, version):
...
-                description_text_elem.text if description_text_elem else "",
-                valid_values,
+                version=version,
+                sobject=sobject,
+                api_name=field_name,
+                label=field.label.text,
+                type=f"{field_type}{length}",
+                help_text=help_text_elem.text if help_text_elem else "",
+                description=description_text_elem.text if description_text_elem else "",
+                valid_values=valid_values,
```
Br4nd0R, line 537: +1

cumulusci/tasks/datadictionary.py
```diff
@@ -434,18 +527,24 @@ def _process_field_element(self, sobject, field, version):
...
             )
             fully_qualified_name = f"{sobject}.{fd.api_name}"
             self.fields[fully_qualified_name].append(fd)
 
+    @staticmethod
```
Br4nd0R, line 542: What's the reason for this being static? It looks like we already have an instance of the task when these calls are being made.

Br4nd0R (approved):
This looks good to me! Only a couple questions/suggestions.

cumulusci/tasks/datadictionary.py
davidmreed: Pydantic models (or dataclasses) >> namedtuples.

cumulusci/tasks/datadictionary.py
davidmreed: `process_bool_arg()` documents that `None` is accepted as falsy but that behavior is deprecated, so this construct ensures we don't pass an explicit or implicit None.

cumulusci/tasks/datadictionary.py
davidmreed: It's not really _needed_ and perhaps is more confusing than it is helpful; it's static simply because it does not access the task's state.

PR #2675, opened 2021-06-10
Fix GitHub authentication for Homebrew releases

davidmreed:
`make release-homebrew` wasn't working due to an issue in the `push-homebrew` script.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/utility/push-homebrew.py
+++ b/utility/push-homebrew.py
@@ -20,7 +20,7 @@ def get_github_user():
         CliRuntime().project_config, CliRuntime().get_keychain_key()
     )
     github_config = keychain.get_service("github")
-    return github_config.username, github_config.password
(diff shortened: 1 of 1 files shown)


utility/push-homebrew.py
```diff
@@ -20,7 +20,7 @@ def get_github_user():
         CliRuntime().project_config, CliRuntime().get_keychain_key()
     )
     github_config = keychain.get_service("github")
-    return github_config.username, github_config.password
+    return github_config.username, (github_config.password or github_config.token)
```
davisagli, line 23: This can just be 
```suggestion
    return github_config.username, github_config.token
```

The deprecated password key is silently copied to token when the service is loaded

davisagli (approved):

PR #2676, opened 2021-06-11
Feature/merge keychains

Br4nd0R:
# Changes
* CumulusCI now automatically recognizes services and orgs configured via environment variables. See the docs [LINK HERE] for more details.

# Issues Closed
* Fixed an issue where scratch orgs failed to be deleted in CI environments.
* Fixed an issue where deleting an org failed to mark the org as deleted on CumulusCI's keychain.
* Fixed an issue where CumulusCI would fail on Linux distributions that were incompatible with the `keyring` package.


cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -147,45 +116,60 @@ def get_default_org(self):
                 return org, org_config
         return None, None
 
+    def get_org(self, name: str):
```
Br4nd0R, line 119: Most of the changes to this file was me moving around the Public API methods to be at the top of their respective sections.

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
```diff
@@ -188,17 +255,117 @@ def test_load_service_files(self, keychain):
...
+
+        service_filepath = Path(
+            keychain.global_config_dir, "services/github/alias.service"
+        )
+        assert service_filepath.is_file()
+
+        # ensure expected file permissions
+        stat_result = os.stat(service_filepath)
+        actual_mode = oct(stat_result.st_mode & 0o777)
+        assert actual_mode == "0o666"
```
Br4nd0R, line 367: From what I can gather on inspection of the file in windows, it seems to have the correct file permissions. I've verified with `cmd.exe` using [icacls](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/icacls#additional-references) which _seems_ like what we want:
```
(cci-venv) D:\Users\bparker\cumulusci>icacls \Users\bparker\AppData\Local\Temp\1\fake_home_tj8bb3in\.cumulusci\services\github\alias.service
\Users\bparker\AppData\Local\Temp\1\fake_home_tj8bb3in\.cumulusci\services\github\alias.service NT AUTHORITY\SYSTEM:(I)(F)
                                                                                                BUILTIN\Administrators:(I)(F)
                                                                                                roundata\bparker:(I)(F)

Successfully processed 1 files; Failed processing 0 files
```
This seems to indicate that there are three "groups" of users that have the `(I)` and `(F)`:
* NT AUTHORITY\SYSTEM
* BUILTIN\Administrators
* roundata\bparker

The permissions are as follows:
* (I) - permission inherited from the parent container
* (F) - full access

I can't find anything that mentions why `os.stat()` would return a value that differs on windows environments.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -95,30 +120,32 @@ def _encrypt_config(self, config):
...
-            )
-        # Convert bytes created in Python 2
-        config_dict = {}
-        for k, v in unpickled.items():
-            if isinstance(k, bytes):
-                k = k.decode("utf-8")
-            if isinstance(v, bytes):
-                v = v.decode("utf-8")
-            config_dict[k] = v
+        if self.key:
```
davisagli, line 123: I guess this is just a matter of taste but I would do:
```
if self.key is None:
    return
```

which makes it clearer that nothing more is going to happen in this case, and avoids extra indentation.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -197,34 +257,69 @@ def _load_org_files(self, dirname: str, constructor=None):
                     constructor(config) if constructor else config
                 )
 
-    def _set_org(self, org_config, global_org):
+    def _set_org(self, org_config, global_org, save=True):
+        if not global_org and not self.project_local_dir:
+            return
```
davisagli, line 262: When does this happen? Should it raise an exception?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -197,34 +257,69 @@ def _load_org_files(self, dirname: str, constructor=None):
...
         if global_org:
             filename = Path(f"{self.global_config_dir}/{name}.org")
         elif self.project_local_dir is None:
             return
         else:
             filename = Path(f"{self.project_local_dir}/{name}.org")
-        with open(filename, "wb") as f_org:
-            f_org.write(encrypted)
+
+        fd = os.open(filename, os.O_WRONLY | os.O_CREAT, SERVICE_ORG_FILE_MODE)
```
davisagli, line 300: https://docs.python.org/3/library/os.html#os.open says that we also need O_BINARY to open the file in binary mode on Windows

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -197,34 +257,69 @@ def _load_org_files(self, dirname: str, constructor=None):
...
-            extra=[name, self],
-            context=f"org config ({name})",
-        )
-        if self.orgs[name].global_org:
-            org.global_org = True
+        try:
+            config = self.orgs[name].data
+            global_org = self.orgs[name].global_org
+        except KeyError:
+            raise OrgNotFound(f"Org with name '{name}' does not exist.")
```
davisagli, line 309: Good, this is a much less confusing error message that the current one.

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -423,36 +515,63 @@ def _rename_alias_in_default_service_file(
...
             self._migrate_services()
+
         self._load_service_files()
 
-    def _set_service(self, service_type, alias, service_config):
+    def _set_service(
+        self, service_type, alias, service_config, save=True, config_encrypted=False
+    ):
+        if "services" not in self.config:
+            self.config["services"] = {}
```
davisagli, line 531: Should we do this once in `__init__`?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -423,36 +515,63 @@ def _rename_alias_in_default_service_file(
...
+    def _save_encrypted_service(self, service_type, alias, encrypted):
+        """Write out the encrypted service to disk."""
         service_path = Path(
             f"{self.global_config_dir}/services/{service_type}/{alias}.service"
         )
         if not service_path.parent.is_dir():
             service_path.parent.mkdir()
-        with open(service_path, "wb") as f:
+
+        fd = os.open(service_path, os.O_WRONLY | os.O_CREAT, SERVICE_ORG_FILE_MODE)
```
davisagli, line 562: and O_BINARY for Windows?

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -423,36 +515,63 @@ def _rename_alias_in_default_service_file(
...
             f.write(encrypted)
 
     def _get_service(self, service_type, alias):
         if service_type == "connected_app" and alias == DEFAULT_CONNECTED_APP_NAME:
             # CumulusCI's default connected app is not encrypted, just return it
             return self.config["services"]["connected_app"][DEFAULT_CONNECTED_APP_NAME]
 
+        # if alias.startswith(self.env_service_alias_prefix):
+        # any environment based services aren't encrypted, just return it
+        # return self.services[service_type][alias]
```
davisagli, line 573: Remove comment if no longer needed

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
```diff
@@ -23,6 +29,15 @@
...
 from cumulusci.core.keychain.encrypted_file_project_keychain import GlobalOrg
+from cumulusci.core.keychain.encrypted_file_project_keychain import (
+    SERVICE_ORG_FILE_MODE,
+)
+from cumulusci.core.tests.utils import EnvironmentVarGuard
+
+
+@pytest.fixture()
+def env():
+    yield EnvironmentVarGuard()
```
davisagli, line 40: Maybe just use it directly without the fixture?

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
```diff
@@ -61,7 +76,33 @@ def test_set_and_get_org__global(self, keychain, org_config):
...
+        with open(filepath, "rb") as f:
+            assert pickle.load(f) == org_config.config
+
+    @mock.patch("cumulusci.core.keychain.encrypted_file_project_keychain.open")
+    def test_save_org_when_no_project_local_dir_present(
+        self, mock_open, keychain, org_config
+    ):
+        with mock.patch.object(EncryptedFileProjectKeychain, "project_local_dir", None):
+            keychain._save_org("alias", org_config, global_org=False)
+        assert mock_open.call_count == 0
```
davisagli, line 101: Is this the right assertion given that we're using os.open?

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
```diff
@@ -92,6 +133,10 @@ def test_remove_org__global_not_found(self, keychain):
         with pytest.raises(OrgNotFound):
             keychain.remove_org("test", True)
 
+    @pytest.mark.skipif(
+        sys.platform.startswith("win"),
+        reason="os.stat() returns a differing value on windows",
+    )
```
davisagli, line 139: Can we skip the specific assertion about file mode (or make what it asserts conditional based on platform) rather than skipping the entire test?

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
```diff
@@ -188,17 +259,117 @@ def test_load_service_files(self, keychain):
...
+    def test_get_service__does_not_exist(self, keychain, service_config):
+        keychain.set_service("github", "alias", service_config)
+        error_message = "No service of type github exists with the name: does-not-exist"
+        with pytest.raises(ServiceNotConfigured, match=error_message):
+            keychain.get_service("github", "does-not-exist")
+
+    @pytest.mark.skipif(
+        sys.platform.startswith("win"),
+        reason="Windows has a different value returned by os.stat",
+    )
```
davisagli, line 334: Can we skip or adjust the assertion about file mode rather than skipping the entire test?

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
```diff
@@ -188,17 +259,117 @@ def test_load_service_files(self, keychain):
...
+
         assert default_github_service.config == {
             **service_config.config,
             "token": "test123",
         }
 
+    @pytest.mark.skipif(
+        not sys.platform.startswith("win"),
+        reason="Windows has a different value returned by os.stat",
+    )
```
davisagli, line 357: Ditto

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
davisagli: I think it's just that os.stat wraps a system call that does something different on Windows.

Thanks for checking the permissions. I'm fine accepting that and not testing the permissions on Windows if there isn't a straightforward way to do so, but I'd rather not skip the rest of the assertions in the same tests.

davisagli (commented):
Really great work, @Br4nd0R!

cumulusci/core/keychain/encrypted_file_project_keychain.py
```diff
@@ -23,9 +23,15 @@
 from cumulusci.core.keychain import BaseProjectKeychain
 from cumulusci.core.keychain.base_project_keychain import DEFAULT_CONNECTED_APP_NAME
 from cumulusci.core.utils import import_class
+from cumulusci.core.utils import import_global
 
 DEFAULT_SERVICES_FILENAME = "DEFAULT_SERVICES.json"
 
+# The file permissions that we want set on all
+# .org and .service files. Equivalent to -rw-------
+SERVICE_ORG_FILE_MODE = 0o600
```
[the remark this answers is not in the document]
Br4nd0R: 🐙

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: I would do that, but we need to hit line 149 and below if they key isn't present. Maybe I can refactor it somehow so that logic at the bottom doesn't need to be in `_decrypt_config()`.

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: Hmm.... not sure if this is even possible. You wouldn't be able to operate on a local org without a CumulusCI project being present, which would necessitate there being a project_local_dir 🤔

cumulusci/core/keychain/encrypted_file_project_keychain.py
Br4nd0R: It would probably be best.

Br4nd0R: Looks like BaseProjectKeychain.__init__() is already doing that for us.

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
Br4nd0R: We use `os.open` to get a file descriptor which we pass to the built-in `open()` function.

cumulusci/core/keychain/tests/test_encrypted_file_project_keychain.py
Br4nd0R: This test and the one above are the same thing, but one is for windows, and the other is not.

davisagli (approved):

PR #2677, opened 2021-06-14
disable prerelease testing

davisagli:
We're past the period where there is an active preview release, so let's not try to test using it.

# Critical Changes

# Changes

# Issues Closed


boakley (approved):
Looks good to me, though I wish there was a better way to do this than to comment out 50 lines of code :-\

Br4nd0R (approved):

PR #2687, opened 2021-06-22
Fix windows CI issue

prescod:


Br4nd0R (approved):

davisagli (changes requested):
I think we should add `pip install pip-tools` as the first command done by `make dev-install` -- as it stands, the instructions in CONTRIBUTING.rst are missing this step.

davisagli (approved):

PR #2692, opened 2021-06-25
Handle, and warn about, package versions specified as numbers

davidmreed:
# Critical Changes

# Changes

# Issues Closed
- We fixed an issue where CumulusCI did not correctly convert a package version specified as a number in YAML to a string. This now raises a warning.

### DIFF
--- a/cumulusci/tasks/salesforce/install_package_version.py
+++ b/cumulusci/tasks/salesforce/install_package_version.py
@@ -104,6 +104,12 @@ class InstallPackageVersion(BaseSalesforceApiTask):
                 self.project_config,
                 get_resolver_stack(self.project_config, "production"),
             )
+        elif isinstance(version, (float, int)):
+            self.logger.warning(
+                f"The `version` option is specified as a number ({version}). "
+                "Please specify as a quoted string to avoid ambiguous results."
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/salesforce/install_package_version.py
```diff
@@ -104,6 +104,12 @@ def _init_options(self, kwargs):
                 self.project_config,
                 get_resolver_stack(self.project_config, "production"),
             )
+        elif isinstance(version, (float, int)):
```
jstvz, line 107: I don't suppose it makes much difference whether we do it here or in `_init_options`, does it?

jstvz (approved):

cumulusci/tasks/salesforce/install_package_version.py
davidmreed: You mean in `_run_task()`? I don't think so. And it always freezes to a string already, so this only affects MetaCI and local runs.

jstvz: No, I was wrong. This _is_ `_init_options()`, just a bit... longer than average.

PR #2694, opened 2021-06-25
fix reporting of oauth errors

davisagli:
Response objects from `requests` evaluate to False in a boolean context if the status code is >= 400. (Who knew?!) So let's be sure to check if the response is None instead of whether it is falsy -- and update the test to use a real response object since the mock didn't behave the same.

# Critical Changes

# Changes

# Issues Closed
- Fixed a bug where OAuth errors were not reported in detail.


jstvz (approved):
TIL

PR #2696, opened 2021-06-25
Feature/robot keyword docs

boakley:
This tweaks the sphinx conf.py file so that Keywords.html can be linked to without downloading to the local machine.

With this change, it's possible to link to a specific library or keyword. For example, to jump to the Open Test Browser keyword you could add the following the end of the URL: `#Salesforce.robot.Open%20Test%20Browser`

I didn't spend much time tweaking robot.rst since Justin is in the process of rewriting it. I also chose not to modify the build script to generate Keywords.html when it generates the other documents since it's not something I planned for the current sprint. 

Here's an example. It probably won't stand the test of time but it works right now: [Open Test Browser](https://bryans-experiment.readthedocs.io/en/latest/Keywords.html#Salesforce.robot.Open%20Test%20Browser)

# Critical Changes

# Changes
* improved our robot documentation so that it's possible to link to keyword documentation instead of having to download it locally

# Issues Closed
* W-9074271


davisagli (commented):
There is another link to Keywords.html in this section which is still downloading: https://bryans-experiment.readthedocs.io/en/latest/robot.html#keyword-documentation

davisagli (approved):

PR #2712, opened 2021-07-06
Feature enhancements for generate_data_dictionary

davidmreed:
# Critical Changes

# Changes

- The `generate_data_dictionary` task now includes Custom Settings, Custom Metadata Types, and Platform Events.
- The `generate_data_dictionary` task now excludes any schema with visibility set to Protected. This behavior can be turned off (including protected schema) with the `include_protected_schema` option.
- The `generate_data_dictionary` task now parses object and field metadata anywhere in a Salesforce DX release other than in the `unpackaged/` directory tree.

# Issues Closed

- We fixed an issue preventing `generate_data_dictionary` from working with four-digit (1.0.0.0) 2GP version numbers.

### DIFF
--- a/cumulusci/tasks/datadictionary.py
+++ b/cumulusci/tasks/datadictionary.py
@@ -100,8 +100,7 @@ class GenerateDataDictionary(BaseGithubTask):
     - Version Picklist Values Last Changed
     - Version Help Text Last Changed
 
-    Both MDAPI and SFDX format releases are supported. However, only force-app/main/default
-    is processed for SFDX projects.
+    Both MDAPI and SFDX format releases are supported.
     """
 
     task_options = {
@@ -126,6 +125,10 @@ class GenerateDataDictionary(BaseGithubTask):
             "description": "Treat the current branch as containing prerelease schema, "
             "and included it as Prerelease in the data dictionary. NOTE: this option "
             "cannot be used with `additional_dependencies` or `include_dependencies`."
+        },
+        "include_protected_schema": {
+            "description": "Include Custom Objects, Custom Settings, and Custom Metadata "
+            "Types that are marked as Protected. Defaults to False."
         },
     }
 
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/datadictionary.py
```diff
@@ -339,35 +347,67 @@ def _should_process_object(
...
+
     def _process_sfdx_release(self, zip_file: ZipFile, version: PackageVersion):
         """Process an SFDX ZIP file for objects and fields"""
         for f in zip_file.namelist():
             path = PurePosixPath(f)
-            if f.startswith("force-app/main/default/objects"):
+            # Be flexible about processing directories in SFDX context.
+            # This may not be optimal if the repo contains multiple
+            # 2GP package subdirectories.
+            if "objects/" in f and not f.startswith("unpackaged/"):
```
jstvz, line 410: Since we know that `sfdx-project.json` is present we have the option of using `packageDirectories` directly, no?

cumulusci/tasks/datadictionary.py
```diff
@@ -339,35 +347,67 @@ def _should_process_object(
...
         The object itself may or may not be included.
 
-        We will track any field on a custom object or standard entity. We will not track
-        fields on Custom Settings, Platform Events, or Custom Metadata Types."""
-        return (sobject_name.endswith("__c") or sobject_name.count("__") == 0) and (
-            element.find("customSettingsType") is None if element else True
+        We will track any field on a custom object, Platform Event,
+        Custom Setting, Custom Metadata Type, or standard entity.
+        We will not track any object's fields that is protected
+        in this package version."""
```
jstvz, line 375:
```suggestion
        We will not track any object's fields that are protected
        in this package version."""
```

jstvz (dismissed):
LGTM. One query re: sfdx-project.json that can be left for a future revision.

cumulusci/tasks/datadictionary.py
davidmreed: Yes. I shied away from it because I'd have to make a call on whether to process only the default package directory, all of them, or provide options to specify, which was more complexity than I wanted to add at the moment. I think it's clearly the right direction to go, though.

jstvz (approved):

PR #2715, opened 2021-07-07
Match community names case-insensitively

jstvz:
This PR:
- Updates the community exists error message to match the new ExperienceBundle message.
- Matches community names case-insensitively.

Fixes an issue discovered by @Julian88Tex in SalesforceFoundation/SAL#2116.

# Critical Changes

# Changes

# Issues Closed
- Fixed an issue where the `create_community` task would throw an exception on an existing community despite `skip_existing` being true.


cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -119,5 +120,5 @@ def _poll_action(self):
 
     def _get_community(self):
         community_list = self.sf.restful("connect/communities")["communities"]
-        communities = {c["name"]: c for c in community_list}
-        return communities.get(self.options["name"])
+        communities = {c["name"].casefold(): c for c in community_list}
```
prescod, line 123: We use .lower() everywhere else. Is there anything special about this context?

jstvz: Nothing special here. I just stumbled on it in the docs and thought "huh, better caseless matching of unicode strings" and added it here.

jstvz: Following up, @prescod. Are you requesting a change here?

prescod (dismissed):
Sorry my comment was pending. I can approve but I'd rather not switch to casefold pervasively unless we are confident that its better (i.e. more like Salesforce).

jstvz:
> Sorry my comment was pending. 

Weird. It actually put the comments in created order, but out of chronological order as displayed.

> I can approve but I'd rather not switch to casefold pervasively unless we are confident that its better (i.e. more like Salesforce).

You're right. I looked it up and this particular resource is calling [`String.equalsIgnoreCase()`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#equalsIgnoreCase(java.lang.String)). 

Updated.

cumulusci/tasks/salesforce/CreateCommunity.py
```diff
@@ -119,5 +120,5 @@ def _poll_action(self):
 
     def _get_community(self):
         community_list = self.sf.restful("connect/communities")["communities"]
-        communities = {c["name"]: c for c in community_list}
-        return communities.get(self.options["name"])
+        communities = {c["name"].lower().upper(): c for c in community_list}
```
Br4nd0R, line 123: Was going to ask why this needed but then read the conversation thread. TIL.

Br4nd0R (approved):
Nice work. All looks good.

cumulusci/tasks/salesforce/CreateCommunity.py
jstvz: Yeah, I went all the way down the rabbit hole on this one.

PR #2720, opened 2021-07-07
Fix create package version dependency output

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue causing `release_2gp_beta` to fail to create a GitHub release with a dependency-parsing error.


jstvz (approved):
LGTM

PR #2730, opened 2021-07-13
Feature/bulk dataload results bug fix

sfdcale:
This PR contains the fix to how Bulk API batch response is parsed. Work Item for this is : W-8880039

If a Bulk API Job has multiple batches, it only parses the failed and processed count from last batch:

image


prescod:
@sfdcale Do you plan to submit this again later?

sfdcale:
@prescod  Created another one and it is #2733

PR #2733, opened 2021-07-13
Feature/bulk dataload results bug fix

sfdcale:
This PR contains the fix to how Bulk API batch response is parsed. Work Item for this is : W-8880039

If a Bulk API Job has multiple batches, it only parses the failed and processed count from first batch:

image


sfdcale:
To better understand what the defect here is, follow these steps:
1.  Rollback the changes I made to the file `cumulusci/tasks/bulkdata/step.py`
2. Run tests from file `cumulusci/tasks/bulkdata/tests/test_step.py`

and the tests will file as current implementation only parses count from first batch.

PR #2738, opened 2021-07-19
Default to not creating unlocked packages for unmanaged deps

davidmreed:
# Critical Changes

- The `create_package_version` task no longer creates Unlocked Packages from the `unpackaged/pre` and `unpackaged/post` directories of dependencies, or local `unpackaged/pre` directories by default. This behavior is now opt-in via the `create_unlocked_dependency_packages` option, which defaults to False. Projects using the old default behavior must explicitly set this option. We believe the new behavior is a more sane default for most 2GP projects.

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -1053,6 +1053,7 @@ flows:
           version_base: latest_github_release
           version_type: minor
           force_upload: True
+          create_unlocked_dependency_packages: False
       2:
         task: github_release
         options:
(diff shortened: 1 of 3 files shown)


Br4nd0R (approved):
This looks great, and thanks for the quick chat to clarify the end goal.

davidmreed:
Closing in favor of https://github.com/SFDO-Tooling/CumulusCI/pull/2741.

PR #2740, opened 2021-07-19
Change default and pinned API versions to 52.0

davidmreed:
# Critical Changes

# Changes

- We updated the default API version to 52.0.

# Issues Closed


Br4nd0R (approved):
This seems tedious. I think we should make a story to create a constant somewhere that all other files reference from.

jofsky (approved):

PR #2741, opened 2021-07-19
Support objectSettings in create_package_version and change default behavior to not create unlocked packages

davidmreed:
# Critical Changes

- The `create_package_version` task no longer creates Unlocked Packages from the `unpackaged/pre` and `unpackaged/post` directories of dependencies, or local `unpackaged/pre` directories by default. This behavior is now opt-in via the `create_unlocked_dependency_packages` option, which defaults to False. Projects using the old default behavior must explicitly set this option. We believe the new behavior is a more sane default for most 2GP projects.

# Changes

- The `create_package_version` task now supports `objectSettings` in the org definition file.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/salesforce/org_settings.py
+++ b/cumulusci/tasks/salesforce/org_settings.py
@@ -1,7 +1,6 @@
 import contextlib
 from cumulusci.tasks.metadata.package import PackageXmlGenerator
 import json
-import os
 import pathlib
 import textwrap
 from typing import Optional
@@ -74,7 +73,7 @@ def build_settings_package(
 ):
     with temporary_dir() as path:
         if settings:
-            os.mkdir("settings")
+            pathlib.Path("settings").mkdir()
             for section, section_settings in settings.items():
                 settings_name = capitalize(section)
                 if section == "orgPreferenceSettings":
@@ -98,7 +97,7 @@ def build_settings_package(
                     )
 
         if object_settings:
-            os.mkdir("objects")
+            pathlib.Path("objects").mkdir()
             for obj_lower_name, this_obj_settings in object_settings.items():
                 object_name = capitalize(obj_lower_name)
                 file_content = _get_object_file(object_name, this_obj_settings)
--- a/cumulusci/tasks/salesforce/org_settings.py
+++ b/cumulusci/tasks/salesforce/org_settings.py
@@ -1,7 +1,10 @@
 import contextlib
+from cumulusci.tasks.metadata.package import PackageXmlGenerator
 import json
 import os
+import pathlib
 import textwrap
+from typing import Optional
 
 from cumulusci.tasks.salesforce import Deploy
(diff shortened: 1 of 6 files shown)


davidmreed:
Depends on https://github.com/SFDO-Tooling/CumulusCI/pull/2738

davidmreed:
E2E testing is in progress

davidmreed:
E2E testing successful.

cumulusci/tasks/salesforce/org_settings.py
```diff
@@ -74,33 +69,106 @@ def capitalize(s):
...
-            settings_name = capitalize(section)
-            if section == "orgPreferenceSettings":
-                values = "    " + "\n    ".join(
-                    line
-                    for line in "\n".join(
-                        ORGPREF.format(name=capitalize(k), value=v)
-                        for k, v in section_settings.items()
-                    ).splitlines()
+        if settings:
+            os.mkdir("settings")
```
Br4nd0R, line 77: I know we typically try to use pathlib over os module for stuff like this. With an `from pathlib import Path` above, this would be:
```suggestion
            Path.mkdir("settings")
```

cumulusci/tasks/salesforce/org_settings.py
```diff
@@ -74,33 +69,106 @@ def capitalize(s):
...
-            )
-            with open(settings_file, "w") as f:
-                f.write(SETTINGS_XML.format(settingsName=settings_name, values=values))
+                with open(settings_file, "w", encoding="utf-8") as f:
+                    f.write(
+                        SETTINGS_XML.format(settingsName=settings_name, values=values)
+                    )
+
+        if object_settings:
+            os.mkdir("objects")
```
Br4nd0R, line 101: Same `pathlib` comment as above.

cumulusci/tasks/salesforce/org_settings.py
```diff
@@ -74,33 +68,106 @@ def capitalize(s):
...
-        for section, section_settings in settings.items():
-            settings_name = capitalize(section)
-            if section == "orgPreferenceSettings":
-                values = "    " + "\n    ".join(
-                    line
-                    for line in "\n".join(
-                        ORGPREF.format(name=capitalize(k), value=v)
-                        for k, v in section_settings.items()
-                    ).splitlines()
+        if settings:
```
Br4nd0R, line 75: Not necessary at all, but it seems like this method could be refactored to call two methods internally with prototypes: `_generate_settings_file(settings: Optional[dict])` and `_generate_object_settings_file(object_settings: Optional[dict])`. Then after those methods, you would have what's currently on lines 111-115.

The first line of each of those methods would be to check for the presence of the argument, and if the value is `None`, then they would just return. This would also help mitigate the multiple levels of indentation occurring.

Again, not necessary at all, I'm just partial to smaller functions myself.

Br4nd0R (approved):

PR #2742, opened 2021-07-19
Built-in connected_app should be the default if no others are present

### LINKED ISSUE #2737: If no service for a connected_app is defined in environment variable an error occurs
**Describe the bug**
If there is no service for a connected_app defined via environment variables and get_service('connected_app') is called, on EnvironmentProjectKeychain an error occurs.

**Context**
- Best guess is this is the PR which causes it: https://github.com/SFDO-Tooling/CumulusCI/pull/2664
- Occurs with sfdx_url authentication method

**Example**
 https://github.com/Salesforce-org-Impact-Labs/01HousingandHomelessness/runs/3090372340?check_suite_focus=true

**Additional information**
In the _load_default_services() method on the EnvironmentProjectKeychain, check for service_type of `connected_app` and then set it to the DEFAULT_CONNECTED_APP_NAME.

Br4nd0R:
# Issues Closed
* Fixed an issue where the the built-in connected app was not accessible when running CumulusCI in a headless environment. (Fixes #2737)

### DIFF
--- a/cumulusci/core/keychain/environment_project_keychain.py
+++ b/cumulusci/core/keychain/environment_project_keychain.py
@@ -59,7 +59,7 @@ class EnvironmentProjectKeychain(BaseProjectKeychain):
         # then set the built-in connected_app as the default.
         # The built-in is always loaded, so 2 or more would indicate
         # the presence of a user-provided connected app.
-        if len(self.config["services"]["connected_app"].keys()) == 1:
+        if len(self.config["services"]["connected_app"]) == 1:
             self._default_services["connected_app"] = DEFAULT_CONNECTED_APP_NAME
 
     def cleanup_org_cache_dirs(self):
--- a/cumulusci/core/keychain/environment_project_keychain.py
+++ b/cumulusci/core/keychain/environment_project_keychain.py
@@ -56,7 +56,7 @@ class EnvironmentProjectKeychain(BaseProjectKeychain):
             self._default_services[service_type] = "env"
 
         # If there are no connected_app services loaded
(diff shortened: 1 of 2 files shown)


cumulusci/core/keychain/environment_project_keychain.py
```diff
@@ -54,5 +55,12 @@ def _load_default_services(self):
         for service_type in self.services:
             self._default_services[service_type] = "env"
 
+        # If there are no connected_app services loaded
+        # then set the built-in connecteid_app as the default.
+        # The built-in is always loaded, so 2 or more would indicate
+        # the presence of a user-provided connected app.
+        if len(self.config["services"]["connected_app"].keys()) == 1:
```
davidmreed, line 62: `len()` can take a `dict` directly, without calling `keys()`.

cumulusci/core/keychain/environment_project_keychain.py
```diff
@@ -54,5 +55,12 @@ def _load_default_services(self):
         for service_type in self.services:
             self._default_services[service_type] = "env"
 
+        # If there are no connected_app services loaded
+        # then set the built-in connecteid_app as the default.
```
davidmreed, line 59:
```suggestion
        # then set the built-in connected_app as the default.
```

cumulusci/core/keychain/environment_project_keychain.py
Br4nd0R: TIL

davidmreed (approved):

PR #2751, opened 2021-07-20
Update PACKAGE_XML_BEFORE__PROFILES

Br4nd0R:
The `test_run_task()` function overwrites the `cumulusci/files/admin_profile.xml` file. This should likely be changed in the future, but the quick fix is to update the `PACKAGE_XML_BEFORE__PROFILES` to match the current contents of that file in the repository.


jstvz (approved):
I am ashamed that I walked past this several times without fixing it. 😔

PR #2754, opened 2021-07-20
Add GitHub response handlers

jstvz:
This PR:

- Updates `validate_service()` to:
  - validate the user's token by calling `/user`, rather than `/rate_limit`, which doesn't require authorization.
  - bypass automatic retries for `401` responses, which meant that invalid tokens could take several minutes to fail.
  - Adds to `service info`:
    - a list of organizations for the authenticated user 
    - a list of organizations for the authenticated user for which SSO is disabled
    - adds a list of scopes authorized for the token
- Utility functions to parse exceptions generated by the github API and return a user-friendly message for the following error conditions:
  - The user's token is valid, but SSO has not been enabled for the token.
    - Generates an error message containing the missing organization IDs (`FIXME`) 
    - Opens a browser to authorize the token.
  - The user's token is not valid (`too many 401 responses`)
    - Generates an error message instead of a traceback
  - The user's token is missing necessary scopes (e.g. `gist`)
    - Generates an error message containing the specific scopes required for the action
- Adds the `catch_common_github_auth_errors` decorator  to catch and process any exception meeting the criteria above:
   ```python
    @catch_common_github_auth_errors
    def get_repo_from_url(self, url):
        owner, name = split_repo_url(url)
        return self.get_github_api(owner, name).repository(owner, name)
   ```
---
# Critical Changes

# Changes
- We added additional information about github tokens to `cci service info github` to help troubleshoot authentication errors.
- If you're a member of any GitHub organizations that enforce Single Sign On (SSO), CumulusCI will warn you if your access token has not been enabled for SSO for any such orgs and open a browser to do so.
# Issues Closed
- We fixed an issue where `cci service connect github` validated invalid access tokens.


cumulusci/cli/service.py
```diff
@@ -157,7 +157,7 @@ def callback(*args, **kwargs):
             validator_path = service_config.get("validator")
             if validator_path:
                 validator = import_global(validator_path)
-                validator(serv_conf)
+                serv_conf = validator(serv_conf)
```
davidmreed, line 160: CHECK: Is the return value semantic of other validator methods compatible with this?

cumulusci/core/github.py
```diff
@@ -268,3 +303,131 @@ def get_version_id_from_tag(repo: Repository, tag_name: str) -> str:
...
+
+    request_url = urlparse(exc.response.url)
+    if not accepted_scopes and request_url.path == "/gists":
+        accepted_scopes = {"gist"}
+
+    missing_scopes = accepted_scopes.difference(token_scopes)
+    if missing_scopes:
+        user_warning = f"Your token may be missing the following scopes: {', '.join(missing_scopes)}\n"
+        # This assumes we're not on enterprise and 'api.github.com' == request_url.hostname
+        user_warning += "Visit https://github.com/settings/tokens to add them."
```
davidmreed, line 370: Should we specify the clickpath instead of the URL to avoid this caveat?

cumulusci/core/github.py
```diff
@@ -268,3 +303,131 @@ def get_version_id_from_tag(repo: Repository, tag_name: str) -> str:
...
+    if exc.response.status_code != 403 or "X-Github-Sso" not in headers:
+        return user_warning
+
+    sso_header = str(headers["X-Github-Sso"] or "")
+    if sso_header.startswith("required; url="):
+        # In this case the message from github is good enough, but we can help
+        # the user by opening a browser to authorize the token.
+        auth_url = sso_header.split("url=", maxsplit=1)[1]
+        user_warning = f"{exc.message}\n{auth_url}"
+        webbrowser.open(auth_url)
```
davidmreed, line 395: What happens if we're running headless?

cumulusci/core/github.py
```diff
@@ -268,3 +303,131 @@ def get_version_id_from_tag(repo: Repository, tag_name: str) -> str:
...
+        # In cases where we don't have complete results we get the
+        # partal-results header, so return the organization IDs. This may or
+        # may not be useful without help from us to lookup the org IDs.
+        unauthorized_org_ids = get_sso_disabled_orgs(exc.response)
+        user_warning = f"Results may be incomplete. You have not granted your Personal Access token access to the following organizations: {unauthorized_org_ids}"
+
+    return user_warning
+
+
+def get_sso_disabled_orgs(response: Response) -> list:
```
davidmreed, line 406: Isn't this type hint syntax limited to Python 3.9+?

cumulusci/core/github.py
```diff
@@ -268,3 +303,131 @@ def get_version_id_from_tag(repo: Repository, tag_name: str) -> str:
...
+    sso_header = response.headers.get("X-Github-Sso")
+    partial_results_prefix = "partial-results; organizations="
+
+    if sso_header and partial_results_prefix in sso_header:
+        disabled_orgs = sso_header[len(partial_results_prefix) :].split(",")
+
+    return disabled_orgs
+
+
+def get_oauth_scopes(response: Response) -> set:
```
davidmreed, line 421: Ditto the above.

cumulusci/core/tests/test_github.py
```diff
@@ -369,3 +373,55 @@ def test_get_version_id_from_tag__dependency_error(self, repo):
...
+
+        resp.status_code = 403
+        exc = ForbiddenError(resp)
+        assert "" == check_github_sso_auth(exc)
+
+    @mock.patch("webbrowser.open")
+    def test_check_github_sso_unauthorized_token(self, browser_open):
+        resp = Response()
+        resp.status_code = 403
+        auth_url = "https://github.com/orgs/foo/sso?authoriziation_request=longhash"
```
davidmreed, line 411:
```suggestion
        auth_url = "https://github.com/orgs/foo/sso?authorization_request=longhash"
```

cumulusci/core/github.py
Br4nd0R: What's the specific syntax that is in question here, the function argument type and return value? We have this in many other places in the code base, so I'm wondering what is different about this?

davidmreed: Lowercase `list` - from [PEP 585](https://www.python.org/dev/peps/pep-0585/). It's supported with a future annotation, though.

Br4nd0R: Interesting, as @davisagli indicated to me that I could use the lowercase `dict` type [here](https://github.com/SFDO-Tooling/CumulusCI/pull/2676#discussion_r662549413).

jstvz: PEP 585 is about using built-ins as a generic (e.g. `list[str]`). Using a built-in as a hint (`-> list:`) is supported in 3.5+. See: [PEP 484](https://www.python.org/dev/peps/pep-0484/#acceptable-type-hints)

cumulusci/cli/service.py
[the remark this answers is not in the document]
jstvz: Good catch. This isn't actually safe since any validator returning `None` would clobber `serv_conf`. I'll add a conditional here.

Now, we don't actually have a `Service` class with a public API, so I'm uncertain whether it is worth calling out as a Critical Change... but no harm done by doing so.

I was tempted to add such a class here, but decided against it to limit the scope of the change. Given our experience with SFDX I'm coming around to a bolder approach.

cumulusci/core/github.py
jstvz: tl;dr: returns `False`

---
Google fail, had to do it the hard way.

Image is `python:3.8-slim-buster`:

```console
❯ docker run -it empty-python
Python 3.8.11 (default, Jul 22 2021, 15:41:12)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import webbrowser
>>> webbrowser.open('google.com')
False
>>>
❯ python
Python 3.9.5 (default, Jun  7 2021, 17:45:25)
[Clang 12.0.0 (clang-1200.0.32.29)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import webbrowser
>>> webbrowser.open('google.com')
True
```

davidmreed (approved):

PR #2755, opened 2021-07-20
Week 29 Updates

Br4nd0R:
Need to test Python 3.4 with latest update to `idna==3.2`.
States that it deprecated support for python 3.4: https://github.com/kjd/idna/releases/tag/v3.2

Edit: We are supporting >=3.6 so we can let this through.


codexmarcel (approved):
LGTM pending checks

PR #2769, opened 2021-07-29
Use ISO-8601 formatted start_time option

jstvz:
# Critical Changes

# Changes
- We've updated the Push Upgrade API (`push_list`, `push_sandbox`, etc) task option `start_time` to accept ISO-8601 formatted datetimes.
# Issues Closed


jofsky (approved):
I like this @jstvz. Are you ok with adding this dependency though to CCI @jstvz  @davisagli?

davisagli (approved):

PR #2771, opened 2021-07-29
Fix lint workflow

jstvz:
Previously, the `lint` job compared HEAD against origin/main on every push, but that didn't necessarily match the context; you could, for example, be in a child branch. Ultimately, it made the most sense for this to be triggered by the pull_request event instead.

This PR also changes the events triggering the feature test workflow to one of the following:

1. Manual runs:
    gh workflow run feature_test.yml
2. Pull requests
3. Commits to `main`

Failing workflow [here](https://github.com/SFDO-Tooling/CumulusCI/pull/2772/checks?check_run_id=3225142344).

---

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/.github/workflows/feature_test.yml
+++ b/.github/workflows/feature_test.yml
@@ -1,9 +1,11 @@
 name: Feature Test
 
 on:
+  workflow_dispatch:
+  pull_request:
+    types: [opened, synchronize, reopened] # Default
   push:
     branches:
-      - feature/**
       - main
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/feature_test.yml
```diff
@@ -34,8 +34,8 @@ jobs:
           pip install pre-commit
       - name: Run pre-commit on changed files
         run: |
-          git fetch --depth=1 origin main
-          pre-commit run --files $(git diff --diff-filter=d --name-only origin/main...HEAD)
+          git fetch origin $GITHUB_HEAD_REF:$GITHUB_HEAD_REF
```
davisagli, line 37: Does this do something different from `--depth=1` or is it just a different way of doing the same thing?

davisagli:
> Ultimately, it probably makes the most sense for this to be triggered by the pull_request event instead.

That makes sense to me.

.github/workflows/feature_test.yml
jstvz: Unfortunately, this also doesn't work because those environment variables aren't present unless the workflow is triggered by the `pull_request` event. If it were, `$GITHUB_BASE_REF` would be equivalent to `origin/main` on PRs against `main`, and `$GITHUB_HEAD_REF` matches `HEAD`.

davisagli: Shall we just switch the workflow to the pull_request event? I doubt we are frequently trying to check build status for a commit that doesn't have a PR yet, and if we want to it's easy to create a draft PR

jstvz: I think so. The `synchronize` activity on that event will trigger new runs after merging in the base branch. 

We also have the option of adding `workflow_dispatch` to allow manual runs as a backstop.

davisagli (approved):
Good stuff!

PR #2775, opened 2021-07-30
Prompt for defaulting new connected service if one already exists

Br4nd0R:
# Changes
* We now prompt users if they would like to set a newly connected service as the new default, if a default service of the same type already exists.


cumulusci/cli/service.py
```diff
@@ -142,6 +142,14 @@ def callback(*args, **kwargs):
                     abort=True,
                 )
 
+            set_as_default = False
```
prescod, line 145:
```suggestion
```

cumulusci/cli/service.py
```diff
@@ -142,6 +142,14 @@ def callback(*args, **kwargs):
...
+            prompt_to_default_service = f"A default service already exists for service type {service_type}. Would you like to set this service as the new default?"
+            default_service_exists = runtime.keychain.get_default_service_name(
+                service_type
+            )
+            if default_service_exists and click.confirm(prompt_to_default_service):
+                set_as_default = True
```
prescod, line 151:
```suggestion
            set_as_default = default_service_exists and click.confirm(prompt_to_default_service)
```

cumulusci/cli/tests/test_service.py
```diff
@@ -156,6 +158,34 @@ def test_service_connect__alias_already_exists(confirm):
     confirm.assert_called_once()
 
 
+@mock.patch("click.confirm")
+def test_service_connect__set_new_service_as_default(confirm):
+    confirm.side_effect = "y"
```
prescod, line 163: Should we test the "n" path as well?

prescod (changes requested):

cumulusci/cli/tests/test_service.py
Br4nd0R: This prompt actually led me to realize that I was mocking incorrectly, such that the return values differed from what click.confirm actually returns (a boolean). Thanks Paul, all fixed now!

cumulusci/cli/service.py
```diff
@@ -142,6 +142,16 @@ def callback(*args, **kwargs):
                     abort=True,
                 )
 
+            prompt_to_default_service = f"A default service already exists for service type {service_type}. Would you like to set this service as the new default?"
+            default_service_exists = (
+                True
+                if runtime.keychain.get_default_service_name(service_type) is not None
+                else False
```
prescod, line 149: Debatable, but i very mildly prefer this:

```suggestion
            default_service_exists = runtime.keychain.get_default_service_name(service_type) is not None
```

But that's totally debatable.

prescod (approved):
One extremely minor suggestion

davisagli (approved):

PR #2783, opened 2021-08-04
New testing infra

prescod:
Upgraded testing infrastructure for CumulusCI.

During review, start with the docs:

https://github.com/SFDO-Tooling/CumulusCI/blob/47a81abdd4e9d5c0599c8308225b378ec7e73857/CONTRIBUTING.rst#org-reliant-automated-tests

Then review how the features are used in CI:

https://github.com/SFDO-Tooling/CumulusCI/compare/feature/vcr-upgrades?expand=1#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R70


.github/workflows/slow_integration_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+  workflow_dispatch:
+
+  # temp for testing. Delete before merging.
+  push:
+    branches:
+      - feature/** # delete this once it works
+
+env:
+  COVERALLS_PARALLEL: true
+  COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
```
davisagli, line 16: These aren't needed since this workflow isn't reporting to coveralls.

CONTRIBUTING.rst
```diff
@@ -147,48 +147,81 @@ Finally, post the release notes to our usual channels:
...
 Do not commit the files ("large_cassettes/\*.yml") to the repository.
 
+Some tests generate even more network traffic data and it isn't practical 
+to use VCR at all. Still, we'd like to run them when we run all of the
+other org-reliant tests with --org. Mark them with ``@pytest.mark.needs_org()``
+and they will run with the VCR tests.
+
+Some tests are so slow that you only want to run them on an opt-in basis.
+Mark these tests with ``@pytest.mark.slow()`` and run them with
+``pytest --run-slow`` or ``pytest --run-slow --orgname <orgname>``.
```
davisagli, line 189: It looks like it's actually named --run-slow-tests?

CONTRIBUTING.rst
```diff
@@ -147,48 +147,81 @@ Finally, post the release notes to our usual channels:
...
+~~~~~~~~~~~~~~~~~~~~~~~~~
+All features should have integration tests which work against
+real orgs or APIs.
+
+You will need to use some the following fixtures in your tests. Search
+the repo to see examples where they are used in context, or to see
+their definitions:
+
+* gh_api - get a fake github API
+* restore_cwd - ensure that a test finishes with the
```
davisagli, line 201: This one has autouse=True so shouldn't be necessary to use it explicitly

CONTRIBUTING.rst
```diff
@@ -147,48 +147,81 @@ Finally, post the release notes to our usual channels:
...
+* project_config - Get the project config for the current working directory
+* org_config - Get the project config for the current working directory
+* create_task - Get a task _factory_ which can be used to construct task instances.
+
+Decorators for tests:
+
+ * pytest.mark.slow(): a slow test that should only be executed when requested with --run-slow-tests
+ * pytest.mark.large_vcr(): a network-based test that generates VCR cassettes too large for version control. Use --org to generate them locally.
+ * pytest.mark.needs_org(): a test that needs an org (or at least access to the network) but should not attempt to store VCR cassettes
+ * org_shape('qa', 'qa_org'): - switch the curren org to an org created with org template "qa" after running flow "qa_org".
```
davisagli, line 221:
```suggestion
 * org_shape('qa', 'qa_org'): - switch the current org to an org created with org template "qa" after running flow "qa_org".
```

Makefile
```diff
@@ -67,6 +67,17 @@ coverage: ## check code coverage quickly with the default Python
...
 
+vcr: # remake VCR cassettes and run other integration tests
+	cci org scratch qa pytest
+	cci org scratch_delete pytest
+	find . -name \Test*.yaml | xargs rm
+	pytest --org qa --run-slow-tests -rs --replace-vcrs
+
+slow_tests: vcr # remake VCR cassettes and run other integration tests
+	cci org scratch_delete pytest
+	pytest integration_tests/ --org pytest -rs
```
davisagli, line 78: This won't work if the `pytest` org hasn't been defined yet. Maybe it should be predefined in CumulusCI's cumulusci.yml instead of via `cci org scratch`?

cumulusci.yml
```diff
@@ -35,6 +35,11 @@ tasks:
               - cumulusci/robotframework/pageobjects/BasePageObjects.py
 
 flows:
+    qa_org:
+        steps:
+            2:
+                flow: None
```
davisagli, line 41: Do we need this? CumulusCI doesn't have any tests which depend on things set up by the qa_org flow, does it?

cumulusci/conftest.py
```diff
@@ -107,3 +107,14 @@ def open_db():
...
+
+@pytest.fixture()
+def delete_data_from_org(create_task):
+    def delete_data_from_org(object_names):
+        from cumulusci.tasks.bulkdata.delete import DeleteData
+
+        t = create_task(DeleteData, {"objects": object_names})
+        t()
+
+    return delete_data_from_org
```
davisagli, line 120: This seems like a really weird way to do this. It's possible to get the project config from the org config (org_config.project_config) so it should be pretty straightforward to write a function which takes an org_config and object_names as parameters but isn't tied to the particular way that orgs get set up in the test infrastructure.

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
```diff
@@ -74,6 +75,25 @@ def temporary_file_path(filename):
         yield path
 
 
+@pytest.fixture()
+def ensure_accounts(create_task, run_code_without_recording, sf):
+    """Delete all accounts and create a certain number of new ones"""
```
davisagli, line 80: So if you accidentally run this against the wrong org, it's destructive. Should we have some check to confirm that the tests can only run against scratch orgs?

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
```diff
@@ -1,27 +1,76 @@
...
 from cumulusci.core.config import TaskConfig
+from cumulusci.core.exceptions import OrgNotFound
 from cumulusci.salesforce_api.utils import get_simple_salesforce_connection
 from cumulusci.tests.util import unmock_env
 
+from .pytest_sf_vcr import READING, RECORDING, set_mode
+
 
 def pytest_addoption(parser, pluginmanager):
+    """Pytest magic method: add --org and --run-slow-tests features"""
```
davisagli, line 21: and --replace-vcrs

cumulusci/tests/pytest_plugins/pytest_sf_vcr.py
```diff
@@ -7,36 +7,117 @@
...
-        decoded = re.sub(
-            r"<sessionId>.*</sessionId>",
-            "<sessionId>**Elided**</sessionId>",
-            decoded,
-        )
-    decoded = re.sub(r"001[\w\d]{15,18}", "001ACCOUNTID", decoded)
+    decoded = _cleanup(decoded)
+    # if "GeocodeAccuracy" in decoded:
+    #     breakpoint()
+
```
davisagli, line 23:
```suggestion
```

cumulusci/tests/pytest_plugins/pytest_sf_vcr.py
```diff
@@ -7,36 +7,117 @@
...
+        r"202\d-\d\d-\d\dT\d\d:\d\d\:\d\d.\d\d\d\+0000",
+        "2021-01-01T01:02:01.000+0000",
+    ),
+    (r'"005[\w\d]{12,15}"', '"0050xUSERID0000000"'),
+    (r'"InstanceName" : "[A-Z]{2,4}\d{1,4}",', '"InstanceName" : "CS420",'),
+    (r"<id>0.*<\/id>", lambda m: "<id>0ANID</id>"),
+    (
+        r"<asyncProcessId>0.*<\/asyncProcessId>",
+        lambda m: "<asyncProcessId>0ANAPPID</asyncProcessId>",
+    ),
```
davisagli, line 47: What are the lambdas for?

cumulusci/tests/pytest_plugins/pytest_sf_vcr.py
```diff
@@ -7,36 +7,117 @@
...
+    (r"\d\d\d\d-\d\d-\d\dT\d\d:\d\d:\d\d.\d\d\dZ", "2021-01-03T01:11:11.420Z"),
+    (r"/User/005[\w\d]{12,15}", "/User/0050xUSERID0000000"),
+    (r"<lastModifiedById>005[\w\d]{12,15}", "<lastModifiedById>005USERID"),
```
davisagli, line 50: I don't like replacing Account and User ids like this; it means we can't distinguish between multiple accounts or users

integration_tests/pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
 testpaths = tests
-addopts =  --typeguard-packages=cumulusci -p cumulusci.tests.pytest_plugins.pytest_sf_orgconnect
```
davisagli, line 3: No more typeguard?

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
```diff
@@ -1,27 +1,76 @@
...
+    classify_and_modify_test(item, marker_names)
+
+
+def sf_pytest_cli_orgname(request):
+    """Helper to see what org the user has ased for"""
     return request.config.getoption("--org")
 
 
+@pytest.fixture(scope="session")
+def user_requested_network_access(request):
```
davidmreed, line 63: The "network access" terminology I find confusing. There doesn't appear to be any distinction between network access and access to an org. Is that accurate?

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
```diff
@@ -35,28 +84,37 @@ def project_config(runtime):
...
+
+
+@pytest.fixture(scope="function")
+def org_config(request, current_org_shape, cli_org_config, fallback_org_config):
+    """Get an org config with an active access token.
+
+    Specify the org name using the --org option when running pytest.
+    Or else it will use a dummy org.
+    """
+    org_config = current_org_shape.org_config or cli_org_config
```
davidmreed, line 107: Unless I'm misunderstanding, this appears to be opposite to the docstring: we use the current org shape if already configured, otherwise the org set on the command line, and lastly a fallback org.

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
```diff
@@ -85,36 +143,123 @@ def create_task(task_class, options=None, project_config=None, org_config=None):
...
+
+@pytest.fixture(
+    scope="function",
+)
+def run_code_without_recording(request, vcr, user_requested_network_access):
+    def really_run_code_without_recording(func):
+        if user_requested_network_access:
+            # Run the setup code, but don't record it
+            with set_mode(RECORDING, False), set_mode(READING, False):
+                return func()
```
davidmreed, line 185: Silently doing nothing when there's no org present seems surprising. Where would this be used such that it could run in a test that isn't being skipped with no org?

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
```diff
@@ -85,36 +143,123 @@ def create_task(task_class, options=None, project_config=None, org_config=None):
...
+        with change_org_shape(current_org_shape, config_name, flow_name) as org_config:
+            yield org_config
+    else:
+        yield None
+
+
+@contextmanager
+def change_org_shape(current_org_shape, config_name: str, flow_name: str):
+    # I don't love that we're using the user's real keychain
+    # but I get weird popups when I use an empty keychain.
```
davidmreed, line 232: Do we leak the org after the test suite runs? I don't see anything that would remove it.

cumulusci/tests/pytest_plugins/pytest_sf_vcr.py
```diff
@@ -7,36 +7,117 @@
...
+    (r'"00D[\w\d]{12,15}"', '"00D0xORGID00000000"'),
+    (r".com//", r".com/"),
+    (r"ersion>\d\d.0<", r"ersion>vxx.0<"),
+    (r"<sessionId>.*</sessionId>", r"<sessionId>**Elided**</sessionId>"),
+    (r"001[\w\d]{12,15}", "0010xACCOUNTID0000"),
+    (r"//.*.my.salesforce.com", "//orgname.my.salesforce.com"),
+    (r"//.*\d+.*.salesforce.com/", "//orgname.my.salesforce.com/"),
+    (
+        r"202\d-\d\d-\d\dT\d\d:\d\d\:\d\d.\d\d\d\+0000",
+        "2021-01-01T01:02:01.000+0000",
```
davidmreed, line 39: Same concern as with Ids. Some tests may be reliant upon dates actually being different.

cumulusci/tests/pytest_plugins/pytest_sf_vcr.py
```diff
@@ -7,36 +7,117 @@
...
+    ),
+    (r'"005[\w\d]{12,15}"', '"0050xUSERID0000000"'),
+    (r'"InstanceName" : "[A-Z]{2,4}\d{1,4}",', '"InstanceName" : "CS420",'),
+    (r"<id>0.*<\/id>", lambda m: "<id>0ANID</id>"),
```
davidmreed, line 43: I'm not sure what this replaces.

cumulusci/tests/pytest_plugins/pytest_sf_vcr.py
```diff
@@ -7,36 +7,117 @@
...
+        recording_mode = True
+        reading_mode = "irrelevant"  # reading from missing file
+    elif user_requested_network_access and vcr_cassette_path.exists():
+        # user wants to keep existing cassette, so disable VCR usage entirely, like:
+        # https://github.com/ktosiek/pytest-vcr/blob/08482cf0724697c14b63ad17752a0f13f7670add/pytest_vcr.py#L56
+        recording_mode = False
+        reading_mode = False
+    elif user_requested_network_access:
+        recording_mode = True
+        reading_mode = "irrelevant"  # file doesn't exist, so reading missing file
```
davidmreed, line 101: I don't love that these values can be either arbitrary strings or Booleans. How about enums?

cumulusci/tests/test_integration_infrastructure.py
```diff
@@ -3,25 +3,119 @@
 import pytest
 
 from cumulusci.tasks.preflight.packages import GetInstalledPackages
+from cumulusci.tasks.salesforce import SOQLQuery
+
+pytestmark = pytest.mark.random_order(disabled=True)
```
davidmreed, line 8: After reading, I see why these tests are order-dependent, but a note might be helpful.

davidmreed (changes requested):

cumulusci.yml
prescod: Well I want to test features that depend on config-ing an org and the org-configuring flows crash on main due to trying to deploy non-existent metadata.

prescod: I could make a new flow, but then the examples will be less "realistic".

davisagli: The deploy_unmanaged flow is supposed to silently do nothing if there's not actually a package directory, so this feels like the test may actually be showing a bug that you're covering up here.

I have mixed feelings about adding this when we don't actually have tests that need it. On the one hand, we're more likely to add tests like that if the infrastructure already exists. On the other hand, it's hard to be confident that we're building the right infrastructure without a concrete use case.

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
davisagli: @prescod this comment was buried by github and I think it's important; did you see it?

cumulusci/conftest.py
davisagli: @prescod I suppose we both share the goal of being able to write tests that are easy to understand, and just have different ideas about the best way to do that. I guess it bothers me because I had to study this fixture for a number of minutes to make sure I understood how it worked. It's not clear that my suggestion is better so I'll drop it for now.

.github/workflows/slow_integration_tests.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python 3.8
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.8
+      - name: pip cache
+        uses: actions/cache@v1
+        with:
+          path: ~\AppData\Local\pip\Cache
```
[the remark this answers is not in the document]
davisagli: @prescod that's a mistake. I'll fix the other ones in a separate PR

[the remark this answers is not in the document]
davisagli: https://github.com/SFDO-Tooling/CumulusCI/pull/2806

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
davisagli: Seems weird to need to specify an org in order to run an integration test for a task that only talks to Marketing Cloud or github, but ok

cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py
davisagli: There's a limit on active scratch orgs in a dev hub. It could be problematic if we get to the point where a run of CCI's integration tests creates a lot of scratch orgs and doesn't clean them up.

Makefile
prescod: fe16f83

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
prescod: That code that translates orgnames to orgconfigs has that check already:

* https://github.com/SFDO-Tooling/CumulusCI/blob/246e75f49e89b3d269a9613d68e932c4363cf9fb/cumulusci/tests/pytest_plugins/pytest_sf_orgconnect.py#L50

integration_tests/pytest.ini
prescod: This is just in the integration_tests directory. Which I should probably get rid of. I've kind of gone back and forth about whether to put the integration tests just in the regular test files but with marks, or give them their own directory so there are less tests being skipped in a typical 'pytest' invocation.

davisagli: Ah okay, I missed that it's in integration_tests

.github/workflows/slow_integration_tests.yml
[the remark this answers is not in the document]
davisagli: @prescod I fixed the existing ones but this still needs to be updated.

PR #2784, opened 2021-08-04
CCI version 3.41

prescod:
Ready for review but not to. merge


HISTORY.rst
```diff
@@ -1,6 +1,45 @@
 =======
 History
 =======
+3.41.0 (2012-08-04)
```
davisagli, line 4:
```suggestion
3.41.0 (2012-08-05)
```

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+-------------------
+
+Changes
+
+* CCI has a new Metadata ETL task that allows adding fields to existing layouts. Sections can be selected using an index (#2766)
```
davisagli, line 9: It has other ways of targeting where to put the field too. I don't think it makes sense to call out section indexes specifically.

Also release notes for tasks should always mention the name of the task config!

```suggestion
* We added a new Metadata ETL task, `add_page_layout_fields`, that allows adding fields to existing layouts. (#2766)
```

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+* We added a task to enable an Einstein prediction (thanks, @erikperkins!) (#2778)
```
davisagli, line 11: @jstvz @davidmreed was it intentional to add this task class without a task config?

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+* We added standard flows for releasing unlocked packages: release_unlocked_beta and release_unlocked_production (#2768)
```
davisagli, line 13:
```suggestion
* We added standard flows for releasing unlocked packages: `release_unlocked_beta` and `release_unlocked_production` (#2768)
```

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+ 
+* We added documentation for using CumulusCI to build managed 2GP packages, 
+  unlocked packages, and extending NPSP and EDA with 2GP packages. (#2768)
```
davisagli, line 16: Link to https://cumulusci.readthedocs.io/en/stable/packaging.html which is where this will be published after the release tag is created.

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+* Contributions to CCI now require verification by isort (#2770)
```
davisagli, line 18: I would leave this out of the release notes; it's not a customer-facing change.

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+ 
+* CumulusCI now supports deploying unmanaged dependencies in SFDX source format. (#2735)
+
+* The `create_package_version` task now handles dependencies that use a `zip_url`. (#2735)
+
+* Updates to Github Actions configuration documentation. Thanks @Julian88Tex (#2773)
+
+* CumulusCI now automatically recognizes services and orgs configured via environment variables. See the `docs <https://github.com/SFDO-Tooling/CumulusCI/blob/481ab27446edbfdb038ea62834a6f276dc53bf3d/docs/headless.rst#L9>`_ for more details. (#2676 and #2776)
```
davisagli, line 26: I would put this in the Critical Changes section because I think it's pertinent to mention that the EnvironementProjectKeychain is deprecated and should no longer be necessary.

There's also a related change I think is worth calling out, which is that it's no longer necessary to explicitly set CUMULUSCI_KEY on Linux.

We should link to the docs on readthedocs (https://cumulusci.readthedocs.io/en/stable/headless.html), not github!

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+* We've updated the Push Upgrade tasks (push_list, push_sandbox, etc) task option `start_time` to accept ISO-8601 formatted datetimes. (#2769)
```
davisagli, line 28:
```suggestion
* We've updated the Push Upgrade tasks (`push_list`, `push_sandbox`, etc) task option `start_time` to accept ISO-8601 formatted datetimes. (#2769)
```

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+* CumulusCI now automatically recognizes enhanced domains in sandboxes when 
+  establishing a connection via JWT. (#2753)
```
davisagli, line 31: I would move this into Issues Closed and rephrase to: "Fixed a couple issues with connecting CumulusCI to sandboxes using enhanced domains." We ended up fixing more than just the JWT issue (see #2765)

We should also include the other change from 2753: You can now specify "sandbox": true on a CUMULUSCI_ORG_* variable in headless environments to indicate that the org you want to connect to is a sandbox. when connecting sandbox orgs in a headless environment

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+Issues Closed
```
davisagli, line 33: PR numbers for these?

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+* Fixed an issue where scratch orgs failed to be deleted in CI environments.
+
+* Fixed an issue where deleting an org failed to mark the org as deleted on CumulusCI's keychain.
+
+* Fixed an issue where CumulusCI would fail on Linux distributions that were incompatible with the `keyring` package.
```
davisagli, line 39: As I said above, put this in Critical Changes and point out that it should no longer be necessary to explicitly set CUMULUSCI_KEY (at least in a CI context where there aren't existing encrypted files to be read using that key)

HISTORY.rst
```diff
@@ -1,6 +1,45 @@
...
+
+Issues Closed
+
+* Fixed an issue where scratch orgs failed to be deleted in CI environments.
+
+* Fixed an issue where deleting an org failed to mark the org as deleted on CumulusCI's keychain.
+
+* Fixed an issue where CumulusCI would fail on Linux distributions that were incompatible with the `keyring` package. 
+
+* We fixed an issue causing release_2gp_production to fail with a dependency parsing error.
```
davisagli, line 41:
```suggestion
* We fixed an issue causing the `release_2gp_production` flow to fail with a dependency parsing error.
```

davisagli (changes requested):

HISTORY.rst
Br4nd0R: sorry for the bad link

HISTORY.rst
prescod: Isn't `CUMULUSCI_ORG_` a new feature in this release? If so, this isn't a change: it's just the behaviour of the new feature.

davisagli: No, it's not a new feature. The new feature is that you can use it without explicitly switching to EnvironmentProjectKeychain.

HISTORY.rst
prescod: If CUMULUSCI_KEY isn't necessary anymore, why do we still issue a warning? I don't think we can publicize a "critical change" that people should strip it out if it will generate a confusing warning?

HISTORY.rst
davisagli: Okay, I guess, but please make sure we end up with links to readthedocs both here and in the github release.

HISTORY.rst
davisagli: The warning is trying to say that we're able to store configuration but it won't be encrypted. I agree we should make it less scary but it's too late for this release. I'm okay with holding off on mentioning that CUMULUSCI_KEY is optional until we do that.

davisagli: In particular we should make sure that we aren't showing a warning that suggests installing keyring.alt on Linux, because I don't think that's actually what we want people to do.

HISTORY.rst
prescod: I think we should almost always add task configs.

davisagli: Yeah, I agree. I'm guessing it was an oversight but asked the question in case there's some special reason not to in this case.

davidmreed: No. I will address this quickly.

davidmreed: https://github.com/SFDO-Tooling/CumulusCI/pull/2786

prescod: And tasks.rst auto-generates now, right?

HISTORY.rst
prescod: Unless you feel very strongly, I'm going to leave it (and clarify a bit). We get enough external contributors of custom tasks that I think it is of interest to people who would read these notes.

HISTORY.rst
prescod: You don't think we should save "Critical Changes" for changes that are required (failures to be backwards compatible) rather than just recommended?

HISTORY.rst
davisagli: @prescod I'm fine with including this sort of thing but I think we should put it in a separate section so that customers aren't scratching their heads trying to figure out if it's relevant to them.

HISTORY.rst
davisagli: This one is grey area, I'll let you make the call.

HISTORY.rst
prescod: 61061226398

HISTORY.rst
prescod: 61061226398

HISTORY.rst
prescod: 61061226398

davisagli (dismissed):

HISTORY.rst
```diff
@@ -1,6 +1,47 @@
...
+
+* Contributions to CCI now require verification by isort, which ensures consistency in the order that imports are used. (#2770)
+ 
+* CumulusCI now supports deploying unmanaged dependencies in SFDX source format. (#2735)
+
+* The `create_package_version` task now handles dependencies that use a `zip_url`. (#2735)
+
+* Updates to Github Actions configuration documentation. Thanks @Julian88Tex (#2773)
+
+* CumulusCI now automatically recognizes services and orgs configured via environment variables. See the `docs <https://github.com/SFDO-Tooling/CumulusCI/blob/481ab27446edbfdb038ea62834a6f276dc53bf3d/docs/headless.rst#L9>`_ for more details. (#2676 and #2776)
```
davidmreed, line 26:
```suggestion
* CumulusCI now automatically recognizes services and orgs configured via environment variables. See the `docs <https://cumulusci.readthedocs.io/en/stable/headless.html>`_ for more details. (#2676 and #2776)
```

davidmreed (changes requested):

davidmreed (approved):

PR #2789, opened 2021-08-05
Bulk dataload results bug fix

prescod:
Mirror PR for #2733. 

Issues Fixed

Correct miscounting of bulk job results. (Thanks @sfdcale!)


cumulusci/tasks/bulkdata/step.py
```diff
@@ -17,6 +16,8 @@
 from cumulusci.core.utils import process_bool_arg
 from cumulusci.tasks.bulkdata.utils import iterate_in_chunks
 
+BULK_BATCH_SIZE = 3
```
davisagli, line 19: @prescod Did you really intend to change this from 10000 to 3 or is that left over from testing?

cumulusci/tasks/bulkdata/load.py
```diff
@@ -135,11 +137,13 @@ def _run_task(self):
                             raise BulkDataException(
                                 f"Step {after_name} did not complete successfully: {','.join(result.job_errors)}"
                             )
+                results[name] = result
         if self.options["set_recently_viewed"]:
             try:
                 self._set_viewed()
             except Exception as e:
                 self.logger.warning(f"Could not set recently viewed because {e}")
+        self.return_values = {"step_results": results}
```
davisagli, line 146: @prescod Returning this info in the return_values is good, but we should probably make sure that it gets serialized with meaningful keys rather than just a tuple.

cumulusci/tasks/bulkdata/step.py
```diff
@@ -91,42 +98,53 @@ def _parse_job_state(self, xml):
             el.text for el in tree.iterfind(".//{%s}stateMessage" % self.bulk.jobNS)
         ]
 
-        failures = tree.find(".//{%s}numberRecordsFailed" % self.bulk.jobNS)
-        record_failure_count = int(failures.text) if failures is not None else 0
-        processed = tree.find(".//{%s}numberRecordsProcessed" % self.bulk.jobNS)
-        records_processed = int(processed.text) if processed is not None else 0
+        # Get how many total records failed acorss all the batches.
```
davisagli, line 101:
```suggestion
        # Get how many total records failed across all the batches.
```

davisagli (changes requested):

cumulusci/tasks/bulkdata/load.py
prescod: @davisagli What serializer do we use and when? Is this related to freezing for MetaDeploy? Or Tubular API? JSON or YAML? Or both?

cumulusci/tasks/bulkdata/step.py
prescod: cc241d743fac9f07aabd9ac527561adcfd1bd950

cumulusci/tasks/bulkdata/load.py
prescod: a7af5aab7adb77

davisagli: It's not related to freezing steps for MetaDeploy. Both MetaCI (https://github.com/SFDO-Tooling/MetaCI/blob/main/metaci/build/flows.py#L33) and tubular (https://github.com/SFDO-Tooling/tubular/blob/main/tubular/worker/runner.py#L96) serialize return_values to JSON.

Would it make sense to switch from the namedtuple to a pydantic model so you can just call `.json()` rather than a custom serializer function? I'll approve anyway...

davisagli (approved):

PR #2792, opened 2021-08-06
Fix tag prefix and name inconsistencies

Br4nd0R:
# Critical Changes
* The `github_release` task now requires the `tag_prefix` option to be passed.
* The `github_copy_subtree` task now requires a `tag_name` option to be passed, and the `version` option has been deprecated.

# Issues Closed
* Fixed an issue where GitHub tags for 2GP package would always include the "release" prefix (even for Beta packages).

## Dev Notes
* @jstvz - We're reworking the `project_config.get_tag_for_version()` method to require both a `tag_prefix` and a `version`. The `PublishSubtree` class is the only other place where this method is referenced but is currently not passing a prefix.  I don't know too much about this task; is there a way we can derive which tag prefix to use—either `beta/` or `release/`—based on the the value of the `create_release` option? If not, is there another way we can infer what prefix is desired? Looks like we do set the `prerelease` value for the release based on `release.prerelease`, but we only have access to that release object after the tag has been created.

### DIFF
--- a/cumulusci/tasks/github/publish.py
+++ b/cumulusci/tasks/github/publish.py
@@ -19,9 +19,8 @@ class PublishSubtree(BaseGithubTask):
             "description": "The branch to update in the target repo",
             "required": True,
         },
-        "version": {
-            "description": "The version number to release. "
-            "Also supports latest and latest_beta to look up the latest releases. "
+        "tag_name": {
+            "description": "The name of the tag that should be associated with this release. "
             "Required if 'ref' is not set."
         },
         "ref": {
@@ -54,13 +53,8 @@ class PublishSubtree(BaseGithubTask):
 
         self.options["renames"] = self._process_renames(self.options.get("renames", []))
 
-        if self.options.get("version") in ("latest", "latest_beta"):
-            get_beta = self.options["version"] == "latest_beta"
-            self.options["version"] = str(
-                self.project_config.get_latest_version(beta=get_beta)
-            )
-        if "ref" not in self.options and "version" not in self.options:
-            raise TaskOptionsError("Either `ref` or `version` option is required")
+        if "ref" not in self.options and "tag_name" not in self.options:
+            raise TaskOptionsError("Either `ref` or `tag_name` option is required")
 
         self.options["create_release"] = process_bool_arg(
             self.options.get("create_release", False)
@@ -122,9 +116,7 @@ class PublishSubtree(BaseGithubTask):
         if "ref" in self.options:
             self.ref = self.options["ref"]
         else:
-            self.tag_name = self.project_config.get_tag_for_version(
-                self.options["version"]
-            )
+            self.tag_name = self.options["tag_name"]
             self.ref = f"tags/{self.tag_name}"
 
     def _download_repo_and_extract(self, path):
@@ -170,8 +162,8 @@ class PublishSubtree(BaseGithubTask):
     def _create_commit(self, path):
         committer = CommitDir(self.target_repo, logger=self.logger)
         message = f"Published content from ref {self.ref}"
-        if "version" in self.options:
-            message += f'\n\nVersion {self.options["version"]}'
+        if "tag_name" in self.options:
+            message += f'\n\nTag {self.options["tag_name"]}'
         return committer(
             path,
             self.options["branch"],
@@ -221,7 +213,7 @@ class PublishSubtree(BaseGithubTask):
             tagger={
                 "name": self.github_config.username,
                 "email": self.github_config.email,
-                "date": "{}Z".format(datetime.utcnow().isoformat()),
+                "date": f"{datetime.utcnow().isoformat()}Z",
             },
             lightweight=False,
         )
@@ -229,7 +221,18 @@ class PublishSubtree(BaseGithubTask):
         # Create the release
         self.target_repo.create_release(
             tag_name=self.tag_name,
-            name=self.options["version"],
+            name=self._get_release_name_from_tag(self.tag_name),
             prerelease=release.prerelease,
             body=release_body,
         )
+
+    def _get_release_name_from_tag(self, tag_name):
+        """Parse the name of the release from the tag name.
+        Example: The tag 'beta/1.117-Beta_2' should
+        return a name of '1.117 (Beta 2)'"""
+        tag_name = tag_name.split("/")[-1]
+        if "Beta" in tag_name:
+            tag_name = tag_name.replace("-", " ").replace("_", " ")
+            part = tag_name.partition("Beta")
+            tag_name = f"{part[0]}({part[1] + part[2]})"
+        return tag_name
--- a/cumulusci/tasks/github/publish.py
+++ b/cumulusci/tasks/github/publish.py
@@ -4,7 +4,11 @@ from pathlib import Path
 
 import github3.exceptions
 
-from cumulusci.core.exceptions import GithubException, TaskOptionsError
+from cumulusci.core.exceptions import (
+    CumulusCIException,
+    GithubException,
+    TaskOptionsError,
+)
 from cumulusci.core.utils import process_bool_arg, process_list_arg
 from cumulusci.tasks.github.base import BaseGithubTask
 from cumulusci.tasks.github.util import CommitDir
@@ -19,8 +23,13 @@ class PublishSubtree(BaseGithubTask):
             "description": "The branch to update in the target repo",
             "required": True,
         },
+        "version": {
+            "description": "(Deprecated) Only the values of 'latest' and 'latest_beta' are acceptable. "
+            "Required if 'ref' is not set. This will override any"
+        },
         "tag_name": {
             "description": "The name of the tag that should be associated with this release. "
+            "Values of 'latest' and 'latest_beta' are also allowed. "
             "Required if 'ref' is not set."
         },
         "ref": {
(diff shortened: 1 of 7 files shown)


jstvz:
> @Br4nd0R: ...is there a way we can derive which tag prefix to use—either `beta/` or `release/`—based on the the value of the `create_release` option? 

Not really. As it stands, the closest you can get is `version`:

https://github.com/SFDO-Tooling/CumulusCI/blob/037065fa181fc0b9816cad7141a2ccd413dc7260/cumulusci/tasks/github/publish.py#L57-L61

(A switch to `get_latest_tag()` could work here instead?)

---

> Looks like we do set the `prerelease` value for the release based on `release.prerelease`, but we only have access to that release object after the tag has been created.

Not quite. If given a `version` the task looks up an existing tag here:

https://github.com/SFDO-Tooling/CumulusCI/blob/037065fa181fc0b9816cad7141a2ccd413dc7260/cumulusci/tasks/github/publish.py#L200

---

>...is there another way we can infer what prefix is desired? 

I don't think so. We were trying to be helpful to the user by not making them specify a tag. This was possible to infer in a 1GP context, but no longer. I'm not seeing another approach besides making a breaking change to require a tag instead of a version number.

Br4nd0R:
@jstvz - It would be appreciated if you could look over my changes to the `PublishSubtree` task. It now requires a `tag_name` instead of a `version` as an option.

cumulusci/tasks/github/publish.py
```diff
@@ -54,13 +53,8 @@ def _init_options(self, kwargs):
 
         self.options["renames"] = self._process_renames(self.options.get("renames", []))
 
-        if self.options.get("version") in ("latest", "latest_beta"):
-            get_beta = self.options["version"] == "latest_beta"
-            self.options["version"] = str(
-                self.project_config.get_latest_version(beta=get_beta)
-            )
-        if "ref" not in self.options and "version" not in self.options:
-            raise TaskOptionsError("Either `ref` or `version` option is required")
```
davisagli, line 63: @Br4nd0R I guess we have to stop supporting `version` set to a specific version number, but the [main use of this in the wild](https://github.com/SalesforceFoundation/sal/blob/a12a3360fe9f7d9b6f07ce52255b003384d050d1/cumulusci.yml#L1159) is using `version: latest` and I don't think we need or want to stop supporting `version: latest` and `version: latest_beta`. Instead of the existing logic (get version using get_latest_version, then convert to a tag name using get_tag_for_version) we can simply use get_latest_tag which also takes a flag for whether to find beta or final tags. /cc @jstvz

cumulusci/tasks/github/publish.py
```diff
@@ -221,15 +213,26 @@ def _create_release(self, path, commit):
             tagger={
                 "name": self.github_config.username,
                 "email": self.github_config.email,
-                "date": "{}Z".format(datetime.utcnow().isoformat()),
+                "date": f"{datetime.utcnow().isoformat()}Z",
```
jstvz, line 216: 🤔: Well, no need to do this, but I wonder if we can just pass `datetime.now(timezone.utc).isoformat()` instead.

cumulusci/tasks/github/publish.py
```diff
@@ -221,15 +213,26 @@ def _create_release(self, path, commit):
...
         )
+
+    def _get_release_name_from_tag(self, tag_name):
+        """Parse the name of the release from the tag name.
+        Example: The tag 'beta/1.117-Beta_2' should
+        return a name of '1.117 (Beta 2)'"""
+        tag_name = tag_name.split("/")[-1]
+        if "Beta" in tag_name:
+            tag_name = tag_name.replace("-", " ").replace("_", " ")
+            part = tag_name.partition("Beta")
```
jstvz, line 236: TIL about `partition`.

jstvz (commented):
Publish looks fine overall, just a comment and a couple of questions.

cumulusci/tasks/github/publish.py
```diff
@@ -221,15 +213,26 @@ def _create_release(self, path, commit):
...
         # Create the release
         self.target_repo.create_release(
             tag_name=self.tag_name,
-            name=self.options["version"],
+            name=self._get_release_name_from_tag(self.tag_name),
             prerelease=release.prerelease,
             body=release_body,
         )
+
+    def _get_release_name_from_tag(self, tag_name):
```
jstvz, line 229: Github ate the other substantive question from the above review: Why wouldn't `self.project_config.get_version_for_tag(tag_name)` work here?

davisagli:
> I agree that we ought to continue supporting the latest/latest_beta shorthand, but I'm not convinced that doing so via version (while disallowing version numbers) is the best way to do so. We'd end up with three mutually exclusive options in that case, and I think it'd be just as easy to support latest in tag_name.

I'd like to make sure that latest/latest_beta still work as values for `version` because it's backwards-compatible and doesn't require coordinating updates to SAL's configuration. But I'd be okay with logging a warning that it's deprecated and preferred to use those values in `tag_name` instead.

jstvz:
> I'd like to make sure that latest/latest_beta still work as values for `version` because it's backwards-compatible and doesn't require coordinating updates to SAL's configuration. 

Fair point. My belief is that using version numbers here was a conceptual error. We're only using those in a handful of tasks that are related to packages. We should have been using tags from the beginning, I think.

> But I'd be okay with logging a warning that it's deprecated and preferred to use those values in `tag_name` instead.

Works for me.

cumulusci/tasks/github/publish.py
```diff
@@ -20,8 +24,12 @@ class PublishSubtree(BaseGithubTask):
             "required": True,
         },
         "version": {
-            "description": "The version number to release. "
-            "Also supports latest and latest_beta to look up the latest releases. "
+            "description": "(Deprecated) Only the values of 'latest' and 'latest_beta' are acceptable. "
+            "Required if 'ref' is not set. This will override any"
```
jstvz, line 28: Got a partial sentence here.

cumulusci/tasks/github/publish.py
```diff
@@ -20,8 +24,12 @@ class PublishSubtree(BaseGithubTask):
             "required": True,
         },
         "version": {
-            "description": "The version number to release. "
-            "Also supports latest and latest_beta to look up the latest releases. "
+            "description": "(Deprecated) Only the values of 'latest' and 'latest_beta' are acceptable. "
```
jstvz, line 27: ±: I would argue in favor of a specific deprecation version, say `>= 3.44.0`?

cumulusci/tasks/github/publish.py
```diff
@@ -170,8 +199,8 @@ def _rename_files(self, zip_dir):
     def _create_commit(self, path):
         committer = CommitDir(self.target_repo, logger=self.logger)
         message = f"Published content from ref {self.ref}"
-        if "version" in self.options:
-            message += f'\n\nVersion {self.options["version"]}'
+        if "tag_name" in self.options:
+            message += f'\n\nTag {self.options["tag_name"]}'
```
jstvz, line 203: We should be specific here:

```suggestion
            message += f'\n\nTag {self.tag_name}'
```
Otherwise we could end up with `Tag latest_beta`.

jstvz (changes requested):
We need to state the order of precedence explicitly if we're not only allowing one of the three options, either by adding `task_docs`, or updating the option descriptions.

cumulusci/tasks/github/publish.py
jstvz: Actually, OK leave as is, but add a `FIXME`.

cumulusci/tasks/github/publish.py
Br4nd0R: Current release is 3.41.0. So should we say `(Deprecated >= 3.42.0)`?

cumulusci/tasks/github/publish.py
Br4nd0R: I can make this change. It's better to use the instance reference to `tag_name` I feel.

jstvz (commented):
One note RE: `warnings` (that I don't feel too strongly about) below.

cumulusci/tasks/github/publish.py
```diff
@@ -54,13 +63,21 @@ def _init_options(self, kwargs):
 
         self.options["renames"] = self._process_renames(self.options.get("renames", []))
 
-        if self.options.get("version") in ("latest", "latest_beta"):
-            get_beta = self.options["version"] == "latest_beta"
-            self.options["version"] = str(
-                self.project_config.get_latest_version(beta=get_beta)
+        if "version" in self.options:  # pragma: no cover
```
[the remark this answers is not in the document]
Br4nd0R: - First execution is with `warnings.warn`. It seems to include the file path and line num where the warning was issued from. 
- Second execution is with `self.logger.warn`.

I prefer `self.logger.warn`. @jstvz - Let me know what you think.

logger_output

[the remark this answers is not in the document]
davisagli: `logger.warn` is a deprecated spelling; `logger.warning` is preferred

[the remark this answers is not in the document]
Br4nd0R: Thanks, will change. I verified that `logger.warning` identical output to `logger.warn`.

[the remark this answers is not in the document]
jstvz: 1. I think you want `DeprecationWarning`, since `UserWarning` is the default.
2. We can handle both interactive and client use by setting `logging.captureWarnings(True)`.

But this is fine as is, IMO.

jstvz (approved):

PR #2805, opened 2021-08-18
Use the JSON result from force:source:convert to determine where the MDAPI package ended up

davisagli:
- Run force:source:convert with the --json flag and parse the result to get the result path
- Also improve error handling: if there was an error, get it from the result and raise it as a Python exception
- De-mockified a test so we have at least one that actually runs sfdx force:source:convert

# Critical Changes

# Changes

# Issues Closed
- Fixed a bug where the `create_package_version` task reported a "Package.xml not found" error when using sfdx 7.112 or greater.

### DIFF
--- a/.github/workflows/feature_test.yml
+++ b/.github/workflows/feature_test.yml
@@ -95,7 +95,9 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install sfdx
-        run: npm install sfdx-cli --global
+        run:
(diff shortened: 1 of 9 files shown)


davidmreed (approved):

davisagli:
Abandoning this for now since I can't figure out how to reliably wait for sfdx to be done accessing the current directory so that we can delete it in Windows.

PR #2806, opened 2021-08-18
fix cache path for ubuntu builds, and broken test with hardcoded date

davisagli:
# Critical Changes

# Changes

# Issues Closed


jstvz (approved):
These are fine as-is, but we should probably switch to `pip cache dir` at some point.

davisagli:
@jstvz I used this to fix the start_time tests that were hardcoded to yesterday's date -- weird coincidence that that happened right after we were working on that code.

davidmreed (approved):

PR #2807, opened 2021-08-18
Resolution strategies for sources

davidmreed:
# Critical Changes

# Changes

- The `sources` feature, which allows CumulusCI projects to consume automation from other projects, now supports `resolution_strategy`, just like dependencies. Sources can now resolve to the same GitHub refs as corresponding dependencies, including branch matching. The default behavior is to use the resolution strategy `production`.

# Issues Closed


cumulusci/core/config/tests/test_config.py
```diff
@@ -619,29 +620,33 @@ def test_get_namespace__not_found(self):
         with self.assertRaises(NamespaceNotFoundError):
             project_config.get_namespace("test")
 
+    def test_get_namespace__bad_spec(self):
+        universal_config = UniversalConfig()
+        project_config = BaseProjectConfig(
+            universal_config, {"sources": {"test": {"foo": "some_nonsense"}}}
+        )
+        with self.assertRaises(ValueError):
```
davisagli, line 628:
```suggestion
        with pytest.raises(ValueError, match="Invalid source spec"):
```

cumulusci/core/config/project_config.py
```diff
@@ -516,42 +522,49 @@ def get_flow(self, name):
...
+            else:
+                break
 
-        `spec` is a dict which contains different keys depending on the type of source:
+        if not parsed_spec:
+            raise ValueError(f"Invalid source spec: {spec}")
 
-        - `github` indicates a GitHubSource
-        - `path` indicates a LocalFolderSource
+        return self.include_source(parsed_spec)
```
davisagli, line 546: include_source is called with a dict (stored on the Step model in the db) in MetaDeploy -- https://github.com/SFDO-Tooling/MetaDeploy/blob/226e32eedd0ba316c3f1f4f43ec774f601485858/metadeploy/api/models.py#L672 -- so it'd be helpful to either keep include_source accepting dict input, or put the dict-to-model logic in its own method that MetaDeploy can use.

Also keep in mind that the old implementation was using the equality of frozendicts to avoid fetching the same source more than once. We need to make sure that if two models are constructed from equivalent dicts, the resulting models are also equivalent.

davidmreed: @davisagli See what you think of the tweak I made to support that use case.

As far as caching/equality - the new version is using `HashableBaseModel` so that it can be a dict key. Those models compare based on their field values, and their hash is derived from the JSON representation of those same values. I _think_ the code therefore makes the same guarantees (and there is a test, `test_include_source__cached()` attempting to demonstrate that), but if I've missed a hole please tell me.

davisagli (approved):

PR #2811, opened 2021-08-18
Support all package install options

davidmreed:
# Critical Changes

- We now support the `activate_remote_site_settings`, `security_type`, `name_conflict_resolution`, and `password` options on both `update_dependencies` (no `password`) and `install_managed`. 

  We also fixed a minor inconsistency in defaulting the `activate_remote_site_settings` (or formerly `activateRSS`) setting. Projects that define custom tasks based on the `InstallPackageVersion` class should ensure they explicitly set the `activate_remote_site_settings` option, or accept the new default of `True`.

  MetaDeploy install plans now _do not_ freeze defaulted package install options. If your install plans are dependent on specific install options, we recommend explicitly specifying them. Install plans without explicit options will use the defaults at the time of execution.

# Changes

# Issues Closed


davidmreed:
Note for discussion: these changes result in MetaDeploy plans freezing _without_ package install options anywhere they're not explicitly specified (that is, allowing the defaults to be used). This is a change from the previous behavior, which was inconsistent: previous versions froze a default of `True` for `activateRSS` on `install_managed` and `install_managed_beta`, and `update_dependencies` froze a default of `FULL` for `security_type`.

I could argue either way whether or not we should have the tasks freeze the current defaults at the time of freezing if no explicit configuration is present.

cumulusci/salesforce_api/package_install.py
```diff
@@ -48,6 +48,43 @@ class PackageInstallOptions(CCIModel):
     password: Optional[str]
     security_type: SecurityType = SecurityType.FULL
```
davisagli, line 50: Should activate_remote_site_settings really be optional? We don't want it to be possible to explicitly set it to null, do we?

cumulusci/salesforce_api/package_install.py
```diff
@@ -48,6 +48,43 @@ class PackageInstallOptions(CCIModel):
...
+            if "name_conflict_resolution" in task_options:
+                options.name_conflict_resolution = NameConflictResolution(
+                    task_options["name_conflict_resolution"]
+                )
+            if "password" in task_options:
+                options.password = task_options["password"]
+        except ValueError as e:
+            raise TaskOptionsError(f"Invalid task options: {e}")
+
+        return options
```
davisagli, line 71: Maybe I'm missing something, but I think this could be done more concisely but equivalently like this:
```
task_options = {**task_options, activate_remote_site_settings=process_bool_arg(
    task_options.get("activate_remote_site_settings", True)
)
options = PackageInstallOptions(**task_options)
```

(By default, pydantic ignores extra fields in input)

cumulusci/tasks/salesforce/install_package_version.py
```diff
@@ -33,9 +33,7 @@ class InstallPackageVersion(BaseSalesforceApiTask):
             "number to the user and in logs. Has no effect otherwise."
         },
         "activateRSS": {
-            "description": "If True, preserve the isActive state of "
-            "Remote Site Settings and Content Security Policy "
-            "in the package. Default: False."
+            "description": "Deprecated. Use activate_remote_site_settings instead."
         },
         "password": {"description": "The package password. Optional."},
```
davisagli, line 38: This is replaced in PACKAGE_INSTALL_TASK_OPTIONS

cumulusci/tasks/salesforce/update_dependencies.py
```diff
@@ -56,6 +56,7 @@ class UpdateDependencies(BaseSalesforceTask):
         "packages_only": {
             "description": "Install only packaged dependencies. Ignore all unmanaged metadata. Defaults to False."
         },
+        **PACKAGE_INSTALL_TASK_OPTIONS,
```
davisagli, line 59: Do we really want to include password here?

cumulusci/salesforce_api/package_install.py
davidmreed: No, we don't - correcting this.

cumulusci/salesforce_api/package_install.py
davidmreed: It is more concise, but I find the explicit form more readable. I also would prefer not to create a second location (in addition to the Pydantic model) where `activate_remote_site_settings` gets defaulted.

cumulusci/tasks/salesforce/update_dependencies.py
davidmreed: Probably not. Correcting.

davisagli (approved):

PR #2814, opened 2021-08-19
Make all datetimes tz-aware within the push tasks

### LINKED ISSUE #2808: Push upgrade tasks' start_time option compares the wrong datetime
**Describe the bug**
Traceback:
```python-traceback
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/cumulusci/cli/cci.py", line 76, in main
    cli(args[1:], standalone_mode=False, obj=runtime)
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 829, in __call__
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 1259, in invoke
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 1066, in invoke
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/click/core.py", line 610, in invoke
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/cumulusci/cli/task.py", line 178, in run_task
    task()
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/cumulusci/core/tasks.py", line 152, in __call__
    self.result = self._run_task()
  File "/Users/jestevez/.pyenv/versions/3.9.5/lib/python3.9/site-packages/cumulusci/tasks/push/tasks.py", line 282, in _run_task
    if start_time < datetime.utcnow():
TypeError: can't compare offset-naive and offset-aware datetimes
```

**Reproduction steps**
1. Pass ISO formatted string `--start_time 2021-08-19T04:00+00:00` to `push_sandbox`

**Gist**
https://gist.github.com/jstvz/c727d4da09bb2dfe86e51ba67066cccb

**Additional information**
`datetime.utcnow()` is offset-naive. Use `datetime.now(timezone.utc)` instead.

**Workaround**
Pass `start_time` without an offset: `2021-08-19T04:00`.

davisagli:
@jstvz thoughts? I'm not sure if this is the best way to fix this.

# Critical Changes

# Changes

# Issues Closed
Fixes #2808


jstvz (changes requested):
This is basically where I landed, but using `datetime.timezone` and sticking to treating naive datetimes as UTC rather than local. 

∆: Either approach to naive datetimes is acceptable, but we need to update the option description to make it explicit.

davisagli:
@jstvz I see that we were treating naive datetimes as UTC even before https://github.com/SFDO-Tooling/CumulusCI/pull/2769 so I'll make that change for backwards compatibility. Good call.

I used dateutil.tz.tzlocal because I figured it'd be best to always have tzinfo objects from dateutil even when we didn't use it for parsing. Is there a reason to prefer datetime.timezone? I could certainly be missing something.

jstvz (approved):
> Is there a reason to prefer datetime.timezone? 

Not really. (TIL: China Standard Time is UTC+8 for the entire country.)

PR #2815, opened 2021-08-19
CumulusCI v3.42.0

Br4nd0R:


HISTORY.rst
```diff
@@ -1,7 +1,29 @@
 =======
 History
 =======
-3.41.0 (2012-08-05)
+
+3.42.0 (2021-08-19)
+-------------------
+Critical Changes
+
+* The ``github_release`` task now requires the ``tag_prefix`` option to be passed. (#2792)
```
davisagli, line 9: I might elaborate on this a bit:
```suggestion
* The ``github_release`` task now requires the ``tag_prefix`` option to be passed, because for 2nd-generation packages we can't tell from the version number whether it is a beta or not. We've updated the standard release flows to set the ``tag_prefix`` appropriately, but if you have custom flows using this task you will need to update them. (#2792)
```

HISTORY.rst
```diff
@@ -1,7 +1,29 @@
...
+ 
+* The ``github_copy_subtree`` task now requires a ``tag_name`` option to be passed. (#2792)
+
+* The ``github_copy_subtree`` task option ``version`` has been deprecated (use the ``tag_name`` option instead). (#2792)
```
davisagli, line 13: Technically it's not required, and it probably makes sense to combine these, something like:

```suggestion
* In order to run the ``github_copy_subtree`` task for a specific package version, you must now use the ``tag_name`` option instead of the ``version`` option. Using the ``version`` option set to ``latest`` or ``latest_beta`` is deprecated; it's preferred to pass these values in the ``tag_name`` option instead. (#2792)
```

HISTORY.rst
```diff
@@ -1,7 +1,29 @@
...
+
+Changes
+
+* ``uninstall_packaged_incremental`` now defaults to ignoring non-deletable ``CustomObjectTranslation`` metadata.
```
davisagli, line 17:
```suggestion
* The ``uninstall_packaged_incremental`` task now defaults to ignoring non-deletable ``CustomObjectTranslation`` metadata.
```

HISTORY.rst
```diff
@@ -1,7 +1,29 @@
...
+  If your project customizes the ``ignore_types`` option on ``uninstall_packaged_incremental``, we recommend you add ``CustomObjectTranslation`` to this option. (#2790)
+
+Issues Fixed
+
+* Fixed an issue where bulk job results were being miscounted. (Thanks @sfdcale!) (#2789)
+
+* Fixed an issue where GitHub tags for 2GP package would always include the "release" prefix (even for Beta packages). (#2792)
```
davisagli, line 24:
```suggestion
* Fixed an issue where GitHub tags for a 2GP package would always include the "release" prefix (even for Beta package versions). (#2792)
```

davisagli (changes requested):

davisagli (approved):

PR #2819, opened 2021-08-23
Feature/ignore row errors

prescod:
# Changes

The `snowfakery` task now supports the `ignore_row_errors` option to continue loading even if there are row errors.


cumulusci/tasks/bulkdata/load.py
```diff
@@ -137,17 +137,24 @@ def _run_task(self):
                             raise BulkDataException(
                                 f"Step {after_name} did not complete successfully: {','.join(result.job_errors)}"
                             )
-                results[name] = result
+                results[name] = StepResultInfo(mapping, result)
```
davisagli, line 140: Why are you putting the mapping here? It seems like an input rather than a result. Step results are stored in the database in MetaCI and I don't think we want to store a copy of the mapping every time a build using this task runs.

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -676,3 +718,11 @@ def relevant_sobjects(self):
         with open(self.mapping_file, encoding="utf-8") as f:
             mapping = yaml.safe_load(f)
             return [m.get("sf_object") for m in mapping.values() if m.get("sf_object")]
+
+
+class RunningTotals:
+    errors: int = 0
+    successes: int = 0
+
+    def __repr__(self):  # pragma: no cover
```
davisagli, line 727: Please go ahead and cover code that is this easy to cover. I've seen too many mistakes in simple string formatting functions.

cumulusci/tasks/bulkdata/load.py
davisagli: Ah, I see below that you're using it to get the sobject name. Could we make the step return just that rather than the entire mapping?

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -313,6 +321,32 @@ def _report_status(
...
+    def update_running_totals(self) -> None:
+        while True:
+            try:
+                results = self.load_data_q.results_reporter.get(block=False)
+            except Empty:
+                break
+            if "results" in results:
+                self.update_running_totals_from_load_step_results(results["results"])
+            elif "error" in results:
+                self.logger.info(f"Error in load: {results}")
```
davisagli, line 333: This should probably be logged at the warning or error level rather than info

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -313,6 +321,32 @@ def _report_status(
...
+            else:  # pragma: no cover
+                self.logger.info(f"Unexpected message from subtask: {results}")
```
davisagli, line 335: Ditto

cumulusci/tasks/bulkdata/load.py
prescod: The result doesn't have the SObject name, so I needed at least that. As far as I knew at the time, storing the whole mapping was essentially free, so I just stored it all in case some future user wants the record_type or bulk_mode or whatever. But I can change it to only store what we know we need.

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -328,8 +362,6 @@ def tick(
         ):  # pragma: no cover
             self.logger.info("Finished Generating")
         # queue is full
-        elif self.data_gen_q.full:
-            self.logger.info("Waiting before datagen (queue full)")
```
davisagli, line 332: Why is this not needed? (and in any case you need to remove the comment on line 364)

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -506,14 +542,15 @@ def _generate_and_load_initial_batch(self, working_directory: T.Optional[Path]):
 
         template_dir = Path(working_directory) / "template_1"
         template_dir.mkdir()
-        self._generate_and_load_batch(
+        results = self._generate_and_load_batch(
             template_dir,
             {
                 "generator_yaml": self.options.get("recipe"),
                 "num_records": 1,  # smallest possible batch to get to paralleizing fast
```
davisagli, line 549:
```suggestion
                "num_records": 1,  # smallest possible batch to get to parallelizing fast
```

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
```diff
@@ -20,11 +21,52 @@
...
+def call_closure():
+    return_values = cycle(iter(FAKE_LOAD_RESULTS))
+
+    def __call__(self, *args, **kwargs):
+        self.return_values = {"step_results": next(return_values)}
+        return __call__.mock(*args, **kwargs)
+
+    __call__.mock = mock.Mock()
+
+    return __call__
```
davisagli, line 61: This is a lot of magic for test code. Please add a comment about how it's here to cycle through the fake results each time the task is called.

I don't understand the part about setting `__call__.mock` and using it within the closure. `__call__` doesn't actually need to return anything, does it?

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
```diff
@@ -239,8 +281,8 @@ def test_record_count(self, snowfakery, mock_load_data):
         with mock.patch.object(task, "logger") as logger:
             task()
         mock_calls_as_string = str(logger.mock_calls)
-        assert "Account: 5 records" in mock_calls_as_string, mock_calls_as_string[-500:]
-        assert "Contact: 5 records" in mock_calls_as_string, mock_calls_as_string[-500:]
+        assert "Account: 5 succe" in mock_calls_as_string, mock_calls_as_string[-500:]
```
davisagli, line 284: "succe"?

davisagli (changes requested):

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
prescod: I want lines like this to work:

        assert len(mock_load_data.mock_calls) == 3
        
 The gist is that I'm sort of wrapping a mock so I can inject a side_effect. The normal side effect techniques are hard to achieve with pure mocks because they don't get bound like bound methods do. So I need a "real function" that wraps the mock.
 
 I can add docs to that effect.

davisagli: It might be clearer as a purpose-built fake object that updates return_values and records the call count rather than trying to wrap the mock. But I understand now and I'll accept it if you add a comment.

davisagli (approved):

PR #2838, opened 2021-08-30
Marketing Cloud: Create Subscriber Attribute, Create User, Update User Role

codexmarcel:
# Critical Changes

# Changes
Several new tasks for interacting with the Marketing Cloud SOAP API:
- `marketing_cloud_create_subscriber_attribute` creates a Subscriber Attribute.
- `marketing_cloud_create_user` creates a new User.
- `marketing_cloud_update_user_role` assigns a Role to an existing User.

# Issues Closed


cumulusci/tasks/marketing_cloud/api.py
```diff
@@ -0,0 +1,205 @@
+import requests
+from lxml import etree
+
+from cumulusci.salesforce_api import mc_soap_envelopes as envelopes
+
+from .base import BaseMarketingCloudTask
+
+
+class MarketingCloudCreateSubscriberAttribute(BaseMarketingCloudTask):
```
davisagli, line 9: I'd be inclined to remove MarketingCloud from the names of these classes; they are already in a Python package named marketing_cloud

cumulusci/tasks/marketing_cloud/api.py
```diff
@@ -0,0 +1,205 @@
...
+            access_token=self.mc_config.access_token,
+            parent_bu_mid=self.options.get("parent_bu_mid"),
+            default_bu_mid=self.options.get("default_bu_mid"),
+            external_key=self.options.get("external_key"),
+            user_name=self.options.get("user_name"),
+            user_email=self.options.get("user_email"),
+            user_password=self.options.get("user_password"),
+            user_username=self.options.get("user_username"),
+            role_id=self.options.get("role_id"),
+        )
```
davisagli, line 107: You could do something like this:
```
from collections import defaultdict
string_options = defaultdict(str)
string_options.update(self.options)
envelope = envelope.format(
    soap_instance_url=self.mc_config.soap_instance_url,
    access_token=self.mc_config.access_token,
    **string_options,
)
```

cumulusci/tasks/marketing_cloud/api.py
```diff
@@ -0,0 +1,205 @@
...
+        # construct request
+        # TO DO: DRY refactor
```
davisagli, line 109: +1

cumulusci/tasks/marketing_cloud/tests/test_api.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    task = create_task(
+        MarketingCloudCreateSubscriberAttribute,
+        {
+            "attribute_name": "test",
+        },
+        project_config=project_config,
+    )
+    task()
```
davisagli, line 65: We should add an assertion:
```
assert task.return_values == {"success": True}
```

and we need similar tests for the other 2 tasks. The setup of the project_config and services can be moved into a shared fixture

davisagli (approved):

PR #2844, opened 2021-09-01
Fix formatting of cumulusci.yml

bethbrains:
- Formatting only; prettier is suggesting these changes

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,26 +1,26 @@
 default_language_version:
-  python: python3
+    python: python3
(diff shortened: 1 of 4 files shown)


davisagli (commented):
Hmm, I just ran prettier on the same file and it didn't suggest the change to list indentation. Do we need some configuration file to make sure that it's applying the same preferences for both of us?

davisagli (approved):

davisagli (approved):

PR #2848, opened 2021-09-01
Run prettier against yaml files in repo

jstvz:
...to reduce diffs.

<details>
<summary>Pointless bike-shedding</summary>
 I understand that `prettier` is supposed to help us avoid quibbling over details like these, but... I'll be that guy, I guess.

I don't think adopting 4 spaces indentation is the way to go here. 2 spaces isn't part of the spec, but it's prettier's default, and is the consensus spacing for the format (Github's workflow docs, for example). 
</details>

# Critical Changes

# Changes

# Issues Closed


davisagli:
> I don't think adopting 4 spaces indentation is the way to go here. 2 spaces isn't part of the spec, but it's prettier's default, and is the consensus spacing for the format (Github's workflow docs, for example).

@bethbrains asked if we could do 4 since 2 looked cramped, and I agreed since I felt the same way. I don't feel strongly about it though.

davisagli (approved):

davisagli:
But it'd matter less if we used tabs... [waits for incoming tomatoes]

[not actually advocating for tabs]

PR #2832, opened 2021-08-27
Capture 1GP version ids for efficiency

davidmreed:
# Critical Changes

# Changes

- CumulusCI uses package version Ids from 1GP releases wherever available, reducing the need to install 1GP packages in an org to build 2GP dependencies.

# Issues Closed

PR #2851, opened 2021-09-02
Consume unlocked, non-namespaced releases

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue causing CumulusCI to fail to install releases that contain an Unlocked Package without a namespace.


davidmreed:
Builds on https://github.com/SFDO-Tooling/CumulusCI/pull/2832

davisagli:
@davidmreed This needs a conflict resolved

cumulusci/core/dependencies/dependencies.py
```diff
@@ -329,8 +329,14 @@ def flatten(self, context: BaseProjectConfig) -> List[Dependency]:
             )
         )
 
-        # Deploy the project, if unmanaged.
-        if not managed:
+        if not self.managed_dependency:
```
davisagli, line 332: Would it be reasonable to rename `managed_dependency` to `packaged_dependency`?

davidmreed: Sure. Done.

bethbrains (approved):

PR #2860, opened 2021-09-07
Catch, retry and if necessary, warn on timeout errors during schema caching

prescod:
Previously, when CumulusCI would try to cache information from some orgs about metadata, it would hang due to an underlying platform bug. Now it has its own timeouts and retries and will report the errors if the retries fail.


cumulusci/utils/http/multi_request.py
```diff
@@ -1,10 +1,13 @@
...
 from itertools import chain
-from typing import Dict, Iterable
 
+from requests.exceptions import ReadTimeout
 from requests_futures.sessions import FuturesSession
 
-from cumulusci.utils.iterators import iterate_in_chunks
+from cumulusci.utils.iterators import iterate_in_chunks, partition
+
+RECOVERABLE_ERRORS = (ReadTimeout, ConnectionError)
```
davisagli, line 10: These are only recoverable if we're doing a non-mutative request like GET. Otherwise we don't know whether the server handled the request or not.

cumulusci/utils/http/multi_request.py
```diff
@@ -21,19 +24,41 @@ def __enter__(self, *args):
...
         return self.session.request(
             method=method,
-            url=self.base_url + path.lstrip("/"),
+            url=self.base_url + url.lstrip("/"),
             headers=headers,
             json=json,
+            timeout=30,
+        )
+
+    def do_requests(self, requests: T.Iterable[T.Dict]):
```
davisagli, line 40:
```suggestion
    def do_requests(self, requests: T.Iterable[T.Dict]):
        """Initiate requests and wait for them to complete.
        
        Returns a tuple with a sequence of successes and a sequence of failures.
        """
```

cumulusci/utils/http/multi_request.py
```diff
@@ -21,19 +24,41 @@ def __enter__(self, *args):
...
+        def catch(future, futures_to_requests):
```
davisagli, line 41: You could move this down below line 49 and let futures_to_requests be in the closure

cumulusci/utils/http/multi_request.py
```diff
@@ -21,19 +24,41 @@ def __enter__(self, *args):
...
-        headers = {**(headers or {}), "Accept-Encoding": "gzip"}
+    def _async_request(
+        self, url: str, method: str, json: object = None, httpHeaders: dict = None
+    ):
+        "Make an async HTTP request and return a future"
+        headers = {**(httpHeaders or {}), "Accept-Encoding": "gzip"}
         return self.session.request(
             method=method,
-            url=self.base_url + path.lstrip("/"),
+            url=self.base_url + url.lstrip("/"),
```
davisagli, line 34: Why did you rename path to url, given that it's a path rather than an url?

cumulusci/utils/http/multi_request.py
```diff
@@ -100,13 +127,62 @@ def __enter__(self, *args):
...
+        self, requests
+    ) -> T.Tuple[T.Sequence, T.Sequence]:  # results, erros
         if not self.psf:
             raise AssertionError(
                 "Session was not opened. Please call open() or use as a context manager"
             )
 
         composite_requests = create_composite_requests(requests, self.chunk_size)
-        composite_results = self.psf.do_requests(composite_requests)
+        composite_results, errors = self.psf.do_requests(composite_requests)
```
davisagli, line 139: It's hard to read psf as anything other than Python Software Foundation

cumulusci/utils/http/multi_request.py
```diff
@@ -100,13 +127,62 @@ def __enter__(self, *args):
...
+        else:
+            singleton_results = []
+            unrecoverable_errors = []
+
+        individual_results = list(individual_results)
+
+        individual_results += list(singleton_results)
+        return individual_results, unrecoverable_errors
+
+    def retry_errors(self, errors):
```
davisagli, line 154: docstring

cumulusci/utils/http/multi_request.py
```diff
@@ -100,13 +127,62 @@ def __enter__(self, *args):
...
+        return singleton_results, unrecoverable_errors + errors
+
+
+def split_requests(composite_request: dict):
+    # need to remove this prefix-pattern because it will be added again laters
+    prefix = "/services/data/vxx.x"
+    single_requests = [r.copy() for r in composite_request["json"]["compositeRequest"]]
+    for request in single_requests:
+        del request["referenceId"]
+        if request["url"].startswith(prefix[0:-4]):
```
davisagli, line 186: Why would it literally start with vxx.x? Is this something you ended up with because of your integration testing's normalization that won't work for real?

cumulusci/tests/util.py
```diff
@@ -261,3 +262,25 @@ def unmock_env():
         return mock_env(homedir, cci_key)
     else:
         return nullcontext()
+
+
+class FakeUnreliableRequestHandler:
```
davisagli, line 267: docstring

cumulusci/utils/http/multi_request.py
prescod: I decided that maybe its pushing the boundary of the size of a a good closure and had very few dependencies on local state, so I just turned it into a normal function. 08b7f7d6e

cumulusci/utils/http/multi_request.py
davisagli: Not a problem I'm asking you to change. Just trying to give insight into why using abbreviations is generally bad.

cumulusci/utils/http/multi_request.py
davisagli: Why do you prefer namedtuples over dataclasses?

cumulusci/utils/http/multi_request.py
davisagli: @prescod no, it doesn't. I don't understand why the composite API requests would have the literal string "vxx.x" instead of a specific version like "v52.0"

cumulusci/utils/http/multi_request.py
prescod: We haven't removed support for 3.6 yet, nor added a dataclasses backport library, as far as I know.

cumulusci/utils/http/multi_request.py
prescod: The composite APIs do have a literal version string in them. The version pattern is used for strlen purposes.

cumulusci/utils/http/multi_request.py
davisagli: Right, makes sense. Should we consider switching to dataclasses once we've dropped 3.6?

cumulusci/utils/http/multi_request.py
davisagli: @prescod I get it now; I was misreading the range passed to startswith as including the version rather than excluding it.

Might be safer to split on / and work with indexed parts so that we don't suddenly break when the API gets to 3 digits.

davisagli:
@prescod a couple of the integration tests failed

cumulusci/utils/http/multi_request.py
prescod: We could have a period of experimenting with DataClasses and seeing whether there are any downsides.

We might discover that there are some cases where we might still like NamedTuples for minor reasons:

1. Immutability can make reasoning slightly easier.
2. We often import "T" and therefore T.NamedTuple can be used with one less import
3. In the Snowfakery context (and perhaps other bulkdata contexts) NamedTuples would probably be less memory and perhaps other perf benefits? 

I mean if you want mutability or the tons of other features of Dataclasses then of course they are better. But if you "would" have used a tuple before either of them existed (simple pairs and triples) and you just want to use names instead of positions, I don't see a harm with using NamedTuples. The forward migration path from NamedTuples to Dataclasses is probably pretty trivial in almost every case, so if your model evolves towards more complexity it isn't hard to switch. But I'm certainly not a NamedTuple zealot and I might find that there's some detail of dataclasses that is so much better that makes them better by default.

davisagli (approved):

davisagli:
@prescod is the remaining integration test failure meaningful?

cumulusci/utils/http/multi_request.py
prescod: Fair enough; 9cd9d1613995810d085984db9a0622a0b25787de

PR #2878, opened 2021-09-16
More descriptive index page for this section

prescod:
Issue: I find this page easier to read if every major topic has a summary on the home page.

Resolution: Now they do.


davisagli (approved):
Good idea

Br4nd0R:
Thanks for this @prescod!

PR #2879, opened 2021-09-16
Feature/case insensitive record types

prescod:
SObject names are generally case insensitive in Salesforce and Snowfakery. Because Snowfakery does not know the "real" names of sobjects, it may generate synthetic record type tables which do not have the same case as the real SObjects. This PR fixes a crash that happens when sobject names use the 'wrong' case.

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -4,6 +4,13 @@ project:
         name: CumulusCI
 
 tasks:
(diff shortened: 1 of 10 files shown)


jstvz:
Closing since it's been more 90 days since the last activity. Feel free to re-open as needed.

PR #2886, opened 2021-09-23
Better header (esp. content-type) handling

prescod:
Changes to our integration test to record Content-Type more accurately.


davisagli (approved):
Do we need to update existing cassettes?

prescod:
Not strictly speaking, because CCI tests don't depend on the bulk API being VCR mocked yet. Snowfakery tests did though (not the whole bulk API but some setup stuff), and they were breaking. If we would rather get the diffs out of the way we could do it now. I don't care much one way or the other.

PR #2887, opened 2021-09-27
Remove terminaltables from CLI table.

jstvz:
This PR replaces the now-archived terminaltables with Rich, and simplifies the CliTable helper class.

---
# Critical Changes

# Changes
- We changed how some the output from some commands and tasks are 
 displayed in the CLI.
# Issues Closed


codexmarcel (dismissed):
🎮

cumulusci/cli/ui.py
```diff
@@ -2,137 +2,78 @@
...
+        orig_box = self._table.box
+        if plain:
+            self._table.box = box.ASCII2
         else:
-            table = str(self)
-        click.echo(table)
+            self._table.box = box_style or orig_box
+        console = Console()
+        console.print(self._table)
+        self._table.box = orig_box
```
davisagli, line 66: I don't see any discernible difference when passing `--plain` to `cci org list` vs not. Small concern in the scheme of things, but am I missing something?

davisagli (commented):
Have we tested the new output on Windows?

davisagli (approved):

cumulusci/cli/ui.py
jstvz: I'm not sure whether the plain option still makes sense. The original use case for the option was windows and for processing a table through stdout, but Rich does the right thing when writing to stdout, and we're using `box.SIMPLE` by default, which works the same across platforms.

jstvz: We'll revisit this as part of the feedback updates.

PR #2888, opened 2021-09-27
Don't raise iteration error in validate_service

jstvz:
# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -112,7 +112,7 @@ def validate_service(options: dict) -> dict:
         # We're checking for a partial-response SSO header and /user/orgs
         # doesn't include one, so we need /user/repos instead.
         repo_generator = gh.repositories()
-        _ = repo_generator.next()
+        _ = next(repo_generator, None)
         repo_response = repo_generator.last_response
         options["scopes"] = get_oauth_scopes(repo_response)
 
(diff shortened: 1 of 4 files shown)


cumulusci/core/tests/cassettes/TestGithub.test_validate_no_repo_exc.yaml
```diff
@@ -0,0 +1,55 @@
...
+          uri: https://api.github.com/user/repos?per_page=100
+      response:
+          body:
+              string: "[]"
+          headers:
+              Content-Type:
+                  - application/json; charset=utf-8
+              X-OAuth-Scopes:
+                  - gist, repo
+              Others: Elided
```
jstvz, line 51: @prescod Regarding our conversation last week, this is the use case I was describing. I'd like sf_vcr to pass any headers I've added myself to the through to the response. Looks like `headers` is getting ignored in its entirety right now.

cumulusci/core/github.py
```diff
@@ -112,7 +112,7 @@ def validate_service(options: dict) -> dict:
         # We're checking for a partial-response SSO header and /user/orgs
         # doesn't include one, so we need /user/repos instead.
         repo_generator = gh.repositories()
-        _ = repo_generator.next()
+        _ = next(repo_generator, None)
```
Br4nd0R, line 115: Don't we want to handle a `StopIteration` exception here?

Br4nd0R (approved):

cumulusci/core/github.py
Br4nd0R: Just kidding, default value is `None` :)

PR #2893, opened 2021-09-28
Remove max-lines option from `cci error info`

Br4nd0R:
This option is likely rarely used if at all. I was thinking it would be more useful than it actually is when I originally implemented the command.

# Critical Changes
- The `--max-lines` option in the `cci error info` command has been deprecated.


davisagli (changes requested):
@Br4nd0R we still want to only print the stack trace, not the entire log, right?

PR #2895, opened 2021-09-28
remove max-lines option

Br4nd0R:
# Critical Changes
* The `--max-lines` option on the `cci error info` command has been deprecated.


davisagli (approved):

davisagli:
> The --max-lines option on the cci error info command has been deprecated.

"Deprecated" implies that it still works but will show a warning and be removed in the future. I would say it's been removed.

PR #2897, opened 2021-09-29
Handle result with status of FATAL_ERROR

Br4nd0R:
# Issues Closed
* The marketing cloud `deploy` task now properly exits when a result status of `FATAL_ERROR` is returned.


cumulusci/tasks/marketing_cloud/deploy.py
```diff
@@ -84,7 +86,7 @@ def _poll_action(self):
         )
         result = safe_json_from_response(response)
         self.logger.info(f"Waiting [{result['status']}]...")
-        if result["status"] == "DONE":
+        if result["status"] in [DEPLOY_FINISHED_STATUS, "FATAL_ERROR"]:
             self.poll_complete = True
             self._validate_response(result)
```
davisagli, line 91: I'd suggest something like this so we can keep `_validate_response` focused on handling the case where the deployment reached DONE status:
```suggestion
        if result["status"] == DEPLOY_FINISHED_STATUS:
            self.poll_complete = True
            self._validate_response(result)
        elif result["status"] == "FATAL_ERROR":
            self.poll_complete = True
            self._report_fatal_error(result)
```

And can we make sure that _report_fatal_error logs the entire `result` (I don't know if it will give us any more information, but if we don't log it we won't find out)

davisagli (approved):

PR #2910, opened 2021-10-06
Specify org in all robot example commands.

umeditor:
# Critical Changes

# Changes

- Specify org in all robot example commands.

# Issues Closed


umeditor:
Going through [the Acceptance Testing with Robot Framework document](https://cumulusci.readthedocs.io/en/latest/robot.html), I ran into a few places where the examples didn't work as no org was specified. This change should allow users to get through the exercises without error.

PR #2925, opened 2021-10-13
Refactor package version handling

davidmreed:
# Critical Changes

# Changes

# Issues Closed
- We fixed an issue causing the `github_release_notes` task to fail on 2GP projects when Issues were closed in a Pull Request.


davidmreed:
> It'd be cool to get dunder methods added to support comparisons, e.g. .parse('1.2.3.5') > ...('1.2.5.5'), but doesn't need to be in this PR

I like this idea a lot - we could then potentially drop most of our uses of `StrictVersion` and `LooseVersion`, which also come with unexpected and subtle behavior differences.

PR #2934, opened 2021-10-18
Filter for `cci task list`

TheBitShepherd:
# Changes
- You can now filter the `cci task list` command to see only relevant results with the `--filter` (`-f`) option.
- `cci task list` has a new `--rainbow` option, because why not 🤷  🌈

### DIFF
--- a/cumulusci/cli/task.py
+++ b/cumulusci/cli/task.py
@@ -2,6 +2,9 @@ import json
 from pathlib import Path
 
 import click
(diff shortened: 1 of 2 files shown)


Br4nd0R:
@jstvz - Meant to open this in draft status, no need to approve. Was wondering if you've given thought to a filter option for `cci task list` like this before? I think it would be even better if we could highlight the filter string in the results.

PR #2941, opened 2021-10-22
Use rich for logging

Br4nd0R:
* We now use a `RichHandler` for logging (and exceptions when logged via `logger.exception(str)`.


jstvz (commented):
Looks good so far. Not approving outright in case you want to try the patch patch below.

Re: MCI/MDP logging... do you have some output to share for those?

---
#### Logging tables

> The only thing I would still like to figure out is if there is a way that we can have CliTable render in a call to logger.info/warn/etc. 

```diff
diff --git a/cumulusci/cli/ui.py b/cumulusci/cli/ui.py
index b6ef124ee..46fa35814 100644
--- a/cumulusci/cli/ui.py
+++ b/cumulusci/cli/ui.py
@@ -7,7 +7,7 @@ Classes:
 import os
 from typing import List
 
-from rich import box, print
+from rich import box, get_console, print
 from rich.console import Console
 from rich.style import Style
 from rich.table import Table
@@ -64,7 +64,7 @@ class CliTable:
             self._table.box = box.ASCII2
         else:
             self._table.box = box_style or orig_box
-        console = Console()
+        console = get_console()
         console.print(self._table)
         self._table.box = orig_box
 
@@ -73,10 +73,11 @@ class CliTable:
         return self._table
 
     def __str__(self):
-        console = Console()
-        with console.capture() as capture:
-            console.print(self._table)
-        return capture.get()
+        from io import StringIO
+        console = Console(file=StringIO())
+        tab_width = console.size.width - 21
+        console.print(self._table, width=tab_width)
+        return console.file.getvalue()
 
 
 def _soql_table(results, truncated):
diff --git a/cumulusci/tasks/salesforce/composite.py b/cumulusci/tasks/salesforce/composite.py
index ffd8ab338..a5a4f724e 100644
--- a/cumulusci/tasks/salesforce/composite.py
+++ b/cumulusci/tasks/salesforce/composite.py
@@ -122,7 +122,7 @@ Example Task Definition
         return status_code >= 200 and status_code < 300
 
     def _log_exception_message(self, subrequests):
-        table_data = [["ReferenceId", "Message"]]
+        table_data = [["ReferenceId", "[green]Message[/]"]]
         table_data.extend(
             [sub["referenceId"], body["message"]]
             for sub in subrequests
@@ -133,6 +133,9 @@ Example Task Definition
             table_data,
         )
         self.logger.error("The request failed with the following message(s):\n\n")
+        self.logger.info(table.table)
+        self.logger.info(str(table))
+        self.logger.info(table)
         table.echo()
 
     def _log_success_message(self, subrequests):
```

...emits the following lines:
```console
❯ cctr deploy_hed_data_1 --org dev
[11/01/21 12:19:38] Getting org info from Salesforce CLI for test-98aoqryfc7ru@example.com
[11/01/21 12:19:40] Beginning task: CompositeApi
                    As user: test-98aoqryfc7ru@example.com
                    In org: 00DP00000041NJ7

                    Processing datasets/hed/accounts.json
                    The request failed with the following message(s):


                    <rich.table.Table object at 0x10990aa90>

                      ReferenceId   Message
                     ────────────────────────────────────────────────────────────────────────────────────
                      uni           Invalid reference specified. No value for schoolRt.records[0].Id
                                    found in schoolRt. Provided referenceId ('schoolRt.records[0].Id')
                                    must start with a letter or a number, and it can contain only
                                    letters, numbers and underscores ('_').



                      ReferenceId   Message
                     ────────────────────────────────────────────────────────────────────────────────────
                      uni           Invalid reference specified. No value for schoolRt.records[0].Id
                                    found in schoolRt. Provided referenceId ('schoolRt.records[0].Id')
                                    must start with a letter or a number, and it can contain only
                                    letters, numbers and underscores ('_').



  ReferenceId   Message
 ─────────────────────────────────────────────────────────────────────────────────────────────────────────
  uni           Invalid reference specified. No value for schoolRt.records[0].Id found in schoolRt.
                Provided referenceId ('schoolRt.records[0].Id') must start with a letter or a number, and
                it can contain only letters, numbers and underscores ('_').
```
---

cumulusci/cli/logger.py
```diff
@@ -4,8 +4,9 @@
 import sys
 import tempfile
 
-import coloredlogs
+import coloredlogs  # noqa: F401
```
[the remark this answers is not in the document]
Br4nd0R: will check

cumulusci/cli/cci.py
```diff
@@ -71,36 +72,46 @@ def main(args=None):
...
             should_show_stacktraces = runtime.universal_config.cli__show_stacktraces
 
-            init_logger(log_requests=debug)
+            init_logger(debug=debug)
             # Hand CLI processing over to click, but handle exceptions
             try:
                 cli(args[1:], standalone_mode=False, obj=runtime)
             except click.Abort:  # Keyboard interrupt
-                show_debug_info() if debug else click.echo("\nAborted!", err=True)
+                console = Console()
```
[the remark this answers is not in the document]
Br4nd0R: In this instance it's ok to make that switch, but the console that is fetched in `handle_exception()` needs `stderr=True` as a kwarg, which `get_console()` doesn't provide us with.

Br4nd0R:
@jstvz - Here is a screenshot I took from my local MetaDeploy. It should look the same as it does currently.
Screen Shot 2021-10-28 at 9 51 54 AM

cumulusci/cli/logger.py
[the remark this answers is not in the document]
Br4nd0R: looks like it

cumulusci/cli/org.py
```diff
@@ -258,9 +259,9 @@ def calculate_org_days(info):
 def org_info(runtime, org_name, print_json):
     org_name, org_config = runtime.get_org(org_name)
     org_config.refresh_oauth_token(runtime.keychain)
-
+    console = Console()
     if print_json:
-        click.echo(
+        rich.print_json(
```
jstvz, line 264: There's an open issue at willmcgugan/rich#1591. Should we hold off?

cumulusci/core/sfdx.py
```diff
@@ -65,6 +66,30 @@ def sfdx(
...
+    """Pretty log option key/value pairs given to sfdx commands.
+    Outputs one option per line."""
+    if not options:
+        logger.info("(No command options)")
+        return
+
+    logger.info("Options:")
+    option_str = ""
+    for idx, value in enumerate(options):
+        if "-" in value:
```
jstvz, line 79: What happens if we pass `force-app`?

jstvz (changes requested):
Would you split the sfdx options logging into a separate PR?

A couple of questions, but aside from that, LGTM.

cumulusci/cli/org.py
Br4nd0R: I feel ok because it looks like it's only an issue when piping to other things (which I know you'd probably be bummed about). However, printing of JSON itself seems to be working. What are your thoughts?

jstvz: Since piping  machine-readable output to `jq` and other tools is the primary use case  for this option, I'm leaning towards being conservative. (That said, piping to jq is already broken because we're logging to stdout.)

---

```console
$ cci org info --json dev | tail -n +2 | jq
parse error: Invalid string: control characters from U+0000 through U+001F must be escaped at line 3, column 3
```

Same command works on main. The `access_token` has a length of 133 characters (including indent), while `print_json` truncates this line to the width of my terminal, 107 characters.

jstvz: Ah, never mind, `console.print(json_str, soft_wrap=True)` fixes truncation. I'll push a fix upstream.

Br4nd0R: Odd, this command works for me on both this branch and main. I'm seeing my access token with 113 chars on each command.

jstvz: Sure, but how wide is your terminal? Repro:
```python-console
>>> from rich import print_json, get_console
>>> import json
>>> test_dict = {'access_token': '................................................................................................................', 'scratch': 100, 'set_password': False}
>>> c = get_console()
>>> c.width
107
>>> rich.print_json(json.dumps(test_dict))
{
  "access_token": "........................................................................................
  "scratch": 100,
  "set_password": false
}
```

Br4nd0R: Ahh, ok I can reproduce that with a small window. I'll go back to printing json with `click.echo()` for the time being and we can keep an eye on the bug.

jstvz (approved):

requirements/prod.txt
```diff
@@ -20,8 +20,6 @@ click==7.1.2
     #   snowfakery
 colorama==0.4.4
     # via rich
-coloredlogs==15.0.1
-    # via -r requirements/prod.in
```
davisagli, line 24: @Br4nd0R Note for next MetaCI update: MetaCI uses coloredlogs (https://github.com/SFDO-Tooling/MetaCI/blob/922df3c24d9d8661c50a241dde7e4b580a823cd0/metaci/cumulusci/logger.py#L8) but doesn't explicitly include it as a dependency. Now that it won't get installed as as transitive dependency of cumulusci, MetaCI will either need to explicitly require it (in prod.in) or stop using it.

PR #2942, opened 2021-10-25
First cut at the "click refresh button" task, implemented in robot.

davidmreed:
We don't have a way to test it yet, but this gives us a good starting
point once we do. If you try to run it, though, it should fail in the
correct way saying it's not on the page it expects to be on.

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -5,11 +5,6 @@ project:
         name: CumulusCI
 
 tasks:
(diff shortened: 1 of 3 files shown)


jstvz:
Closing since it's been more 90 days since the last activity. Feel free to re-open as needed.

PR #2946, opened 2021-10-26
Add 'cci plan info' command

boakley:
see embedded FIXME in plan.py for issues that need to be resolved.

@jstvz this is just a draft PR, but I want to make sure I'm heading in the right direction before getting much more done. My main question is whether or not I'm computing the steps correctly, and whether or not I'm computing the step preflight table correctly.

I've manually tested this in the SAL repository, and it at least displays output without crashing. I don't know if the output is correct since I'm not at all familiar with the plans and steps. 

# Critical Changes

# Changes
- We added a new command, `cci plan info`. Similar to `cci task info`, this command displays detailed information about a MetaDeploy `plan`, and includes a `--messages` option to display user-facing text.
# Issues Closed


jstvz (approved):
Thanks, @boakley. I took a stab at editing the change note, so take a look and tweak as needed.

cumulusci/utils/metadeploy.py
```diff
@@ -0,0 +1,24 @@
...
+        if step.skip:
+            continue
+        with cd(step.project_config.repo_root):
+            task = step.task_class(
+                step.project_config,
+                TaskConfig(step.task_config),
+                name=step.task_name,
+            )
+            steps.extend(task.freeze(step))
+    return steps
```
davisagli, line 24: @boakley @jstvz To avoid circular import problems, cumulusci/utils is designated for utility modules that stand alone and do not have dependencies on other parts of CumulusCI. This module would belong better in cumlusci/core somewhere.

davisagli: @boakley I think it makes sense to stay as a new, separate module, cumulusci/core/metadeploy.py. It's a different domain of functionality than the existing functions in cumulusci/core/utils.py

PR #2954, opened 2021-11-02
Pretty log sfdx command options

Br4nd0R:
# Changes
* The `sfdx()` method now neatly logs command option/value pairs below the sfdx command.


cumulusci/core/sfdx.py
```diff
@@ -35,19 +35,20 @@ def sfdx(
...
         for arg in args:
-            command += " " + shell_quote(arg)
+            sfdx_cmd += " " + shell_quote(arg)
     if username:
-        command += f" -u {shell_quote(username)}"
+        sfdx_cmd += f" -u {shell_quote(username)}"
     if log_note:
-        logger.info(f"{log_note} with command: {command}")
+        logger.info(f"{log_note} with command: sfdx {command}")
+        plog_command_options(args)
```
jstvz, line 46: Your indentation has this beneath the conditional, so it will only run when log_note is passed. We're only doing this in a handful of places, so it won't run for most commands. Is this what you want?

cumulusci/core/sfdx.py
```diff
@@ -65,6 +66,30 @@ def sfdx(
     return p
 
 
+def plog_command_options(options: T.List[str]) -> None:
```
jstvz, line 69: How about using `args` or `arguments` instead of options to maintain consistency with the `args` parameter above?

jstvz (commented):
This is trickier than it appears. From a scan of the references, most callers build the command using an f-string:

https://github.com/SFDO-Tooling/CumulusCI/blob/46ae50dcbe404f443fe9da7d70074cb398ab47a1/cumulusci/core/config/sfdx_org_config.py#L169

Without using `shlex.split` or its equivalent we won't have any options (args) to print. The other tricky bit is that the most common way to invoke SFDX, `SFDXOrgTask`, doesn't use `sfdx` at all, it inherits from `SFDXBaseTask`:

https://github.com/SFDO-Tooling/CumulusCI/blob/46ae50dcbe404f443fe9da7d70074cb398ab47a1/cumulusci/tasks/sfdx.py#L23-L40

...which inherits from `Command`, which has its own `sarge.Command`:

https://github.com/SFDO-Tooling/CumulusCI/blob/46ae50dcbe404f443fe9da7d70074cb398ab47a1/cumulusci/tasks/command.py#L108-L122

(🐢³)

The path forward here isn't immediately clear to me, so we should probably discuss this before we meet next week.

Br4nd0R:
@jstvz - Thanks for the explanation on complexity here. I think this breaks cleanly into two chunks of work:

#. Make the pretty logging functionality work when options are present in the command (as a single string) in the `sfdx()` method.
#. Make calls that go through `SFDXBaseTask` work for pretty printing their command/options.

PR #2955, opened 2021-11-02
Add error msg when retrieving lightweight tag

jstvz:
When attempting to publish a lightweight tag with `metadeploy_publish` a 404 is raised. This commit adds a message informing the user that a lightweight tag was passed, and refactors the `metadeploy_publish` task to use the `get_tag_by_name` method from core.github.

# Critical Changes

# Changes

# Issues Closed


cumulusci/core/tests/test_github.py
```diff
@@ -329,6 +329,28 @@ def test_get_tag_by_name__404(self, repo):
...
+        responses.add(
+            "GET",
+            "https://api.github.com/repos/TestOwner/TestRepo/git/tags/tag_SHA",
+            json=self._get_expected_tag("beta/1.0", "tag_SHA"),
+            status=404,
+        )
+        with pytest.raises(DependencyLookupError) as exc:
+            get_tag_by_name(repo, "tag_SHA")
+
+        assert "not an annotated tag" in str(exc)
```
Br4nd0R, line 352: I like this better than having to write out a larger string to match the whole error message with the `match` kwarg for `pytest.raises()`.

Br4nd0R (approved):
LGTM!

davisagli (commented):
Could we simply use the commit that is referenced by the lightweight tag? Or is there some reason we _need_ an annotated tag?

jstvz:
We're extracting a version number via `parse_tag`, so: kinda.

Looks like he assumption is that `tag` is a release tag. There's a decent chance that a lightweight tag isn't correctly formatted and would fail in `_find_or_create_version()`.

We should either accept lightweight tags, or update the option description to make the need for a *release* tag explicit.

PR #2956, opened 2021-11-02
fix log handler for urllib3

Br4nd0R:
# Changes
* Fixes an issue where a log handler wasn't being properly applied to `urllib3` when the `--debug` flag is passed.


jstvz (approved):
Tested OK

PR #2962, opened 2021-11-08
Snowfakery 2.2 support in CCI

prescod:
# Changes

[Snowfakery 2.2](https://github.com/SFDO-Tooling/Snowfakery/releases/tag/v2.2) is now included with CumulusCI. New features include unique IDs and Numeric Counters.

### DIFF
--- a/cumulusci/tasks/bulkdata/generate_from_yaml.py
+++ b/cumulusci/tasks/bulkdata/generate_from_yaml.py
@@ -64,6 +64,7 @@ class GenerateDataFromYaml(BaseGenerateDataTask):
 
     def __init__(self, *args, **kwargs):
         self.vars = {}
+        self.plugin_options = {}
         super().__init__(*args, **kwargs)
 
     def _init_options(self, kwargs):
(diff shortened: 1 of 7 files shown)


cumulusci/tasks/bulkdata/generate_from_yaml.py
```diff
@@ -27,6 +27,9 @@ class GenerateDataFromYaml(BaseGenerateDataTask):
         "vars": {
             "description": "Pass values to override options in the format VAR1:foo,VAR2:bar"
         },
+        "plugin_options": {
+            "description": "Pass values to override plugin options in the format VAR1:foo,VAR2:bar"
```
davisagli, line 31: Don't you need to run this through process_list_of_pairs_dict_arg in order to support that format on the CLI in addition to mappings in yaml?

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -503,6 +507,7 @@ def data_generator_opts(self, working_dir, *args, **kwargs):
             "continuation_file": wd.continuation_file,
             "num_records_tablename": self.run_until.sobject_name or COUNT_REPS,
             "vars": self.recipe_options,
+            "plugin_options": plugin_options,
```
davisagli, line 510: Just wanted to double check that it's intentional this is configurable in the old task but hardcoded in the new one.

cumulusci/tasks/bulkdata/generate_from_yaml.py
prescod: Good point. I was only thinking of this class as an implementation detail of the other one, but of course it is its own task that people can and do use. Fixed in 5e7182fa2

cumulusci/tasks/bulkdata/snowfakery.py
prescod: It's configurable in the old task as an implementation detail of making it available to the new one. Making it configurable here is a reasonable feature request and I can do that before the next CCI release but probably shouldn't scope creep today.

davisagli (approved):

PR #2971, opened 2021-11-09
Week 45 Updates

Br4nd0R:


cumulusci/tasks/bulkdata/step.py
```diff
@@ -75,7 +75,7 @@ def download_file(uri, bulk_api):
         resp = requests.get(uri, headers=bulk_api.headers(), stream=True)
         resp.raise_for_status()
         f = os.fdopen(handle, "wb")
-        for chunk in resp.iter_content(chunk_size=None):
```
Br4nd0R, line 78: @davisagli - cumulusci/tasks/tests/test_step.py::TestDownloadFile::test_download_file() was hanging after updates to the `responses` package. @codexmarcel and I looked at [the source code](https://github.com/psf/requests/blob/b0e025ade7ed30ed53ab61f542779af7e024932e/requests/models.py#L780)for this method and suspect this might be a bug? Having a chunk size of none doesn't seem to make sense.

cumulusci/tests/test_schema.py
```diff
@@ -13,14 +13,15 @@ class TestSchema:
...
+        from yaml import Loader
 
         schemapath = (
             cumulusci_test_repo_root / "cumulusci/schema/cumulusci.jsonschema.json"
         )
         with open(schemapath) as f:
             schema = json.load(f)
         with open(cumulusci_test_repo_root / "cumulusci.yml") as f:
-            data = yaml.load(f)
+            data = yaml.load(f, Loader=Loader)
```
Br4nd0R, line 24: @prescod - It looks like a `Loader` is [now required](https://pyyaml.org/wiki/PyYAMLDocumentation) here? This seems quite silly to impose on end users.

cumulusci/tasks/bulkdata/step.py
davisagli: The docs for requests say: "A value of None will function differently depending on the value of stream. stream=True will read data as it arrives in whatever size the chunks are received." So the use of None is intentional here to make sure that we read in all data that has been received so far, not just one byte at a time (chunk_size defaults to 1). I wouldn't like to make this change without checking its impact on performance.

This may need further investigation to figure out if there's a bug in responses that we should report.

cumulusci/tests/test_schema.py
davisagli: Shouldn't we be using `yaml.safe_load` here?

requirements/prod.txt
```diff
@@ -140,7 +168,7 @@ salesforce-bulk==2.2.0
     # via -r requirements/prod.in
 sarge==0.1.6
     # via -r requirements/prod.in
-selenium==3.141.0
+selenium==4.0.0
```
davisagli, line 171: @Br4nd0R Let's not update selenium until @boakley has a chance to evaluate it.

davisagli: Actually we clearly can't update to selenium 4 without further work. For one thing, it no longer supports Python 3.6 (you can see those builds failed). Also all of the robot tests are failing.

cumulusci/tasks/bulkdata/step.py
Br4nd0R: I'll take a look this morning to see if I can repro in a simpler script.

requirements/prod.txt
```diff
@@ -154,9 +163,9 @@ six==1.16.0
     #   python-dateutil
     #   salesforce-bulk
     #   snowfakery
-snowfakery==2.1
+snowfakery==2.2
```
davisagli, line 166: This needs to happen along with corresponding changes in #2962 right @prescod?

cumulusci/tasks/bulkdata/step.py
Br4nd0R: Here is some (slightly) simplified code to repro the the failure. When debugging I'm seeing that [this while loop never exits](https://github.com/urllib3/urllib3/blob/fac7db19a1e8e15e5b302e2203ce35cd0485bc88/src/urllib3/response.py#L726).
```
import os
import pathlib
import requests
import responses
import tempfile


def download_file(uri, handle, chunk_size):
    """Download the Bulk API result file for a single batch,
    and remove it when the context manager exits."""
    resp = requests.get(uri, headers={}, stream=True)
    resp.raise_for_status()
    f = os.fdopen(handle, "wb")

    for chunk in resp.iter_content(chunk_size=chunk_size):
        f.write(chunk)

    f.close()
    return f


@responses.activate
def test_download_file__chunk_size_one():
    url = "https://example.com"
    responses.add(method="GET", url=url, body=b"TEST\xe2\x80\x94")

    (handle, path) = tempfile.mkstemp(text=False)
    f = download_file(url, handle, chunk_size=1)
    pathlib.Path(path).unlink()


@responses.activate
def test_download_file__chunk_size_none():
    url = "https://example.com"
    responses.add(method="GET", url=url, body=b"TEST\xe2\x80\x94")

    (handle, path) = tempfile.mkstemp(text=False)
    f = download_file(url, handle, chunk_size=None)
    pathlib.Path(path).unlink()
```

davisagli: @Br4nd0R I wonder if that is fixed by this change in responses that hasn't been released yet: https://github.com/getsentry/responses/pull/422

cumulusci/tests/test_schema.py
prescod: This is just a test file but yeah, yaml.safe_load will probably fix the error.

requirements/prod.txt
prescod: Sorry yes, I got focused on the Windows installer work and it slipped my mind that I need to finish #2962 . I'll switch to that now. @Br4nd0R

Br4nd0R: Per my discussion with @prescod, I'm pinning snowfakery to 2.1 and Paul will undo the pin on a subsequent PR.

cumulusci/tasks/bulkdata/step.py
Br4nd0R: Pinning to 0.14.0 for now.

davisagli (approved):
Approving, assuming builds pass

PR #2974, opened 2021-11-10
Add a `--version` option

Br4nd0R:
Just looking to be kind to our users here. This is a common option across CLIs and it should work just the same as the `cci version` command.

# Changes
* Version information for CumulusCI can now be obtained through a `--version` option via `cci --version`. This output is the same as the `cci version` command.


cumulusci/cli/cci.py
```diff
@@ -136,34 +137,41 @@ def show_debug_info():
...
         )
-        click.echo(f"To upgrade, run `{get_cci_upgrade_command()}`")
-        click.echo(
+        console.print(f"To upgrade, run `{get_cci_upgrade_command()}`")
+        console.print(
             f"Release notes: https://github.com/SFDO-Tooling/CumulusCI/releases/tag/v{str(latest_version)}"
         )
     else:
-        click.echo("You have the latest version of CumulusCI.")
+        console.print("You have the latest version of CumulusCI.")
```
boakley, line 161: 👍 I like that you not only get the version number, but you also get an assurance that it's the latest version.

cumulusci/cli/tests/test_cci.py
```diff
@@ -396,23 +396,24 @@ def test_cli():
     "cumulusci.cli.cci.get_latest_final_version",
     mock.Mock(return_value=pkg_resources.parse_version("100")),
 )
-@mock.patch("click.echo")
-def test_version(echo):
+def test_version(capsys):
     run_click_command(cci.version)
-    assert cumulusci.__version__ in echo.call_args_list[1][0][0]
+    console_output = capsys.readouterr().out
+    assert cumulusci.__version__ in console_output
```
boakley, line 402: It might be good to make this a bit more precise, such as `assert f"CumulusCI version:  {cumulusci.__version__)" in  console_output`. While it seems highly unlikely, if someone did a refactor late at night they could accidentally swap version numbers between python and cci and this test wouldn't catch it.

cumulusci/cli/cci.py
```diff
@@ -136,34 +137,41 @@ def show_debug_info():
...
+    console.print()
 
-    click.echo()
     current_version = get_installed_version()
     latest_version = get_latest_final_version()
     if latest_version > current_version:
-        click.echo(
-            f"There is a newer version of CumulusCI available ({str(latest_version)})."
+        console.print(
+            f"[yellow flashing]There is a newer version of CumulusCI available ({str(latest_version)})."
```
boakley, line 154: I don't see a test for this line of code.

boakley (changes requested):

cumulusci/cli/cci.py
jstvz, line 154: We should avoid flashing text:

> Flashing is generally a bad idea. It can cause all sorts of issues, from seizures to motion sickness. If you absolutely must have a flashing element there are a few things to consider.

https://accessibility.18f.gov/flashing/

```suggestion
            f"[yellow]There is a newer version of CumulusCI available ({str(latest_version)})."
```

---

💭: we'll want to set up shared styles so warnings and errors are consistent across commands.

cumulusci/cli/tests/test_cci.py
Br4nd0R: Ya this could definitely be made better. I always wonder if there is a "best practice" for assertions on lots of different string that could potentially be displayed to a user during program execution.

cumulusci/cli/cci.py
Br4nd0R: We don't have an explicit assert for this string, but we do have coverage of this if block by mocking `get_latest_final_version` in the [test_version](https://github.com/SFDO-Tooling/CumulusCI/blob/344c6913a70d0a7d71ecced047653c0d7383e974/cumulusci/cli/tests/test_cci.py#L399) method.

jstvz (changes requested):
Good improvement.  There are accessibility issues, though.

cumulusci/cli/cci.py
```diff
@@ -136,34 +137,41 @@ def show_debug_info():
...
-def version():
-    click.echo("CumulusCI version: ", nl=False)
-    click.echo(click.style(cumulusci.__version__, bold=True), nl=False)
-    click.echo(f" ({sys.argv[0]})")
-    click.echo(f"Python version: {sys.version.split()[0]}", nl=False)
-    click.echo(f" ({sys.executable})")
+def show_version_info():
+    console = rich.get_console()
+    console.print(
+        f"[blue bold]CumulusCI version: [blink] {cumulusci.__version__}[/] ({sys.argv[0]})"
```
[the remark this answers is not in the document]
Br4nd0R: I was just messing around with the console markup and actually meant to remove this stuff. I actually couldn't get `blink` to work for me 🤷 .

cumulusci/cli/cci.py
Br4nd0R: I don't think `flashing` was valid console markup anyway 😂 .

Br4nd0R: The `test_cci.py::test_version()` method does cover this if block, but probably a good call to add an assert for at least one string in this block.

cumulusci/cli/cci.py
```diff
@@ -136,34 +137,47 @@ def show_debug_info():
...
+        f"https://github.com/SFDO-Tooling/CumulusCI/releases/tag/v{latest_version}"
+    )
+    console = rich.get_console()
+    console.print(
+        f"See the latest CumulusCI Release Notes: [link={release_notes_link}]{release_notes_link}[/link]"
+    )
+
+
 @click.group("main", help="")
+@click.version_option(show_version_info(), message="")
```
davisagli, line 173: This doesn't look right. It runs show_version_info() at import time, so it is displayed for every invocation of cci whether or not `--version` was specified.

davisagli (changes requested):

cumulusci/cli/cci.py
Br4nd0R: good catch 👍

davisagli (approved):
LGTM

boakley (approved):
Looks good to me

jstvz (approved):

PR #2976, opened 2021-11-12
v3.48.1: Remove backports-datetime

prescod:
Remove an unneeded dependency which caused installation problems.


jstvz (dismissed):

cumulusci/version.txt
```diff
@@ -1 +1 @@
-3.48.0
+3.48.0.dev0
```
davisagli, line 1:
```suggestion
3.48.1.dev0
```

3.48.0.dev0 is less than 3.48.0

davisagli (changes requested):

jofsky (approved):

PR #2980, opened 2021-11-15
Fix manifest, and run unit tests using sdist

davisagli:
# Critical Changes

# Changes

# Issues Closed


tox.ini
```diff
@@ -1,13 +1,15 @@
...
+[gh-actions]
+python =
+    3.6: py36
+    3.7: py37
+    3.8: py38
+    3.9: py39
+    3.10: py310
 
+[testenv]
+commands = coverage run --source={envsitepackagesdir}/cumulusci {envbindir}/pytest {envsitepackagesdir}/cumulusci []
```
jstvz, line 13: FYI: Something's off with the `windows-latest` builds... workflow passes, but the install command is failing because pytest isn't in the path. Do we need something like [this](https://github.com/ymyzk/tox-gh-actions/blob/master/ARCHITECTURE.md#using-factors-to-decide-environments), or is it just a PATH issue?

davisagli: @jstvz thanks for pointing that out; I'll see what I can figure out

davisagli: @jstvz I'm trying a switch to pytest-cov https://github.com/SFDO-Tooling/CumulusCI/pull/2980/commits/33544e96b519c2869dc2b9be24a8bea7c8c0bf3b

PR #2981, opened 2021-11-15
Fix `dx_convert_from`

Br4nd0R:
* We now always remove the `src/` folder and its contents before calling `sfdx force:source:convert -d src`.

# Critical Changes
* The `dx_convert_from` task now uses a custom Python task class instead of `cumulusci.tasks.sfdx.SFDXBaseTask`. In most cases this will have no visible impact, but if you have customized this task config to change the command option, make sure you have also explicitly set its `class_path` to `cumulusci.tasks.sfdx.SFDXBaseTask`.

# Issues Closed
* Fixed an issue where deleted components could potentially still deploy into persistent orgs.


Br4nd0R:
@davisagli - This technically meets the AC for the story, but wondering if there are scenarios that @davidjray and I may have overlooked. If this approach doesn't work, our next option could be to make `dx_convert_from` and first-class task and perform the needed logic before the call to `force:source:convert`.

davisagli:
@Br4nd0R @davidjray It's clever (and I learned something new; I didn't realize it was possible to run multiple commands with the Command task like this). But it won't work on Windows (no rm command). I think it'd be best to give dx_convert_from its own task class.

cumulusci/tasks/tests/test_dx_convert_from.py
```diff
@@ -0,0 +1,67 @@
...
+    assert not dir_structure.exists()
+    assert not src_file1_path.exists()
+    assert not src_file2_path.exists()
+    sarge.Command.assert_called_once_with(
+        "sfdx force:source:convert -d src",
+        cwd=".",
+        env=ANY,
+        shell=True,
+        stdout=ANY,
+        stderr=ANY,
```
Br4nd0R, line 65: Today, we learned about [ANY](https://docs.python.org/3/library/unittest.mock.html#any) :)

cumulusci/cumulusci.yml
```diff
@@ -239,9 +239,9 @@ tasks:
         group: Salesforce DX
     dx_convert_from:
         description: Converts force-app directory in sfdx format into metadata format under src
-        class_path: cumulusci.tasks.sfdx.SFDXBaseTask
+        class_path: cumulusci.tasks.dx_convert_from.DxConvertFrom
         options:
-            command: "force:source:convert -d src"
+            command: "force:source:convert"
```
davisagli, line 244: I don't think we can make this change to add a separate src_dir option in a backwards-compatible way. Someone may have put this in a project cumulusci.yml:

```
tasks:
    dx_convert_from:
        options:
            command: "force:source:convert -d custom_path"
```

and with the new implementation they would end up with 2 paths specified in the sfdx command.

davisagli: Ah, but now I see how it has to be separate in order to easily find it so we know what path to delete. Okay, let's go ahead and make that change but be sure to call it out in the critical changes section of the release notes.

cumulusci/tasks/dx_convert_from.py
```diff
@@ -0,0 +1,41 @@
...
+
+from cumulusci.tasks.sfdx import SFDXBaseTask
+
+
+class DxConvertFrom(SFDXBaseTask):
+    task_options = {
+        "command": {
+            "description": "The full command to run with the sfdx cli.",
+            "required": True,
+        },
```
davisagli, line 11: I think we may as well remove the `command` option since we're making another backwards-incompatible change anyway

cumulusci/tasks/dx_convert_from.py
```diff
@@ -0,0 +1,41 @@
...
+
+        # append the -d option to sfdx force:source:convert
+        self.options[
+            "command"
+        ] = f"{self.options['command']} -d {self.options['src_dir']}"
+
+    def _run_task(self):
+        src_dir = Path(self.options["src_dir"])
+        if src_dir.exists():
+            self._clear_directory(src_dir)
```
davisagli, line 31: You can use [shutil.rmtree](https://docs.python.org/3/library/shutil.html#shutil.rmtree) for this (I _think_ the sfdx command will create the src directory if it doesn't exist)

cumulusci/tasks/tests/test_dx_convert_from.py
```diff
@@ -0,0 +1,67 @@
...
+    )
+    Path("src").rmdir()
```
davisagli, line 67: The problem with doing cleanup this way is that is won't happen if the test fails before this line. I tend to like using the `temporary_dir()` context manager for building test data on the filesystem so that it gets automatically deleted if the test exits early.

cumulusci/tasks/dx_convert_from.py
Br4nd0R: If we remove the `command` option, from `task_options`, then would we just keep the current logic in `_init_options()` but now hardcode a value for `self.options["command"] = f"force:source:convert -d {self.options['src_dir']}"` so that the parent tasks can access that value?

davisagli: Yeah, that's basically what I had in mind

Br4nd0R:
I'm not sure if the wording in the "Critical Changes" section encompasses everything we want to say. Feel free to edit.

cumulusci/cumulusci.yml
davisagli: @Br4nd0R we should remove the `command` option here

cumulusci/tasks/dx_convert_from.py
```diff
@@ -0,0 +1,27 @@
...
+        "extra": {"description": "Append additional options to the command"},
+        "src_dir": {
+            "description": "The path to the src directory where converted contents will be stored. Defaults to src/"
+        },
+    }
+
+    def _init_options(self, kwargs):
+        super()._init_options(kwargs)
+        if "src_dir" not in self.options:
+            self.options["src_dir"] = "src"
```
davisagli, line 18: I would put this default in cumulusci.yml instead of Python so that it shows up in `cci task info dx_convert_from`

davisagli (changes requested):
Added a couple more small requests

cumulusci/tasks/tests/test_dx_convert_from.py
```diff
@@ -0,0 +1,62 @@
...
+    universal_config = UniversalConfig()
+    project_config = BaseProjectConfig(universal_config, config={"no_yaml": True})
+    project_config.project__name = "TestProject"
+    return project_config
+
+
+@pytest.fixture
+def task_config():
+    return TaskConfig(
+        {"options": {"command": "force:source:convert", "src_dir": "src"}}
```
davisagli, line 26: @Br4nd0R shouldn't need the command, right?

davisagli (approved):

PR #2992, opened 2021-11-16
Centralize XML parsing and avoid XXE

prescod:
Per the Python documentation, XXE is not a problem for the stdlib: 

https://docs.python.org/3/library/xml.html#xml-vulnerabilities

lxml does have an XXE problem:

https://lxml.de/FAQ.html#how-do-i-use-lxml-safely-as-a-web-service-endpoint

### DIFF
--- a/cumulusci/tasks/bulkdata/step.py
+++ b/cumulusci/tasks/bulkdata/step.py
@@ -9,13 +9,13 @@ from contextlib import contextmanager
 from enum import Enum
 from typing import Any, Dict, List, NamedTuple, Optional
 
-import lxml.etree as ET
 import requests
 
 from cumulusci.core.exceptions import BulkDataException
 from cumulusci.core.utils import process_bool_arg
 from cumulusci.tasks.bulkdata.utils import iterate_in_chunks
 from cumulusci.utils.classutils import namedtuple_as_simple_dict
+from cumulusci.utils.xml import elementtree_parse_string
 
 DEFAULT_BULK_BATCH_SIZE = 10_000
 DEFAULT_REST_BATCH_SIZE = 200
@@ -96,9 +96,9 @@ class BulkJobMixin:
         response.raise_for_status()
         return self._parse_job_state(response.content)
 
-    def _parse_job_state(self, xml):
+    def _parse_job_state(self, xml: str):
         """Parse the Bulk API return value and generate a summary status record for the job."""
-        tree = ET.fromstring(xml)
+        tree = elementtree_parse_string(xml)
         statuses = [el.text for el in tree.iterfind(".//{%s}state" % self.bulk.jobNS)]
         state_messages = [
             el.text for el in tree.iterfind(".//{%s}stateMessage" % self.bulk.jobNS)
--- a/cumulusci/tasks/bulkdata/step.py
+++ b/cumulusci/tasks/bulkdata/step.py
@@ -15,7 +15,7 @@ from cumulusci.core.exceptions import BulkDataException
 from cumulusci.core.utils import process_bool_arg
 from cumulusci.tasks.bulkdata.utils import iterate_in_chunks
 from cumulusci.utils.classutils import namedtuple_as_simple_dict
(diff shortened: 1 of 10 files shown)


cumulusci/tasks/bulkdata/step.py
```diff
@@ -9,13 +9,13 @@
...
 from typing import Any, Dict, List, NamedTuple, Optional
 
-import lxml.etree as ET
 import requests
 
 from cumulusci.core.exceptions import BulkDataException
 from cumulusci.core.utils import process_bool_arg
 from cumulusci.tasks.bulkdata.utils import iterate_in_chunks
 from cumulusci.utils.classutils import namedtuple_as_simple_dict
+from cumulusci.utils.xml import elementtree_parse_string
```
davisagli, line 18: You've replaced lxml with xml.etree here

cumulusci/tasks/marketing_cloud/api.py
```diff
@@ -1,7 +1,7 @@
 import requests
-from lxml import etree
 
 from cumulusci.salesforce_api import mc_soap_envelopes as envelopes
+from cumulusci.utils.xml import elementtree_parse_string
```
davisagli, line 4: You've replaced lxml with xml.etree here

cumulusci/utils/xml/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def elementtree_parse_file(path: T.Union[str, Path]) -> etree.ElementTree:
+    """Parse a file from filename or Path using Python stdlib"""
+    try:
+        tree = etree.parse(path)
+    except etree.ParseError as err:
+        err.filename = path
+        raise err
+    assert isinstance(tree, etree.ElementTree), type(tree)
```
davisagli, line 17: When is it not a tree?

cumulusci/utils/xml/metadata_tree.py
```diff
@@ -37,15 +38,15 @@ def parse(source):
     """
     if hasattr(source, "open"):  # for pathlib.Path objects
         with source.open() as stream:
-            doc = etree.parse(stream)
+            doc = lxml_parse_file(stream)
```
davisagli, line 41: the type annotations for lxml_parse_file say it accepts a str or a Path; looks like that should also include some sort of file-like?

davisagli (changes requested):
@prescod just wanted to check about a few things...

cumulusci/utils/xml/metadata_tree.py
prescod: Updated and tested.

prescod: b105bd070

cumulusci/utils/xml/__init__.py
prescod: b105bd070

cumulusci/tasks/bulkdata/step.py
davisagli: @prescod hmm. What's the motivation? The two etrees are not exactly the same and my impression has been that lxml provides a more ergonomic and complete API, with an added bonus of speed. I think I'd need to take a much closer look at the differences and at whether we are exposing lxml internals anywhere that others might have built on.

In any case it feels like a slightly riskier change and might be good to do separately from this one.

prescod: I'll back it out, but I think that using two different APIs is just slightly confusing, it's another dependency (a binary one!) and so forth.

Based just on guesstimating the implementation details it isn't clear that lxml is faster, actually. They are both based on c-libraries underneath and I think that lxml's data structures are more elaborate to support more sophisticated use cases, so it may actually be slower. The LXML docs kind of imply it often is.

But either way it's a pretty low priority. and I was on the fence about whether to take the risk, so I can back it out.

If you agree that we should move towards the simpler, batteries-included API, I'll make a note like that in the docs so that people know to pick the simpler one if they can.

davisagli (approved):

PR #2993, opened 2021-11-17
Add version check

prescod:
# Changes

Python versions less than 3.8 now give a warning.

"Sorry! Your Python version is not supported. Please upgrade to Python 3.9."

### DIFF
--- a/cumulusci/cli/utils.py
+++ b/cumulusci/cli/utils.py
@@ -1,6 +1,7 @@
 import contextlib
 import os
 import re
+import sys
 import time
 from collections import defaultdict
 
@@ -12,6 +13,8 @@ from cumulusci import __version__
 from cumulusci.core.config import UniversalConfig
 from cumulusci.utils import get_cci_upgrade_command
 from cumulusci.utils.http.requests_utils import safe_json_from_response
+
+LOWEST_SUPPORTED_VERSION = (3, 8, 0)
 
 
 def group_items(items):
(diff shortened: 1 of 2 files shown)


cumulusci/cli/utils.py
```diff
@@ -93,6 +96,12 @@ def check_latest_version():
                 err=True,
             )
 
+        if sys.version_info < LOWEST_SUPPORTED_VERSION:
+            click.echo(
+                "Sorry! Your Python version is not supported. Please upgrade to Python 3.9.",
```
pconrad-sfdo, line 101: Language looks good. My only question is whether it makes sense to use `"Please upgrade to Python " + LOWEST_SUPPORTED_VERSION + "."`

pconrad-sfdo (approved):

cumulusci/cli/utils.py
davisagli: @pconrad-sfdo That's a good thought, but LOWEST_SUPPORTED_VERSION is formatted as a tuple so it can be compared to sys.version_info

@prescod maybe we should define the lowest version in cumulusci/__init__.py and use it as a source for https://github.com/SFDO-Tooling/CumulusCI/blob/main/setup.py#L73 as well as here?

prescod: @davisagli : There are two "lowest versions" in play. A previous version of the branch actually checked both but I ripped that out.

There's "probably does not even work right lowest" (which for us is Python 3.6, right now).

And there's "lowest that probably works but we want you to upgrade in the next X weeks" (which is 3.8 for us).

I ripped out the check for "probably does not even work right" because that one is checked by PIP and I couldn't trigger it because I couldn't get my Python environment itself into such a state.

I think that we only want the two "lowest" numbers to diverge for a few weeks every year or two, but they do diverge, as is the case right now.

@pconrad-sfdo : Similarly, the "best recommended version" is not usually the "lowest supported version", for several reasons like: higher versions may be more secure and performant. And also if we let users leapfrog then they are forced to upgrade less frequently. e.g. "we do support 3.8 but if you go right to 3.9 then you won't need to upgrade when we drop 3.8"

davisagli: @prescod right, makes sense and I was glossing over that difference.

PR #3004, opened 2021-12-06
Feature/robot custom locators bugfix

boakley:
This fixes the bug described in W-10187485, which caused SeleniumLibrary's `get webelements` keyword  to return a
list with a single `None` element rather than an empty list when no elements were found.

Although the SeleniumLibrary documentation says that a custom locator keyword should return a single web element, on close inspection of the code it appears that it should instead return a list. I contacted the developers of SeleniumLibrary and they agreed with me that the documentation seemed to be incorrect.

# Critical Changes

# Changes
* fixed a bug that prevented `Get webelements` from returning an empty list if the locator was a custom locator created via  the `register_locators` function of `cumulusci.robotframework.locator_manager`

# Issues Closed

### DIFF
--- a/cumulusci/robotframework/locator_manager.py
+++ b/cumulusci/robotframework/locator_manager.py
@@ -97,16 +97,16 @@ def locate_element(prefix, parent, locator, tag, constraints):
     logger.info(f"locator: '{prefix}:{locator}' => '{loc}'")
 
     try:
-        element = selenium.get_webelement(loc)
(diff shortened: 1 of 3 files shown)


jofsky (dismissed):

boakley:
@jofsky I had to push a small fix for a test that only failed on firefox. I should have caught this before pushing, sorry! I could use a rubber stamp so we can close this out. Thanks!

davisagli (approved):

PR #3007, opened 2021-12-07
Fix race condition in test framework

prescod:
Fix a race condition in test framework

The old code mutated the "last" mock call. If two threads create mock calls at the same time, the data got attached to the wrong call.

```
        self.return_values = {"step_results": next(return_values)}
        rc = __call__.mock(*args, **kwargs)
        # remember the values that would have been loaded for later inspection in tests
        values_loaded = db_values_from_db_url(self.options["database_url"])
        __call__.mock.mock_calls[-1].values_loaded = values_loaded
```

The new code is more "functional" style: no mutations or dependence on shared variables.

```
        values_loaded = db_values_from_db_url(self.options["database_url"])
        kwargs = {**kwargs, "values_loaded": values_loaded}
        rc = __call__.mock(*args, **kwargs)
```

The other changes are side effects of this change.

### DIFF
--- a/cumulusci/tasks/bulkdata/tests/test_snowfakery.py
+++ b/cumulusci/tasks/bulkdata/tests/test_snowfakery.py
@@ -78,17 +78,16 @@ def call_closure():
     """Simulate a cycle of load results without doing a real load."""
     return_values = cycle(iter(FAKE_LOAD_RESULTS))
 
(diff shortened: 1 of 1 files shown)


boakley (approved):
I'm not entirely clear how this fix works, but I've run this suite of tests a few dozen times and the tests never failed due to a race condition, so I'm happy with this change.

PR #3008, opened 2021-12-08
Use Python > 3.7  for readthedoc builds

Br4nd0R:
# Dev Changes
* The CumulusCI repo now includes a `readthedocs.yml` file for configuring builds in readthedocs.


boakley (approved):
Looks good to me. I verified that the docs built on my machine.

PR #3011, opened 2021-12-13
Fix `cci task list --json`

Br4nd0R:
# Issues Closed
* Fixed an issue where the `--json` flag was not outputting properly formatted `JSON` with the `cci task list` command.


Br4nd0R:
I'm wondering if the `test_task_list__json()` method is even worth keeping around now since the new test case actually validates that the parsed output is what is expected (instead of an assert on a mock object).

jstvz (approved):

PR #3012, opened 2021-12-14
Update deps to clear security warning

prescod:
Update deps to clear security warning

We don't actually have a vulnerability, but we might as well clear it on main.


requirements/dev.txt
```diff
@@ -81,8 +81,6 @@ importlib-metadata==4.8.2
     #   -c requirements/prod.txt
     #   keyring
     #   twine
-importlib-resources==5.4.0
-    # via jsonschema
```
davisagli, line 85: Why did this go away?

pconrad-sfdo (approved):

davisagli (approved):

PR #3014, opened 2021-12-16
Feature/check custom settings

erikperkins:
# Critical Changes
The `check_org_settings_value` task uses the metadata tooling API, which does not support custom objects. However, many of our packages, such as NPSP, use custom settings objects. Here we add support for preflight checks for custom settings objects through the standard API, and apply that support to check for Enhanced Recurring Donations.

# Changes

# Issues Closed


cumulusci/cumulusci.iml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
erikperkins, line 1: I don't know how this ended up here. To be removed.

cumulusci/tasks/preflight/tests/test_settings.py
```diff
@@ -1,5 +1,5 @@
 from unittest.mock import MagicMock
```
erikperkins, line 1: This file seems like it should be promoted to a complete test class. Is there a good reason why it isn't one?

cumulusci/cumulusci.iml
erikperkins: Removed.

davisagli:
@erikperkins Thanks for working on this. Is there any existing discussion of this enhancement that would provide additional context for how you intend to use the check, and why you chose to add a new task rather than adjusting the existing one to handle both standard settings and custom settings?

My gut feeling is that it might make more sense to have a single task that uses the Tooling API when the entity doesn't end with `__c`, or the normal REST API when it does.

I guess using this check with custom settings will always need to be used in conjunction with another check that verifies the package providing the custom settings object is installed.

cumulusci/cumulusci.yml
```diff
@@ -98,6 +102,14 @@ tasks:
...
             settings_field: IsEnhancedNotesEnabled
             value: True
+    check_rd2_enabled:
+        description: Preflight check to validate that Enhanced Recurring Donations are enabled.
+        group: Salesforce Preflight Checks
+        class_path: cumulusci.tasks.preflight.settings.CheckCustomSettingsValue
+        options:
+            settings_type: npe03__Recurring_Donations_Settings__c
+            settings_field: npsp__IsRecurringDonations2Enabled__c
+            value: True
```
davisagli, line 112: I don't think a task config that is preconfigured to check for an NPSP feature belongs in CumulusCI, which is a general-purpose framework designed to be used with many different products. I would define it in cumulusci.yml for the project that needs to perform the check instead.

cumulusci/tasks/preflight/tests/test_settings.py
davisagli: pytest can run tests whether or not they are in a test class, and our coding style doesn't state a preference. What benefit do you see from putting them in a class?

erikperkins: Since there's no stylistic preference, then it doesn't matter. I've usually seen tests collected into classes, but either way is fine.

cumulusci/cumulusci.yml
erikperkins: Good point. Removed in ff5c58b7e0cb6ce10660362bc5df05d310feee13

erikperkins:
@davisagli There isn't specific documentation for the need for this sort of check. In NPSP, there is a [check for Enhanced Recurring Donations](https://github.com/SalesforceFoundation/NPSP/blob/b4a95f24a278c282052bda6151eada1d397c8dc2/cumulusci.yml#L1734) which (improperly) relies on a locally defined task. For Insights Analytics, we need a [preflight check](https://github.com/SalesforceFoundation/InsightsAnalytics/pull/83#discussion_r758531138) for Enhanced Recurring Donations. It seemed obvious to build on the existing settings check.

As for introducing a separate check for custom settings, the idea was basically separation of concerns for custom settings objects. However, I agree with your point about simply extending the existing check.

Addressed in 8b1e10b229061969def43fd0db645d6d199c2f8e

PR #3016, opened 2021-12-20
User channels including tests

prescod:
Using channel declarations in load.yml, users can do synthetic dataloads across multiple user accounts at once, which is faster for some very large orgs.


cumulusci/tasks/bulkdata/snowfakery_utils/subtask_configurator.py
```diff
@@ -0,0 +1,54 @@
...
+            "generator_yaml": str(self.recipe),
+            "database_url": wd.database_url,
+            "num_records": batch_size,
+            "reset_oids": False,
+            "continuation_file": wd.continuation_file,
+            "num_records_tablename": self.run_until.sobject_name or COUNT_REPS,
+            "vars": recipe_options,
+        }
+
+    def data_loader_opts(self, working_dir: Path):
```
jofsky, line 42: Docstring?

jofsky (dismissed):

cumulusci/core/utils.py
```diff
@@ -96,13 +98,20 @@ def process_glob_list_arg(arg):
 
 def process_list_arg(arg):
     """Parse a string into a list separated by commas with whitespace stripped"""
+    if isinstance(arg, Path):
+        arg = str(arg)
```
davisagli, line 102: @prescod can you help me understand this addition? I'm failing to think of a scenario where a Path instance would end up in a task option parsed from the CLI or loaded from yaml.

cumulusci/core/utils.py
```diff
@@ -96,13 +98,20 @@ def process_glob_list_arg(arg):
...
         return arg
     elif isinstance(arg, str):
         args = []
         for part in arg.split(","):
             args.append(part.strip())
         return args
+    else:
+        getLogger(__file__).info(
+            f"Unknown argument type `{type(arg)}` for type `{arg}`"
+        )
```
davisagli, line 114: This looks like a confusing message to log. To the user, they are configuring options, not arguments. The option value is referred to as "type" here. And if they got the type wrong, they'd probably like to know which option it was. I like the general idea of warning about unexpected types, but could we take a step back and do some team work on ACs for this before implementing?

cumulusci/tasks/bulkdata/snowfakery.py
```diff
@@ -631,133 +587,106 @@ def _cleanup_object_tables(self, engine, metadata):
...
+        )
+        msg = f"Multiple channel declarations: {files}"
+        raise TaskOptionsError(msg)
+    elif len(rules_lists) == 1:
+        return rules_lists[0].rules
+    else:
+        return []
+
+
+class ChannelConfig(T.NamedTuple):
```
davisagli, line 639: @prescod I probably won't remember what a channel is in 6 months. Could you add some docstrings to the top-level functions and classes added in this PR?

cumulusci/tasks/bulkdata/tests/test_snowfakery.py
```diff
@@ -74,25 +77,56 @@ def table_values(connection, table):
...
+    """Simulates load results without doing a real load."""
+
+    # these are all used as mutable class variables
+    mock_calls: list  # similar to how a mock.Mock() object works
+    fake_return_values: T.Iterator
+    fake_exception_on_request = -1
+    lock = Lock()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
davisagli, line 90: What is the purpose of this `__init__`?

cumulusci/core/utils.py
prescod: Tasks can be loaded from CLI or YAML but this code is also executed when they are called as an API.

Unit tests are an example, but other places we call them as an API are the `snowfakery` task which calls sub-tasks and Robot keywords that create tasks.

davisagli: @prescod ok. I like the idea of making options that accept string paths also accept pathlib instances, but this seems incomplete or maybe the wrong place to do it since it only achieves that for options that use parse_list_arg. I'm okay with leaving this in but it seems like something that should be addressed as part of switching to pydantic validation of options.

prescod: So is it okay if I leave it until we do the Grand Options Refactoring, because it will add risk to this PR if I take it out.

cumulusci/core/utils.py
prescod: I agree it can be tightened up and I did do some of that. But it's a code path that is hit so infrequently that it didn't become an issue in the last 4 years. I don't think we should open up a big conversation with a lot of people for it. It's also going to be replaced with the grand argument refactoring at some point.

prescod: 100% chance if we open it up for a larger conversation the logical thing to say is "shouldn't we tell them the arg name" which is not possible in a backwards compatible context. We need to refactor to make this message "good".

davisagli (approved):

PR #3026, opened 2022-01-07
Config-None Cleanup

prescod:
Step towards fixing the Config-None Problem:

https://salesforce.quip.com/VURuAh8VSd1V

1. declare all config properties
2. warn about use of undeclared properties
3. add lookup() method for looking up arbitrary properties without warning
4. add env-var to translate warnings into exceptions
5. Create a RemoteProjectConfig class
 
# Changes

Programs using CumulusCI as a library may generate warnings when getting values from config objects. A new method `config.lookup` is the preferred way to look up these values.


cumulusci/cli/tests/test_org.py
```diff
@@ -801,7 +802,13 @@ def test_org_prune(self, echo):
...
             "persistent",
         ]
-        runtime.project_config.orgs__scratch = {"shape1": True, "shape2": True}
+
+        runtime.project_config.lookup = MockLookup(
+            orgs__scratch={
+                "shape1": True,
+                "shape2": True,
+            },
+        )
```
davisagli, line 811: I'd strongly encourage rewriting the tests to use real runtime, keychain, and project config objects instead of mocks. I worked on this for 1 or 2 of the CLI modules already (the same ones that got updated to use run_cli_command instead of run_click_command) but didn't get to this one yet.

cumulusci/cli/tests/test_service.py
```diff
@@ -103,7 +103,10 @@ def test_service_connect__attr_with_default_value():
     # but input of an empty line accepts the default.
     assert "attr (example) [PRESET]: " in result.output
     service_config = runtime.keychain.get_service("test", "test-alias")
-    assert service_config.attr == "PRESET"
+    with mock.patch("warnings.warn") as w:
+        assert service_config.lookup("attr") == "PRESET"
+        assert service_config.attr == "PRESET"
+        assert len(w.mock_calls) == 1
```
davisagli, line 109: Maybe use pytest.warns? https://docs.pytest.org/en/6.2.x/warnings.html#warns

cumulusci/core/config/__init__.py
```diff
@@ -11,19 +11,37 @@
...
 
 class FlowConfig(BaseConfig):
     """A flow with its configuration merged"""
 
-    pass
+    description: str
+    steps: list
+    group: str
+    checks: list
+    project_config: BaseConfig
```
davisagli, line 27: ProjectConfig, not BaseConfig, no?

cumulusci/core/config/__init__.py
```diff
@@ -11,19 +11,37 @@
...
+    url: str
+    username: str
+    password: str
+    token: str
+    email: str
+    client_id: str
+    client_secret: str
+    token_uri: str
+    callback_url: str
+    login_url: str
```
davisagli, line 43: Ooh, I don't like this. The valid service attributes depend on what type of service it is. And new service types with new attributes can be defined in cumulusci.yml

cumulusci/core/config/__init__.py
```diff
@@ -32,6 +50,15 @@ def __init__(self, config, name=None, keychain=None):
 class TaskConfig(BaseConfig):
     """A task with its configuration merged"""
 
+    options: str
+    class_path: str
+    description: str
+    group: str
+    ui_options: str
```
davisagli, line 57: dict

cumulusci/core/config/__init__.py
```diff
@@ -32,6 +50,15 @@ def __init__(self, config, name=None, keychain=None):
...
+    name: str
+    checks: str
```
davisagli, line 59: list

cumulusci/core/config/__init__.py
```diff
@@ -32,6 +50,15 @@ def __init__(self, config, name=None, keychain=None):
...
+    project_config: str
```
davisagli, line 60: ProjectConfig

cumulusci/core/config/base_config.py
```diff
@@ -23,7 +30,47 @@ def _load_config(self):
...
         pass
 
+    @classmethod
+    def allowed_names(cls) -> set:
+        properties = set(dir(cls))
+        annotations = set(getattr(cls, "__annotations__", {}))
+        return properties.union(annotations)
+
+    # long term plan is to get rid of this
     def __getattr__(self, name):
```
davisagli, line 40: A docstring explaining the policy we're trying to enforce would be helpful.

cumulusci/core/config/base_config.py
```diff
@@ -23,7 +30,47 @@ def _load_config(self):
...
+        if not name.startswith("_"):
+            first_part = name.split("__")[0]
+            if (
+                first_part
+                not in self.all_allowed_names()
+                # and first_part not in self.config
+            ):
+                warnings.warn(
+                    f"__getattr__ on Configs is deprecated: `{first_part}` on `{self.__class__.__name__}`",
```
davisagli, line 49: This is likely to be a confusing warning, since the person seeing it is likely either a user with no awareness of how cci works internally, or a cci dev who may not understand they were using `__getattr__`.

cumulusci/core/config/base_config.py
```diff
@@ -23,7 +30,47 @@ def _load_config(self):
...
+                    DeprecationWarning,
+                )
+
+                assert (
+                    not STRICT_GETATTR
+                ), f"__getattr__ on Configs is deprecated: `{first_part}` on `{self.__class__.__name__}`"
```
davisagli, line 55: So with STRICT_GETATTR there's both a warning and an AssertionError?

cumulusci/core/config/base_config.py
```diff
@@ -23,7 +30,47 @@ def _load_config(self):
...
+        return self.lookup(name)
+
+    @classmethod
+    @lru_cache
+    def all_allowed_names(cls):
```
davisagli, line 60: Probably best to call them _allowed_names and _all_allowed_names

cumulusci/core/config/base_task_flow_config.py
```diff
@@ -72,3 +76,10 @@ def get_suggested_name(self, name, steps):
             return f'. Did you mean "{match_list[0]}"?'
         else:
             return ""
+
+
+from .base_task_flow_config import *  # noqa
+
+warn(
+    "Deprecation warning: Please import from .base_task_flow_config rather than .BaseTaskFlowConfig"
+)
```
davisagli, line 85: Huh? This is base_task_flow_config

cumulusci/core/config/marketing_cloud_service_config.py
```diff
@@ -10,6 +10,12 @@
 
 
 class MarketingCloudServiceConfig(OAuth2ServiceConfig):
+
+    refresh_token: str
+    oauth2_client: str
+    soap_instance_url: str
+    rest_instance_url: str
```
davisagli, line 17: I think a marketing_cloud service has more than these.

cumulusci/core/config/org_config.py
```diff
@@ -31,6 +33,39 @@
...
+    userinfo: str
+    id: str
+    expires: str
+    active: str
+    default: str
+    expired: bool
+    client_id: str
+    refresh_token: str
+    client_secret: str
+    connected_app: str
```
davisagli, line 65: Some of these are valid/expected for ScratchOrgConfigs but not OrgConfigs. Some of the types look wrong. And really an OrgConfig may have any attribute that was returned by the oauth auth flow.

davisagli: Also I think it'll lead to confusion that some of these are expected to be None sometimes but are not declared as Optional

cumulusci/core/config/project_config.py
```diff
@@ -42,7 +43,23 @@
 )
 
 
-class BaseProjectConfig(BaseTaskFlowConfig):
+class ProjectConfigPropertiesMixin(BaseConfig):
```
davisagli, line 46: Why didn't you just put them directly in BaseProjectConfig?

cumulusci/core/config/project_config.py
```diff
@@ -42,7 +43,23 @@
...
+    cumulusci: dict
+    orgs: dict
+    minimum_cumulusci_version: str
+    sources: dict
+    flows: dict
+    plans: dict
+    tasks: dict
+    repo_owner: str
+    repo_name: str
+    dev_config: dict  # this is not documented and should be deprecated
```
davisagli, line 59: Where is it used?

cumulusci/core/config/project_config.py
```diff
@@ -605,3 +622,7 @@ def open_cache(self, cache_name):
         with open_fs_resource(self.cache_dir / cache_name) as cache_dir:
             cache_dir.mkdir(exist_ok=True, parents=True)
             yield cache_dir
+
+
+class RemoteProjectConfig(BaseConfig):
+    project: dict
```
davisagli, line 628: Why do we need a separate RemoteProjectConfig?

cumulusci/core/config/tests/test_config.py
```diff
@@ -47,64 +47,78 @@
...
         assert config.foo is None
+        assert len(warn.mock_calls) == 1
 
     def test_getattr_child_key(self):
-        config = BaseConfig()
+        config = FakeConfig()
         config.config = {"foo": {"bar": "baz"}}
         assert config.foo__bar == "baz"
 
+    def test_strict_getattr(self):
```
davisagli, line 72: Why did you change the existing tests to use FakeConfig? It makes it much less clear that the goal here is to test BaseConfig.

cumulusci/core/config/universal_config.py
```diff
@@ -2,16 +2,23 @@
...
+    ProjectConfigPropertiesMixin,
+)
 from cumulusci.core.utils import merge_config
 from cumulusci.utils.yaml.cumulusci_yml import cci_safe_load
 
 __location__ = os.path.dirname(os.path.realpath(__file__))
 
 
-class UniversalConfig(BaseTaskFlowConfig):
+class UniversalConfig(BaseTaskFlowConfig, ProjectConfigPropertiesMixin):
```
davisagli, line 15: Oh, now I see why you created ProjectConfigPropertiesMixin. Well, give it a docstring to explain!

cumulusci/core/config/universal_config.py
```diff
@@ -2,16 +2,23 @@
...
     """Base class for the global config which contains all configuration not specific to projects"""
 
+    universal_config_obj: dict  # undocumented
```
davisagli, line 18: universal_config_obj is used on BaseProjectConfig, not on UniversalConfig

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -28,12 +28,15 @@
...
+    orgs: dict
+    app: object  # what is this?
+    services: dict
 
     def __init__(self, project_config, key):
         super(BaseProjectKeychain, self).__init__()
         self.config = {
             "orgs": {},
-            "app": None,
+            "app": None,  # what is this?
```
davisagli, line 39: It's where connected app info was kept before the concept of services was invented. I doubt it still needs to be here.

davisagli (commented):
I think the overall goal here has some merit but there are a bunch of details I'm not sure about, especially when it comes to actually evaluating the type annotations as type annotations rather than as a way of configuring how `__getattr__` works.

cumulusci/core/config/project_config.py
prescod: It seems to have a single property. What other type would be preferable?

cumulusci/core/config/tests/test_config.py
prescod: Because BaseConfig doesn't have a property called "foo".

cumulusci/cli/tests/test_org.py
prescod: Okay but that would be a separate code cleanup I guess?

cumulusci/cli/tests/test_service.py
prescod: Done in ee235fe01

cumulusci/core/config/__init__.py
prescod: Done in ee235fe01

cumulusci/core/config/__init__.py
prescod: Fixed in ee235fe01

cumulusci/core/config/__init__.py
prescod: Fixed in ee235fe01

cumulusci/core/config/__init__.py
prescod: Fixed in ee235fe01

cumulusci/core/config/base_config.py
prescod: Fixed in ee235fe01

cumulusci/core/config/base_config.py
```diff
@@ -23,7 +30,47 @@ def _load_config(self):
...
+        return properties.union(annotations)
+
+    # long term plan is to get rid of this
     def __getattr__(self, name):
+        if not name.startswith("_"):
+            first_part = name.split("__")[0]
+            if (
+                first_part
+                not in self.all_allowed_names()
+                # and first_part not in self.config
```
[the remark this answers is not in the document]
prescod: Fixed in ee235fe01

cumulusci/core/config/base_config.py
prescod: Fixed in ee235fe01

cumulusci/core/config/base_task_flow_config.py
prescod: Merge error caused by renaming files. Fixed now.

cumulusci/core/config/org_config.py
```diff
@@ -20,6 +20,8 @@
 from cumulusci.utils.fileutils import open_fs_resource
 from cumulusci.utils.http.requests_utils import safe_json_from_response
 
+from .org_config import *  # noqa
```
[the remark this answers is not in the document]
prescod: Merge error fixed now.

cumulusci/core/config/project_config.py
prescod: Documented now.

cumulusci/core/config/project_config.py
```diff
@@ -42,7 +43,23 @@
...
+    cumulusci: dict
+    orgs: dict
+    minimum_cumulusci_version: str
+    sources: dict
+    flows: dict
+    plans: dict
+    tasks: dict
+    repo_owner: str
+    repo_name: str
```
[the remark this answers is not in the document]
prescod:
```
    def get_repo(self):
        repo = self.get_github_api(self.repo_owner, self.repo_name).repository(
            self.repo_owner, self.repo_name
        )
```

cumulusci/core/config/universal_config.py
prescod: Fixed now.

cumulusci/core/config/universal_config.py
prescod: Moved in ee235fe01

cumulusci/core/config/__init__.py
davisagli: @prescod I didn't mean to imply that solving all that is the only acceptable way to make incremental improvements here. But I wanted to point out where I see potential for things to not work out cleanly. And I do feel rather strongly that adding _incorrect_ type annotations would be a step backward. If we can't take the time to get the type annotations correct right now, maybe it would be better to use a different mechanism for specifying what names are expected (like just a set assigned to a class attribute) for the time being.

With regard to point 4 about dynamically generating type annotations, that sounds pretty messy and pretty far from the notion of type annotations as something that can be statically analyzed. Instead of that, it might make sense to move in the direction of making each type of service use its own ServiceConfig subclass that declares its attributes as a pydantic model. That could replace the existing configuration of service attributes in yaml; I'm not sure that has much benefit. Not as part of this PR though!

cumulusci/core/config/project_config.py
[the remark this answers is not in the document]
davisagli: Okay but those are properties of BaseProjectConfig and will not go through `__getattr__`, right?

https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/config/project_config.py#L283
https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/config/project_config.py#L307

cumulusci/core/config/project_config.py
davisagli: Can that use .lookup() instead?

cumulusci/core/config/universal_config.py
davisagli: I see you moved it, but it's still the wrong type. It's a UniversalConfig instance, not a dict

cumulusci/core/config/project_config.py
prescod: It can, but why treat it differently from everything else? It's a valid property which should either be deprecated or documented and hiding it in the code will delay rather than accelerate that process.

cumulusci/core/config/project_config.py
prescod: Okay I'll change it to return a BaseProjectConfig. Why is BaseProjectConfig called that when there don't seem to be classes which inherit from it? Are we using "Base" to mean something different than Base Class?

prescod: Actually we need RemoteProjectConfig because BaseProjectConfig has initialization code that expects a local config.

But I can broaden RemoteProjectConfig's list of properties to match BaseProjectConfig: 9f8860355

davisagli: @prescod what are you hitting exactly? I think BaseProjectConfig avoids trying to read local config if a config dict was passed in when it's initialized.

cumulusci/core/config/__init__.py
prescod: Okay I added an opt-in type checking mode to verify that the types are at least correct enough to pass our test suite. 059de0e8a

cumulusci/core/config/project_config.py
prescod: For whoever reads this in the future: the root of the confusion is that BaseProject takes config as its first init argument and BaseProjectConfig takes it as its second. So config was being defaulted to None and then {} and then attempted to be loaded.

prescod: Fixed in 40254a0ad9d505b65131346980ceea4588b11067

cumulusci/core/config/base_config.py
```diff
@@ -23,7 +43,46 @@ def _load_config(self):
         """Subclasses may override this method to initialize :py:attr:`~config`"""
         pass
 
+    @classmethod
+    def _allowed_names(cls) -> T.Dict[str, type]:
+        return getattr(cls, "__annotations__", {})
```
Br4nd0R, line 48: It took me a while but I finally figured out what `__annotations__` is. 😅

Br4nd0R (approved):
This looks great @prescod!

PR #3028, opened 2022-01-07
Rename config files to match current coding standards

prescod:
Rename config files to:

1. Reflect current naming standards.
2. Be easier to disambiguate from their same-named classes, for mocking and importing purposes


davidmreed (changes requested):
The code looks good, but this change breaks MetaCI, so I'm going to put a "request changes" until we sort that out.

Integration tests:
- [x] MetaDeploy
- [ ] MetaCI
- [ ] Metecho

prescod:
@davidmreed The MetaCI change seems to be as shallow as we'd expect, based on my local testing. We can triple-check by making a dev release and linking to that on CI and staging environments.

cumulusci/version.txt
```diff
@@ -1 +1 @@
-3.51.1
+3.51.1.dev1
```
davisagli, line 1: This should be 3.52.0.dev0 -- 3.51.1.dev1 comes before 3.51.1 the way that PyPI sorts version numbers.

cumulusci/core/config/__init__.py
```diff
@@ -35,22 +35,20 @@ class TaskConfig(BaseConfig):
...
-from cumulusci.core.config.BaseTaskFlowConfig import BaseTaskFlowConfig
-
+from cumulusci.core.config.base_task_flow_config import BaseTaskFlowConfig
 
 # inherit from BaseTaskFlowConfig
 from cumulusci.core.config.project_config import BaseProjectConfig
 
+# inherit from SfdxOrgConfig
+from cumulusci.core.config.scratch_org_config import ScratchOrgConfig
+
```
davisagli, line 45: Why did you move this one? I guess as long as there aren't import cycle problems it's fine, but ScratchOrgConfig extends SfdxOrgConfig so it made sense to me for them to be listed in the previous order.

prescod:
`isort` does that @David Glick ***@***.***>

davisagli:
eek, we probably need to disable isort for `__init__.py` files

prescod:
@davisagli It looks like we do have such a setting but my editor doesn't know about it.

In-file declarations are probably more reliable?

So 

1. do you want me to "unsort" this particular file?

2. do you want us to add isort disabling comments too each __init__.py file?

davisagli:
@prescod I don't think any immediate action is required as long as it's still possible to run cci commands without hitting ImportErrors. But adding the comment to disable isort in `__init__.py` files that have convenience imports sounds like a good idea to me.

davisagli:
I believe this is simply waiting for a fix in version.txt and final approval by @davidmreed

prescod:
Okay thanks @davisagli . Updated to 3.52.0.dev0 . Although I am unclear what happens if we get the sort order wrong for dev releases. Nothing wrong with getting it "right" of course, just curious if there is anything that happens if you get it wrong.

@davidmreed Ready for your re-review.

davisagli:
@prescod Quite possible it wouldn't actually matter in practice, but I can think of 2 situations where it might. It would make a difference in whether the new dev release gets installed when someone pip installs cumulusci with the `--pre` flag but no specific version specified. It would also interfere with a downstream dependency specifying the dev release as a minimum required version (e.g. `cumulusci>=3.51.0.dev0` would allow 3.51.0 but `cumulusci>=3.52.0.dev0` wouldn't. I think)

davidmreed (approved):

PR #3033, opened 2022-01-13
Require keychain when generating flow docs

Br4nd0R:
# Changes
- The `cci flow doc` command now only shows CumulusCI's standard flows unless the `--project` option is specified.

# Issued Closed
- Fixed a bug where the `cci flow doc` command would break when using cross-project flows.

### DIFF
--- a/cumulusci/cli/flow.py
+++ b/cumulusci/cli/flow.py
@@ -21,7 +21,7 @@ def flow():
 
 
 @flow.command(name="doc", help="Exports RST format documentation for all flows")
-@pass_runtime(require_keychain=True)
+@pass_runtime(require_project=False)
 def flow_doc(runtime):
     flow_info_path = Path(__file__, "..", "..", "..", "docs", "flows.yml").resolve()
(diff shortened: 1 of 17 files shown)


cumulusci/cli/flow.py
```diff
@@ -29,7 +29,7 @@ def flow_doc(runtime):
     click.echo(flow_ref_title_and_intro(flow_info["intro_blurb"]))
     flow_info_groups = list(flow_info["groups"].keys())
 
-    flows = runtime.get_available_flows()
+    flows = runtime.get_available_flows(project=False)
```
davisagli, line 32: Hmm, previously it was possible to run `cci flow doc` in a CumulusCI project to get output including project-specific flows. I know that's not what we want for CumulusCI's own documentation, but I'm not sure we should remove that capability for other projects. It at least warrants a release note.

Br4nd0R: Would a good middle ground be to provide the `cci flow doc` command a `--project` flag? When passed, it would also include documentation for project specific flows?

davisagli (approved):

PR #3034, opened 2022-01-15
Make cross-project step preflights work

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We fixed an issue preventing step-level preflight checks from working correctly in MetaDeploy when run as a cross-project flow.
- The `metadeploy_publish` task now displays a clear error message if you supply the wrong API URL.

### DIFF
--- a/cumulusci/core/flowrunner.py
+++ b/cumulusci/core/flowrunner.py
@@ -809,7 +809,7 @@ class CachedTaskRunner(object):
             task_name=self.task_name,
             task_config=task_config,
             task_class=task_class,
-            project_config=self.cache.flow.project_config,
+            project_config=self.cache.project_config,
         )
         self.cache.flow.callbacks.pre_task(step)
         result = TaskRunner(step, self.cache.flow.org_config, self.cache.flow).run_step(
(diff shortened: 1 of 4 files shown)


cumulusci/core/flowrunner.py
```diff
@@ -790,7 +802,7 @@ def __call__(self, **options):
         if cache_key in self.cache.results:
             return self.cache.results[cache_key].return_values
 
-        task_config = self.cache.flow.project_config.tasks[self.task_name]
+        task_config = self.cache.project_config.tasks[self.task_name]
```
davisagli, line 805: @davidmreed we should also pass the step-specific project config below on line 812

davisagli (changes requested):
Good analysis and fix. I think I see one more place where we need to pass the step-specific project_config

cumulusci/core/flowrunner.py
davidmreed: 👍 Good catch.

davisagli (approved):

davisagli:
Hmm, what's up with the failed test in py3.8?

davidmreed:
It passed on a retry; looks like maybe a flappy Snowfakery user channels test. @prescod Do you recognize it?

davidmreed:
I'll go ahead and merge since I don't see any plausible link to this work.

PR #3039, opened 2022-01-19
Marketing Cloud: Get User Info Task

Br4nd0R:
# Changes
* A new task is available for getting user information from the Marketing Cloud REST API `/userinfo` endpoint: `marketing_cloud_get_user_info`.

# Dev Changes
* Created a new `mc_constants.py` file for shared constants across MC files.
* Refactored a couple of pytest fixtures into a mc specific `conftest.py`.

### DIFF
--- a/cumulusci/tasks/marketing_cloud/get_user_info.py
+++ b/cumulusci/tasks/marketing_cloud/get_user_info.py
@@ -1,3 +1,5 @@
+import json
+
 import requests
 
 from cumulusci.utils.http.requests_utils import safe_json_from_response
@@ -27,7 +29,7 @@ class GetUserInfoTask(BaseMarketingCloudTask):
         payload = self._sanitize_payload(payload)
 
         self.logger.info("Successfully fetched user info.")
-        self.logger.info(payload)
+        self.logger.info(json.dumps(payload, indent=4))
 
         self.return_values = payload
 
--- a/cumulusci/tasks/marketing_cloud/get_user_info.py
+++ b/cumulusci/tasks/marketing_cloud/get_user_info.py
@@ -1,5 +1,3 @@
-from logging import getLogger
-
 import requests
 
 from cumulusci.utils.http.requests_utils import safe_json_from_response
@@ -11,8 +9,6 @@ from .mc_constants import MC_API_VERSION
 class GetUserInfoTask(BaseMarketingCloudTask):
     """Retrieves user info of pertaining to the currently logged in user.
     Sanitizes and returns the payload in self.return_values."""
-
-    logger = getLogger(__name__)
 
     def _run_task(self):
         endpoint = f"https://{self.mc_config.tssd}.auth.marketingcloudapis.com/{MC_API_VERSION}/userinfo"
@@ -27,9 +23,13 @@ class GetUserInfoTask(BaseMarketingCloudTask):
             self.logger.error(f"Exception occurred fetching user info: {response.text}")
             raise
 
-        self.logger.info("Successfully fetched user info")
         payload = safe_json_from_response(response)
-        self.return_values = self._sanitize_payload(payload)
+        payload = self._sanitize_payload(payload)
+
+        self.logger.info("Successfully fetched user info.")
+        self.logger.info(payload)
+
+        self.return_values = payload
 
     def _sanitize_payload(self, payload: dict) -> dict:
         """Removes any sensitive or non-pertinent informaiton from the payload.
--- a/cumulusci/tasks/marketing_cloud/get_user_info.py
+++ b/cumulusci/tasks/marketing_cloud/get_user_info.py
@@ -38,10 +38,12 @@ class GetUserInfoTask(BaseMarketingCloudTask):
         This currently includes the following top-level portions of the response:
         (1) rest
         (2) application
+        (3) permissions
 
         See the following for a comprehensive response example:
         https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/getUserInfo.html
         """
         del payload["rest"]
         del payload["application"]
+        del payload["permissions"]
         return payload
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/marketing_cloud/get_user_info.py
```diff
@@ -0,0 +1,42 @@
...
+    def _run_task(self):
+        endpoint = f"https://{self.mc_config.tssd}.auth.marketingcloudapis.com/{MC_API_VERSION}/userinfo"
+        headers = {
+            "Authorization": f"Bearer {self.mc_config.access_token}",
+            "Content-Type": "application/json",
+        }
+        try:
+            response = requests.get(endpoint, headers=headers)
+        except requests.exceptions.HTTPError as e:
+            self.logger.error(f"Error fetching user info: {e}")
```
Br4nd0R, line 25: Should we log here, or should we just let error handling in `cci.py` work its magic?

cumulusci/tasks/marketing_cloud/get_user_info.py
```diff
@@ -0,0 +1,42 @@
...
+        """Removes any sensitive or non-pertinent informaiton from the payload.
+        This currently includes the following top-level portions of the response:
+        (1) rest
+        (2) application
+
+        See the following for a comprehensive response example:
+        https://developer.salesforce.com/docs/marketing/marketing-cloud/guide/getUserInfo.html
+        """
+        del payload["rest"]
+        del payload["application"]
```
Br4nd0R, line 41: These seemed specific to the installed package, and thus, not very relevant to the user information.

cumulusci/tasks/marketing_cloud/get_user_info.py
davisagli: I would be sure to log something here which includes response.text so we can see the error as reported by Marketing Cloud.

Also don't forget to call response.raise_for_status() to handle HTTP responses with a non-200 status.

davisagli: Actually please use the safe_json_from_response helper

Br4nd0R: I was thinking that the try/except block would be in-place-of `raise_for_status()`, I'm happy to change it though. Are you thinking it would cover a larger swath of exception scenarios?

cumulusci/tasks/marketing_cloud/tests/test_get_user_info.py
```diff
@@ -0,0 +1,130 @@
...
+def test_get_user_info__HTTPError(get_user_info_task, user_info_payload):
+    responses.add(
+        "POST",
+        "https://tssd.auth.marketingcloudapis.com/v2/token",
+        json={"access_token": "ACCESS_TOKEN"},
+    )
+    responses.add(
+        "GET",
+        f"https://TSSD.auth.marketingcloudapis.com/{MC_API_VERSION}/userinfo",
+        body=requests.exceptions.HTTPError("bad request"),
```
davisagli, line 112: If you're trying to simulate a 400 Bad Request response, I think you want status=400 and body= a string, not an error instance

cumulusci/tasks/marketing_cloud/get_user_info.py
davisagli: requests.get doesn't raise an HTTPError when it gets a non-200 response. It returns the response object with the status that was received.

cumulusci/tasks/marketing_cloud/get_user_info.py
```diff
@@ -0,0 +1,45 @@
...
+
+from .base import BaseMarketingCloudTask
+from .mc_constants import MC_API_VERSION
+
+
+class GetUserInfoTask(BaseMarketingCloudTask):
+    """Retrieves user info of pertaining to the currently logged in user.
+    Sanitizes and returns the payload in self.return_values."""
+
+    logger = getLogger(__name__)
```
davisagli, line 15: You don't need or want this; task instances get initialized with a logger that is derived from the flow in some cases: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/core/tasks.py#L87

cumulusci/tasks/marketing_cloud/get_user_info.py
```diff
@@ -0,0 +1,45 @@
...
+        try:
+            response = requests.get(endpoint, headers=headers)
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            self.logger.error(f"Exception occurred fetching user info: {response.text}")
+            raise
+
+        self.logger.info("Successfully fetched user info")
+        payload = safe_json_from_response(response)
+        self.return_values = self._sanitize_payload(payload)
```
davisagli, line 32: Should we also log the payload? The current output from running the task on the CLI is not very enlightening:

```
% cci task run marketing_cloud_get_user_info                                                                                                                                                                                 [01/19/22 16:30:42] Beginning task: GetUserInfoTask                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                 
[01/19/22 16:30:43] Successfully fetched user info
```

davisagli (approved):

cumulusci/tasks/marketing_cloud/tests/test_api.py
```diff
@@ -1,57 +1,18 @@
...
 
 @responses.activate
-def test_marketing_cloud_create_subscriber_attribute_task(create_task, project_config):
+def test_marketing_cloud_create_subscriber_attribute_task(
+    create_task, mc_project_config
+):
     responses.add(
         "POST",
-        "https://tssd.auth.marketingcloudapis.com/v2/token",
+        f"https://tssd.auth.marketingcloudapis.com/{MC_API_VERSION}/token",
```
codexmarcel, line 15: nice

cumulusci/tasks/marketing_cloud/mc_constants.py
```diff
@@ -0,0 +1 @@
+MC_API_VERSION = "v2"
```
codexmarcel, line 1: good idea

PR #3040, opened 2022-01-19
Adjust marketing_cloud_create_user task

davisagli:
# Critical Changes

# Changes
* The `marketing_cloud_create_user` task now creates an unlocked user with a notification email address so that it is possible for the user to log in. This task also has a new option, `activate_if_existing`, which can be set to True to ensure that if the user already exists in an inactive state, it will be activated.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/marketing_cloud/base.py
+++ b/cumulusci/tasks/marketing_cloud/base.py
@@ -17,7 +17,7 @@ class BaseMarketingCloudTask(BaseTask):
         super()._init_task()
         self.mc_config = self.project_config.keychain.get_service("marketing_cloud")
 
-    def _check_response(self, response):
+    def _check_soap_response(self, response):
         """Make sure the response indicates success."""
         response.raise_for_status()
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/marketing_cloud/base.py
```diff
@@ -15,3 +16,18 @@ class BaseMarketingCloudTask(BaseTask):
     def _init_task(self):
         super()._init_task()
         self.mc_config = self.project_config.keychain.get_service("marketing_cloud")
+
+    def _check_response(self, response):
+        """Make sure the response indicates success."""
```
Br4nd0R, line 21: If this is SOAP API specific maybe this should be `_check_soap_response()` instead?

Br4nd0R (approved):
Looks good. Just the one comment from me :)

cumulusci/tasks/marketing_cloud/base.py
davisagli: Good call; done

PR #3043, opened 2022-01-20
fix overriding flow steps

Br4nd0R:
# Changes

* Flow steps can now be replaced with the same syntax for all step types! (Current step types are: 'task' or 'flow'). See the [replacing a flow step](https://cumulusci.readthedocs.io/en/latest/config.html?highlight=override#replace-a-flow-step) for more details. CumulusCI is _still _compatible__ with the old syntax which required setting the current step type to 'None' when replacing with a step of a differing type.

# Issues Closed

* Fixed and issue where overriding flow steps was not working as documented.

### QA Performed

- [x] `cci flow run` works as expected when overriding flow steps of different types.
- [x] `cci flow info` shows the expected output (correct task options when overrides present).
- [x] Flow step overrides of the same step type continue to work as expected.
- [x] Still compatible with old flow step replace syntax.


cumulusci/core/tests/test_flowrunner.py
```diff
@@ -86,11 +86,9 @@ def test_each_flow(self):
                 flow_config = self.project_config.get_flow(flow_name)
                 flow = FlowCoordinator(self.project_config, flow_config, name=flow_name)
             except Exception as exc:
```
jstvz, line 88: 🧐

jstvz (approved):
Some weirdness here. If I apply:
```diff
diff --git a/cumulusci.yml b/cumulusci.yml
index 812e8394c..e32599968 100644
--- a/cumulusci.yml
+++ b/cumulusci.yml
@@ -41,6 +41,11 @@ tasks:
                 - cumulusci/robotframework/pageobjects/BasePageObjects.py
 
 flows:
+    dev_org:
+        steps:
+            1:
+                flow: None
+                task: util_sleep
     qa_org:
         steps:
             2:
```

I get the following `flow info` output:
```console
❯ ccfi dev_org
Description: Set up an org as a development environment for unmanaged metadata

Flow Steps
1) task: util_sleep [from current folder]
   options:
       seconds: 5
       update_dependencies: {'resolution_strategy': 'preproduction'}
2) flow: deploy_unmanaged
[...]
```
_but_... the step executes correctly:
```console
------------------------------------------------------------
Running task: util_sleep
------------------------------------------------------------

Options:
  seconds: 5
Beginning task: Sleep

Sleeping for 5 seconds
```
---
Besides that, everything works as expected. May be worth documenting the unexpected options, but otherwise LGTM.

cumulusci/core/tests/test_flowrunner.py
```diff
@@ -613,6 +611,30 @@ def change_username(keychain):
...
+        overrides a flow step that is currently a task with a sub-flow."""
+
+        self.project_config.config["flows"]["test"] = {
+            "description": "Replace a task with a flow",
+            "steps": {1: {"task": "pass_name", "flow": "None"}},
+        }
+        flow_config = self.project_config.get_flow("test")
+        # No assert, we just don't want an exception thrown
+        FlowCoordinator(self.project_config, flow_config)
+
```
Br4nd0R, line 637: Note that the (already existing) test method `test_init_ambiguous_step()` covers the case where both `task` and `flow` come in with a value of `"None"`.

cumulusci/core/tests/test_flowrunner.py
Br4nd0R: Not sure what this is about... I was just subbing in f-strings 😂

Br4nd0R:
@jstvz - This should be ready for your review/qa with the new (simpler) syntax for replacing a flow step.

@pconrad-sfdo - Minor doc adjustments on this one under the `config.rst` file.

cumulusci/core/flowrunner.py
```diff
@@ -541,12 +541,8 @@ def _visit_step(
...
             parent_ui_options = {}
 
-        # Step Validation
-        # - A step is either a task OR a flow.
-        if all(k in step_config for k in ("flow", "task")):
-            raise FlowConfigError(
-                f"Step {number} is configured as both a flow AND a task. \n\t{step_config}."
-            )
+        # This should never happen because of cleanup in core/utils/merge_config()
+        assert not all(step_type in step_config for step_type in ("task", "flow"))
```
jstvz, line 545: Can simplify this with a set:
```suggestion
        # This should never happen because of cleanup in core/utils/merge_config()
        assert step_config.keys() != {"task", "flow"}
```

cumulusci/core/tests/test_flowrunner.py
```diff
@@ -280,11 +278,6 @@ def test_init__nested_options(self):
         flow = FlowCoordinator(self.project_config, flow_config)
         self.assertEqual("bar", flow.steps[0].task_config["options"]["foo"])
 
-    def test_init_ambiguous_step(self):
-        flow_config = FlowConfig({"steps": {1: {"task": "None", "flow": "None"}}})
-        with self.assertRaises(FlowConfigError):
-            FlowCoordinator(self.project_config, flow_config, name="test")
-
```
jstvz, line 287: We'll want some version of this test here to verify the behavior is as expected.

jstvz (changes requested):
Hit a snag here with the old behavior. If that's an intentional breaking change we'll need to throw a better exception message and be clearer about it in the release notes.

We'll need some tests either way.

cumulusci/core/utils.py
```diff
@@ -151,15 +151,101 @@ def decode_to_unicode(content):
...
+    both a "task" and a "flow" listed in a flow step after merging the configs with `dictmerge()`."""
+    if "flows" not in config_to_override or "flows" not in overridding_config:
+        return
+
+    for flow, flow_config in overridding_config["flows"].items():
+        for (
+            step_num,
+            overridding_step_config,
+        ) in flow_config["steps"].items():
+            if config_has_flow_and_step_num(config_to_override, flow, step_num):
```
Br4nd0R, line 225: extract these to capture return value in a properly named variable.

cumulusci/core/flowrunner.py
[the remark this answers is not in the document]
Br4nd0R: Good call out here. I believe I can tweak the logic to also handle this case of the "old" syntax for replacing a step.

cumulusci/core/tests/test_flowrunner.py
Br4nd0R: added test for ambiguous step to `test_utils_merge_config.py`

cumulusci/core/flowrunner.py
Br4nd0R: implemented

cumulusci/core/utils.py
```diff
@@ -151,15 +155,123 @@ def decode_to_unicode(content):
...
+
+    The while loop below is how we account for these scenarios.
+    """
+    config_precedence_order = [
+        "additional_yaml",
+        "project_local_config",
+        "project_config",
+        "global_config",
+        "universal_config",
+    ]
```
davisagli, line 194: Why are you using a hardcoded list here and not simply processing all configs in the order they are listed? `merge_config` specifies that the configs are in priority order (highest precedence first).

cumulusci/core/utils.py
```diff
@@ -151,15 +155,123 @@ def decode_to_unicode(content):
...
+    while len(config_precedence_order) > 1:
+        overridding_config = config_precedence_order[0]
```
davisagli, line 196: Spelling should be corrected to overriding everywhere this occurs

cumulusci/core/utils.py
```diff
@@ -151,15 +155,123 @@ def decode_to_unicode(content):
...
 
 def merge_config(configs):
-    """recursively deep-merge the configs into one another (highest priority comes first)"""
-    new_config = {}
+    """
+    First remove any flow steps that are being overridden so that there are no conflicts
+    or different step types after merging. Then recursively deep-merge the configs into
+    one another (highest priority comes first)
+    """
+    configs = cleanup_flow_step_override_conflicts(configs)
```
davisagli, line 163: We might want to deepcopy `configs` first to make sure that we don't mutate the structures that were passed in. I forget the scenario but I remember there were some bugs in the past that came from dictmerge not being careful to copy on write.

Br4nd0R: done.

cumulusci/core/utils.py
Br4nd0R: thanks for the catch. 🐟

jstvz (approved):
My smoke tests are passing now

PR #3049, opened 2022-01-26
Add Snowfakery datetime handling.

prescod:
# The Problem

Using Snowfakery I am trying to insert a record with a datetime field populated with ${{fake.DateTimeThisYear}}. I am getting this error:

Error: Error on record with id 1: JSON_PARSER_ERROR: Cannot deserialize instance of datetime from VALUE_STRING value 2022-01-01 

00:37:15 or request may be missing a required field at [Source: N/A; line: -1, column: -1] ()

https://trailhead.salesforce.com/trailblazer-community/feed/0D54S00000EAHP9SAP

# Solution

CumulusCI's Snowfakery task  now outputs faked datetimes in a Salesforce-compatible format without additional configuration or code.

### DIFF
--- a/cumulusci/tasks/bulkdata/generate_from_yaml.py
+++ b/cumulusci/tasks/bulkdata/generate_from_yaml.py
@@ -1,10 +1,14 @@
 import os
 import shutil
 from contextlib import contextmanager
+from datetime import datetime
 from pathlib import Path
 from typing import Optional
 
-from snowfakery import generate_data
+from snowfakery import SnowfakeryApplication, generate_data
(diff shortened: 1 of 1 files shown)


davisagli (commented):
Like I said in slack, this feels like something that should be implemented in Snowfakery as an option for how to serialize datetimes, rather than in a Salesforce-specific layer.

Another point is that if there's a need to debug snowfakery's output (like if we end up with timezone-aware Python datetimes for some reason, which it looks like this doesn't handle correctly) then it'll be quite confusing if switching to a different output stream to do that changes what's being written.

PR #3053, opened 2022-01-31
Add 'log' method to robot logginer inside 'Run task' keyword

boakley:
See W-10503175
In short, a while ago we replaced the normal task logger with the robot logger in the 'Run Task' keyword. Unfortunately, the robot logger is missing the 'log' method that normal loggers have.

This monkey patches the robot logger to add the missing function, which takes a numerical level and maps it to one of the few levels supported by robot.  I'm not sure this is the best solution; we probably should create a proper logger that passes information to the robot logger, but this seemed like the fix that required the fewest lines of code and the least amount of testing. 

# Critical Changes
* Fixes a bug that might in rare cases cause the `Run Task` and `Run Task Class` robot keywords to throw an error like `AttributeError: module 'robot.api.logger' has no attribute 'log'`

# Changes

# Issues Closed


jstvz (approved):
LGTM. 

---
There's one more method on `logging`, `critical`, that doesn't map to a method in `robot.api.logger`... _but_ it's relatively uncommon (1/10th as many hits as `logger.warning` in github code search).

Br4nd0R (approved):
LGTM as well!

PR #3055, opened 2022-02-01
Support OrgConfigs using a non-default connected_app service

davisagli:
# Critical Changes

# Changes
- Orgs connected using `cci org connect` now keep track of which `connected_app` service was used to connect them, and will use the same `connected_app` service when getting a fresh access token, even if it isn't currently the default `connected_app` service. This makes it easier to use CumulusCI with multiple `connected_app` services. (Orgs that were connected before this update will continue to use the default `connected_app` service.)
- When connecting a custom `connected_app` service, a default `login_url` can be specified (to use a login URL other than https://login.salesforce.com when connecting orgs using this `connected_app` service).

# Issues Closed
- Fixed a bug in the JSON format output of `cci service list --json`


cumulusci/cli/cci.py
```diff
@@ -120,7 +120,7 @@ def handle_exception(
             traceback.print_exc(file=log_file)  # log stacktrace silently
 
     if should_show_stacktraces and not isinstance(error, USAGE_ERRORS):
-        raise error
+        error_console.print_exception()
```
davisagli, line 123: Unrelated fix to make sure we get rich's nice formatting of stacktraces.

cumulusci/cli/service.py
```diff
@@ -35,7 +34,7 @@ def service_list(runtime, plain, print_json):
 
     console = Console()
     if print_json:
-        console.print(json.dumps(services))
+        console.print_json(data=services)
```
davisagli, line 37: Unrelated bug I found while adjusting test_service.py -- the old way could break lines in the middle of a string, so that the output wasn't parsable as JSON

cumulusci/cli/tests/utils.py
```diff
@@ -13,6 +15,20 @@ def run_click_command(cmd, *args, **kw):
         return cmd.callback(*args, **kw)
 
 
+def run_cli_command(*args, runtime=None, input=None, **kw):
+    """Run a click command with arg parsing and injected CCI runtime object."""
```
davisagli, line 19: Added this as a better way to test click commands, including verifying how the args get parsed and easier ways to provide input and check output. The goal is to gradually switch existing tests over to this and then get rid of run_click_command above.

cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -179,7 +182,7 @@ def set_service(
         self,
         service_type: str,
         alias: str,
-        service_config: Dict,
+        service_config: ServiceConfig,
```
davisagli, line 185: Mistake in the type annotations that I found while adjusting tests

davisagli:
@pconrad-sfdo Please review my changes in docs/connected_orgs.rst as well as the release notes in the PR description.

Br4nd0R (approved):
This looks great. I still haven't QA'ed but am ok with merging assuming you've tested it out.

cumulusci/cli/org.py
```diff
@@ -180,8 +202,15 @@ def org_connect(runtime, org_name, sandbox, login_url, default, global_org):
             "Use the my.salesforce.com version instead"
         )
 
-    sf_client = setup_client(runtime, login_url, sandbox)
-    connect_org_to_keychain(sf_client, runtime, global_org, org_name)
+    connected_app_name = (
+        connected_app_name or runtime.keychain.get_default_service_name("connected_app")
+    )
+    click.echo(f"Connecting org using the {connected_app_name} connected app...")
```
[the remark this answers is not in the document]
davisagli: @Br4nd0R I don't have a good understanding of how they compare. Feels like something we should tackle separately across the various commands?

jstvz (approved):

PR #3059, opened 2022-02-02
Fix Snowfakery's Unique ID feature

prescod:
Snowfakery's Unique ID feature depends upon a parent process passing a uniquifying "context identifier" called a "pid" (because it is the ID for this Snowfakery process).

I dug into why the test case didn't catch the problem but the fix for the test is larger and I don't want to slow down this PR, but I'm working on a separate PR to fix that.

The old code was here:

https://github.com/SFDO-Tooling/CumulusCI/blob/f6777e673eca9978a84f70af365835d307885abb/cumulusci/tasks/bulkdata/snowfakery.py#L498

### DIFF
--- a/cumulusci/tasks/bulkdata/snowfakery_utils/subtask_configurator.py
+++ b/cumulusci/tasks/bulkdata/snowfakery_utils/subtask_configurator.py
@@ -34,7 +34,7 @@ class SubtaskConfigurator:
         #
         # big_ids ensures that the pid is incorporated into the ids.
         plugin_options = {
-            "pid": parts[0],
+            "pid": str(wd.index),
             "big_ids": "True",
         }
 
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/bulkdata/snowfakery_utils/subtask_configurator.py
```diff
@@ -29,6 +29,15 @@ def data_generator_opts(self, working_dir, recipe_options=None, *args, **kwargs)
         parts = name.rsplit("_", 1)
         batch_size = int(parts[-1])
 
+        # The pid is the tool that ensures that every uniqueid is
+        # unique even across different processes.
+        #
+        # big_ids ensures that the pid is incorporated into the ids.
+        plugin_options = {
+            "pid": str(wd.index),
+            "big_ids": "True",
```
jofsky, line 38: Should we not use a name relating to pid?

jofsky (approved):

cumulusci/tasks/bulkdata/snowfakery_utils/subtask_configurator.py
prescod: @jofsky yes we could consider renaming in the future, although this name is documented for now.

PR #3064, opened 2022-02-08
Clean up message printing from Snowfakery in CCI update_data task

prescod:
CCI prints out "info messages" about CSV files. Those info messages are confusing in the case where the CSV files are just tempfiles as they are in the update_data task context.

Snowfakery allows applications to inject their own message printing, so this message printing filters those out.


cumulusci/tasks/bulkdata/update_data.py
```diff
@@ -224,3 +225,16 @@ def _object_description(self, obj):
...
+class CumulusCIUpdatesApplication(SnowfakeryApplication):
+    """Takes over Snowfakery logging so CumulusCI can control it"""
+
+    def __init__(self, logger) -> None:
+        self.logger = logger
+        super().__init__()
+
+    def echo(self, message, *args, **kwargs):
+        # skip CSV creation messages
+        if not message.startswith("Created "):
```
davisagli, line 239: This seems a bit hacky and prone to unintentionally suppressing messages about other things being created. Wouldn't it make more sense to pass a flag to snowfakery which tells it not to emit the message in the first place?

prescod: @davisagli Either way the parent app and Snowfakery would need some way of communicating which messages to send or not send. I did tighten up the test so that it's essentially impossible for it to match anything I don't want it to.

cumulusci/tasks/bulkdata/update_data.py
```diff
@@ -224,3 +226,19 @@ def _object_description(self, obj):
...
+
+    def __init__(self, logger) -> None:
+        self.logger = logger
+        super().__init__()
+
+    def echo(self, message, *args, **kwargs):
+        # skip CSV creation messages
+        if not self.MATCHER.match(message):
+            print(repr(message))
+            self.logger.info(message)
```
davisagli, line 244: Shouldn't we leave whether to print up to how the logging system is configured?

prescod: Sorry I was debugging the regexp and forgot to take out the print statement.

a69eb911

davisagli (approved):

PR #3066, opened 2022-02-09
Switch mc package deploy task to new endpoints

Br4nd0R:
# Changes
* The `deploy_marketing_cloud_package` task now utilizes the new API endpoint for deployments.

## Dev Notes
Per a discussion I had with Sam Varga, all stacks should now be swapped over to using these new endpoints:

> All stacks are live now from our side. The MC R2 release actually got pushed back to the 18th, but with your server-side integration that doesn't matter. That's only for the UI oauth flow endpoints.

I have tested and confirmed the deploy task works with the new endpoints for our learning account which resides on S4.

### DIFF
--- a/cumulusci/tasks/marketing_cloud/deploy.py
+++ b/cumulusci/tasks/marketing_cloud/deploy.py
@@ -72,8 +72,14 @@ class MarketingCloudDeployTask(BaseMarketingCloudTask):
             "Authorization": f"Bearer {self.mc_config.access_token}",
             "SFMC-TSSD": self.mc_config.tssd,
         }
-        stack_key = get_mc_stack_key(self.mc_config.tssd, self.mc_config.access_token)
-        self.endpoint = self.options.get("endpoint") or MCPM_ENDPOINT.format(stack_key)
+        custom_endpoint = self.options.get("endpoint")
+        self.endpoint = (
+            custom_endpoint
+            if custom_endpoint
+            else MCPM_ENDPOINT.format(
+                get_mc_stack_key(self.mc_config.tssd, self.mc_config.access_token)
+            )
+        )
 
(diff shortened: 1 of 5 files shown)


cumulusci/tasks/marketing_cloud/deploy.py
```diff
@@ -73,12 +72,23 @@ def _run_task(self):
...
             "SFMC-TSSD": self.mc_config.tssd,
         }
+        custom_endpoint = self.options.get("endpoint")
+        self.endpoint = (
+            custom_endpoint
+            if custom_endpoint
+            else MCPM_ENDPOINT.format(
+                get_mc_stack_key(self.mc_config.tssd, self.mc_config.access_token)
+            )
+        )
```
Br4nd0R, line 82: I know this is a little gross. I'm trying to cheat so I don't need an if/else block which would require another test.

cumulusci/tasks/marketing_cloud/tests/test_deploy.py
```diff
@@ -52,21 +56,31 @@ def task_without_custom_inputs(project_config):
...
 
+@pytest.fixture
+def mocked_responses():
+    with responses.RequestsMock() as rsps:
+        rsps.add(
+            "GET",
+            f"https://{TEST_TSSD}.auth.marketingcloudapis.com/{MC_API_VERSION}/userinfo",
+            json={"organization": {"stack_key": STACK_KEY}},
+        )
+        yield rsps
```
Br4nd0R, line 71: TIL: mocked responses as pytest fixtures

davisagli (approved):
Looks good and seems to work.

codexmarcel (approved):
@davisagli's approval is authoritative since I paired on this with @Br4nd0R.

PR #3070, opened 2022-02-12
Make org_list resilient and make error messages better

prescod:
# Changes

cci commands are more resilient in the face of corrupted or impossible to decode org config files.


cumulusci/cli/org.py
```diff
@@ -348,12 +348,22 @@ def org_info(runtime, org_name, print_json):
 )
 @pass_runtime(require_project=False, require_keychain=True)
 def org_list(runtime, json_flag, plain):
+    def _get_org_safe(org):
+        try:
+            return runtime.keychain.get_org(org)
+        except Exception as e:
```
jstvz, line 354: Why not narrow which exceptions we catch here, e.g. `except KeychainKeyNotFound` here or `UnicodeDecodeError` below?

jstvz (approved):

PR #3073, opened 2022-02-14
update dependencies including github3.py

davisagli:
Also includes some tidying of github actions configuration

# Critical Changes

# Changes

# Issues Closed


davisagli:
Looking at the tests that broke...

setup.py
```diff
@@ -33,13 +33,15 @@ def find_packages(path=["."], prefix=""):
             continue
         else:
             req = req.split(" ")[0]
+            # Work around normalized name of github3.py distribution
+            req = req.replace("github3-py", "github3.py")
```
jstvz, line 37: Huh. Figured setuptools/wheel would do the right thing here.

requirements/prod.txt
```diff
@@ -30,18 +30,15 @@ cryptography==36.0.1
...
     # via -r requirements/prod.in
 faker==12.0.0
     # via
     #   -r requirements/prod.in
     #   snowfakery
-fs==2.4.14
+fs==2.4.15
     # via -r requirements/prod.in
-github3.py==3.0.0
+github3-py==3.1.0
```
jstvz, line 41: 😑

jstvz (approved):

requirements/prod.txt
davisagli: I think it's from this change in pip-tools: https://github.com/jazzband/pip-tools/pull/1572

davisagli: I think it's from this change in pip-tools: https://github.com/jazzband/pip-tools/pull/1572

setup.py
davisagli: Yeah -- I think there's probably a bug somewhere in there, but I hit my timebox for locating it and decided to do this instead.

PR #3074, opened 2022-02-14
Derive the MC userinfo endpoint from the marketing_cloud service's rest_instance_url

davisagli:
instead of from the oauth2_client service's auth_uri.
The latter is not available in the context of a tubular worker.

Other changes:
- Moved get_user_info to be a method of the MarketingCloudServiceConfig, because it seemed to fit nicely there
- Adjusted tests to use a real MarketingCloudServiceConfig instead of a mock to make sure we're actually testing it

No release notes because this is a fix for an unreleased bug.


codexmarcel (approved):
nice sleuthing

PR #3076, opened 2022-02-15
Fix test_record_count test method by using queue.Queue instead of multiprocessing.Queue

prescod:
# Changes for Release Notes:

None.

Fix test_record_count test method by using queue.Queue instead of multiprocessing.Queue

Attached file shows strange behaviours of multiprocessing.Queue. The downside of queue.Queue is that it only works for threads, not processes. Luckily Snowfakery currently only needs results reported "in realtime" from threads, so queue.Queue will work for our current needs.

[testqueues.py.txt](https://github.com/SFDO-Tooling/CumulusCI/files/8065064/testqueues.py.txt)

### DIFF
--- a/cumulusci/utils/parallel/task_worker_queues/parallel_worker.py
+++ b/cumulusci/utils/parallel/task_worker_queues/parallel_worker.py
@@ -135,20 +135,22 @@ class TaskWorker:
                 self.subtask()
                 logger.info(str(self.subtask.return_values))
                 logger.info("SubTask Success!")
-                self.results_reporter.put(
-                    {
-                        "status": "success",
-                        "results": self.subtask.return_values,
-                        "directory": str(self.working_dir),
-                    }
-                )
+                if self.results_reporter:
+                    self.results_reporter.put(
+                        {
+                            "status": "success",
+                            "results": self.subtask.return_values,
+                            "directory": str(self.working_dir),
+                        }
+                    )
             except BaseException as e:
                 logger.info(f"Failure detected: {e}")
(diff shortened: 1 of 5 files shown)


cumulusci/utils/parallel/task_worker_queues/parallel_worker.py
```diff
@@ -135,20 +135,22 @@ def run(self):
...
+                    )
             except BaseException as e:
                 logger.info(f"Failure detected: {e}")
                 self.save_exception(e)
                 self.failures_dir.mkdir(exist_ok=True)
                 logfile.close()
                 shutil.move(str(self.working_dir), str(self.failures_dir))
-                self.results_reporter.put({"status": "error", "error": str(e)})
+                if self.results_reporter:
+                    self.results_reporter.put({"status": "error", "error": str(e)})
```
davisagli, line 153: It's fine as is, but you could avoid all these conditionals by implementing a NullResultsReporter that doesn't actually do anything with the messages passed to it

davisagli (approved):

cumulusci/utils/parallel/task_worker_queues/parallel_worker.py
prescod: I did debate that internally but I thought it might be equally confusing if someone was reading the code and it looked like something was being put into the queue but it wasn't "appearing" anywhere. Since there are only two (?) places with the conditional, I thought it was clearer this way.

davisagli (approved):

PR #3083, opened 2022-02-16
Fix test by removing filesystem-related race conditioon

prescod:
Fix a bug where too few records are processed due to a race condition where files are moved around while the controller process is looking at them.

Use a lock to avoid that. My previous solution to this problem had performance consequences.

Thousands of iterations of the flaky test have been run.


davisagli:
Proposed release note for the Issues Closed section: "Fixed an edge case in the `snowfakery` task where it might stop before processing all records."

PR #3086, opened 2022-02-16
Fix windows path

jstvz:
Changes the command run on windows to fix a path issue. ) ~I thought we already had a 3PP for pytest-cov, but we don't so it's a draft for now~.

- [x] Split coveralls into a second step

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/cli/tests/test_service.py
+++ b/cumulusci/cli/tests/test_service.py
@@ -1,4 +1,5 @@
 import json
+from unittest import mock
 
 import click
 import pytest
@@ -10,7 +11,8 @@ from cumulusci.core.tests.utils import EnvironmentVarGuard
 from .utils import run_cli_command
 
 
-def test_service_list():
+@mock.patch("cumulusci.cli.service.CliTable")
+def test_service_list(table_mock):
     runtime = BaseCumulusCI(
         config={
             "services": {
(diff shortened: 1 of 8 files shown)


prescod (dismissed):
Do you think this will fail more noisily now or is that another step?

jstvz:
> Do you think this will fail more noisily now or is that another step?

@prescod Failing noisily now. Looks like this change revealed [three test failures](https://github.com/SFDO-Tooling/CumulusCI/runs/5224051383?check_suite_focus=true), so good catch.

@boakley There are two failures in `TestRobotLibDoc`, `test_to_tuples` and `test_csv`... looks like we're getting full paths. I can take a look, but pinging you in case you see an obvious fix.

prescod (dismissed):

cumulusci/cli/tests/test_service.py
```diff
@@ -26,19 +28,18 @@ def test_service_list():
...
+    table_mock.assert_called_once_with(
+        [
+            ["Default", "Type", "Name", "Description"],
+            [True, "bad", "bad_alias", "Unconfigured Service"],
+            [False, "something_else", "", "something else"],
+            [False, "test", "test2_alias", "Test Service"],
+            [True, "test", "test_alias", "Test Service"],
+        ],
+        title="Services",
+        dim_rows=[2],
```
davisagli, line 42: I had intentionally changed it away from this approach because I couldn't tell by looking at the test whether it produces correct output.

I guess changing it back is motivated by the output not being the same on Windows? In which case this is probably the most pragmatic solution. But another option might be to run the output through a normalizer before making an assertion about it.

davisagli (dismissed):

davisagli (dismissed):

davisagli (approved):

davisagli (approved):

PR #3092, opened 2022-02-23
Flatten cache directory structure

davidmreed:
# Critical Changes

# Changes

# Issues Closed

- We improved our caching mechanisms to better handle projects whose cross-project sources themselves have cross-project sources.


cumulusci/core/source/github.py
```diff
@@ -124,18 +125,12 @@ def resolve(self):
...
-                self.project_config.cache_dir
-                / "projects"
-                / self.repo_name
-                / self.commit
-            )
-        if not path.exists():
-            path.mkdir(parents=True)
+        # fs.path, which we use in open_cache, doesn't work with Pathlib path objects
+        with self.project_config.open_cache(
+            fs.path.join("projects", self.repo_name, self.commit)
```
jstvz, line 132: I don't think this comment is accurate. I see `open_cache` is using `project_config.cache_dir`, which is a `Path` instance, and the function signature for `open_fs_resource` accepts `Union[str, Path, "FSResource"]`.

jstvz (changes requested):
LGTM, ± for a comment mismatch.

 Smoke test OK:
```console
❯ tree -dL 4 .cci
.cci
└── projects
    ├── CaseMan
    │   └── 6f7381f8c2177fd3f6f05bab687f6134d39f6ec9
    │       ├── datasets
    │       └── unpackaged
    ├── CaseMan-Exp-Cloud
    │   └── 8318cce0036a57651e9cc2ae50c16f87381dd618
    │       ├── datasets
    │       └── unpackaged
    └── pmm
        └── 28001085c95b56da53edacbcaaa7625b4c81d59c
            ├── data
            └── unpackaged

35 directories
```

cumulusci/core/source/github.py
```diff
@@ -146,16 +141,15 @@ def fetch(self, path=None):
...
         project_config = self.project_config.construct_subproject_config(
             repo_info={
                 "root": os.path.realpath(path),
                 "owner": self.repo_owner,
                 "name": self.repo_name,
                 "url": self.url,
                 "commit": self.commit,
-            }
+            },
+            cache_dir=self.project_config.cache_dir,
```
davisagli, line 152: Should we put this in the implementation of `construct_subproject_config` so that all subprojects inherit the cache dir, or is there a reason we only want that to happen for subprojects constructed from a github source? The local folder source (folder.py) also calls construct_subproject_config

cumulusci/core/source/github.py
davidmreed: I'm trying to remember what breakage I found - I think it's actually that it doesn't work with Pathlib paths where one or more path components do not yet exist (?). I'll remove the comment in any case.

cumulusci/core/source/github.py
davidmreed: THat's a good point. There's even an edge case where a local folder source could itself have a GitHub source that needs access to the parent cache.

davidmreed: Done (plus test changes) in [58f733b](https://github.com/SFDO-Tooling/CumulusCI/pull/3092/commits/58f733b1f35bbc6a82f488639316104d5ccf9e14)

jstvz (approved):

PR #3093, opened 2022-02-23
Add interactive mode for update_dependencies

davidmreed:
# Critical Changes

# Changes

 - The `update_dependencies` and `install_managed` tasks accept the options `interactive` and `base-package-url-format`, which pause the automation and wait for the user to confirm the list of packages to be installed.

# Issues Closed


cumulusci/tasks/salesforce/install_package_version.py
```diff
@@ -145,30 +165,45 @@ def _run_task(self):
...
+
+        if self.options.get("interactive", False):
+            if dep.version_id:
+                package_desc = self.options["base_package_url_format"].format(
+                    dep.version_id
+                )
+            else:
+                package_desc = str(dep)
+            self.logger.info("Package to install: {}".format(package_desc))
+            if not click.confirm("Continue to install dependencies?", default=True):
```
davisagli, line 184: Should we try to validate the `interactive` option to throw an error if it's True but sys.stdin.isatty() is False?

davidmreed: Seems reasonable, but I don't know if there are any Windows platform-specific issues with detecting an interactive environment. Any idea?

davidmreed: Some quick googling suggests it false-positives on Windows.

davisagli: Oh yeah, glad you thought to check that

jofsky (approved):
Looks good!

PR #3098, opened 2022-02-28
Feature/robot go to page fix

boakley:
The `Go to page` keyword passes all of its arguments to the method `_go_to_page` for a page object. This enhances our built-in generic page objects for Listing, Detail, and Home pages to accept a locator. This locator is then passed on to the keyword `Wait until loading is complete` instead of the default locator used by that keyword.

# Critical Changes

# Changes
* Robot tests can now pass a locator when calling `Go To Page` for one of our bulit-in page objects (Listing, Detail, Home), and the keyword will wait until that locator is visible. This is mostly useful for custom pages where you need to wait for an iframe or custom component to be available.

# Issues Closed


cumulusci/robotframework/tests/test_pageobjects.py
```diff
@@ -36,6 +41,11 @@
 # peasy to import by file path
 importer = robot.utils.Importer()
 
+# These tests don't need an arg, so mocking it out will prevent some
```
prescod, line 44:
```suggestion
# These tests don't need an org, so mocking it out will prevent some
```

prescod (commented):
One small comment. The rest looks fine.

boakley:
> One small comment. The rest looks fine.

That was an eagle-eyed observation. Thanks for catching that!

prescod (approved):

PR #3101, opened 2022-03-01
Black is the new black.

prescod:
Reformat files to black 22.1.0

No relevance to release notes.


jstvz (approved):
💭: We should probably add a `.git-ignore-revs` to the repo at some point.

PR #3102, opened 2022-03-01
Code cleanups: remove org_config in MP and sf_version

prescod:
1. Don't pass org_config to mapping_parser functions. Pass salesforce_client instead. The only reason they need an org_config is for the salesforce_client, and passing in an existing one is cleaner, more performant and easier to mock.

2. Start the process of centralizing the "preferred" Salesforce version for test cases in a single place.

Nothing relevant to release notes.

### DIFF
--- a/cumulusci/tasks/bulkdata/tests/utils.py
+++ b/cumulusci/tasks/bulkdata/tests/utils.py
@@ -7,13 +7,18 @@ from cumulusci.tasks.bulkdata.step import (
 )
 from cumulusci.tests import util as cci_test_utils
 
+current_sf_version = "48.0"
+
 
 def _make_task(task_class, task_config):
     task_config = TaskConfig(task_config)
     universal_config = UniversalConfig()
     project_config = BaseProjectConfig(
         universal_config,
-        config={"noyaml": True, "project": {"package": {"api_version": "46.0"}}},
+        config={
+            "noyaml": True,
+            "project": {"package": {"api_version": current_sf_version}},
+        },
     )
     keychain = BaseProjectKeychain(project_config, "")
     project_config.set_keychain(keychain)
@@ -31,6 +36,7 @@ class FakeBulkAPI:
 
     next_job_id = 0
     next_batch_id = 0
+    endpoint = f"https://example.my.salesforce.com/services/async/{current_sf_version}"
 
     @classmethod
     def create_job(cls, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


cumulusci/tasks/bulkdata/tests/utils.py
```diff
@@ -7,13 +7,18 @@
 )
 from cumulusci.tests import util as cci_test_utils
 
+current_sf_version = "48.0"
```
davisagli, line 10: Can we call this CURRENT_SF_VERSION to match our usual convention for names that are used like constants?

davisagli: And actually, for maximum clarity, how about CURRENT_SF_API_VERSION to be clear it's not a release number?

cumulusci/tests/util.py
```diff
@@ -19,7 +19,7 @@
     UniversalConfig,
 )
 from cumulusci.core.keychain import BaseProjectKeychain
-from cumulusci.tasks.bulkdata.tests import utils as bulkdata_utils
+from cumulusci.tasks.bulkdata.tests.utils import FakeBulkAPI, current_sf_version
```
davisagli, line 22: Can we move those here? The current organization feels backwards since these are not intended for use only by the bulkdata tests.

davisagli:
There's a lint error. Maybe a version mismatch between black in requirements.txt and black in .pre-commit.yaml?

cumulusci/tests/util.py
prescod: It's a mutual import issue that I didn't want to fight with.

prescod: I can remove the circular import and refactor things in a separate PR.

cumulusci/tasks/bulkdata/tests/utils.py
prescod: @davisagli Renamed to CURRENT_SF_API_VERSION

cumulusci/tasks/bulkdata/tests/utils.py
```diff
@@ -7,13 +7,18 @@
 )
 from cumulusci.tests import util as cci_test_utils
 
+CURRENT_SF_API_VERSION = "48.0"
```
davisagli, line 10: @prescod should we update it to the actual current version, 54.0? Or do that separately?

davisagli (approved):

PR #3105, opened 2022-03-03
Add support for additional Unlocked Package install options

davidmreed:
# Critical Changes

# Changes

- We added the `apex_compile_type` and `upgrade_type` options for performing Unlocked Package installations.

# Issues Closed


cumulusci/salesforce_api/tests/test_package_install.py
```diff
@@ -183,4 +185,5 @@ def test_package_install_options_from_task_options():
         name_conflict_resolution=NameConflictResolution.RENAME,
         password="foo",
         security_type=SecurityType.PUSH,
+        apex_compile_type=ApexCompileType.PACKAGE,
     )
```
davisagli, line 189: Include upgrade_type in the test too?

davisagli (changes requested):

cumulusci/salesforce_api/tests/test_package_install.py
davidmreed: I deliberately left it out to validate that there wasn't any negative behavior with its nullability.

davidmreed: Perhaps I should make that test explicit.

davisagli: Ah okay. Wouldn't hurt to have both tests but we don't need to hold this up.

davisagli (dismissed):
@davidmreed approved, but did you want to update version.txt to make a dev release?

davisagli (approved):

PR #3109, opened 2022-03-09
Snowfakery crash when loading into "directly connected" ("persistent") orgs

prescod:
Snowfakery had a bug introduced when the get_service calling code changed in 40520bee4c1c .

This PR changes the method signature for SubprocessKeyChain.get_service()

It's hard to replicate in integration tests because it depends on a connected app-based org_config as opposed to the sfd-based org_config

W-10811651

# Issues Closed

Fixed a bug running the `snowfakery` task in parallel mode with orgs connected using a Connected App.

### DIFF
--- a/cumulusci/utils/parallel/task_worker_queues/tests/test_parallel_worker.py
+++ b/cumulusci/utils/parallel/task_worker_queues/tests/test_parallel_worker.py
@@ -12,7 +12,7 @@ from cumulusci.core.config import BaseProjectConfig, OrgConfig, UniversalConfig
 from cumulusci.tasks.util import Sleep
 from cumulusci.utils.parallel.task_worker_queues.parallel_worker import (
     ParallelWorker,
-    SubprocessKeyChain,
+    SubprocessKeychain,
     TaskWorker,
     WorkerConfig,
 )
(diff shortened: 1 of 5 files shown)


davisagli (changes requested):
`alias` has actually been part of the signature for `get_service` since support for default services was added in https://github.com/SFDO-Tooling/CumulusCI/pull/2499 -- before SubprocessKeyChain was implemented! But it makes sense that we didn't run into it until OrgConfig started passing a connected_app alias in _refresh_token.

I'll approve this once you update the PR description to write a release note.

Would it make sense to move this keychain class into cumulusci.core.keychain next to the other keychains to make sure we don't overlook it when doing future work on keychains?

cumulusci/utils/parallel/task_worker_queues/tests/test_parallel_worker.py
```diff
@@ -343,9 +343,9 @@ def test_worker__cannot_move_to_outdir(self):
 # Also we usually mock refresh_oauth_token which is what would
 # invoke this. In other words, using integration tests to cover this
 # function is far easier than using unit test.
-class TestSubprocessKeyChain:
+class TestSubprocessKeychain:
```
davisagli, line 346: Ideally this test would move into a test module in the same package as the code being tested, but it hardly seems worth the effort.

davisagli (approved):

davisagli (approved):

davisagli:
@prescod I edited the release note a bit and made sure it's indicated under the Issued Closed heading.

PR #3113, opened 2022-03-10
Consolidate robot_ui jobs

jstvz:
Refactors the robot_ui_* jobs to use a matrix in the same way that unit_test does.

Another change could be to use sfdx via actions/setup-node, using caching to shave a few seconds from its installation step.


davisagli:
Nice, I didn't realize it was possible to use a matrix strategy with custom axes/variables.

davisagli (approved):

PR #3114, opened 2022-03-10
Add a way to verify certs using system trust stores instead of certifi

davisagli:
# Critical Changes

# Changes
- CumulusCI now provides an option to verify server certificates using root CA certs loaded from the system, instead of the default set of CA certs bundled with the Python requests library. This is considered experimental, so must be opted in by setting the CUMULUSCI_SYSTEM_CERTS environment variable to "True"

# Issues Closed


davidmreed:
@davisagli Do you already have a core-developer collaborator on this one, or would it be helpful to ask a friend to help us validate the behavior hands-on?

PR #3117, opened 2022-03-15
rewrote the logic for the custom locator "label:"

boakley:
This fixes the following work items:
* [W-10809197](https://gus.lightning.force.com/a07EE00000sHyKqYAK) - "label:" locator strategy doesn't work for labels with spans
* [W-9070673](https://gus.lightning.force.com/a07AH0000009R9yYAE) - Performance improvement in `Input form data` for non-lightning components

The logic for transforming a locator such as `label:First Name` has changed. Previously the code would look for a lightning component with a single XPath, and fall back to finding an input or textarea only once that xpath timed out.  Now it looks for the label first, and from that it tries to find the component it is associated with. This is the
root of the speed performance.

Also, the xpath for finding the related input item has been tweaked to work for more styles of markup. I've added a suite of tests for the variations I've observed in the wild.

# Critical Changes

# Changes
* Improvements to the custom locator strategy used by the robot framework keyword `input form data` to find form fields and lightning components based on label text. This locator strategy (`label:`) is now up to 10x faster at finding non-lightning form components, and is able to find elements with more complicated labels (eg: labels with text inside of nested spans).

# Issues Closed


davisagli:
@boakley could you please edit the release note to include a mention of Robot Framework so that the context is clearer?

davidjray (approved):
LGTM

PR #3120, opened 2022-03-18
Feature/robot pageobject alias

boakley:
This is the implementation for [W-9966502](https://gus.lightning.force.com/a07EE00000Bm2xIYAR)

I originally thought this story was going to add a new feature, but after digging into the code I realized we had the feature all along, but it was slightly buggy. This fixes the bug and then adds some improvements to the public facing docs to make it more clear how to use this feature.

# Critical Changes

# Changes
* fixed a bug that prevented the use of aliases for robot framework page objects
* improved the documentation for using robot framework page objects and the ``pageobject`` decorator
# Issues Closed


docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
 The ``pageobject`` Decorator
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Page objects are normal Python classes that use the ``pageobject`` decorator provided by CumulusCI. Unlike traditional Robot Framework keyword libraries, you can define multiple sets of keywords in a single file.
+Page objects are normal Python classes that use the ``pageobject`` decorator provided by CumulusCI. Unlike traditional Robot Framework keyword libraries, you can easily define and use keywords in multiple classes within a single file.
```
pconrad-sfdo, line 120: I recommend removing "easily." For one thing, how easily? Adverbs make things less specific. For another, for that one reader out there who doesn't find it easy, we're kind of making them feel bad :D

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+    from cumulusci.robotframework.pageobjects import ListingPage, pageobject
 
+    @pageobject(page_type='Listing', object_name='CustomObject__c')
+    class CustomObjectListingPage(ListingPage):
+        ...
+
+Using object aliases
+""""""""""""""""""""
+
+Within a test, if you want to refer to the page object with a more human-readable name such as ``Custom Object`` rather than ``CustomObject__c`` you can do so by setting ``object_name`` to ``Custom Object`` and then defining ``_object_name`` in the class, like in the following example.
```
pconrad-sfdo, line 137:
```suggestion
Within a test, if you want to refer to the page object with a more human-readable name such as ``Custom Object`` rather than ``CustomObject__c`` you can do so by setting ``object_name`` to ``Custom Object`` and then defining ``_object_name`` in the class, as in the following example.
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+.. code-block:: python
+
+    from cumulusci.robotframework.pageobjects import ListingPage, pageobject
+
+    @pageobject(page_type = 'Listing', object_name = 'My Object')
+    class CustomObjectListingPage(ListingPage):
+        _object_name = 'MyObject__c'
+        ...
+
+By using an alias, you can reference the page object with either the alias or the actual object name. For example, the following two uses of ``Go to page`` are identical:
```
pconrad-sfdo, line 148:
```suggestion
By using an alias, you can reference the page object with either the alias or the actual object name. For example, if ``object_name`` is set as described above, the following two uses of ``Go to page`` are identical:
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+
+    Go to page  Listing  My Object
+    Go to page  Listing  MyObject__c
 
 
 Page Object Base Classes
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-Presently, CumulusCI provides the following base classes, which should be used for all classes that use the ``pageobject`` decorator:
+Presently, CumulusCI provides the following base classes, which should be used for all classes that use the ``pageobject`` decorator. These base classes may be imported from ``cumulusci.robotframework.pageobjects``.
```
pconrad-sfdo, line 159:
```suggestion
CumulusCI provides the following base classes, which should be used for all classes that use the ``pageobject`` decorator. You can import these base classes from ``cumulusci.robotframework.pageobjects``.
```
If you don't like "You can import," I'm okay with "These base classes can be imported."

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
 
-* `cumulusci.robotframework.pageobjects.BasePage <Keywords.html#file-cumulusci/robotframework/pageobjects/BasePageObjects.py>`_ A generic base class used by the other base classes. Use the ``BasePage`` class for creating custom page objects when none of the other base classes make sense.
-    * The ``BasePage`` adds the ``Log current page object`` keyword to every page object. This keyword is most useful when debugging tests. It will add to the log information about the currently loaded page object.
+* `cumulusci.robotframework.pageobjects.BasePage <Keywords.html#file-cumulusci/robotframework/pageobjects/BasePageObjects.py>`_ A generic base class used by the other pageobject classes. Use the ``BasePage`` class for creating custom page objects when none of the other base classes make sense.
```
pconrad-sfdo, line 161: What are the circumstances that determine when none/some of the other base classes make sense?

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
 * ``cumulusci.robotframework.pageobjects.DetailPage``: A class for a page object that represents a detail page.
 * ``cumulusci.robotframework.pageobjects.HomePage``: A class for a page object that represents a home page.
 * ``cumulusci.robotframework.pageobjects.ListingPage``: A class for a page object that represents a listing page.
 * ``cumulusci.robotframework.pageobject.NewModal``: A class for a page object that represents the "new object" modal.
 * ``cumulusci.robotframework.pageobject.ObjectManagerPage``: A class for interacting with the object manager.
 
 
+Common page object attributes
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+When using the decorator and inheriting from one of the page object base classes, your class will inherit the following attributes and properties.
```
pconrad-sfdo, line 172:
```suggestion
When using the decorator and inheriting from one of the page object base classes, your class inherits the following attributes and properties.
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+
+* ``self._object_name``: The name of the object related to the class. If the class does not define this property, it will be set to the value provide as the  ``object_name`` parameter to the ``pageobject`` decorator. Note: do not add the namespace prefix in the decorator. This attribute will automatically add the prefix from the ``cumulusci.yml`` file when necessary.
```
pconrad-sfdo, line 174:
```suggestion
* ``self._object_name``: The name of the object related to the class. If the class does not define this property, it is set to the value provided as the  ``object_name`` parameter to the ``pageobject`` decorator. Note: do not add the namespace prefix in the decorator. This attribute automatically adds the prefix from the ``cumulusci.yml`` file when necessary.
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+* ``self.object_name``: this is a property that combines the ``_object_name`` attribute along with the namespace returned by the ``get namespace prefix`` keyword from the CumulusCI library. If there is no namespace this returns the value of the ``_object_name`` attribute.
```
pconrad-sfdo, line 175:
```suggestion
* ``self.object_name``: A property that combines the ``_object_name`` attribute with the namespace returned by the ``get namespace prefix`` keyword from the CumulusCI library. If there is no namespace, this returns the value of the ``_object_name`` attribute.
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+* ``self.builtin``: A reference to the Robot Framework ``BuiltIn`` library that can be used to directly call built-in keywords. Any built-in keyword can be called by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.builtin.log`` and ``self.builtin.get_variable_value``).
```
pconrad-sfdo, line 176:
```suggestion
* ``self.builtin``: A reference to the Robot Framework ``BuiltIn`` library that you can use to directly call built-in keywords. You can call any built-in keyword by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.builtin.log`` and ``self.builtin.get_variable_value``).
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+* ``self.cumulusci``: A reference to the CumulusCI keyword library. Call any keyword in this library by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.cumulusci.get_org_info``).
```
pconrad-sfdo, line 177:
```suggestion
* ``self.cumulusci``: A reference to the CumulusCI keyword library. You can call any keyword in this library by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.cumulusci.get_org_info``).
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+* ``self.salesforce``: A reference to the Salesforce keyword library. Call any keyword in this library by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.salesforce.wait_until_loading_is_complete``).
```
pconrad-sfdo, line 178:
```suggestion
* ``self.salesforce``: A reference to the Salesforce keyword library. You can call any keyword in this library by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.salesforce.wait_until_loading_is_complete``).
```

docs/robot_advanced.rst
```diff
@@ -117,32 +117,68 @@ With the ``PageObjects`` library, you can define classes that represent page obj
...
+* ``self.selenium``: A reference to SeleniumLibrary. Call any keyword in this library by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.selenim.wait_until_page_contains_element``).
```
pconrad-sfdo, line 179:
```suggestion
* ``self.selenium``: A reference to SeleniumLibrary. You can call any keyword in this library by converting the name to all lowercase, and replacing all spaces with underscores (such as ``self.selenim.wait_until_page_contains_element``).
```

docs/robot_advanced.rst
boakley: Thanks for the suggestion!

docs/robot_advanced.rst
boakley: Thanks for pointing this out. I've tried rewording it to make it slightly more clear. My assumption is that if someone is experienced enough to be creating custom page objects, they'll understand the concept of base classes.

prescod (approved):

PR #3126, opened 2022-03-23
Load certs from the mac keychain but don't use securetransport

davisagli:
[update to #3114; no additional release note needed]

# Critical Changes

# Changes

# Issues Closed

### DIFF
--- a/cumulusci/utils/http/requests_utils.py
+++ b/cumulusci/utils/http/requests_utils.py
@@ -1,4 +1,6 @@
+import functools
 import os
+import subprocess
 import sys
 from json import JSONDecodeError
 
@@ -30,8 +32,7 @@ def init_requests_trust():
       (perhaps via the REQUESTS_CA_BUNDLE environment variable),
       we'll continue to honor that.
 
-    - On macOS, verify certificates via the SecureTransport API,
-      which is aware of the system keychain.
+    - On macOS, fetch CA certs from the system keychain.
 
     - On other platforms, verify certificates using urllib3's default approach,
       which loads CAs using SSLContext.load_default_certs()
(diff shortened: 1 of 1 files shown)


cumulusci/utils/http/requests_utils.py
```diff
@@ -66,3 +76,23 @@ def cert_verify(self, conn, *args, **kw):
             conn.ca_certs = None
 
     HTTPAdapter.cert_verify = cert_verify
+
+
+def get_macos_ca_certs() -> str:
+    # get certs from both the default keychains and the SystemRootCertificates keychain
```
prescod, line 82: Docstring instead?

cumulusci/utils/http/requests_utils.py
```diff
@@ -66,3 +76,23 @@ def cert_verify(self, conn, *args, **kw):
...
+            "-p",
+        ]
+        if path:
+            args.append(path)
+        result += subprocess.run(
+            args,
+            stdout=subprocess.PIPE,
+            encoding="latin-1",
+        ).stdout
+    return result
```
prescod, line 98:
```suggestion
def get_cert_from_path(path: str = None):
    """Get cert from a path or the default if path is None"""
    args = [
        "security",
        "find-certificate",
        "-a",
        "-p",
    ]
    if path:
        args.append(path)
    return subprocess.run(
        args,
        stdout=subprocess.PIPE,
        encoding="latin-1",
    ).stdout


def get_macos_ca_certs() -> str:
    """Get certs from both the default keychains and the SystemRootCertificates keychain"""
    return get_cert_from_path(None) + get_cert_from_path(
        "/System/Library/Keychains/SystemRootCertificates.keychain"
    )

```

prescod (approved):
Made some suggestions, but not a show-stopper.

davidmreed (dismissed):

davidmreed (approved):

PR #3129, opened 2022-03-24
Deprecate dx_push task

prescod:
# Critical Changes

The `dx_push` task is deprecated and should not be used anymore. There is currently no plan to remove it, however.


cumulusci/cumulusci.yml
```diff
@@ -259,7 +259,7 @@ tasks:
             command: "force:source:pull"
         group: Salesforce DX
     dx_push:
-        description: Uses sfdx to push the force-app directory metadata into a scratch org
+        description: "DEPRECATED: Use the `deploy` command to deploy metadata instead. Use the `dx` command if you nned to run `force:source:push` for some reason."
```
pconrad-sfdo, line 262:
```suggestion
        description: "DEPRECATED: Use the `deploy` command to deploy metadata instead. Only use the `dx` command if you need to run `force:source:push` for some reason."
```

prescod: The typo is fixed in 61abe4dc3.

The bit about "Only use" is probably based on a misunderstanding.

davidmreed:
I am not convinced we should deprecate this task. I'd like to talk to more product developers first. It has a very legitimate use case - making it easy to do a source push, which is much faster than a full deploy, during development.

prescod:
@davidmreed : that use-case wasn't mentioned in Slack but I don't care one way or the other as long as the documentation is clear. If you want to document it as being of use in that narrow situation that's fine, but the documentation does not make clear when to use "deploy" and when to use "dx_push" and this PR was prompted by a support request by a confused user.

davidmreed:
Closing in favor of projected enhancements to the documentation.

PR #3130, opened 2022-03-25
Automate API version updates

jstvz:
Add a new workflow 'chores' containing two jobs:

- `check_api_versions`: polls the sfdo devhub for the latest API
   version, then stores that and the current default version from
   `cumulusci/cumulusci.yml` in its output.
- `update_api_versions`: if the outputs for the check job do not match,
writes the new version to `cumulusci/cumulusci.yml`, commits, pushes,
and creates a PR with those changes.

This job is scheduled to run Sundays at 00:00 UTC.

# Critical Changes

# Changes

# Issues Closed


jstvz:
Used the branch name target on accident. Closing in favor of #3132

PR #3132, opened 2022-03-25
Add automated api version updates via Github Actions

jstvz:
Add a new workflow 'chores' containing two jobs:

- `check_api_versions`: polls the sfdo devhub for the latest API
   version, then stores that and the current default version from
   `cumulusci/cumulusci.yml` in its output.
- `update_api_versions`: if the outputs for the check job do not match,
writes the new version to `cumulusci/cumulusci.yml`, commits, pushes,
and creates a PR with those changes.

This job is scheduled to run Sundays at 00:00 UTC.

In addition, the `release_test` workflow now

- is not triggered on every push to feature branches
- may be manually run
- is triggered from the "Release cumulusci" workflow when `version.txt` is included in a `push` event

### DIFF
--- a/.github/workflows/release_test.yml
+++ b/.github/workflows/release_test.yml
@@ -1,9 +1,9 @@
 name: Release Test
 
 on:
+    workflow_dispatch:
     push:
         branches:
-            - feature/**
             - main
     pull_request:
         types: [opened, synchronize, reopened] # Default
--- a/.github/workflows/release_test.yml
+++ b/.github/workflows/release_test.yml
@@ -1,10 +1,13 @@
 name: Release Test
 
 on:
+    workflow_call:
     workflow_dispatch:
     push:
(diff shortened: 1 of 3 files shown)


.github/workflows/release_test.yml
```diff
@@ -1,9 +1,9 @@
 name: Release Test
 
 on:
+    workflow_dispatch:
     push:
         branches:
-            - feature/**
```
davidmreed, line 6: Is it possible to make the Release workflow dependent on this one, so that we don't have to remember to check it when we do a CCI release?

davidmreed (approved):
Elegant. LGTM

.github/workflows/release_test.yml
jstvz: See 3f09e8f

PR #3135, opened 2022-03-25
Add new CumulusCI Flow diagram

davidmreed:
# Critical Changes

# Changes

# Issues Closed


davidjray:
@davidmreed would it be possible to make the colored "package" icons unique in some way? Since the diagram is using only color to specify what type of build it could be difficult to parse information for readers with color-blindness.

docs/cumulusci_flow.rst
```diff
@@ -12,9 +12,11 @@ There are several key reasons we like using CumulusCI Flow:
 CumulusCI Flow is implemented in the standard library flows provided by CumulusCI, but the approach to working
 with a GitHub repository does not require the use of CumulusCI.
 
-.. image:: images/salesforce-org-process.png
+The diagram below depicts the branch and merge structure of CumulusCI Flow in its fullest form, as applied to a long-running release. The diagram also depicts the use of the Salesforce.org :doc:`Testing with Second-Generation Packaging <2gp_testing>` process. Your practice of CumulusCI Flow might include some or all of these features.
+
+.. image:: images/cumulusci_flow.png
```
davidjray, line 17: @davidmreed  would it be possible to make the colored "package" icons unique in some way? Since the diagram is using only color to specify what type of build it could be difficult to parse information for readers with color-blindness.

davidmreed: _Great_ question. Let me revise that. Thanks for calling it out.

PR #3139, opened 2022-03-30
Implement ETL Upserts

prescod:
# Changes

CumulusCI now supports ETL Upserts, as described in the documentation. This means that users can do upserts on fields or groups of fields that Salesforce normally would not allow


datasets/upsert/upsert_complex_example.sql
```diff
@@ -0,0 +1,56 @@
...
+	"FirstName" VARCHAR(255), 
+	"LastName" VARCHAR(255), 
+	"Email" VARCHAR(255),
+	PRIMARY KEY (id)
+);
+-- Change emails based on FirstName/LastName
+INSERT INTO "Contact" VALUES(1,'Michael','Bluth','Nichael.Bluth@example.com');
+INSERT INTO "Contact" VALUES(2,'GOB','Bluth','GeorgeOscar.Bluth@example.com');
+INSERT INTO "Contact" VALUES(3,'Lindsay','Bluth','lindsay.bluth@example.com');
+INSERT INTO "Contact" VALUES(4,'Annyong','Bluth','annyong.bluth@example.com');
```
davidmreed, line 19: 👋 Hello

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -0,0 +1,127 @@
...
+    q = select(*relevant_data_columns)
+
+    q = q.outerjoin(upsert_key_table, and_(*join_on_clauses))
+    return q
+
+
+def _create_empty_upsert_key_table(
+    tablename, metadata, fieldnames: T.List[str]
+) -> Table:
+    """Create an empty table matching the upsert semantics"""
```
davidmreed, line 64: Could this use `create_table_if_needed()`?

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -0,0 +1,127 @@
...
+    """Create a table with keys and IDs from Salesforce"""
+    tablename = f"upsert_{sf_object}_{'_'.join(keys)}"
+    if "Id" not in (key.title() for key in keys):
+        keys = ("Id",) + keys
+
+    table = _create_empty_upsert_key_table(tablename, metadata, keys)
+
+    soql_query = f"select {','.join(keys)} from {sf_object}"
+    try:
+        table = sf_query_to_table(
```
davidmreed, line 92: This is an elegant utility. Nice. 👍

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        non_id_fields = [
+            fieldname for fieldname in fieldnames if fieldname.lower() != "id"
+        ]
+        assert len(fields) == len(non_id_fields) + 1
+        t = Table(tablename, metadata, *fields, UniqueConstraint(*non_id_fields))
+        t.create(metadata.bind)
+        return t
+
+
+def _create_upsert_key_table(
```
davidmreed, line 76: This function name could more clearly state that we are actually performing a (potentially very lengthy) extraction. `_extract_upsert_key_data()`, perhaps?

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -0,0 +1,127 @@
...
+        (
+            upsert_key_table.columns[column_name]
+            == update_data_table.columns[column_name]
+        )
+        for column_name in mapping.update_key
+    ]
+
+    q = select(*relevant_data_columns)
+
+    q = q.outerjoin(upsert_key_table, and_(*join_on_clauses))
```
davidmreed, line 57: 👍 I like this approach a lot.

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -0,0 +1,127 @@
...
+            context=context,
+            query=soql_query,
+            metadata=metadata,
+            connection=connection,
+        )
+    except IntegrityError as e:
+        message = str(e)
+        if "UNIQUE constraint failed" not in message:  # pragma: no cover
+            raise e
+        raise BulkDataException(f"Duplicate values for upsert key:\n {e.params}") from e
```
davidmreed, line 106: Potential future enhancement: at present, this will only advise the user of the first duplicate key; there could be many.

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -0,0 +1,127 @@
...
+
+    return table
+
+
+def needs_etl_upsert(mapping: MappingStep, sf: Salesforce):
+    """Is this an upsert that Salesforce cannot do nativvly?"""
```
davidmreed, line 112:
```suggestion
    """Is this an upsert that Salesforce cannot do natively?"""
```

cumulusci/tasks/bulkdata/upsert_utils.py
davidmreed: Given the potentially time-consuming nature of this operation, could we add a log line so the user knows what work we're doing?

davidmreed (changes requested):
I really like this implementation. Just inquiring about a handful of little items.

davidmreed (approved):

PR #3145, opened 2022-04-05
link missing task or flow when merging configs

Br4nd0R:
# Changes
* Fixed an issue where CumulusCI was incorrectly merging task option overrides.

### QA Notes
I used the following in the standard lib while testing:
```
official_testing_flow:
        group: foo
        description: Does some stuff
        steps:
            1:
                task: util_sleep
            2:
                task: util_sleep
            3:
                task: util_sleep
   ```
   Then I modified the project `cumulusci.yml` to include:
   ```
   flows:
    official_testing_flow:
        steps:
            1:
                # task: util_sleep
                options:
                    seconds: 1
```
* Also verified setting the flow step to a value of `None` still works.

### DIFF
--- a/cumulusci/core/tests/test_utils_merge_config.py
+++ b/cumulusci/core/tests/test_utils_merge_config.py
@@ -216,7 +216,9 @@ def test_cleanup_flow_step_override_conflicts__multiple_overrides_of_alternating
     expected_universal_config["flows"]["steps_all_tasks"]["steps"][3] = {}
 
     expected_project_config = copy.deepcopy(project_config)
-    expected_project_config["flows"]["steps_all_tasks"]["steps"][3] = {}
+    expected_project_config["flows"]["steps_all_tasks"]["steps"][3] = {
+        "flow": "custom_flow_two"
+    }
 
     expected_project_local_config = copy.deepcopy(project_local_config)
     expected_project_local_config["flows"]["steps_all_tasks"]["steps"][3] = {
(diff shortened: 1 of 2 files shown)


cumulusci/core/tests/test_utils_merge_config.py
```diff
@@ -260,3 +262,15 @@ def test_cleanup_flow_step_override_conflicts__obsolete_flow_with_tasks(
     clean_configs = utils.cleanup_flow_step_override_conflicts(configs)
 
     assert expected_project_config == clean_configs["project_config"]
+
+
+def test_link_missing_task_or_flow():
```
prescod, line 267: I'm not really a fan of this kind of testing. I think the automated tests should do what you described in the PR cmments. As a maintainer I don't care much that the new function `link_missing_task_or_flow` does the right thing. I care that CCI configs merge properly. I'd like to see some CCI configs merge. Then if I change the underlying implementation later, I'll still know that CCI configs merge properly.

For example, here is how we test the CCI validating linter.

https://github.com/SFDO-Tooling/CumulusCI/blob/8d46f48da6850d4b979d7f36f01f4b20103ed236/cumulusci/utils/yaml/tests/test_cumulusci_yml.py#L97

prescod: IMO: A good rule of thumb is whether there's a clear line between the WI ACs and the test cases.

cumulusci/core/utils.py
```diff
@@ -239,10 +239,27 @@ def remove_overridden_flow_steps_in_config(
...
+def link_missing_task_or_flow(
+    step_config_to_override: dict, overriding_step_config: dict
+):
+    """If the incoming override does not have task/flow defined then inherit from
+    the flow step that we're overridding."""
+    if "flow" not in overriding_step_config and "task" not in overriding_step_config:
+        if "task" in step_config_to_override:
+            overriding_step_config["task"] = step_config_to_override["task"]
+        elif "flow" in step_config_to_override:
+            overriding_step_config["flow"] = step_config_to_override["flow"]
```
davidmreed, line 260: 👍

cumulusci/core/tests/test_utils_merge_config.py
davidmreed: I agree that we should build new tests to catch this type of problem, and I will write a WI to do so (along with documenting this specific wrinkle of flow customization). However, I don't want to hold up @Br4nd0R's fix to this bug to do that, as there is serious production impact we need to cure.

davidmreed (approved):

PR #3149, opened 2022-04-07
Add property in OrgConfig to check for advanced survey features

davidmreed:
- We added a property, `is_survey_advanced_features_enabled`, to the `org_config` entity available in `when` clauses.


jstvz (approved):
Was going to suggest documenting this, but I don't think we're documenting any of the other org_config properties...

PR #3150, opened 2022-04-11
Pass secrets into release_test

jstvz:
When we made the release test reuseable we missed that secrets need to
be defined in the workflow_call defn. [docs]

[docs]: https://docs.github.com/en/actions/using-workflows/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow

-------

# Critical Changes

# Changes

# Issues Closed


Br4nd0R (approved):
Thanks @jstvz, this looks good.

PR #3154, opened 2022-04-11
Errors in cleanup should not prevent the orderly completion of the task

prescod:
If we don't properly nest context managers, Windows will not allow tempfiles to be cleaned up. This crashes the program. I think that sometimes it happens because of an underlying exception which is swallowed and replaced by this spurious one. 

Either way: I don't think that a failure to clean up a tempdir should crash the whole process.

Making a Draft until I investigate more with a user.


cumulusci/utils/__init__.py
```diff
@@ -536,7 +537,10 @@ def temporary_dir(chdir=True):
             yield d
     finally:
         if os.path.exists(d):
-            shutil.rmtree(d)
+            try:
+                shutil.rmtree(d)
+            except Exception as e:  # pragma: no cover
+                warnings.warn(f"Cannot remove temporary directory: {d}: {e}")
```
davidmreed, line 543: I'm not very familiar with this module - what does it do relative to `logging.getLogger(...).warn()`?

prescod: I guess the main difference is that it always warns whereas the logger depends on your logger config. It'ss simpler and low-risk in interactive scenarios.

warnings.warn would probably not output info that you'd see in MetaCI. I think MetaCI has hooked into the logging framework. But I am also not 100% sure that MetaCI will capture logs sent to these free-standing loggers as opposed to self.logger. It probably does but I'm not sure.

I could go either way if you have a strong preference.

davidmreed: Using self.logger seems like "least surprise" to me. I'd hate to be wondering in six months why some log or another, maybe on MetaCI, doesn't contain the warning we expect.

prescod: Okay, I switched to logging.

davidmreed (approved):

PR #3159, opened 2022-04-15
Fix windows update_depdencies problem

prescod:
# Issues Closed

The update_dependencies task could crash on Windows if the a depended-upon project required converting from SFDX to MDAPI. This has been fixed.


davidmreed (approved):
Nice clean fix to nasty problem. LGTM.

Issue #3047, opened 2022-01-25
Open Test Browser TypeError: 'NoneType' object is not callable when using useralias param


ZephyrusMB:
### Describe the bug

Hi,
Trying to open a test browser with a different user of a persistent org gives me this error **TypeError: 'NoneType' object is not callable**. If I use a scratch org the keyword works perfectly and it allows me to open 2 browsers with different accounts. To authenticate i use  'sfdx auth:web:login' for both the persistents accounts, while for the scratch org ones  i generated a password and then created a new profile with sfdx following this [youtube video](https://www.youtube.com/watch?v=sqKb7QbyNJ0).
Here the robot file i execute
```
`*** Settings ***

Resource        cumulusci/robotframework/Salesforce.robot 
Library         cumulusci.robotframework.PageObjects 

Test Teardown    Delete Records and Close Browser
*** Test Cases ***
Utente admin
    Open Test Browser       useralias=Chatter  
Utente UTest
    Open Test Browser            useralias=UTest`

```
Here the log with the same error in each case 
![image](https://user-images.githubusercontent.com/81178328/151055720-562d2a0b-a839-45e7-ab75-00d37187fff2.png)

### Reproduction steps

1.connect to a persistent org with `sfdx auth:web:login` with the first account
2.connect to a persistent org with `sfdx auth:web:login` with the second account
3.execute the file robot with `cci task run robot  --suites pathToFile.robot`
4.get the TypeError: 'NoneType' object is not callable

### Your CumulusCI and Python versions

CumulusCI version: 3.51.0 (C:\Users\${user}\.local\bin\cci)
Python version: 3.9.8 (C:\Users\${users}\.local\pipx\venvs\cumulusci\Scripts\python.exe)

### Operating System

Microsoft Windows 10.0.19042

### Windows environment

Command Prompt

### CumulusCI installation method

pipx

davisagli:
The `useralias` for the `Open Test Browser` keyword refers to an org in the CumulusCI keychain, not the sfdx keychain. In other words the alias needs to show up in `cci org list` in order for it to work. For a persistent org, that means you would need to connect the org using `cci org connect` rather than using sfdx. Unfortunately we don't currently provide a way to import a persistent org from sfdx into the cci keychain.

davisagli:
[...]

PR #3169, opened 2022-04-20
Support `useralias` in `Open Test Browser` for connected orgs

boakley:
Update the `login url` keyword to be able to get the URL for a user in a connected persistent org. See [W-10502445](https://gus.lightning.force.com/a07EE00000jGUN1YAO) and issue #3047

The root of the problem is that the OrgConfig class doesn't define the method `get_access_token` since we don't have per-user access tokens in the config object for this type of org.  Instead, we have a single access token for the org as a whole. 

The solution is to scan all connected orgs for an org that is associated with the given user and then return the access token for that org.

# Critical Changes

# Changes
* The `open test browser` keyword now works when providing the `useralias` argument, when the org is a connected persistent org.
* The `login url` keyword now works when passing in a username or alias for a user in a connected persistent org.
# Issues Closed
* #3047

### DIFF
--- a/cumulusci/robotframework/CumulusCI.py
+++ b/cumulusci/robotframework/CumulusCI.py
@@ -4,7 +4,7 @@ import robot.api.logger
 from robot.libraries.BuiltIn import BuiltIn
 
 from cumulusci.cli.runtime import CliRuntime
-from cumulusci.core.config import TaskConfig
+from cumulusci.core.config import ScratchOrgConfig, TaskConfig
 from cumulusci.core.exceptions import TaskOptionsError
 from cumulusci.core.tasks import CURRENT_TASK
 from cumulusci.core.utils import import_global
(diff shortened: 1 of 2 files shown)


cumulusci/robotframework/CumulusCI.py
```diff
@@ -116,10 +116,17 @@ def login_url(self, org=None, **userfields):
...
         org = self.org if org is None else self.keychain.get_org(org)
+
         if userfields:
-            access_token = org.get_access_token(**userfields)
-            login_url = f"{org.instance_url}/secur/frontdoor.jsp?sid={access_token}"
-            return login_url
+            if org.get_access_token is not None:
+                # connected persistent org configs won't have the get_access_token method
+                access_token = org.get_access_token(**userfields)
+                login_url = f"{org.instance_url}/secur/frontdoor.jsp?sid={access_token}"
```
prescod, line 124: Can these two lines be merged with the same ones in the next branch?

cumulusci/robotframework/CumulusCI.py
```diff
@@ -265,6 +272,48 @@ def _run_task(self, task):
...
+        be used. If not, this function will do a query to find a username
+        that matches the given parameters.
+
+        """
+
+        username = userfields.get("username", None)
+        if username is None:
+            where = [f"{key}='{value}'" for key, value in userfields.items()]
+            query = f"SELECT Username FROM User WHERE {' AND '.join(where)}"
+            result = base_org.salesforce_client.query_all(query).get("records", [])
```
prescod, line 295: Minor issue but no real reason to use "query_all" (versus "query") here. Not sure if that has some small perf implication.

prescod (commented):
`_find_access_token`  has great tests.

Should there be tests of the two new branches of login_url? Are they covered somehow already???

Should there be a Robot test that executes against a real org?

cumulusci/robotframework/CumulusCI.py
boakley: I've refactored this section slightly to move the common code outside of the `if` statements.

cumulusci/robotframework/CumulusCI.py
boakley: Good catch. I don't know why I used `query_all` here. Probably copied it from somewhere else without thinking. I've switched the code to use `query` instead of `query_all`

prescod (approved):

PR #3173, opened 2022-04-27
Fix bug when setting the 'options' option in the robot task

boakley:
Prior to this fix, doing `cci task run robot -options=outputdir:/tmp` would result in a cryptic error message `AttributeError: 'str' object has no attribute 'setdefault'` This PR fixes the code that caused that error message.

The root of the problem is that the robot task requires that the option value is a dictionary, but we weren't converting the string from the command line into a dictionary. Later, we would try to run a dictionary method on the string resulting in this error message.

This fix involves running the command line option through an existing function specifically for dictionary-based options, `process_list_of_pairs_dict_arg`. 

This was originally [reported in the trailblazer community]( https://trailhead.salesforce.com/trailblazer-community/feed/0D54S00000GgRH9SAN)

# Critical Changes

# Changes
* Fixed a bug that prevented the `options` option of the robot task from being set on the command line (eg: `cci task run robot --options=outputdir:/tmp`)

# Issues Closed


davidjray (approved):
Tested by running against a product repo using `cci task run robot --suites robot/SAL/tests/browser/advisee_snapshot --options outputdir:"robot/TMP/results"` and the results were placed in the proper directory

PR #3180, opened 2022-05-03
Clean up error messaging for deleted org configs

prescod:
Minor improvements to error messaging after an org-config is deleted from cumulusci.yml


jstvz:
Duplicates #3469

PR #3183, opened 2022-05-09
Refactor loaddata and fix bugs in ETL Upsert

prescod:
This PR is dependent on #3184 . Please review that first.

# Critical Changes

# Changes

# Refactoring

_query_db has been totally refactored to use a sort of Strategy pattern.

This necessitated lots of test changes.

Snowfakery ETL is one of the participants in that pattern.

Snowfakery ETL tests were added.

# Issues Closed

### DIFF
--- a/cumulusci/tasks/bulkdata/mapping_parser.py
+++ b/cumulusci/tasks/bulkdata/mapping_parser.py
@@ -4,7 +4,7 @@ from enum import Enum
 from functools import lru_cache
 from logging import getLogger
 from pathlib import Path
-from typing import IO, Any, Callable, Dict, List, Mapping, Optional, Union
+from typing import IO, Any, Callable, Dict, List, Mapping, Optional, Tuple, Union
 
 from pydantic import Field, ValidationError, root_validator, validator
 from requests.structures import CaseInsensitiveDict as RequestsCaseInsensitiveDict
@@ -248,9 +248,9 @@ class MappingStep(CCIDictModel):
         if values is None:
             values = {}
         if type(values) is list:
-            return {elem: elem for elem in values}
+            values = {elem: elem for elem in values}
 
-        return values
+        return CaseInsensitiveDict(values)
 
     @root_validator
     @classmethod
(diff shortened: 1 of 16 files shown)


cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -496,6 +495,21 @@ def strip(element: str):
         ):
             return False
 
+        # inject namespaces intoo the update_key
```
davidmreed, line 498:
```suggestion
        # inject namespaces into the update_key
```

cumulusci/tasks/bulkdata/mapping_parser.py
```diff
@@ -496,6 +495,21 @@ def strip(element: str):
         ):
             return False
 
+        # inject namespaces into the update_key
+        if self.update_key:
+            assert isinstance(self.update_key, Tuple)
```
davidmreed, line 500: Confirming: `update_key` is a tuple even if it's a single field?

cumulusci/tasks/bulkdata/tests/test_upsert.py
```diff
@@ -608,6 +608,166 @@ def test_simple_upsert__bad_key(
...
+                "sql_path": cumulusci_test_repo_root
+                / "datasets/upsert/upsert_example_2.sql",
+                "mapping": cumulusci_test_repo_root
+                / "datasets/upsert/upsert_smart__native_field.yml",
+                "ignore_row_errors": True,
+                "set_recently_viewed": False,
+            },
+        )
+        task._update_credentials = mock.Mock()
+        task()
```
davidmreed, line 661: Since we don't have any explicit asserts in these three tests, how about using the `responses` context-manager style so it will assert for us that all expected HTTP calls are made?

davidmreed (commented):
This is nice and clean. Same thoughts as the other PR - smoke testing with real mappings and existing DBs will add confidence. A couple of questions below.

cumulusci/tasks/bulkdata/upsert_utils.py
```diff
@@ -128,3 +86,45 @@ def needs_etl_upsert(mapping: MappingStep, sf: Salesforce):
         or describe_data[key]["idLookup"]
     )
     return not is_sf_builtin_key
+
+
+class AddUpsertsToQuery(LoadQueryExtender):
+    @property
```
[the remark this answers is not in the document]
prescod: I think this method is actually very cheap. It doesn't do the extract data, just referred to the table that should already exist.

cumulusci/tasks/bulkdata/mapping_parser.py
prescod: Correct. There's a method that guarantees it and this check double-checks that it wasn't turned off by accident.

https://github.com/SFDO-Tooling/CumulusCI/blob/2b27f19e63157941b58ad4929ef2bce8bfd59528/cumulusci/tasks/bulkdata/mapping_parser.py#L112

cumulusci/tasks/bulkdata/tests/test_upsert.py
prescod: I think I added the appropriate assertions in eb2f6b9. How does that look to you?

davidmreed: Looks good to me.

davidmreed (approved):

PR #3184, opened 2022-05-10
Refactor _query_db in load.py

prescod:
Refactorings:

- Move test_load away from heavy mocking
- Refactor query_db in load.py


cumulusci/tasks/bulkdata/tests/test_load.py
```diff
@@ -488,144 +488,76 @@ def test_get_statics_record_type_not_matched(self):
...
-            sf_object="Account",
-            table="accounts",
-            action=DataOperationType.UPDATE,
-            fields={"Id": "sf_id", "Name": "name"},
-            lookups={
-                "ParentId": MappingLookup(
-                    table="accounts", key_field="parent_id", name="ParentId"
-                )
-            },
+    def test_query_db__joins_self_lookups(self):
```
davidmreed, line 491: This testing strategy is nice because of the elision of a thousand mocks. Do you think there is potential fragility to changes in SQLAlchemy, though? I can't say that I know how stable their SQL marshalling is.

davidmreed: I don't think it's a deal-breaker in any case, but wanted to think it through.

cumulusci/tasks/bulkdata/tests/test_load.py
```diff
@@ -668,22 +600,31 @@ def test_query_db__person_accounts_enabled__contact_mapping(self, aliased):
             },
         )
 
-        task._query_db(mapping)
+        query = task._query_db(mapping)
+        assert (
+            task.session.query == task.session.query
+        )  # check that query chaining from above worked.
```
davidmreed, line 606: I'm not following this tautological assertion

davidmreed (approved):
I have a couple of non-blocking queries. I'm going ahead with approving as this looks very solid and massively improves the clarity of `load.py`.

I do think this change is high-risk and hard to fully evaluate via code review due to the complexity of the module. Can we/have you already tested with some real-world mapping files? Grants and Case Management might be good testbeds if needed.

cumulusci/tasks/bulkdata/tests/test_load.py
prescod: Yes I'm a bit worried about that too but it seemed that:

a) I need to either make the code sensitive to changes in SQL Alchemy or changes in our code's implementation details.

b) if a test breaks it's easier to eyeball two SQL statements to see if they are logically equivalent than weird mocks.

PR #3192, opened 2022-05-11
Workaround for pip-tools issue upstream

jstvz:
First guess was to pin pip to 22.0.4, but I wondered: What exactly is
pip-sync doing here? There's no virtualenv to synchronize since we get a
fresh environment in each workflow run.

### DIFF
--- a/.github/workflows/feature_test.yml
+++ b/.github/workflows/feature_test.yml
@@ -34,7 +34,7 @@ jobs:
                   cache: pip
                   cache-dependency-path: "requirements/*.txt"
             - name: Install dependencies
-              run: make dev-install
+              run: pip install -r requirements_dev.txt
             - name: Build Docs
               run: make docs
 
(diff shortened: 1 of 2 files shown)


davisagli:
@jstvz Are you sure? Even with the pip cache from actions/setup-python?

davisagli:
> Even with the pip cache from actions/setup-python?

Oh, I suppose that's a cache of downloaded distributions, not an installed virtualenv

jstvz:
@Br4nd0R Would you sanity check this tomorrow?  (The release will be blocked otherwise)

If we do need to run `pip-sync` in the runner after all we can just pin `pip`:
```diff
diff --git a/Makefile b/Makefile
index e3981f0f0..9b6dcb7bc 100644
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ update-deps:
 	pip-compile --upgrade requirements/dev.in
 
 dev-install:
-	python -m pip install --upgrade pip pip-tools setuptools
+	python -m pip install --upgrade pip==22.0.4 pip-tools setuptools
 	pip-sync requirements/*.txt
 	python -m pip install -e .
 
```

Br4nd0R (approved):

PR #3193, opened 2022-05-12
Don't require that `project__git__repo_url` be set to publish a MetaDeploy plan

jstvz:
Because we infer `repo_url` based on `repo_name` and `repo_owner` elsewhere in the same code path it makes sense to remove an undocumented dependency that is tripping up users.

---
# Critical Changes

# Changes

- Setting `repo_url` in your project config is no longer required to publish a MetaDeploy plan. 

# Issues Closed


jstvz:
🚮

PR #3195, opened 2022-05-12
Snowfakery 3.1

prescod:
A bunch of other dependencies floated at the same time.

`click` and `pycrypto` are the only ones I think we need to keep an eye on.


requirements/prod.txt
```diff
@@ -14,11 +14,11 @@ certifi==2021.10.8
     #   snowfakery
 cffi==1.15.0
     # via cryptography
-charset-normalizer==2.0.11
+charset-normalizer==2.0.12
     # via
     #   requests
     #   snowfakery
-click==8.0.3
+click==8.1.3
```
jstvz, line 21: ❔: Are we trying to get 3.1 into the next CumulusCI release?

jstvz (approved):
---
Emoji Key

requirements/prod.txt
jstvz: Never mind, forgot what repo this was in.

jstvz:
I missed a `make dev-install`, would you update that here as well?: https://github.com/SFDO-Tooling/CumulusCI/blob/b84369edc018fe46a5638f2c7379f9225865f152/.github/workflows/slow_integration_tests.yml#L26-L27

Issue #1054, opened 2019-04-24
Throw error from metadeploy_publish if no git repo in cumulusci.yml


jlantz:
Currently, if you try to run the metadeploy_publish task in a repo that doesn't have project__git__repo_url configured, you get a nasty traceback (below).  We should throw a nicer error with helpful text about what to do.

Traceback (most recent call last):
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/bin/cci", line 11, in <module>
load_entry_point('cumulusci==2.4.2', 'console_scripts', 'cci')()
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/click/core.py", line 764, in __call__
return self.main(args, *kwargs)
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/click/core.py", line 717, in main
rv = self.invoke(ctx)
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
return _process_result(sub_ctx.command.invoke(sub_ctx))
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
return _process_result(sub_ctx.command.invoke(sub_ctx))
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/click/core.py", line 956, in invoke
return ctx.invoke(self.callback, **ctx.params)
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/click/core.py", line 555, in invoke
return callback(args, *kwargs)
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/cumulusci/cli/cci.py", line 219, in new_func
func(config, args, *kw)
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/cumulusci/cli/cci.py", line 1094, in task_run
handle_exception_debug(config, debug, no_prompt=no_prompt)
File "/usr/local/Cellar/cumulusci/2.4.2/libexec/lib/python3.7/site-packages/cumulusci/cli/cci.py", line 1077, in task_run
task()
[...]

PR #3200, opened 2022-05-17
Stop requiring project__git__repo_url to publish

jstvz:
- Setting `repo_url` in your project config is no longer required to run `metadeploy_publish`.


cumulusci/tasks/tests/test_metadeploy.py
```diff
@@ -145,7 +144,7 @@ def test_run_task(self):
 
         responses.add(
             "GET",
-            "https://metadeploy/products?repo_url=EXISTING_REPO",
+            "https://metadeploy/products?repo_url=https%3A%2F%2Fgithub.com%2FTestOwner%2FTestRepo",
```
jofsky, line 147: @jstvz I don't think it's expecting the result to be html escaped. Although I don't think this is a bad time to add an html.escape() addition to our repo URL https://github.com/SFDO-Tooling/CumulusCI/pull/3200/files#diff-97be17eb03341832c29a8f929a2358faf7a3157182956ce96eb7170cfcdea1e4L242 let me know what you think.

jofsky, line 147: Lazy fix for test to pass but I'd like to discuss above comment @jstvz

```suggestion
            "https://metadeploy/products?repo_url=https://github.com/TestOwner/TestRepo",
```

jstvz: Fixed in 91a993f

jstvz: It's not clear because of the old test value for `repo_url`, but MetaDeploy is expecting a URL encoded string as the query parameter. Updated the test to match.

jofsky (approved):

jstvz:
Fixes #1054

PR #3204, opened 2022-05-20
Document the release option for sources

### LINKED ISSUE #2439: Missing documentation for sources: release:
I see this syntax in a project:

```
sources:
    other_proj:
        github: https://github.com/SalesforceFoundation/other-project
        release: latest_beta
```

But I don't see the word "release:" in this document:

https://cumulusci.readthedocs.io/en/latest/config.html#tasks-and-flows-from-a-different-project

jstvz:
Fixes #2439


docs/config.rst
```diff
@@ -377,17 +377,19 @@ By default, CumulusCI uses the resolution strategy ``production``, which will fe
...
     sources:
+        eda:
+            github: https://github.com/SalesforceFoundation/EDA
+            release: latest
         npsp:
             github: https://github.com/SalesforceFoundation/NPSP
             tag: rel/3.163
 
-or a specific ``commit`` or ``branch``. We recommend that most projects, however, use a resolution strategy.
+where ``release`` is one of ``latest``, ``previous``, or ``latest_beta``. You can also select a specific ``commit`` or ``branch``. We recommend that most projects, however, use a resolution strategy.
```
pconrad-sfdo, line 392: I'm not always a fan of the "you can do this... [example] ...as long as xyz is true" construction. I'm not going to strike it down here, but I'll lightly recommend the following instead:

```
You can fetch a specific ``tag`` or ``release``, where  ``release`` is one of ``latest``, ``previous``, or ``latest_beta``. Example:

[Example]

You can also select a specific ``commit`` or ``branch``. Example:

[Example]

However, we recommend that most projects use a resolution strategy, because [reason]
```

davidmreed (approved):

PR #3206, opened 2022-05-23
Auto-update dependencies

Br4nd0R:
Nothing for release notes.

## Dev Notes
* New workflow to automatically update cumulusci python dependencies  weekly.


.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+                  git config user.name github-actions
+                  git config user.email github-actions@github.com
+                  git switch -c "dependency-updates-week-${date +%u}"
+                  git add requirements/*.txt 
+                  git commit -m "update dependencies (automated)"
+                  git push origin "dependency-updates-week${date +%u}"
+            - env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              run: |
+                  gh pr create --fill --label 'auto-pr'
```
jstvz, line 37: Current behavior is to have the `dependencies` label for dependabot's PRs. Suggest we keep it, via:
```suggestion
                  gh pr create --fill --label 'auto-pr,dependencies'
```

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,37 @@
...
+            - name: Update dependencies
+              run: make update-deps
+            - name: Commit changes and open PR
+              run: |
+                  git config user.name github-actions
+                  git config user.email github-actions@github.com
+                  git switch -c "dependency-updates-week-${date +%u}"
+                  git add requirements/*.txt 
+                  git commit -m "update dependencies (automated)"
+                  git push origin "dependency-updates-week${date +%u}"
```
jstvz, line 33: ±: Syntax is slightly different:
```suggestion
                  export WEEK="W$(date +%V)"
                  git switch -c "dependency-updates-week-$WEEK"
                  git add requirements/*.txt 
                  git commit -m "$WEEK dependency updates (automated)"
                  git push origin "dependency-updates-week-$WEEK"
```

(Adding `$WEEK` to the commit message, since it'll be used as-is by `gh` in the next step.)

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,37 @@
+name: Update dependencies
+
+on:
+    workflow_dispatch:
+    schedule:
+        # at 00:00 on Monday (UTC)
+        - cron: "0 0 * * MON"
```
jstvz, line 7: ❓: How about adding this to the existing `chores.yml`? If this needs to run on Monday, update that file so that both jobs run on Monday instead of Sunday.

jstvz (changes requested):
Slight tweak to your shell commands needed.

Might be able to test this from the CLI. If that doesn't work, can add a `pull_request` trigger and create a dummy PR against this branch.

---
<details>
<summary>Emoji Key</summary>


</details>

.github/workflows/update_dependencies.yml
Br4nd0R: I initially thought that I would add it to `chores.yml` but I'm thinking it's going to be pointless to have this run every week when we really only need it based on the first week of a sprint. Since that schedule would be difficult to generate, what if we only had this use `workflow_dispatch` so that who ever wants to update dependencies can just come in and click a button? Otherwise, we will be in a perpetual game of "close the stale dependency update pull requests".

jstvz: Works for me.

.github/workflows/update_dependencies.yml
```diff
@@ -0,0 +1,38 @@
...
+                  python-version: 3.8
+                  cache: pip
+                  cache-dependency-path: "requirements/*.txt"
+            - name: Update dependencies
+              run: |
+                  pip install pip-tools
+                  make update-deps
+            - name: Commit changes and open PR
+              run: |
+                  export WEEK="w$(date +%V)"
```
jstvz, line 28:
```suggestion
                  export WEEK="W$(date +%V)"
```
I will die on [this hill](https://en.wikipedia.org/wiki/ISO_week_date).

jstvz:
@Br4nd0R Oh, and less facetiously: If you end up wanting to reuse this you can put it in [SFDO-Tooling/.github](https://github.com/SFDO-Tooling/.github/tree/main/.github/workflows) and call it with:

https://github.com/SFDO-Tooling/CumulusCI/blob/9ef0c828a92b0ebfa06c8693a79505e26bc65bbb/.github/workflows/feature_test.yml#L16-L19

jstvz (approved):

PR #3209, opened 2022-05-26
Merge yaml config with config present from org file

Br4nd0R:
# Issues Closed
* Fixed a bug where scratch org attributes configured via `cumulusci.yml` were not taking effect.

## Dev Notes
I initially thought that I make this change more central and have the config from yaml merged when the keychain is loaded (`_load_keychain()`). However, I forgot that "loading" only loads the encrypted contents onto the keychain. We only ever decrypt exactly what is being asked for (for a given org or service). Because of this, the best place to make this change is the `get_org(org_name)` method.


davidjray (approved):
LGTM

PR #3223, opened 2022-06-01
Unlocked resolution strategy and supporting flows

davidmreed:
# Critical Changes

# Changes

- We added a new dependency resolution strategy, `unlocked`. Like our existing dependency strategies using parallel 2GP package builds, this strategy installs parallel unlocked package builds.
- We added the flows `build_unlocked_test_package`, `install_unlocked_commit`, and `qa_org_unlocked` to support use of parallel unlocked packages.

# Issues Closed


prescod (approved):
Your walkthrough with me was fairly compelling. I've taken a second look but can't add much to that.

PR #3246, opened 2022-06-20
Add include directive to bring HISTORY.rst into the docs

pconrad-sfdo:


jstvz (changes requested):
We'll also need to convert the README and the `long_description_content_type` [below](https://github.com/SFDO-Tooling/CumulusCI/pull/3246/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7L44-L45). 

The two files are concatenated together to generate the [PyPI package description](https://pypi.org/project/cumulusci/).

---

I can handle the `content_type` if you handle the README conversion, @pconrad-sfdo.

PR #3250, opened 2022-06-22
Systemetize module-moved warnings

prescod:
Add a `warn_moved` function to standardize deprecation warnings.


cumulusci/utils/deprecation.py
```diff
@@ -0,0 +1,13 @@
...
+
+def warn_moved(new, old):
+    msg = f"Deprecation warning: Please import from {new} rather than {old}"
+
+    # this is a hard-error if encountered in testing context
+    if "PYTEST_CURRENT_TEST" in os.environ:
+        assert 0, msg
+
+    # otherwise warning
+    warn(msg)
```
jstvz, line 13: ❔: I'm unclear on why we do it this way. Why emit a `UserWarning` instead of a `DeprecationWarning` or `ImportWarning`?

cumulusci/utils/deprecation.py
```diff
@@ -0,0 +1,13 @@
+import os
+from warnings import warn
+
+
+def warn_moved(new, old):
+    msg = f"Deprecation warning: Please import from {new} rather than {old}"
+
+    # this is a hard-error if encountered in testing context
+    if "PYTEST_CURRENT_TEST" in os.environ:
+        assert 0, msg
```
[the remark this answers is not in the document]
prescod: It would break downstream code if that code needed to be fixed. I'm not convinced that's a bad thing given the totality of our context (how many people use CCI as a library, how many of them are not on our team, how many are likely using the old paths, ...)

cumulusci/utils/deprecation.py
jstvz: Okay, went back and looked at [PEP 565](https://peps.python.org/pep-0565/#additional-use-case-for-futurewarning):

> `FutureWarning`: reported by default for all code. The intended audience is users of applications written in Python, rather than other Python developers (e.g. warning about use of a deprecated setting in a configuration file format).

Seeing that, how about emitting a `FutureWarning`, and setting `pytest.filterwarnings = error::FutureWarning:cumulusci`? Those get displayed to the user, and we can fail our tests without breaking theirs.

prescod: Wouldn't it be safer to Subclass so some other team's FutureWarning won't break *our* tests?

jstvz: That _would_ be annoying. 😏 

We can subclass if you want, but we can also match the message and warning:
```toml
# action:message:category:module:line
[tool.pytest.ini_options]
filterwarnings = [
    'error:Please import from :FutureWarning'
]
```

Up to you.

prescod: In my defence, does anyone actually embed CCI other than us and maybe Jason?

jstvz (approved):

PR #3253, opened 2022-06-22
Import persistent orgs from the SFDX keychain

davidmreed:
CumulusCI now supports importing persistent orgs from the SFDX keychain with the `cci org import` command.

### DIFF
--- a/cumulusci/cli/org.py
+++ b/cumulusci/cli/org.py
@@ -10,6 +10,7 @@ from rich.console import Console
 
 from cumulusci.cli.ui import CliTable, SimpleSalesforceUIHelpers
 from cumulusci.core.config import OrgConfig, ScratchOrgConfig
+from cumulusci.core.config.sfdx_org_config import SfdxOrgConfig
 from cumulusci.core.exceptions import OrgNotFound
 from cumulusci.oauth.client import (
     PROD_LOGIN_URL,
@@ -20,7 +21,7 @@ from cumulusci.oauth.client import (
 from cumulusci.salesforce_api.utils import get_simple_salesforce_connection
 from cumulusci.utils import parse_api_datetime
 
-from .runtime import pass_runtime
+from .runtime import CliRuntime, pass_runtime
 
 
 @click.group("org", help="Commands for connecting and interacting with Salesforce orgs")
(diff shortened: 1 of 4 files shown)


cumulusci/cli/org.py
```diff
@@ -240,32 +235,56 @@ def org_default(runtime, org_name, unset):
...
-    info = scratch_org_config.sfdx_info
-    if not info.get("created_date"):
-        raise click.UsageError(
-            "cci org import only works for locally created "
-            "scratch orgs.\nUse `cci org connect` for other orgs."
+    # Determine if we received a locally-created scratch org
+    # or some other org (which we'll treat as persistent)
+
+    info = org_config.sfdx_info
+    if info.get("created_date"):
```
Br4nd0R, line 256: I paused over this enough times that I'm thinking something like the following may be better. The real question is, does that method make sense as part of `SfdxOrgConfig` 🤔 . At least abstracting the logic into a method would allow for a docstring about how/why we determine this the way we do. I'm probably overthinking it at this point 😂 .
```suggestion
    is_scratch = org_config.is_scratch_org()
    if is_scratch:
```

cumulusci/cli/tests/test_org.py
```diff
@@ -403,16 +410,98 @@ def test_org_import__persistent_org(self, cmd):
...
-        ):
+        with mock.patch("click.echo", out.append):
             run_click_command(
                 org.org_import,
                 username_or_alias="test@test.org",
                 org_name="test",
                 runtime=runtime,
             )
+            runtime.keychain.set_org.assert_called_once()
+            created_org = runtime.keychain.set_org.call_args[0][0]
```
Br4nd0R, line 442: I think something like this allows us to get rid of line 442:
```suggestion
            runtime.keychain.set_org.assert_called_once_with(created_org, global_org=mock.ANY, save=mock.ANY)
```

cumulusci/cli/tests/test_org.py
```diff
@@ -403,16 +410,98 @@ def test_org_import__persistent_org(self, cmd):
...
+        out = []
+        with mock.patch("click.echo", out.append):
+            run_click_command(
+                org.org_import,
+                username_or_alias="test@test.org",
+                org_name="test",
+                runtime=runtime,
+            )
+            runtime.keychain.set_org.assert_called_once()
+            created_org = runtime.keychain.set_org.call_args[0][0]
```
Br4nd0R, line 497: Same note as above. Utilize [mock.ANY](https://docs.python.org/3/library/unittest.mock.html#any) in conjunction with `assert_called_once_with()`.

cumulusci/cli/org.py
davidmreed: I shied away from making changes to the method set or properties of the OrgConfig because the existing code is _so_ nasty with the combination of methods, properties, and config values - some of which override each other. I want to tear it all out and refactor it, but not in this PR!

PR #3255, opened 2022-06-24
Provide more context for git errors

prescod:
We have some excellent error checking infrastructure for github errors, but it didn't show enough context for the error I ran into. I had to launch a debugger to figure out what repo I did not have access to (due to nested dependencies).

Nothing for release notes.

### DIFF
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -1,12 +1,11 @@
+import functools
 import io
 import os
 import re
 import time
-import typing as T
 import webbrowser
-from contextlib import contextmanager
 from string import Template
-from typing import Optional, Union
+from typing import Callable, Optional, Union
 from urllib.parse import urlparse
 
 import github3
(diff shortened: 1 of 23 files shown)


cumulusci/core/github.py
```diff
@@ -510,23 +511,16 @@ def get_oauth_scopes(response: Response) -> set:
...
+@contextmanager
+def catch_common_github_auth_errors(context: T.Any):
+    try:
+        yield
+    except (ResponseError, TransportError) as exc:
+        error_msg = format_github3_exception(exc)
+        if error_msg:
+            raise GithubApiError(f"{context}\n{error_msg}")
+        else:
+            raise
```
jstvz, line 523: What about an alternate approach?: Update `format_github3_exception` to append `f"\n\nURI: {exc.request.url}` to `user_warning`. This would give the requested resource in the body of the error message without requiring the caller to specify a context.

jstvz (commented):
I like the additional information provided here.

cumulusci/core/github.py
prescod:
```
(Pdb) exc.request.url
*** AttributeError: 'ForbiddenError' object has no attribute 'request'
(Pdb) dir(exc)
['__cause__', '__class__', '__context__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__setstate__', '__sizeof__', '__str__', '__subclasshook__', '__suppress_context__', '__traceback__', '__weakref__', 'args', 'code', 'errors', 'message', 'msg', 'response', 'with_traceback']
(Pdb) vars(exc)
{'response': <Response [403]>, 'code': 403, 'errors': [], 'msg': '[No message]'}
(Pdb) exc.response.url
(Pdb) vars(exc.response)
{'_content': None, '_content_consumed': True, '_next': None, 'status_code': 403, 'headers': {'X-Github-Sso': 'partial-results; organizations=0810298,20348880'}, 'raw': None, 'url': None, 'encoding': None, 'history': [], 'reason': None, 'cookies': <RequestsCookieJar[]>, 'elapsed': datetime.timedelta(0), 'request': None}
```

jstvz: (I mean, we'd have to update the tests. 😄 )

Using it here:

https://github.com/SFDO-Tooling/CumulusCI/blob/7f797c9dbbf2cbfabaabf9530cd449c3742067da/cumulusci/core/github.py#L438-L440

_But_... looks like that `check_github_scopes` isn't getting tested.... so should put that on the todo list as well.

prescod: Even in real-world use there is no URL. That's how I ended up down this path:

```
cci flow run dev_org --org dev
...
                    Resolving dependencies...                                                                                     
> /Users/pprescod/code/CumulusCI/cumulusci/core/github.py(398)format_github3_exception()
(Pdb) vars(exc)
{'response': <Response [404]>, 'code': 404, 'errors': [], 'msg': 'Not Found'}
```

jstvz (approved):
Thanks

PR #3259, opened 2022-06-27
Fixes(?) to _get_required_permission_types

prescod:
I'm quite unclear on this code and hesitant to change it, but I think it may hide a bug.

If you remove the mocks that force `_validate_field_dict` to return the "right" value, it will return the "wrong" value. It's unclear to me why we would need the mocks except to work around a bug. Removing the mocks surfaced the bug, so I tried to fix it. But I'm not sure of my fix. The relationship between "operation" and "action" is fuzzy.

@davidmreed


prescod:
The tests don't pass but I'm unsure of the right path forward.

PR #3261, opened 2022-06-28
Extract MetaDeploy API from publish task

jstvz:
`make_api_session`:
- Extract the logic for setting the base URL into a subclass of
  requests.Session. This allows:
  - Read the api url and auth token directly from the ServiceConfig,
  - Automatically prepend base_url and set auth token,
  - First step in allowing calls from outside the task.

`MetaDeployApi`:
- Interface for the MetaDeploy REST API.

Tests for `metadeploy_publish` are left intact to keep them in place
as integration tests.

### DIFF
--- a/cumulusci/core/metadeploy/api.py
+++ b/cumulusci/core/metadeploy/api.py
@@ -1,0 +1,88 @@
+"""Calls to the MetaDeploy REST API."""
+
+import functools
+from typing import Optional, Union
+
+from requests.models import Response
+from requests.sessions import Session
+
+from cumulusci.core.config import ServiceConfig
+from cumulusci.core.exceptions import CumulusCIException
+from cumulusci.utils.http.requests_utils import safe_json_from_response
+
+
+def make_api_session(metadeploy_service: ServiceConfig) -> Session:
+    """
+    Given a MetaDeploy ServiceConfig, returns a requests.Session with
+    authorization headers and service.base_url set.
+    """
+    base_url: str = metadeploy_service.url
+
+    patched_session = Session()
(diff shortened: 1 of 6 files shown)


cumulusci/core/metadeploy/tests/test_api.py
```diff
@@ -0,0 +1,275 @@
...
+        "POST",
+        "https://metadeploy/api/rest",
+        json=expected_result,
+        match=[auth_header_matcher],
+    )
+    service: ServiceConfig = ServiceConfig(
+        {"url": "https://metadeploy/api", "token": DEFAULT_TOKEN}
+    )
+    api_session: Session = make_api_session(service)
+    resp: Response = api_session.get("/rest")
```
Br4nd0R, line 48: So if I'm reading this correctly you can use either the absolute _or_ relative url with session passed back from `make_api_session()`?

cumulusci/core/metadeploy/api.py
```diff
@@ -0,0 +1,88 @@
...
+        return func(method, base_url, *args, **kwargs)
+
+    patched_session.request = functools.partial(
+        patched_request, base_url, patched_session.request
+    )
+
+    return patched_session
+
+
+class MetaDeployApi:
```
Br4nd0R, line 36: Super clean, I like this a lot.

Br4nd0R (dismissed):

Br4nd0R (approved):

Br4nd0R (dismissed):

bethbrains (approved):

PR #3262, opened 2022-06-28
Decorate create_gist to catch scope errors

jstvz:


cumulusci/core/tests/test_github.py
```diff
@@ -515,6 +535,17 @@ def test_format_gh3_exc_retry(self):
         exc = TransportError(base_exc)
         assert UNAUTHORIZED_WARNING == format_github3_exception(exc)
 
+    def test_format_url_from_exc(self):
```
prescod, line 538: Is there a code path that is ONLY tested with this function, or is this a secondary test?

Because I would far prefer to test that github3 is compatible with our function than to test that the function is compatible with our current understanding of github3's exception strategy.

prescod: Does `test_create_gist_no_scope` test the same code path?

jstvz: Okay, I misunderstood what you meant by "real `TransportError`". I'll update.

> Does `test_create_gist_no_scope` test the same code path?

No, it covers the special case where requests made using a token without a `gist` are missing the `X-Accepted-OAuth-Scopes` header:

https://github.com/SFDO-Tooling/CumulusCI/blob/b7362d3945ee0ae9b4187c8927b41afd9d5a33e9/cumulusci/core/github.py#L421-L429

cumulusci/core/tests/test_github.py
```diff
@@ -532,6 +564,32 @@ def test_func():
...
+        mock_rsp.body = retry_err
+
+        @catch_common_github_auth_errors
+        def test_func():
+            GitHub().rate_limit()
+
+        with pytest.raises(
+            GithubApiError, match="^https://api.github.com/rate_limit\n"
+        ):
+            test_func()
```
jstvz, line 591: @prescod This work for you?

prescod (approved):

PR #3263, opened 2022-06-29
Add concurrency key to Release Test workflow

davidmreed:


prescod (dismissed):

davidmreed:
@prescod I tweaked this to remove the `push` trigger for the Release Test job, so it'll only run on open/updated PRs now. Hopefully that will reduce contention; GitHub also presented a confusing suggestion that the job was triggered twice on the previous commit.

PR #3264, opened 2022-06-30
Add count_sobjects function

prescod:
It uses parallelization and SOQL to count what sobjects are present.

### DIFF
--- a/cumulusci/utils/salesforce/count_sobjects.py
+++ b/cumulusci/utils/salesforce/count_sobjects.py
@@ -1,0 +1,48 @@
+import typing as T
+
+from simple_salesforce import Salesforce
+
+from cumulusci.utils.http.multi_request import CompositeParallelSalesforce
+from cumulusci.utils.iterators import partition
+
+
+class ObjectCount(T.NamedTuple):
+    counts: T.Dict[str, int]
+    transport_errors: T.Sequence[dict]  # e.g. couldn't reach server
+    salesforce_errors: T.Sequence[dict]  # e.g. 404, 401
+
+
+def count_sobjects(sf: Salesforce, objs: T.Sequence[str]) -> ObjectCount:
+    """Quickly count SObjects using SOQL and Parallelization"""
+    with CompositeParallelSalesforce(sf, max_workers=8, chunk_size=5) as cpsf:
+        responses, transport_errors = cpsf.do_composite_requests(
+            (
+                {
+                    "method": "GET",
+                    "url": f"/services/data/v{sf.sf_version}/query/?q=select count() from {obj}",
+                    "referenceId": f"ref{obj}",
+                }
+                for obj in objs
+            )
+        )
+        salesforce_errors, successes = partition(
+            lambda response: response["httpStatusCode"] == 200, responses
+        )
+
+        def normalize_errors(error):
+            if isinstance(error, dict):
+                return error
+            elif hasattr(error, "_asdict"):
+                return error._asdict()
+            else:
+                raise vars(error)
+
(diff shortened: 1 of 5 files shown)


cumulusci/utils/salesforce/count_sobjects.py
```diff
@@ -0,0 +1,46 @@
+import typing as T
+
+from simple_salesforce import Salesforce
+
+from cumulusci.utils.http.multi_request import CompositeParallelSalesforce
+from cumulusci.utils.iterators import partition
+
+
+class ObjectCount(T.NamedTuple):
```
davidmreed, line 9: What's the intent behind subclassing `T.NamedTuple`? Preferable to a plain old object? (Someday we'll be able to use dataclasses).

cumulusci/utils/salesforce/count_sobjects.py
```diff
@@ -0,0 +1,46 @@
...
+                    "referenceId": f"ref{obj}",
+                }
+                for obj in objs
+            )
+        )
+        salesforce_errors, successes = partition(
+            lambda response: response["httpStatusCode"] == 200, responses
+        )
+
+        transport_errors = tuple(error._asdict() for error in transport_errors)
```
davidmreed, line 32: `_asdict()` isn't public API. Acknowledging that it's less elegant, how about
```suggestion
        transport_errors = tuple({"exception": error.exception, "request": error.request} for error in transport_errors)
```

cumulusci/utils/salesforce/count_sobjects.py
```diff
@@ -0,0 +1,46 @@
...
+
+def count_sobjects(sf: Salesforce, objs: T.Sequence[str]) -> ObjectCount:
+    """Quickly count SObjects using SOQL and Parallelization"""
+    with CompositeParallelSalesforce(sf, max_workers=8, chunk_size=5) as cpsf:
+        responses, transport_errors = cpsf.do_composite_requests(
+            (
+                {
+                    "method": "GET",
+                    "url": f"/services/data/v{sf.sf_version}/query/?q=select count() from {obj}",
+                    "referenceId": f"ref{obj}",
```
davidmreed, line 23: Does Composite actually require that reference ids start with `ref`? Cannot use the raw sObject name?

davidmreed (changes requested):

cumulusci/utils/salesforce/count_sobjects.py
davisagli: I promise I'm not still reading CumulusCI pull requests closely, but I still get notifications and saw this go by. `_asdict` _is_ documented public API, despite the underscore: https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict -- namedtuple uses the underscore for its public methods to avoid the potential for collision with namedtuple field names

davidmreed: Oh good grief. Thanks for the heads-up, @davisagli.

cumulusci/utils/salesforce/count_sobjects.py
prescod: It really is a named tuple. A tuple of return values that I gave names to for readability purposes. Dataclassses would be way overkill, performance-wise.

cumulusci/utils/salesforce/count_sobjects.py
prescod: No it's not a rule, just a convention used widely in the docs. Makes it slightly easier to read wire JSON if it is clear when we are referring to an SObject and when to a referenced record. Not opposed to getting rid of it if you would rather save the bytes.

davidmreed (approved):

PR #3265, opened 2022-06-30
Filtering for Schema caching and querying

prescod:
It is now possible for a consumer of the Schema querying/caching feature to specify criteria such as "I'm only interested in  deepClonable objects" or "I'm not interested in objects that start with the word Macro" or "I'm only interested in objects populated with data" and so forth.


cumulusci/salesforce_api/org_schema.py
```diff
@@ -348,3 +499,31 @@ def deep_describe(
             for response in responses
         )
         yield from responses
+
+
+def ignore_based_on_properties(obj: dict, filters: T.Sequence[Filters]):
+    if all(obj.get(filter.name, True) for filter in filters):
+        return False
+    else:
+        return True
```
davidmreed, line 508:
```suggestion
    return not all(obj.get(filter.name, True) for filter in filters)
```

cumulusci/salesforce_api/org_schema.py
```diff
@@ -348,3 +499,31 @@ def deep_describe(
...
+NOT_COUNTABLE = (
+    "ContentDocumentLink",  # ContentDocumentLink requires a filter by a single Id on ContentDocumentId or LinkedEntityId
+    "ContentFolder%",  # Implementation restriction: ContentFolderItem requires a filter by Id or ParentContentFolderId
+    "IdeaComment",  # you must filter using the following syntax: CommunityId = [single ID],
+    "Vote",  # you must filter using the following syntax: ParentId = [single ID],
+    "RecordActionHistory",  # Gack: 1133111327-118855 (1126216936)
+)
+
+
+NOT_EXTRACTABLE = NOT_COUNTABLE + (
```
davidmreed, line 520: I think this list is risky. It's valid to name a custom object `User_Permission__c`, for example, and if I read right this would result in it being excluded.

Share records _are_ migrateable but we will not handle them well for other reasons (polymorphic relations). 

`UserRecordAccess` and similar are more like a view than a real object. Can we exclude them based on either explicit names or some other facet of their schema to avoid collateral damage?

prescod: Yes, you're right. This was hacky stuff I did while I was exploring and forgot to clean up later. I will blow these up to explicit lists.

prescod: @davidmreed Your comment was addressed in 99e1f3d

cumulusci/salesforce_api/filterable_objects.py
```diff
@@ -0,0 +1,260 @@
...
+#     "OmniSupervisorConfigUser",
+#     "OpportunityShare",
+#     "OrderShare",
+#     "OutgoingEmail",
+#     "OutgoingEmailRelation",
+# )
+
+# And this code:
+#
+# patterns_to_ignore = NOT_EXTRACTABLE_WITHOUT_NOT_COUNTABpytestLE
```
davidmreed, line 250:
```suggestion
# patterns_to_ignore = NOT_EXTRACTABLE_WITHOUT_NOT_COUNTABLE
```

cumulusci/salesforce_api/filterable_objects.py
```diff
@@ -0,0 +1,260 @@
...
+    "ContentFolder",  # Similar to above
+    "ContentFolderLink",  # Similar to above
+    "ContentFolderMember",  # Similar to above
+    "IdeaComment",  # you must filter using the following syntax: CommunityId = [single ID],
+    "Vote",  # you must filter using the following syntax: ParentId = [single ID],
+    "RecordActionHistory",  # Gack: 1133111327-118855 (1126216936)
+)
+
+
+NOT_EXTRACTABLE = NOT_COUNTABLE + (
```
davidmreed, line 13: I'm fine with going forward using this to support the feature, but I think we'll need to revise it before GA. Every object with a non-`Public` sharing model will create a new `__Share` object, so we'll need a different rubric to recognize them dynamically.

davidmreed (approved):

cumulusci/salesforce_api/filterable_objects.py
prescod: What do you think about 4726c33945

cumulusci/salesforce_api/filterable_objects.py
```diff
@@ -0,0 +1,261 @@
...
+    "%__Share",
```
davidmreed, line 14: Assuming these regexes are applied using a whole-string match, I think we can just use

```suggestion
    "%Share",
```

and remove the specific entries below. Share objects always end in `Share` for standard objects, and `__Share` for custom objects.

cumulusci/salesforce_api/org_schema.py
```diff
@@ -229,27 +314,59 @@ def close(self) -> None:
...
             schema_path.unlink()
 
+        if Filters.populated in filters:
+            filters.add(Filters.queryable)
+            filters.add(Filters.retrieveable)
+            # experiment with removing this limitation by using limit 1 query instead
+            assert include_counts, "Filters.populated depends on include_counts"
+            patterns_to_ignore += NOT_COUNTABLE
+
+        if Filters.extractable in filters:
```
davidmreed, line 365: Doesn't need to be in version 1, but we might want to add a "loadable" criterion (`queryable` + `retrieveable` + `createable`)

davidmreed (approved):
Going ahead and approving so you can move forward; one suggested tweak to the `Share` pattern below.

PR #3276, opened 2022-07-07
Move to PEP-517-ish setup.cfg

jstvz:
Need to maintain setup.py for editable installs.

Substantive changes:
- _Update package author metadata_
- Added classifiers
- Added project_urls
- No longer concatenate README and HISTORY for long_description

Probably worth investigating alternate build systems if we ever get the chance


jstvz:
Need to RTFM, closing for now.

PR #3287, opened 2022-07-13
Optionally support self signed SSL certs for GitHub Enterprise connections

davidjray:


cumulusci/core/keychain/base_project_keychain.py
```diff
@@ -274,6 +274,12 @@ def list_services(self):
                 services[s_type].append(name)
         return services
 
+    def get_services_for_type(self, service_type: str) -> list:
```
davidjray, line 277: Turn out this IS needed to access unencrypted service properties

davisagli:
I would strongly discourage providing this option. Remember, disabling certificate verification doesn't mean only that a self-signed certificate is okay. It means that any man-in-the-middle attacker who presents a certificate claiming to be the target server will be trusted.

davidjray:
Closing in favor of setting `CUMULUSCI_SYSTEM_CERTS` see: https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/utils/http/requests_utils.py#L23

davidjray:
Thanks for the feedback @davisagli !

PR #3295, opened 2022-07-20
Mask sensitive attributes in cci service info output

bethbrains:


cumulusci/cli/service.py
```diff
@@ -283,7 +283,10 @@ def service_info(runtime, service_type, service_name, plain):
         service_data = [["Key", "Value"]]
         service_data.extend(
             [
-                [click.style(k, bold=True), str(v)]
+                [
+                    click.style(k, bold=True),
+                    (v[:5] + (len(v[5:]) * "*") if k == "token" else str(v)),
```
davidmreed: Hard-coding the name here will cure the immediate P1 bug, but it's fragile and breaks the abstraction for Services. 

I suggest instead that we get the project config (`runtime.project_config or runtime.universal_config`). Then we can get the service definition (see `cumulusci/cumulusci.yml` around line 1306 and `service.py` around line 27-68) and find the `attribute` for this key. 

We can add to each attribute a new k/v pair in YAML: `secure: bool`.

Then we mask any attribute marked secure, which should include the GitHub token but also any future service with the same issue.

davidmreed: OK, you pushed a commit doing essentially this within seconds of me submitting the review. Nice, @bethbrains

davisagli:
Check what the similar setting for task options is called. IIRC it was sensitive rather than secure

David

davidmreed (approved):

jstvz (changes requested):
Run `make schema` to update the JSON schema and resolve the test failures.

cumulusci/core/github.py
```diff
@@ -132,7 +132,9 @@ def validate_service(options: dict) -> dict:
         repo_generator = gh.repositories()
         _ = next(repo_generator, None)
         repo_response = repo_generator.last_response
-        options["scopes"] = get_oauth_scopes(repo_response)
+        options["scopes"] = ", ".join(
+            v for v in sorted(get_oauth_scopes(repo_response))
+        )
```
jstvz, line 137:
```suggestion
        options["scopes"] = ", ".join(sorted(get_oauth_scopes(repo_response)))
```

jstvz (approved):

PR #3298, opened 2022-07-20
Validate MetaDeploy plans in cumulusci.yml

jstvz:
Validate that MetaDeploy plans only specify a single primary and secondary plan.

Because `plan.tier` defaults to `primary` in the API, we set the same here, with the expectation that users that do not explicitly set `tier` will see these warnings before plan validation is enforced by `cci plan publish`.

Other callers can validate project_configs with:

```python
with pytest.raises(ValueError, match='Only one plan'):
    CumulusCIRoot.validate_plan_tiers(plans_with_multiple_primary)
```

### DIFF
--- a/cumulusci/utils/yaml/tests/test_cumulusci_yml.py
+++ b/cumulusci/utils/yaml/tests/test_cumulusci_yml.py
@@ -225,3 +225,62 @@ def test_github_source():
         GitHubSourceModel(github="https://github.com/Test/TestRepo").resolution_strategy
         == "production"
     )
+
+
+def test_single_primary_plan(caplog):
+    yaml = """
+plans:
+  first:
+    tier: primary
+  second:
+    tier: primary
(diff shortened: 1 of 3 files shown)


cumulusci/utils/yaml/tests/test_cumulusci_yml.py
```diff
@@ -225,3 +225,39 @@ def test_github_source():
         GitHubSourceModel(github="https://github.com/Test/TestRepo").resolution_strategy
         == "production"
     )
+
+
+def test_single_primary_plan(caplog):
```
bethbrains, line 230: Can we add a golden path test as well, please?

bethbrains (changes requested):

bethbrains (dismissed):

bethbrains (approved):

PR #3309, opened 2022-07-26
Github Enterprise support task fix

davidjray:
Fix `install_package_version` task to not hardcode github.com, which helps with GitHub enterprise compatibility

### DIFF
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -298,7 +298,7 @@ class BaseProjectConfig(BaseTaskFlowConfig, ProjectConfigPropertiesMixin):
                 return str(path)
 
     @property
-    def repo_domain(self):
+    def repo_domain(self) -> Optional[str]:
         domain = self.repo_info.get("domain")
 
         if domain:
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -35,7 +35,7 @@ from cumulusci.core.github import (
 from cumulusci.core.source import GitHubSource, LocalFolderSource, NullSource
 from cumulusci.core.utils import merge_config
 from cumulusci.utils.fileutils import open_fs_resource
-from cumulusci.utils.git import current_branch, git_path, split_repo_url
+from cumulusci.utils.git import current_branch, git_path, parse_repo_url, split_repo_url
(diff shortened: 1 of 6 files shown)


cumulusci/core/github.py
```diff
@@ -117,8 +117,8 @@ def get_github_api_for_repo(keychain, repo_url, session=None):
     elif GITHUB_TOKEN:
         gh.login(token=GITHUB_TOKEN)
```
davidjray, line 118: Leaving this untouched because of backwards compatibility. This will still also support github enterprise connections, but only for single connections. The modifications we are adding overall to support GitHub Enterprise Server, will allow greater flexibility without changing these lines..

cumulusci/core/config/project_config.py
```diff
@@ -297,6 +297,20 @@ def repo_root(self):
             if (path / ".git").is_dir():
                 return str(path)
 
+    @property
+    def repo_domain(self):
```
davidmreed, line 301:
```suggestion
    def repo_domain(self) -> Optional[str]:
```

cumulusci/core/github.py
```diff
@@ -134,16 +134,18 @@ def get_auth_from_service(host, keychain) -> tuple:
...
         if list(service_by_host.keys()).count(host) == 0:
             raise ServiceNotConfigured(
                 f"No Github Enterprise service configured for domain {host}."
             )
 
         service_config = service_by_host[host]
 
-    token = service_config.password or service_config.token
-    return service_config.username, token
+    # Basic Auth no longer supported on github.com, so only returnign token
```
davidmreed, line 145:
```suggestion
    # Basic Auth no longer supported on github.com, so only returning token
```

cumulusci/tasks/salesforce/install_package_version.py
```diff
@@ -91,7 +91,7 @@ def _init_options(self, kwargs):
             ]
 
         dependency = None
-        github = f"https://github.com/{self.project_config.repo_owner}/{self.project_config.repo_name}"
+        github = f"https://{self.project_config.repo_domain}/{self.project_config.repo_owner}/{self.project_config.repo_name}"
```
davidmreed, line 94: Ah, yes, this is rather nasty; thanks for finding it and fixing.

davidmreed (approved):

PR #3316, opened 2022-08-04
Fix get_latest_prerelease on github enterprise

jstvz:
Github3.py sets `session.base_url` to `f'http(s)://{hostname}/api/v3` for GitHubEnterprise sessions, so `build_url` returns the wrong GraphQL endpoint. This results in an inscrutable `KeyError` when accessing the query results. This commit removes the suffix and manually constructs the URL.

### DIFF
--- a/cumulusci/core/github.py
+++ b/cumulusci/core/github.py
@@ -301,14 +301,19 @@ def get_latest_prerelease(repo: Repository) -> Optional[Release]:
     ).substitute(dict(owner=repo.owner, name=repo.name))
 
     session: GitHubSession = repo.session
-    url: str = session.build_url("graphql")
+    # HACK: This is a kludgy workaround because GitHub Enterprise Server
+    # base_urls in github3.py end in `/api/v3`.
+    host = (
+        session.base_url[: -len("/v3")]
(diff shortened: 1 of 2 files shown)


cumulusci/core/github.py
```diff
@@ -301,14 +301,19 @@ def get_latest_prerelease(repo: Repository) -> Optional[Release]:
...
+    )
+    url: str = f"{host}/graphql"
     response: Response = session.request("POST", url, json={"query": QUERY})
     response_dict: dict = response.json()
 
-    release_tags = response_dict["data"]["repository"]["releases"]["nodes"]
-    if release_tags:
-        release = repo.release_from_tag(release_tags[0]["tagName"])
-        return release
+    if release_tags := response_dict["data"]["repository"]["releases"]["nodes"]:
```
davidjray, line 315: :walrus:

davidjray (approved):
LGTM

cumulusci/core/github.py
jstvz: > Walrus?
> 🦭 is best I could do

–Unicode Consortium, probably

PR #3317, opened 2022-08-09
Better SSL Certificate Error Message

davidjray:


davisagli:
Is it worth trying to clarify that the certificate might not actually be self-signed, but could be signed by a certificate authority that isn't trusted?

davidjray:
> Is it worth trying to clarify that the certificate might not actually be self-signed, but could be signed by a certificate authority that isn't trusted?
I'm trying to keep the message short, but I do like the idea since it clarifies why the env is a possible solution.

cumulusci/core/github.py
```diff
@@ -444,6 +453,12 @@ def format_github3_exception(exc: Union[ResponseError, TransportError]) -> str:
         sso_error_msg = check_github_sso_auth(exc)
         user_warning = scope_error_msg + sso_error_msg
 
+    if isinstance(exc, ConnectionError):
+        if "self signed certificate" in str(exc.exception):
+            user_warning = SELF_SIGNED_WARNING
+        else:
+            user_warning = exc.msg
```
jstvz, line 460: ❓ : Would it make sense to return `None` in the case where we have a `ConnectionError` that isn't a self signed certificate? In that case we don't have any advice to offer the user as it may not be related to the GitHub API (maybe they're on a spotty mobile network, etc).

cumulusci/core/github.py
```diff
@@ -570,6 +585,11 @@ def catch_common_github_auth_errors(func: Callable) -> Callable:
     def inner(*args, **kwargs):
         try:
             return func(*args, **kwargs)
+        except (ConnectionError) as exc:
+            if error_msg := format_github3_exception(exc):
```
jstvz, line 589: `(:3 っ)っ`

(Last one, I swear)

cumulusci/core/github.py
```diff
@@ -444,6 +453,12 @@ def format_github3_exception(exc: Union[ResponseError, TransportError]) -> str:
         sso_error_msg = check_github_sso_auth(exc)
         user_warning = scope_error_msg + sso_error_msg
 
+    if isinstance(exc, ConnectionError):
+        if "self signed certificate" in str(exc.exception):
```
jstvz, line 457: 💭 If I'm understanding correctly, this _should_ be platform independent. If I'm not, we may see some other self-signed error messages that won't match. I think that's a reasonable trade-off.

jstvz (changes requested):
One blocking query.

---
<details>
<summary>Emoji Key</summary>


</details>

cumulusci/core/github.py
```diff
@@ -444,6 +453,12 @@ def format_github3_exception(exc: Union[ResponseError, TransportError]) -> str:
...
+            user_warning = SELF_SIGNED_WARNING
+        else:
+            return None  # type: ignore
```
jstvz, line 460: Ah, I was too... `Literal`. Should have said empty string
```suggestion
            return ''
```

jstvz (approved):

PR #3318, opened 2022-08-10
Publish MetaDeploy plans in a deterministic order

bethbrains:
Plans are published in a deterministic order: sorted by tier, and then by the order they appear in the `cumulusci.yml` file.
Clarified error message that appears when a user tries to include more than one primary or more than one secondary Plan in their `cumulusci.yml` file.

### DIFF
--- a/cumulusci/tasks/metadeploy.py
+++ b/cumulusci/tasks/metadeploy.py
@@ -97,7 +97,7 @@ class Publish(BaseMetaDeployTask):
                 plan_name: self.project_config.lookup(f"plans__{plan_name}")
             }
         else:
-            self.plan_configs = self.project_config.plans
+            self.plan_configs = self._sort_plans()
 
         self.labels = read_default_labels(self.labels_path)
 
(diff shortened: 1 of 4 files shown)


cumulusci/tasks/metadeploy.py
```diff
@@ -175,7 +176,9 @@ def _publish_plan(self, product, version, plan_name, plan_config, steps):
             product, plan_name, plan_config
         )
 
-        parsed_plan = Plan.parse_obj(self.project_config.config["plans"][plan_name])
+        order_key = plan_config.pop("order_key")
+        parsed_plan = Plan.parse_obj(plan_config)
+        # TODO: clean this up to not use both Plan and MetaDeployPlan
```
jstvz, line 181: 💭 Yeah, the interaction between these two needs to be revisited. Maybe a class method on `MetaDeployPlan` that'll do the data munging or something.

cumulusci/tasks/metadeploy.py
```diff
@@ -131,7 +131,8 @@ def _run_task(self):
             project_config.set_keychain(self.project_config.keychain)
 
             # Create each plan
-            for plan_name, plan_config in self.plan_configs.items():
+            for i, (plan_name, plan_config) in enumerate(self.plan_configs.items()):
+                plan_config["order_key"] = i
```
jstvz, line 135: I accept responsibility for this temporary kludge.

jstvz (approved):

PR #3324, opened 2022-08-17
Generate mappings and next generation datasets

prescod:
Infrastructure required to extract from orgs without creating a mapping file (by auto-generating a mapping file internally).

This PR is essentially a lot of algorithms for looking at a real org Schema + Extract Declarations and generating two mapping files, one for Extracting data and one for Loading data.

No user-interfaces exist yet for these new features.

A new abstraction called a Dataset will go in another PR.


Br4nd0R (approved):
@prescod - Code looks great. People are going to to really enjoy this feature!

PR #3335, opened 2022-08-18
Add task set_object_settings to enable and disable object settings

jlantz:
# Changes
* New Metadata ETL task `set_object_settings` to enable and disable settings on standard and custom objects


cumulusci/tasks/metadata_etl/objects.py
```diff
@@ -0,0 +1,65 @@
+from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.core.utils import process_list_arg
+from cumulusci.tasks.metadata_etl import MetadataSingleEntityTransformTask
+from cumulusci.utils.xml.metadata_tree import MetadataElement
+
+OBJ_SETTINGS = [
```
davidmreed, line 6: I like this design choice a lot.

cumulusci/tasks/metadata_etl/objects.py
```diff
@@ -0,0 +1,65 @@
...
+class SetObjectSettings(MetadataSingleEntityTransformTask):
+    entity = "CustomObject"
+    task_options = {
+        "enable": {
+            "description": f"Array of object settings to enable. Uses the setting name.  Available values: {OBJ_SETTINGS_STR}"
+        },
+        "disable": {
+            "description": f"Array of object settings to disable. Uses the setting name.  Available values: {OBJ_SETTINGS_STR}"
+        },
+        # Only include api_names option since namespace injection isn't supported
```
davidmreed, line 29: Let's get namespace injection supported here - it could be needed for projects that wish to mutate their own settings.

cumulusci/tasks/metadata_etl/objects.py
```diff
@@ -0,0 +1,65 @@
...
+            raise TaskOptionsError(
+                "Invalid settings: {}.  Valid settings are: ".format(
+                    ", ".join(invalid), OBJ_SETTINGS_STR
+                )
+            )
+        return settings
+
+    def _transform_entity(self, metadata: MetadataElement, api_name: str):
+        for setting in self.options["enable"]:
+            elem = metadata.find("enable" + setting)
```
davidmreed, line 60: I'm not 100% sure that all of the settings are always present in every object metadata file. Do you know? Could we easily make this resilient to the element being missing?

cumulusci/tasks/metadata_etl/objects.py
davidmreed: There's an example of supporting injection on the `CustomObject` entity in `SetOrgWideDefaults`.

davidmreed (changes requested):

cumulusci/tasks/metadata_etl/objects.py
jlantz: I'll add a test for that and make sure we handle it correctly

cumulusci/tasks/metadata_etl/objects.py
jlantz: OK, I wasn't sure if there was a good use case for modifying the project's objects via this task.  I guess I could see a case where you want, for example, history disabled in the packaged object but then a solution configuration might want to enable it.

I'll get that added in and tested.

cumulusci/tasks/metadata_etl/objects.py
jlantz: Yeah, I originally had it where you have to pass `enableReports` instead of just `Reports`.  The drawback is that you have to list all the choices in the description for the option because I don't know of any shorter way to even explain what the value should be "take the element name like enableReports and remove the enable" seems confusing ;)

cumulusci/tasks/metadata_etl/objects.py
jlantz: I think this was easier than I thought.  Since the elements in the object aren't ever namespaced, I think the base class handles namespace injection in the api_names for me automatically.  I just added those options back.

cumulusci/tasks/metadata_etl/objects.py
jlantz: It wasn't handled in the code before.  I changed it to create the element if it's a valid setting and added a test for that.

jlantz:
@davidmreed Ready for re-review

davidmreed (approved):
Looks great! 🎉

prescod:
Superseded by #3364 due to github hating open source contributors (?)

PR #3339, opened 2022-08-19
Remove reference to homebrew from documentation.

prescod:
This documentation was not correct and not helpful.


davisagli:
It's there because the docs then say to use brew to install pipx

prescod:
Good point, @davisagli ! 

And unfortunately that's how pipx itself suggests you install it.

PR #3342, opened 2022-08-22
Add 3.11 Release Candidate to feature test matrix

jstvz:


jstvz:
Proximate cause is python-greenlet/greenlet#305 upstream.

PR #3344, opened 2022-08-23
Specify Python minor version in homebrew formula

jstvz:
Users were intermittently reporting that upgrades failed with some
variation of the following error:

```python-traceback
File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.8/lib/python3.8/site-packages/setuptools/build_meta.py", line 142, in run_setup
     exec(compile(code, __file__, 'exec'), locals())
   File "setup.py", line 14, in <module>
     from setuptools_rust import RustExtension
   File "/private/tmp/pip-build-env-gh8ot2a3/overlay/lib/python3.8/site-packages/setuptools_rust/__init__.py", line 1, in <module>
     from .build import build_rust
   File "/private/tmp/pip-build-env-gh8ot2a3/overlay/lib/python3.8/site-packages/setuptools_rust/build.py", line 23, in <module>
     from setuptools.command.build import build as CommandBuild  # type: ignore[import]
 ModuleNotFoundError: No module named 'setuptools.command.build'
```
This changeset includes two substantive changes to how the formula is
generated:

1. The major.minor version (`python3.9`of the python executable is used instead of `python3`). This prevents the installation from inadvertently using the system Python. This occurs because Homebrew only symlinks a `python3` executable for the latest version. See Homebrew/homebrew-core#107517.
3. Adds `--allow-unsafe` to the `pip-compile` step. This is a noop because we aren't pinning setuptools.

### DIFF
--- a/utility/build-homebrew.sh
+++ b/utility/build-homebrew.sh
@@ -21,7 +21,9 @@ echo " "
 echo "=> Compiling pip requirements including hashes..."
 echo " "
 echo "cumulusci==$PACKAGE_VERSION" > "$REQS_IN_FILE"
-pip-compile "$REQS_IN_FILE" -o "$REQS_FILE" --generate-hashes
+# For a longish thread on why `allow-unsafe`, see https://github.com/pypa/pip/issues/6459
+# Here, we're adding it to collect setuptools-rust for the cryptography library.
+pip-compile "$REQS_IN_FILE" -o "$REQS_FILE" --generate-hashes --allow-unsafe
 
 echo " "
 echo "=> Writing Homebrew Formula to ${OUT_FILE}..."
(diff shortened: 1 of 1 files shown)


utility/build-homebrew.sh
```diff
@@ -34,8 +36,10 @@ class Cumulusci < Formula
   homepage "https://github.com/SFDO-Tooling/CumulusCI"
   url $PACKAGE_URL
   sha256 $PACKAGE_SHA
-  head "https://github.com/SFDO-Tooling/CumulusCI.git"
+  license "BSD-3-Clause"
+  head "https://github.com/SFDO-Tooling/CumulusCI.git", branch: "main"
 
+  depends_on "rust" => :build # cryptography -> setuptools-rust
```
prescod, line 42:
```suggestion
```

prescod (changes requested):
Let's try it without Rust. I think based on our conversation that we don't need it.

utility/build-homebrew.sh
jstvz: ✍🏾 : The idea here is that we don't need this because we're installing this using pip, which makes use of binary wheels to avoid the need to compile dependencies like cffi/cryptography.

prescod (approved):

PR #3345, opened 2022-08-26
CCI projects can now run tasks from their "sources" even when they are custom tasks coded in Python

prescod:
If a flow uses tasks that are implemented in custom Python code
then you may instruct CumulusCI to allow that code to run:

```yaml
sources:
    eda:
        github: https://github.com/SalesforceFoundation/EDA
        allow_remote_code: True
```


davidmreed:
Two thoughts:

1. We need to carefully evaluate the changes to security posture we are introducing here. We might want to introduce new markup on `source` to indicate whether or not the remote repo is allowed to execute custom Python.
2. We need to do E2E tests with MetaDeploy and Metecho to ensure that the web apps are fully compatible.

prescod:
@davidmreed . Yes, you're right. At this point this is considered a "spike" and your idea about having an opt-in or opt-out is good.

cumulusci/core/config/project_config.py
```diff
@@ -641,8 +654,29 @@ def include_source(
             project_config.source = source
             self.included_sources[spec] = project_config
 
+            # If I can't load remote code, make sure that my
```
jofsky, line 657: Thank you for this comment really helps :)

cumulusci/core/exceptions.py
```diff
@@ -134,6 +134,12 @@ class TaskOptionsError(CumulusCIUsageError):
     pass
 
 
+class TaskImportError(CumulusCIException):
+    """Raise when a task's classcan't be loaded"""
```
jofsky, line 138:
```suggestion
    """Raise when a task's class can't be loaded"""
```

cumulusci/core/source/__init__.py
```diff
@@ -3,7 +3,11 @@
 
 
 class NullSource:
```
jofsky, line 5: Can you educate me more on this? The Class name makes it evident however would love to know more for my knowledge, why not handle this in cumulusci defaults? or init_options?

cumulusci/core/tests/untrusted_repo_parent/cumulusci.yml
```diff
@@ -0,0 +1,22 @@
...
+    example_task:
+        description: An example project-level task that does nothing
+        class_path: tasks.untrusted_parent.ExampleTask
+
+flows:
+    untrusted:
+        steps:
+            1:
+                task: example_task
+
```
jofsky, line 20: Is this spacing intentional?

jofsky (commented):
Have a few questions regarding implementation. Love this!

cumulusci/core/tests/untrusted_repo_parent/cumulusci.yml
prescod: I'm not sure if it changed since you commented. I do sometimes use a newline between flow steps. It would be best to be consistent but I didn't notice until you mentioned it.

cumulusci/core/source/__init__.py
prescod: Yeah, let' discuss tomorrow!

jofsky (approved):

yippie:
Is there an ETA on this? This is a super important feature to be able to manage our projects in any kind of reasonable way.

prescod:
> Is there an ETA on this? This is a super important feature to be able to manage our projects in any kind of reasonable way.

I will check with the team doing the review.

jofsky (dismissed):

prescod (commented):
Bump.

TheBitShepherd (approved):
Nice work on this. Tests look sufficient, let's ship!

PR #3359, opened 2022-09-14
Update readthedocs configuration

jstvz:
Update author name to "Cumulus Suite Development Team"


jofsky (dismissed):

kdvenkata:
@jstvz @jofsky  Like to see Salesforce mentioned the author section. Can we add Salesforce explicitly ? Also what is the cadence of the pdf generation ?

jstvz:
@kdvenkata: 
> Like to see Salesforce mentioned the author section. Can we add Salesforce explicitly ? Also what is the cadence of the pdf generation ?

The docs are copyrighted by Salesforce here:

https://github.com/SFDO-Tooling/CumulusCI/blob/52c9f831cd966c839a7b585dec453df0b8f62d82/docs/conf.py#L78

However, Sphinx' LaTeX output doesn't include the copyright statement ([issue](https://github.com/sphinx-doc/sphinx/issues/2876)), so we'd have to add it via a LaTeX macro. Alternatively, we can replace this `author` value or add an additional `author` ([docs](https://www.sphinx-doc.org/en/master/usage/configuration.html?highlight=sphinxbackoftitlepage#confval-latex_documents)).

Since this change would be reflected in the `stable` docs after this PR is merged I would like to go ahead and merge this as-is while we figure which approach makes sense. Any objection?

prescod (approved):
@jstvz I'll approve on the condition that you make a WI or start a Slack thread to figure out the other issue.

Issue #3283, opened 2022-07-12
json parser error when empty string passed for date field


GeekStewie:
### Describe the bug

I have a mapping file where the Birth Date is defined for the object Contact. After I have extracted the data from an org and it has created the .sql file, using the extract_dataset task, I have some dates populated and some are blank (this was the expected output).

However when I then run load_dataset, I get the following error for the rows with no Birth Date defined.

`Error on record with id 28: JSON_PARSER_ERROR: Cannot deserialize instance of date from VALUE_STRING value  or request may be missing a required field at [Source: N/A; line: -1, column: -1][line:-1, column:-1] ()  `

Adding a random date to these records, removes the errors and allows the records to insert but I would expect to be able to pass in a blank value for a date if it's a non-required field.

### Reproduction steps

Add some Contact records to an org, make sure some of them have a birth date and some do not.

Create a mapping file which extracts fields from Contact. Ensure the standard field for Birth Date has been added.

Extract the data.

Load the data into a different org.

The error should show for rows without a value specified for Birth Date.

### Your CumulusCI and Python versions

CumulusCI version: 3.61.1 (/usr/local/Cellar/cumulusci/3.61.1/libexec/bin/cci)
Python version: 3.9.13 (/usr/local/opt/python@3.9/bin/python3.9)

### Operating System

MacOS 12.4

### CumulusCI installation method

brew

prescod:
@GeekStewie  Can you attach or send me an example SQL/mapping file that causes this problem? My first attempts to reproduce have failed.

prescod:
Does your mapping file have reference to the relative date feature?

GeekStewie:
@prescod I have created an example org where I have replicated the issue. See DM on Slack.

prescod:
Fixed in #3361

jstvz:
Closing as part of a backlog triage pass against v4.10.0 / 5.0.0.dev1.

Reason: pr-resolved-#3361

If this is still reproducible against v4.10.0 (the current latest stable) or against v5.0.0.devN on the dev branch, please reopen with a fresh `cci version` block and reproduction steps. We track v5 development on the `dev` branch.

PR #3361, opened 2022-09-16
Support updates and upserts with blank dates represented by strings

prescod:
A dataset like this:

```
BEGIN TRANSACTION;
CREATE TABLE "Contact" (
	id INTEGER NOT NULL, 
	"FirstName" VARCHAR(255), 
	"LastName" VARCHAR(255), 
	"Birthdate" VARCHAR(255), 
	"External_ID__c" VARCHAR(255), 
	"IsPersonAccount" VARCHAR(255), 
	PRIMARY KEY (id)
);
INSERT INTO "Contact" VALUES(1,'AAA','BBBB','','C2','False');
COMMIT;
```

Previously generated an error like this:

```
                    WARNING  Error on record with id 1: JSON_PARSER_ERROR: Cannot deserialize   utils.py:141
                             instance of date from VALUE_STRING value  or request may be                    
                             missing a required field at [Source: N/A; line: -1, column:                    
                             -1][line:-1, column:-1] ()                                                     
```


prescod:
Fixes #3283

cumulusci/tasks/bulkdata/tests/test_step.py
```diff
@@ -1386,3 +1386,66 @@ def test_get_dml_operation__bad_api(self, rest_dml, bulk_dml):
...
+            "fields": [
+                {"name": "Birthdate", "type": "date"},
+                {"name": "IsHappy", "type": "boolean"},
+                {"name": "Name", "type": "string"},
+            ],
+        }
+
+        step = get_dml_operation(
+            sobject="Test__c",
+            operation=DataOperationType.UPSERT,
```
boakley, line 1436: `_record_to_json` has a branch that includes both UPDATE and UPSERT but this test only covers UPSERT. Should there be a test for UPDATE?

boakley (approved):
This looks pretty good to me. I made one small comment, but it's not a show-stopper.

boakley (approved):
Looks good to me. I've never used `@pytest.mark.parameterize` before, now I have a reason to read up on it.

PR #3367, opened 2022-09-21
Fix CCI --version flag

prescod:


jstvz (commented):
What do you think of this approach? More verbose, but you don't have to understand how string formatting and magic methods work.

Either way, a brief docstring explaining why couldn't hurt.

jstvz (approved):

PR #3368, opened 2022-09-23
Add AIApplicationConfig to metadata_map

TheBitShepherd:
Title says it all. [Docs](https://developer.salesforce.com/docs/atlas.en-us.238.0.api_meta.meta/api_meta/meta_aiapplicationconfig.htm) for reference.


jstvz (approved):
LGTM

We're missing a few more metadata types based on the [changes to SFDX's metadataRegistry.json](https://github.com/forcedotcom/source-deploy-retrieve/compare/8b5b50..main#diff-af7a780a5aa66c4831707c3ccd4e24d4f88a6639346bc0b1897c287efcbe14c9) file since the last time we updated this file.

Let's merge this and cut a dev release to unblock users, but we'll want to revisit this problem and implement a permanent fix by autogenerating this file (or consuming metadataRegistry.json directly).

PR #3374, opened 2022-09-28
w39 dependency updates

TheBitShepherd:


prescod (approved):

requirements/prod.txt
```diff
@@ -29,42 +29,35 @@ cryptography==38.0.1
...
+idna==3.4
     # via
     #   requests
     #   snowfakery
 importlib-metadata==4.12.0
     # via keyring
-jeepney==0.8.0
-    # via
-    #   keyring
-    #   secretstorage
```
jstvz, line 58: Transitive dependency that gets pulled in depending on whether `pip-compile` is run on Linux.

jstvz (approved):

PR #3379, opened 2022-09-30
Use parent_tree in CommitDir  to fix timeouts with large repositories

jstvz:
Users have reported 502 errors (i.e., timeouts) when using CommitDir to create a commit when the repository is above a certain size. Because CommitDir uses GitHub's [Git Database API] directly to create blob and tree objects before authoring a commit, it made too many calls and triggered rate limiting. For very large repositories (e.g., NPSP), even very small changes will throw an exception because CommitDir constructs the entire Tree and creates a blob for each file that is new or modified.

This PR resolves this issue with two changes:

1. CommitDir now creates a smaller tree object by omitting unchanged files. By using the `parent_tree` property in the `create_tree` request, we only need to include the blobs that are new, modified, or deleted, meaning the overall tree is much smaller.
2. CommitDir no longer calls the `blob` endpoint for each new or modified file before constructing the new tree. Github will accept property `contents` with the text contents of a file, so only binary files need to create a blob first. This saves a number of API calls and avoids triggering secondary rate limits.

[Git Database API]: https://docs.github.com/en/rest/git

Fixes SFDO-Tooling/Metecho#1303

### DIFF
--- a/cumulusci/tasks/github/util.py
+++ b/cumulusci/tasks/github/util.py
@@ -1,8 +1,11 @@
 import base64
 import hashlib
-import io
 import logging
 import os
+from pathlib import Path
+from typing import Optional, Tuple
+
+from github3.git import Commit, Tree
 
 from cumulusci.core.exceptions import GithubException
 
@@ -57,10 +60,11 @@ class CommitDir(object):
         self.new_tree_list = [item for item in self.new_tree_list if item]
         self._add_new_files_to_tree(self.new_tree_list)
 
-        if tree_unchanged := self._summarize_changes(self.new_tree_list):
+        if not self.new_tree_list:
             self.logger.warning("No changes found, aborting commit")
             return self.parent_commit
 
+        self._summarize_changes(self.tree, self.new_tree_list)
         new_tree = self._create_tree(self.new_tree_list)
         new_commit = self._create_commit(commit_message, new_tree)
         self._update_head(new_commit)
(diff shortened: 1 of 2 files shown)


davidjray (approved):
Pretty cool fix. Nice Work!

cumulusci/tasks/github/util.py
```diff
@@ -57,123 +60,128 @@ def __call__(
...
                             content = f.read()
-                        repo_path = (self.repo_dir + "/") if self.repo_dir else ""
+                        repo_path = f"{self.repo_dir}/" if self.repo_dir else ""
                         new_item = {
-                            "path": "{}{}".format(
-                                repo_path, local_file_subpath.replace(os.sep, "/")
-                            ),
-                            "mode": "100644",
+                            "path": f'{repo_path}{local_file_subpath.replace(os.sep, "/")}',
+                            "mode": "100644",  # FIXME: This is wrong
```
davidmreed, line 128: Why is this wrong? Should we add a WI to correct it?

jstvz: > Why is this wrong?

This assumes that the item (`blob`) is a file, but it could have the executable bit (`100755`) or be a symlink (`120000`), or even a submodule (`160000`). More [here](https://stackoverflow.com/a/69341082).

> Should we add a WI to correct it?

Probably not. This implementation needs these fixes (at least):

1.  The above file mode issue.
2.  It doesn't respect `.gitignore`. We'd need to implement this or add a [library](https://github.com/mherrmann/gitignore_parser/blob/master/gitignore_parser.py).
3.  It ignores any files or directories that start with a dot (`.forceignore`). Simple enough, but we'd need to add an option the consuming task. It'd need to continue ignoring them by default to avoid surprising users accidentally committing code.
4.  It doesn't handle submodules correctly.

None of these are showstoppers for the one task that uses the class, but are probably more impactful downstream in Metecho. Fixing all of these is doable, but I doubt the ROI is worth it, since we can just use Git (via subprocess or a library).

davidjray (approved):

cumulusci/tasks/github/util.py
davidmreed: Sounds like a design spike is in order for the broad-based fix - could you add that WI?

PR #3384, opened 2022-10-06
Auto-load dataset matching org shape name

bethbrains:
If `load_dataset` is called without any path options, it will automatically use a dataset that matches the org shape, if one exists. For example, a `dev` org will automatically use a dataset that exists at `datasets/dev/`. Within that folder, two files must exist, also matching the org shape name: `dev.mapping.yml` and `dev.dataset.sql`, in this example. If the directory or files do not exist and no paths options were specified, the task will look for `datasets/mapping.yml` and `datasets/dataset.sql` by default. When the `org_shape_match_only` option is `True`, this overrides any path options and default files and looks _only_ for a dataset directory that matches the org shape name. The `org_shape_match_only` option defaults to `False`.

In addition, `load_dataset` is included in `config_dev`, `config_qa`, and `config_managed`, so it is automatically called when running most org setup flows. In this context, it runs with `org_shape_match_only` set to `True`, to avoid double loading for backwards compatibility with customer flows that are already customized to call `load_dataset`.

### DIFF
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -134,7 +134,7 @@ class LoadData(SqlAlchemyMixin, BaseSalesforceApiTask):
         self.has_dataset = True
 
     def _find_matching_dataset(self) -> T.Optional[T.Tuple[str, str]]:
-        org_shape = self.org_config.lookup("config_name")
+        org_shape = self.org_config.lookup("config_name") if self.org_config else None
         if not org_shape:
             return None  # persistent org
         dataset_folder = Path("datasets", org_shape)
--- a/cumulusci/tasks/bulkdata/load.py
+++ b/cumulusci/tasks/bulkdata/load.py
@@ -117,10 +117,21 @@ class LoadData(SqlAlchemyMixin, BaseSalesforceApiTask):
         )
 
     def _init_dataset(self):
-        org_shape_match_only = process_bool_arg(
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/bulkdata/load.py
```diff
@@ -120,7 +116,70 @@ def _init_options(self, kwargs):
...
+        if self.options.get("database_url"):
+            # prefer database_url if it's set
+            self.options["sql_path"] = None
+        elif self.options.get("sql_path"):
+            self.options.setdefault("mapping", "datasets/mapping.yml")
+        elif self.options.get("mapping"):
+            self.options.setdefault("sql_path", "datasets/sample.sql")
+        elif found_dataset := (
+            self._find_matching_dataset()
+            or (self._find_default_dataset() if not org_shape_match_only else None)
```
davidmreed, line 138: How about a walrus _or_ nested boolean logic, but not both? This is hard to follow.

cumulusci/tasks/bulkdata/load.py
```diff
@@ -120,7 +116,70 @@ def _init_options(self, kwargs):
             self.options.get("set_recently_viewed", True)
         )
 
+    def _init_dataset(self):
```
davidmreed, line 119: I had to read the docs below to figure out what the intended control flow was in this method. A few comments explaining the order of fallback and how `org_shape_match_only` is intended to work would be helpful.

davidmreed (approved):

PR #3387, opened 2022-10-07
Add new keyword "Wait until salesforce is ready"

boakley:
Work item [W-11241613](https://gus.lightning.force.com/a07EE00000yQFvbYAG)

This adds a new playwright keyword `Wait until salesforce is ready`, and calls it from `Open Test Browser`

This uses `Wait until loading is complete` but also adds an additional check for landing on a classic page.

This also adds a new 'wait' parameter to the playwright version of `Open test browser`, for parity with the selenium version. Plus, it makes testing this particular feature easier.

The wait times are admittedly a bit arbitrary, but they seem to work. If I reduce them much, they will fail more often when Salesforce is running slower than usual.

I don't recall if I'm still supposed to check in the documentation or if we regenerate it when we regenerate the rest of the docs, so I'm including it here.

# Changes

* There is a new playright-based keyword `Wait until salesforce is ready`. This keyword is automatically called by the `Open Test Browser` keyword in the SalesforcePlaywright library. In addition to waiting for the page to be rendered, it will auto attempt to detect a classic page on initial render, and automatically switch to a lightning page if one is detected.

### DIFF
--- a/cumulusci/robotframework/SalesforcePlaywright.py
+++ b/cumulusci/robotframework/SalesforcePlaywright.py
@@ -136,15 +136,10 @@ class SalesforcePlaywright(FakerMixin, BaseLibrary):
         )
         self.browser.get_elements(locator)
         self.browser.execute_javascript(function=WAIT_FOR_AURA_SCRIPT)
-
-        # this seems to fail once in a while for an unknown reason, so we'll
-        # try twice. This seems to be more effective than calling the
-        # function once for a longer timeout
-        try:
-            self.browser.wait_until_network_is_idle(timeout)
-        except Exception as e:
-            self.builtin.log(f"caught error waiting for idle: {e}", "DEBUG")
-            self.browser.wait_until_network_is_idle(timeout)
+        # An old knowledge article recommends waiting a second. I don't
+        # like it, but it seems to help. We should do a wait instead,
+        # but I can't figure out what to wait on.
+        time.sleep(1)
 
     @capture_screenshot_on_error
     def wait_until_salesforce_is_ready(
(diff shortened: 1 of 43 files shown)


cumulusci/robotframework/SalesforcePlaywright.py
```diff
@@ -126,14 +133,83 @@ class 'slds-template__container', but a different locator can
...
+        keyword will actually wait for half of the timeout before
+        starting checks for edge cases.
+
+        """
+
+        timeout_seconds = timestr_to_secs(timeout)
+        start_time = time.time()
+
+        locator = locator if locator else "div.slds-template__container"
+        expected_url = rf"/{self.cumulusci.org.lightning_base_url}\/lightning\/.*/"
```
TheBitShepherd, line 168: ooo, the rare and ever elusive `raw format string` in the wild!

cumulusci/robotframework/SalesforcePlaywright.py
```diff
@@ -126,14 +133,83 @@ class 'slds-template__container', but a different locator can
...
+        The timeout parameter is taken as a rough guideline. This
+        keyword will actually wait for half of the timeout before
+        starting checks for edge cases.
+
+        """
+
+        timeout_seconds = timestr_to_secs(timeout)
+        start_time = time.time()
+
+        locator = locator if locator else "div.slds-template__container"
```
TheBitShepherd, line 167: Could also use the `or` here to keep things even more concise
```suggestion
        locator = locator or "div.slds-template__container"
```

TheBitShepherd (approved):
Nice job @boakley!

cumulusci/robotframework/SalesforcePlaywright.py
boakley: I think this is the first time I've ever used it.

PR #3394, opened 2022-10-13
Improve handling of unknown statuses in mc deploy task

TheBitShepherd:
## Issues Fixed

* Fixed an issue where unknown deploy statuses could cause polling to hang in the `deploy_marketing_cloud_package` task.


cumulusci/tasks/marketing_cloud/deploy.py
```diff
@@ -98,14 +101,29 @@ def _poll_action(self):
...
             f"{self.endpoint}/deployments/{self.job_id}", headers=self.headers
         )
-        result = safe_json_from_response(response)
-        self.logger.info(f"Waiting [{result['status']}]...")
-        if result["status"] == DEPLOY_FINISHED_STATUS:
+        response_data = safe_json_from_response(response)
+        deploy_status = response_data["status"]
+        self.logger.info(f"Deployment status is: {deploy_status}")
+
+        if deploy_status != IN_PROGRESS_STATUS:
```
TheBitShepherd, line 108: Instead of looking for completed or errored statuses to determine when a job is done, we now consider a deployment "finished" when the status changes from `IN_PROGRESS_STATUS`

cumulusci/tasks/marketing_cloud/deploy.py
```diff
@@ -171,10 +189,11 @@ def _validate_response(self, deploy_info: dict):
 
         self.logger.info("Deployment completed successfully.")
 
-    def _report_fatal_error(self, result: dict):
+    def _report_error(self, response_data: dict):
+        deploy_status = response_data["status"]
         self.logger.error(
-            f"> {DEPLOY_ERROR_STATUS} received. Dumping response from Marketing Cloud:\n{result}"
+            f"Received status of: {deploy_status}. Received the following data from Marketing Cloud:\n{response_data}^"
```
davidmreed, line 195: Just validating: there's no indication that this response could contain confidential data, correct?

davidmreed (approved):
Like the resiliency here!

cumulusci/tasks/marketing_cloud/deploy.py
TheBitShepherd: I believe I looked at that with Marcel last spring when we implemented this task, but I don't recall what the actual response body for an errored deployment looks like.

PR #3396, opened 2022-10-14
New release 3.67.1

boakley:
Added javascript files that were missing.


davidmreed (approved):

jstvz (approved):
@boakley Brew issue doesn't look like it requires a hotfix. Approving the patch as-is.

PR #3402, opened 2022-10-18
Add type annotations to core classes

davidmreed:
This PR adds type annotations for several core classes. There are a couple of small code changes to serve this objective, but there's intended to be no change in behavior.


cumulusci/core/flowrunner.py
```diff
@@ -140,13 +160,32 @@ def __repr__(self):
...
+        result: Any,
+        return_values: Any,
+        exception: Optional[Exception],
+    ):
+        self.step_num = step_num
+        self.task_name = task_name
+        self.path = path
+        self.result = result
+        self.return_values = return_values
+        self.exception = exception
```
jstvz, line 185: ↔️: [`typing.NamedTuple`](https://docs.python.org/3.8/library/typing.html#typing.NamedTuple)
 ```python
class StepResult(typing.NamedTuple):
    step_num: StepVersion
    task_name: str
    path: str
    result: Any
    return_values: Any
    exception: Optional[Exception]
```

File: cumulusci/core/flowrunner.py [hunk 2]
```diff
@@ -51,26 +51,34 @@
...
 from operator import attrgetter
+from typing import (
+    Any,
+    DefaultDict,
+    Dict,
+    Optional,
+    Tuple,
+    Type,
+    TYPE_CHECKING,
+    List,
```
jstvz, line 67: Suggestions:
- `pre-commit install --install-hooks`
- `git commit --no-verify -m 'wip'` if you get annoyed.
- Also: https://marketplace.visualstudio.com/items?itemName=ms-python.isort

jstvz (commented):
Nice.

cumulusci/core/flowrunner.py
davidmreed: TIL. That's clean. Will use that approach to avoid adding code.

cumulusci/core/runtime.py
```diff
@@ -38,30 +38,30 @@ def __init__(self, *args, load_keychain=True, **kwargs):
...
     @property
-    def universal_config_cls(self):
+    def universal_config_cls(self) -> Type:
         klass = self.get_universal_config_class()
         return klass or self.universal_config_class
 
     @abstractmethod
-    def get_universal_config_class(self):
+    def get_universal_config_class(self) -> Optional[Type]:
         return None
```
jstvz, line 47: I think `pass` or `...` works with `Type` here ([but not](https://github.com/python/mypy/blob/master/docs/source/class_basics.rst#abstract-base-classes-and-multiple-inheritance) with `super().get_universal...`) but no objection if we come back to it later

jstvz (approved):
🧗🏾

PR #3408, opened 2022-10-20
Playwright debugger

boakley:
This adds support for the "breakpoint" command when using playwright, which previously only worked with Selenium. 

To test this, make sure you run `cci robot install_playwright` to make sure you have all of the playwright code installed.

To test, create a test that looks like this:

```
*** Settings ***
Resource  cumulusci/robotframework/SalesforcePlaywright.robot

Suite Setup     Open test browser
Suite Teardown  Delete Records and Close Browser

*** Test Cases ***
Example
    ${foo}=  set variable  hello, world
    breakpoint
```

You can then run the test and enable the debugger from the command line like so:

```
$ cci task run robot --robot_debug=True  --suites path/to/the/test.robot
```

Once the browser opens, you should eventually get a prompt in the console. You can then type `locate_elements  xpath=//button` (two spaces after `locate_elements`) , and it should highlight with a yellow outline all of the buttons on the page:

```
:::
::: Welcome to rdb, the Robot Framework debugger
:::

Type help or ? to list commands.

> Debugger.Example
-> <Keyword: cumulusci.robotframework.SalesforcePlaywright.Breakpoint>
rdb> locate_elements  xpath=//button
51 elements found
rdb> 
```

To remove the highlight, issue the command `reset_elements`:

```
rdb> reset_elements
```


cumulusci/tasks/robotframework/tests/test_browser_proxies.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def test_highlight_injected_css(self):
+        """Verify we attempt to inject css into the DOM when highlighting an element"""
+
+        self.selenium.get_webelements = mock.Mock(return_value=[])
+        self.proxy.highlight_elements("button")
+
+        self.selenium.driver.execute_script.assert_any_call(initialize_highlight_js)
+
+    def test_highligth_element(self):
```
davidmreed, line 59:
```suggestion
    def test_highlight_element(self):
```

cumulusci/tasks/robotframework/tests/test_browser_proxies.py
```diff
@@ -0,0 +1,137 @@
...
+
+        ... but only Browser has been loaded and SeleniumLibrary has not
+        """
+        self.cli.builtin.get_library_instance = mock.Mock(
+            return_value={
+                "SeleniumLibrary": mock.Mock(),
+                "Browser": mock.Mock(),
+            }
+        )
+        assert isinstance(self.cli.webbrowser, SeleniumProxy)
```
davidmreed, line 49: I don't understand this assertion vis-a-vis the docstring. Is the docstring a copy/paste mistake?

davidmreed (approved):

davidmreed (dismissed):

cumulusci/tasks/robotframework/tests/test_browser_proxies.py
boakley: Ah, good catch! The docstring is incorrect. I'll fix it.

davidmreed (approved):

PR #3410, opened 2022-10-21
Bump robotframework dependency to 6

boakley:
This bumps the robotframework dependency to use the latest version, which at the time I write this is 6.0.

This required a small change to our libdoc task, but other than that everything passed with flying colors. Color me pleasantly surprised

# CRITICAL CHANGES
- robotframework has been upgraded from version 4.1.3 to version 6.0. As with most robotframework updates, there are a host of new features and a few deprecated features and backwards incompatibilities. For a complete list consult the [robotframework release notes](https://github.com/robotframework/robotframework/releases)


requirements/dev.txt
```diff
@@ -1,5 +1,5 @@
 #
-# This file is autogenerated by pip-compile with python 3.8
```
davidmreed, line 2: These dep changes don't look right. Can you regenerate on Python 3.8?

davidmreed (changes requested):

requirements/dev.txt
boakley: yeah, I can do that.

davidmreed:
There's something not right with the dependency updates on this branch, and I haven't figured out why yet. It looks like `pip-sync` has decided that `secretstorage` and `cryptography` are no longer transitive dependencies, along with all of their own deps. That is unlikely to be right!

boakley:
This PR has been superseded by [PR 3417](https://github.com/SFDO-Tooling/CumulusCI/pull/3417)

PR #3439, opened 2022-11-15
Use Twine check to check release during release_test

prescod:
One more check we can do to make sure that we are ready for
release.


TheBitShepherd (approved):

jstvz (commented):
This workflow is already a bit slow. What do you think of moving it to slow_integration_tests.yml as its own job?

PR #3442, opened 2022-11-21
Spike: Most tasks now infer their `api_version` from the org, if the cumulusci.yml does not specify it.

prescod:
Spike (weekend work)

PR is a mess...has some extra files etc.

Tasks now look for their `api_version` in the following places:

 1. In the `cumulusci.yml` project.package.api_version
 2. In the org itself (latest version supported by the org)

If the task has a legitimate need to get an API Version without
access to an org, it can look at a global called
DEFAULT_SF_API_VERSION

The two exceptions:

1: `metadeploy_publish` may need to supply a version to
     it's task being frozen.

2: `UpdatePackageXml` hard-codes a version into XML
   even when run outside of an org context. (arguably,
   it could be changed to require an org, and this is
   a 99.9% backwards-compatible change)

This change will probably have a very small performance
implication. org_config.latest_api_version is cached
so it shouldn't be too bad.


cumulusci/salesforce_api/tests/test_metadata.py
```diff
@@ -99,14 +99,20 @@ def _mock_call_mdapi(self, api, response, status_code=None):
         )
         return response
 
+    from pysnooper import snoop
```
jstvz, line 102: 🤨

cumulusci/salesforce_api/tests/test_metadata.py
```diff
@@ -128,6 +134,7 @@ def test_build_endpoint_url_mydomain(self):
         }
         task = self._create_task(org_config=org_config)
         api = self._create_instance(task)
+        print(api._build_endpoint_url())
```
jstvz, line 137: Extraneous `print`

cumulusci/tasks/salesforce/activate_flow.py
```diff
@@ -20,7 +20,8 @@ def _init_options(self, kwargs):
         self.options["developer_names"] = process_list_arg(
             self.options.get("developer_names")
         )
-        self.api_version = "43.0"
+        # TODO: IS THIS RIGHT????
+        self._api_version = "43.0"
```
jstvz, line 24: Looks like this was just the API version that was current when the task was written. Needs a smoke test.

jstvz (commented):
Made a first pass. I think this approach works in principle, but I'd like to discuss further.

PR #3451, opened 2022-11-29
Vendor LooseVersion

prescod:
For a year or more, the Python interpreter has warned us that they intend to remove the module distutils.version. We use the class LooseVersion from that module. This has two consequences:

1. At some point CCI will break because it depends on a module that will disappear (probably Python 3.12).
2. In the meantime, our Python 3.10 and 3.11 test output is [dominated by warnings](https://github.com/SFDO-Tooling/CumulusCI/actions/runs/3536227946/jobs/5994878911#step:5:257). This hides other warnings which might be more urgent.

The class we need has no dependencies and is licensed with one of the most [liberal licences](https://en.wikipedia.org/wiki/Python_Software_Foundation_License) available. We already depend on this license to be able to deploy many PSF-licensed modules with CCI.

W-12132460


jstvz:
I noticed we have an [approved 3PP request](https://gus.lightning.force.com/lightning/r/ADM_Third_Party_Software__c/a0qEE0000000p49YAA/view) from your installer work last year. By my reading of the Open Source Policy we're in compliance.

TheBitShepherd (approved):

PR #3455, opened 2022-11-30
Add two tasks called `load_sample_data` and `capture_sample_data`.

prescod:
W-12082779

Adds two tasks called `load_sample_data` and `capture_sample_data`.

Critical Change
===============

A few weeks ago we soft-launched a `load_data` feature that matched the dataset loaded to the org_shape.

If you had discovered and used the `load_data` option called `org_shape_match_only`, you should switch to the new task because eventually we will remove the option and only use the new task.


cumulusci/tasks/bulkdata/load.py
```diff
@@ -135,6 +135,10 @@ def _init_dataset(self):
             self.options["mapping"] = None
             self.options["sql_path"] = None
             self.options["database_url"] = None
+            self.logger.warning(
+                "The `default_dataset_only` option has been deprecated. "
+                "Please switch to the `load_sample_data` task."
```
TheBitShepherd, line 140: 👍

TheBitShepherd (approved):
One correction I caught in the docs, otherwise this looks good!

TheBitShepherd (approved):

PR #3473, opened 2022-12-17
Refresh all objects in org schema cache

prescod:
Previously the org schema was versioned as a unified whole. Not one sobject at a time. This worked until I added the feature of allowing one to download a subset of the schema. Now some objects might be refreshed on Monday, some on Tuesday, a different set on Wednesday etc. To allow this, all objects are independently versioned (with a last_modified_date) now. If it is needed by the application it is refreshed and if it isn't, it isn't.

This bug was hidden because of the ways we were using `get_org_schema` before we added `capture_sample_data`. Now we really need the schema to be complete and up-to-date.

Tasks that use `get_org_schema` include `load_data` (for the Recently Used checking) and `generate_mapping`.

### DIFF
--- a/cumulusci/conftest.py
+++ b/cumulusci/conftest.py
@@ -14,9 +14,16 @@ from sqlalchemy.orm import sessionmaker
 
 from cumulusci.core.github import get_github_api
 from cumulusci.salesforce_api.org_schema_models import Base
+from cumulusci.tasks.bulkdata.tests.integration_test_utils import (
+    ensure_accounts,
+    ensure_records,
+)
 from cumulusci.tasks.salesforce.tests.util import create_task_fixture
 from cumulusci.tests.pytest_plugins.pytest_sf_vcr import salesforce_vcr, vcr_config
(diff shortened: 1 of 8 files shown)


cumulusci/conftest.py
```diff
@@ -14,10 +14,17 @@
...
+from cumulusci.tasks.bulkdata.tests.integration_test_utils import (
+    ensure_accounts,
+    ensure_records,
+)
 from cumulusci.tasks.salesforce.tests.util import create_task_fixture
 from cumulusci.tests.pytest_plugins.pytest_sf_vcr import salesforce_vcr, vcr_config
 from cumulusci.tests.util import DummyKeychain, DummyOrgConfig, mock_env
 
+ensure_accounts = ensure_accounts
+ensure_records = ensure_records
```
TheBitShepherd, line 26: What is this doing?

TheBitShepherd (approved):

cumulusci/conftest.py
prescod: Shuts up flake8

PR #3476, opened 2022-12-21
End-to-end release automation

jstvz:
## Un-pin our dependencies

We've reached consensus that using pip-tools to pin our direct and transitive
dependencies has been problematic. We often encounter dependency conflicts
between our first-party (`snowfakery`), third-party, and downstream
dependencies. Moreover, it also results in additional churn from Dependabot and
delays in mitigating vulnerabilities.

## Use Hatch as build system

Highlights:

- Replaces setuptools and twine.
- Removes `MANIFEST.in`, `setup.{cfg,py}`
- Replaces `version.txt` with `__about__.py` for declarative version bumping
- Preserves pip-tools for dependency pinning to preserve users' existing workflows.
- Explicitly adds the `--resolver=backtracking` flag to pip-tools
- Adds `hatch-fancy-pypi-readme` to handle PyPI `long_description`

We adopted `pyproject.toml` for tool configuration a while back, but we never
adopted it for package metadata, build system, or dependency specifications (see
[PEP 517], [PEP 518], and [PEP 621]).

Hatch is a "modern, extensible Python project manager." Unlike Poetry,
it uses standard PEP 621 dependency specifications, and strict adherence
to the packaging PEPs is a stated goal of the project.[^1]

Because the PyPA rejected the reproducible lockfile PEP, Hatch has not yet
implemented dependency resolution and locking. We've left pip-tools in place to
cover this gap.

## Release automation workflows

Adds the pre-release.yaml workflow that, when passed a version bump:

1. Bumps the project version using `hatch version`
2. Automatically generate release notes, using GitHub's REST API (`releases/generate-notes`)
3. Inserts the release notes into docs/history.md
4. Creates a branch and opens a pull request against `main`

Re-writes the release.yml workflow to:

1. Use `hatch` to build the package 
2. Publishes the package to PyPI, concatenating the latest release notes and the
   README in the project description
3. Creates a GitHub release (and corresponding tag), using the latest release
   notes as the release body.

[^1]: The full rationale for picking Hatch will be detailed in a forthcoming ADR. 

[PEP 517]: https://peps.python.org/pep-0517/
[PEP 518]: https://peps.python.org/pep-0518/
[PEP 621]: https://peps.python.org/pep-0621/


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,69 @@
...
+              env:
+                  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              run: |
+                  PREVIOUS_VERSION=$(gh release view --json tagName --jq .tagName)
+                  NEXT_VERSION="v$(hatch version)"
+                  echo "## $NEXT_VERSION ($(date -I))" > changelog.md
+                  gh api \
+                    --method POST \
+                    -H "Accept: application/vnd.github.v3+json" \
+                    /repos/jstvz/CumulusCI/releases/generate-notes \
```
prescod, line 49: @jstvz Filesystem reference?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,69 @@
...
+                    -f previous_tag_name=$PREVIOUS_VERSION \
+                    -f target_commitish='main' \
+                    -f tag_name=$NEXT_VERSION \
+                    --jq '.body' |
+                    sed -e 's_\(https.*\/\)\([0-9]*\)$_[#\2](\1\2)_' \
+                        -e 's_by @\(.*\) in_by [@\1](https://github.com/\1)_' >> changelog.md
+                  python utility/update-history.py
```
prescod, line 56: This should probably happen in release.yml after the changelog.md has been edited.

prescod:
Looks like a major step forward. I suggested some changes above and I per our meeting, I think that we should publish two packages, one with pins and one without. Perhaps "cumulusci" and "cumulusci-lib".

TheBitShepherd (approved):

.github/workflows/pre-release.yml
jstvz: Thanks, fixed.

(Left a reference to my testing fork.)

.github/workflows/pre-release.yml
jstvz: Could be clearer, but `changelog.md` is a temporary file that isn't committed. Only `docs/history.md` (alongside `cumulusci/__about__.py`) will be committed; the history file serves as the source of truth for the PyPI `long_description` and GitHub release notes body.

jstvz:
@TheBitShepherd Would you take a second pass at this? I fixed the REST resource Paul spotted.

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,67 @@
...
+                  gh api \
+                    --method POST \
+                    -H "Accept: application/vnd.github.v3+json" \
+                    /repos/SFDO-Tooling/CumulusCI/releases/generate-notes \
+                    -f previous_tag_name=$PREVIOUS_VERSION \
+                    -f target_commitish='main' \
+                    -f tag_name=$NEXT_VERSION \
+                    --jq '.body' |
+                    sed -e 's_\(https.*\/\)\([0-9]*\)$_[#\2](\1\2)_' \
+                        -e 's_by @\(.*\) in_by [@\1](https://github.com/\1) in_' >> changelog.md
```
TheBitShepherd, line 53: 😎

PR #3481, opened 2023-01-06
`cci service update`

TheBitShepherd:
# Changes

* `cci` has a new command: `cci service update`. Users can now update individual service attributes with `cci service update` instead of having to completely remove and re-add the service. You can utilize the `--attribute` option to set specific service attributes in a headless environment (i.e. `--attribute name value`).


cumulusci/cli/service.py
```diff
@@ -321,18 +351,30 @@ def get_service_data(service_config, sensitive_attributes) -> list:
     return service_data
 
 
-def get_sensitive_service_attributes(runtime, service_type) -> list:
+def get_service_attributes(
+    runtime: CliRuntime, service_type: str, sensitive: bool = False
```
prescod, line 355: Maybe the parameter should be called only_sensitive_attributes to differentiate from the equally plausible include_sensitive_attributes?

prescod (dismissed):

PR #3486, opened 2023-01-10
Add type assurances to metadeploy.py

prescod:
These type assertions and other code help the type checker to validate this file.

Also changed it to test with dry_run on and also off, rather than only hard-coded.


cumulusci/tasks/metadeploy.py
```diff
@@ -24,6 +24,7 @@ class BaseMetaDeployTask(BaseTask):
     """Base class for tasks that talk to MetaDeploy's API."""
 
     def _init_task(self):
+        assert self.project_config.keychain
```
prescod, line 27: This line just proves that the next line won't crash, so it should be backwards compatible.

cumulusci/tasks/metadeploy.py
```diff
@@ -112,10 +113,13 @@ def _run_task(self):
             version = self._find_or_create_version(product)
             if self.labels_path and "slug" in product:
                 self._publish_labels(product["slug"])
+        else:
+            version = {"id": "fake_for_dryrun", "url": "fake_for_dryrun"}
```
prescod, line 117: This sets up a variable that will never be read because otherwise it would have caused a crash in the old code. So it should be backwards compatible.

cumulusci/tasks/metadeploy.py
```diff
@@ -112,10 +113,13 @@ def _run_task(self):
...
 
         # Check out the specified tag
         gh = self.project_config.get_github_api()
         repo = gh.repository(repo_owner, repo_name)
+        assert repo
```
prescod, line 122: If repo is None, later code would crash, so this should be backwards compatible.

cumulusci/tasks/metadeploy.py
```diff
@@ -136,6 +140,7 @@ def _run_task(self):
                     "commit": self.commit,
                 },
             )
+            assert self.project_config.keychain
```
prescod, line 143: As above.

cumulusci/tasks/metadeploy.py
```diff
@@ -268,6 +275,8 @@ def _convert_org_providers_to_plan_equivalent(
             org_providers = "Both"
         elif providers == ["devhub"]:
             org_providers = "Scratch"
+        else:  # pragma: no cover
```
prescod, line 278: This ensures that a None can never escape this function. According to the type signature, it never should have.

But @TheBitShepherd should verify that all cases were supposed to be covered here.

PR #3487, opened 2023-01-10
Run robot UI tests as integration tests.

prescod:
Robut UI tests are integration tests, so we can run them when we run other integration tests,
which is less often than unit tests.


jstvz (changes requested):
```
The workflow is not valid. .github/workflows/feature_test.yml (Line: 116, Col: 40): Job 'coveralls_done' depends on unknown job 'robot_ui'.
```

Needs an update on line 116 to remove a reference to `robot_ui`.

jstvz (approved):

PR #3493, opened 2023-01-17
Run flow with option for project custom attributes

RupertBarrow:
This adds the "-op" option to the "cci flow run" command, to pass a project__custom__attribute value.
For example : 
```
cci flow run myflow -o project__custom__destination US
```

It may be preferable to use a different option than "-op", but I thought that "-p" would be confusing because often used to pass a path.


RupertBarrow:
Sorry @prescod, I need to rework this.
I will close this PR and open another later.

PR #3494, opened 2023-01-17
Close all test zip files

prescod:


jstvz (approved):
@prescod I'm curious, what prompted this? I thought `ZipFile.close()` only mattered if you're writing an archive.

prescod:
@jstvz : When you turn on ultra-strict warnings-as-errors, unclosed resources become an issue. I would like us to have the option to run warnings-as-errors (either by default, or as an opt-in), so I cleaned this up.

PR #3501, opened 2023-01-19
Test that release can be installed

prescod:
Per the title, verify that any built artifacts can be installed.

When I tried to install the wheel from main, they would not install due to a typo which I separately reported. "CumulusCI-Test"

This reminded me that because I have run into similar problems in Snowfakery, I have a build there that tests that it can be installed.

Unless I am mistaken, the line that is failing in this test is

```
                  pip show cumulusci
```

In other word, it i proving that the main branch does not build a wheel that can be installed and inspected under the name `cumulusci`

If my hypothesis is correct, fixing that bug on main will also fix this PR and then it can be merged.

Or...it may reveal other problems. I think there may be a problem with the sdist.


prescod:
Sorry this should have been a draft.

jstvz (dismissed):
The failure in release flows indicates that the workflow is now working as designed. Thanks Paul!

jstvz (approved):

PR #3502, opened 2023-01-19
Stricter check on what counts as a dataset

prescod:
Fix an error where a project had a /datasets/qa/ subdirectory which was not intended to be a real dataset. It was always the intent that such directories would be ignored but that code was not ported from `load.py` to `load_sample_data.py`.

### DIFF
--- a/cumulusci/tasks/sample_data/load_sample_data.py
+++ b/cumulusci/tasks/sample_data/load_sample_data.py
@@ -42,7 +42,7 @@ class LoadSampleData(BaseSalesforceApiTask):
             self.org_config,
             schema,
         ) as dataset:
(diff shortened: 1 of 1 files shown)


cumulusci/tasks/sample_data/load_sample_data.py
```diff
@@ -42,7 +42,7 @@ def _run_task(self):
             self.org_config,
             schema,
         ) as dataset:
-            self.return_values = dataset.load(self.options, self.logger)
+            self.return_values = dataset.load(self.options, self.logger) or {}
```
prescod, line 45: Fix a typing warning.

TheBitShepherd (approved):

allisonletts:
Fixes #3505

PR #3503, opened 2023-01-20
Remove deprecated Playwright keywords

jstvz:
We missed a major release of `robotframework-browser` ([v15]), which removed the deprecated `execute_javascript` keyword. This resulted in test failures that we initially believed were timeout related because the `AttributeError` was being caught by the bare `except` and re-raised with a misleading error message.

This commit resolves the failures by:

1. Switching to the `evaluate_javascript` keyword
2. Raising the timeout exception using the `raise ... from exc`

[v15]: https://github.com/MarketSquare/robotframework-browser/releases/tag/v15.0.0

### DIFF
--- a/cumulusci/robotframework/SalesforcePlaywright.py
+++ b/cumulusci/robotframework/SalesforcePlaywright.py
@@ -31,12 +31,12 @@ class SalesforcePlaywright(FakerMixin, BaseLibrary):
         This expects the url to contain an id that matches [a-zA-Z0-9]{15,18}
         """
         OID_REGEX = r"^(%2F)?([a-zA-Z0-9]{15,18})$"
-        url = self.browser.execute_javascript("window.location.href")
+        url = self.browser.evaluate_javascript(None, "window.location.href")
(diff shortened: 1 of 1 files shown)


cumulusci/robotframework/SalesforcePlaywright.py
```diff
@@ -173,14 +174,14 @@ def wait_until_salesforce_is_ready(
...
-            except Exception:
+            except Exception as exc:
                 # dang. Maybe we landed somewhere unexpected?
                 if self._check_for_classic():
                     continue
 
                 if time.time() - start_time > timeout_seconds:
                     self.browser.take_screenshot()
-                    raise Exception("Timed out waiting for a lightning page")
+                    raise Exception("Timed out waiting for a lightning page") from exc
```
prescod, line 184: How does it get into this block if there is no timeout?

Is the underlying exception CAUSING a timeout?

Or is it failing instantly and yet ending up in this block for some reason.

prescod (approved):

PR #3517, opened 2023-02-03
Exclude label from release notes generator

jstvz:
I'm pretty sure this changed silently, but this is the [documented] way to
do it.

[documented]: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#example-configurations


jstvz:
@davidjray Better fix for the issue we saw yesterday. Would you please review?

prescod (approved):

PR #3521, opened 2023-02-05
📝 add a link to Connect Persistent Orgs video tutorial

Julian88Tex:
📝 add a link to Connect Persistent Orgs video tutorial

(hope to make some more of these 😄 )


prescod:
Great idea @Julian88Tex !

Do you know the source of the exceptions that show up? Is that a bug we should look at? Assuming we can talk management into allowing us to link to videos, we probably would want to show it without the exceptions.

Julian88Tex:
@prescod you saw! I tried to click so fast that you wouldn't notice. I'm not positive but I suspect it had something to do with running this in a GitHub Codespace. I was hoping that this would give me a "cleaner" environment, but something wasn't fully set up. I'll give it another go and add another video without the exception.

jstvz:

If the video link is still valid and useful, a fresh patch against `main` would be welcome.

PR #3528, opened 2023-02-08
Force inclusion of pytest plugins in wheels

jstvz:
Snowfakery depends on some of CumulusCI's pytest plugins for integration
tests. This adds about 175 kB to the size of the wheel.


jstvz:
Downstream tests against this branch pass: https://github.com/SFDO-Tooling/Snowfakery/actions/runs/4137190741/jobs/7152227573

prescod (approved):

PR #3529, opened 2023-02-09
Mask invalid oauth token

jstvz:
Apparently this triggered a security scan. Replaces the string with 'x'
and removes a character to fool the regex.


prescod:
Per our discussion, this is not a real, live token, but deleting it will reduce confusion.

prescod (approved):

PR #3559, opened 2023-03-09
Feature/metadata api support

yippie:
simple-salesforce dependency in cci is very old (from 2019) Since then they have added a very handy and extensive mdapi feature.

## Changes
- bump simple-salesforce version from 1.11.4 to 1.12.3 (latest)
- added new self.mdapi to BaseSalesforceApiTask which has the new simple-salesforce mdapi connection

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -36,3 +36,4 @@ For example:
 
 * Ed Rivas (jerivas)
 * Gustavo Tandeciarz (dcinzona)
+* Kai Amundsen (yippie)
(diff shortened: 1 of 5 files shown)


jstvz (commented):
@Yippie Thanks for the PR!

I agree that it's overdue to update this dependency, and your PR is almost ready to merge. However, since it's been a while, I'll need to do some extra testing and review what's changed upstream. I also need to look at simple-salesforce/simple-salesforce#751

In the meantime, I’ve rebased your changes onto our [feature/metadata-api-support](https://github.com/SFDO-Tooling/CumulusCI/tree/feature/metadata-api-support) branch. Could you please review those changes and update your branch to match?

yippie:
This change got released in 4.8

PR #3562, opened 2023-03-15
Add documentation for Plans and UI options

yippie:
Most of this is not my work but I really could use the documentation so bumping this.

Added the ui_options list with options that I could find while reviewing source code.

### DIFF
--- a/cumulusci/cli/help/plan.md
+++ b/cumulusci/cli/help/plan.md
@@ -1,0 +1,192 @@
+# Plan Configuration
+
+An installer is CumulusCI automation that is published to an instance of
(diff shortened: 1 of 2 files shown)


davidmreed:
@jstvz I don't know all the context around the code changes here, which I think came from your work on the `plan` commands. Should we take this PR forward or carve off just the doc changes? We have a WI on the backlog to get the source doc that became `plan.md` published.

jstvz:
Thank you for your work on this PR. You point out an important documentation gap we need to close.

I'm closing this PR. These changes you've proposed come from a terminal documentation proof of concept that isn't on our roadmap. We acknowledge the need for improved documentation and plan to address these gaps in upcoming releases.

PR #3565, opened 2023-03-17
support prior return values

yippie:
Using prior values in options is wonderful. The "when" clause in flows, however, is not so much as it is VERY limited what you can base it off of.

To make things much more powerful with minimal work, I have extended the prior value syntax support into when clause of the flow runner.


davidmreed:
@yippie Thanks for the contribution! I was honestly surprised we didn't have this already; it's a good addition. Could you add a new test to validate the change? Let me know if I can provide guidance on our testing strategies.

jstvz:

Prior return values are already accessible in `when` clauses via the existing options-substitution path. If explicit `when`-clause support is still wanted, a fresh PR with tests against `main` (for a v4.x bug fix) or `dev` (for v5-only work) would be welcome.

PR #3566, opened 2023-03-17
Fix retrieve unpackaged so it is usable in metadeploy

yippie:
retrieve unpackaged messes with the package_xml in init_options which breaks its use in Metadeploy (because _init_options ends up being called twice)

Re-write options so it is not overwriting the valid value with an invalid one.


davidmreed:
Thanks for fixing that astoundingly weird code path, @yippie. Since per your report freezing this task was already broken, I don't think there are any backwards-compatibility issues to worry about here. Will approve once tests pass.

jstvz (approved):
Oof. Thanks for the fix.

PR #3567, opened 2023-03-27
Include installation key in SubscriberPackageVersion queries when provided.

zenibako:
If you attempt to use an installation key with the `create_package_version` and `promote_package_version` tasks (particularly with the `build_unlocked_test_package` flow), you will be blocked by a Salesforce SOQL error that demands you include the Installation key in the where clause of the query.

This PR updates these tasks and enables the `build_unlocked_test_package` flow to work when an installation key is configured for its tasks. It also adds `install_key` to `promote_package_version` as an option.


cumulusci/tasks/salesforce/promote_package_version.py
```diff
@@ -302,10 +306,24 @@ def _query_SubscriberPackageVersion(self, spv_id: str) -> Optional[Dict]:
...
 
+    def _get_spv_where_clause(self, spv_id: str) -> str:
+        """Get the where clause for a SubscriberPackageVersion query
+
+        Does not include the WHERE.
+        Includes the installation key if provided.
+        """
+        where_clause = f"Id='{spv_id}'"
+
+        if "install_key" in self.options:
```
davidmreed, line 321:
```suggestion
        if self.options.get("install_key"):
```

@zenibako I think there is a tricky edge case here. `"install_key"` may be present in `self.options`, but be `None`. This would happen in the circumstance that a flow containing this task were frozen to be run in MetaDeploy or similar. That'd cause the query to return no results.

cumulusci/tasks/create_package_version.py
```diff
@@ -266,6 +267,20 @@ def _run_task(self):
         self.logger.info(f"  Version Number: {self.return_values['version_number']}")
         self.logger.info(f"  Dependencies: {self.return_values['dependencies']}")
 
+    def _get_spv_where_clause(self, spv_id: str) -> str:
```
davidmreed, line 270: It's a bummer to have this code repeated in two different tasks. What do you think about breaking it out into a utility method, `format_subscriber_package_version_where_clause(spv_id: str, install_key: Optional[str]) -> str`?

davidmreed (changes requested):
@zenibako This looks great. I apologize for the lengthy delay in reviewing your PR. I have a couple of edge-casey comments. I would also love it if you could add a test for these new code paths (uploading & promoting with an install key).

cumulusci/utils/salesforce/tests/test_soql.py
```diff
@@ -0,0 +1,27 @@
...
+
+from cumulusci.utils.salesforce.soql import format_subscriber_package_version_where_clause
+
+
+class TestSoql:
+    spv_id = "04t000000000000"
+
+    @pytest.mark.vcr()
+    def test_format_subscriber_package_version_where_clause_simple(self):
+        where_clause = format_subscriber_package_version_where_clause(self.spv_id)
```
davidmreed, line 12: This would work in JavaScript, but the Python type-checker wants either an explicit positional argument or a default value in the function definition. (See suggestion above)

zenibako: Thanks! I fixed this locally after figuring out how to launch `pytest`, but hadn't pushed it yet. I'm definitely a little fresh to Python so I appreciate the guidance!

zenibako:
@davidmreed do I need to add a test for `create_package_version` with an `install_key` path? I see that the default task options include `install_key`, but I haven't figured out how to see code coverage with Pytest, so not sure if it's covered or not.

davidmreed:
@zenibako I've opened an internal PR as #3654 to get this merged and included in the next release.

Thank you again for your contribution. I apologize for the unwarrantedly long time to review this PR. We look forward to more collaboration in the future!

PR #3569, opened 2023-04-04
The `capture_sample_data` task now looks for a `datasetname.load.yml` file to influence how it generates its .mapping file

prescod:
CumulusCI has a feature called a ".[load.yml](https://cumulusci.readthedocs.io/en/stable/data.html#controlling-the-loading-process)" file which is used when generating mapping files in the context of loading Snowfakery recipes. This allows you to, for example, control the API used, change the load order, etc. Basically it allows you to override default loading behaviours.

Our new `capture_sample_data` task also generates mapping files, and therefore could benefit from the same kinds of overrides. It was always anticipated that the `load.yml` would also be relevant in this context, but it was not implemented until now.

This was a problem when an internal team needed to change the load order of two objects. Now users can do that.

### DIFF
--- a/cumulusci/tasks/sample_data/capture_sample_data.py
+++ b/cumulusci/tasks/sample_data/capture_sample_data.py
@@ -3,6 +3,7 @@ from pathlib import Path
 from cumulusci.core.config.org_config import OrgConfig
 from cumulusci.core.datasets import Dataset
 from cumulusci.core.exceptions import TaskOptionsError
+from cumulusci.salesforce_api.filterable_objects import OPT_IN_ONLY
 from cumulusci.salesforce_api.org_schema import Filters, get_org_schema
 from cumulusci.tasks.salesforce.BaseSalesforceApiTask import BaseSalesforceApiTask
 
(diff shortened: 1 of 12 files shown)


cumulusci/tasks/sample_data/capture_sample_data.py
```diff
@@ -19,7 +20,15 @@ class CaptureSampleData(BaseSalesforceApiTask):
...
         "extraction_definition": {
-            "description": "A file describing what to be extracted."
+            "description": "A file describing what to be extracted. "
+            "Defaults to `datasets/{datasetname}/{datasetname}.extract.yml` if it exists."
+        },
+        "loading_rules": {
+            "description": (
+                "Path to .load.yml file containing rules to use when loading the mapping. "
+                "Defaults to`datasets/{datasetname}/{datasetname}.load.yml` if it exists. "
+                "Multiple files can be comma separated."
```
davidmreed, line 30: Is this line accurate? I don't see any indication we load multiple loading rule files.

docs/data/loading_rules.md
```diff
@@ -0,0 +1,50 @@
...
+  load_after: Contact
+
+- sf_object: Opportunity
+  load_after: Contact
+```
+
+If you wish to share a loading file between multiple recipes or
+datasets, you can refer to it with the `--loading_rules` option.
+That will override the default filename (`<recipename>.load.yml`).
+If you want both, or any combination of multiple files, you can do
```
davidmreed, line 49: Accurate?

davidmreed (approved):

PR #3572, opened 2023-04-05
Correct indention of example github_automerge_feature

Julian88Tex:


Julian88Tex:
I would have saved two hours today if I was just looking at this closely. I'm surprised I didn't get a syntax error every time I tried to run a CCI command or at least do the info command for `github_automerge_feature`.

prescod:
Julian, I do see a warning:

image

image

Julian88Tex:
I actually just started seeing this today so please disregard the part about no warning 🤷🏽. Maybe I was using an older CCI version on a different computer.

PR #3574, opened 2023-04-07
Typing for `ScratchOrgConfig`

TheBitShepherd:
##Dev Notes
* Include more typing information on `ScratchOrgConfig` class.
* Some small refactors, and adjustments to docstrings.


cumulusci/core/config/scratch_org_config.py
```diff
@@ -22,7 +22,7 @@ class ScratchOrgConfig(SfdxOrgConfig):
     # default = None  # what is this?
     instance: str
     password_failed: bool
-    devhub: str
+    devhub_username: str
```
davidmreed, line 25: I am not certain this change is safe. Wouldn't this parameter be persisted in keychains?

I can't off the top of my head assess whether it'll impact how Metecho and MetaDeploy use scratch orgs. We probably should do end-to-end testing.

davidmreed (changes requested):

cumulusci/core/config/scratch_org_config.py
TheBitShepherd: I didn't see any explicit references to `.devhub` in either metadeploy or metecho, but I can switch this back. E2E testing is a little more work than I intended this change to be.

davidmreed (approved):

PR #3590, opened 2023-04-28
Correctly parse `--org-wide-defaults` option when passed via `cci`

TheBitShepherd:
# Bug Fixes
- Fixed a bug where passing the `--org-wide-defaults` option on the `set-organization-wide-defaults` task wasn't being properly parsed when passed from the CLI.


TheBitShepherd:
This is actually WAD. We are [updating the task documentation](https://github.com/SFDO-Tooling/CumulusCI/pull/3591) to more clearly call out how we expect this option to be configured.

PR #3606, opened 2023-06-01
Update fields option description and api_names options description

Julian88Tex:
provides more clarity that the fields option is meant to be populated directly in the yml and has a particular format.


cumulusci/tasks/metadata_etl/help_text.py
```diff
@@ -37,7 +50,7 @@ def _init_options(self, kwargs):
             )
         if not all(["api_name" in entry for entry in self.options["fields"]]):
             raise TaskOptionsError(
-                "The 'api_name' key is required on all entry values."
+                "This option is auto-populated based on the 'fields' option. It represents a list of API names of entities to affect. This option is not intended to be directly provided by the user."
```
davidmreed, line 53: @Julian88Tex I think this change is misplaced. I see what you're aiming to do but throwing an error if `api_names` is populated would need to be before line 37. This error gets thrown if the dicts are not structured correctly.

(And we should really revise how this whole thing works, but that's a topic for another day).

davidmreed (changes requested):

cumulusci/tasks/metadata_etl/help_text.py
Julian88Tex: @davidmreed oops yeah no idea what I was thinking there. Looks like the description is actually in the base.py file: https://github.com/SFDO-Tooling/CumulusCI/blob/e5e9d8b2341b644d0b1e1284513b613c8445d3ab/cumulusci/tasks/metadata_etl/base.py#L250

So yeah doesn't make sense to change it there but probably remove showing this option altogether, but yeah probably ab bigger change and out-of-scope for the intention of this change.

Removed it and updated :)

Julian88Tex:
@davidmreed since builds are failing on forked repo, I'd be happy to recreate this change in CCI's repo if you can/want to make me a contributor :-P

jstvz:
Thanks for the contribution and apologies for the long delay. This PR is stale enough that we are closing it to reduce queue noise. If the docs change is still useful, a fresh PR against `main` would be welcome.

PR #3623, opened 2023-07-17
Querying PermissionSetGroups to get list of assigned permission sets

jkasturi-sf:


cumulusci/tasks/preflight/permsets.py
```diff
@@ -3,10 +3,18 @@
...
-        permsets = "\n".join(self.return_values)
-        self.logger.info(f"Found permission sets assigned:\n{permsets}")
+        query = f"SELECT PermissionSet.Name,PermissionSetGroupId FROM PermissionSetAssignment WHERE AssigneeId = '{self.org_config.user_id}'"
+
+        self.permsets = []
+        for result in self.sf.query_all(query)["records"]:
+            if result["PermissionSet"]["Name"] not in self.permsets:
+                self.permsets.append(result["PermissionSet"]["Name"])
+
+            if result["PermissionSetGroupId"] is not None:
```
davidmreed, line 13: I don't think this will catch everything, @jkasturi-sf. 

What if the user is assigned a PSG, but they are not also individually assigned one of the component PS? This query will never run in that case.

I believe we need to separately query PSG assignments and combine all of the PS elements.

davidmreed: Comment withdrawn. I totally misunderstood what you were doing here and how this schema works. Thank you for clarifying!

davidmreed (approved):

davidmreed (approved):

PR #3625, opened 2023-07-18
Update transforms.py

GeekStewie:
- Updated FindReplaceIdSpec to get the first column of the select statement instead of ID


cumulusci/core/source_transforms/transforms.py
```diff
@@ -2,6 +2,7 @@
 import functools
```
davidmreed, line 2: Can you also update the tests?

davidmreed (changes requested):

davidmreed:
@GeekStewie My apologies, I thought that I had submitted a review for you and it was sitting in Pending state.

jstvz:

If the underlying issue still affects you, we would welcome a fresh PR against the appropriate branch (`main` for a v4.x bug fix, `dev` for v5-only work).

PR #3630, opened 2023-07-28
ADR: Migrate from SFDX to SF CLI

jstvz:
Salesforce has announced the deprecation of SFDX (v7) and is now
focusing on the newer SF CLI (v2). They've said that SF CLI (v2) is
smart enough to understand all SFDX commands, as well as new SF
commands. However, our web applications and other integrations depend on
SFDX. SFDX is end-of-life, so we need to cut over and make some
improvements by using the new, better features. This ADR proposes that
we migrate to the SF CLI.


docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
```diff
@@ -0,0 +1,75 @@
...
+    - Bad: Delays the inevitable. If they drop deprecated commands without us noticing, it would break functionality for our users.
+    - Bad: Could delay us learning and using new features of the SF CLI.
+
+### Decision Outcome
+
+Here's how we plan to make this change:
+
+-   **Maintain Backwards Compatibility:** Our priority is to ensure that all existing processes continue to function as expected. We will not modify the existing SFDX commands in our project until SF CLI announces the version in which these commands will be deprecated.
+
+-   **Delegate maintaining backwards compatibility to the SF CLI:** We will continue using existing commands as written and rely on SF CLI to interpret these commands. We think this will make the change easier and less disruptive.
```
davidmreed, line 45: I want to make sure I understand these two points together. Are you saying that we will _invoke_ the tool as `sf`, but we will pass SFDX-style commands (such as `force:org:display`) until they are explicitly deprecated?

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
```diff
@@ -0,0 +1,75 @@
...
+
+-   **Warn users that are on SFDX (v7) that they need to upgrade:** We'll check the installed version of the CLI and warn users that future versions of CumulusCI will drop support.
```
davidmreed, line 47: How does this align with the answer to the question above - are you proposing here that we will still invoke `sfdx` as such? If we have already migrated some commands to `sf do some org thing`, but not others, does that open the possibility of very different user outcomes based on what flow they ran?

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
```diff
@@ -0,0 +1,75 @@
...
+
+-   **Update our integration tests to include `stable-rc`:** We'll revisit PR #3558 to make sure all tasks that use the CLI are tested in our workflow, and that we're testing the weekly release candidate channel (`stable-rc`).
+
+-   **Update the Falcon web applications to depend on SF as soon as practical:** We know we need to switch to SF CLI, and we want our Falcon web apps to benefit from its latest updates and features as soon as possible. We'll also install all necessary JIT plugins in the apps, so we don't have to install them over and over during builds.
+
+-   **Update the SFDX Heroku buildpack to depend on SF as soon as practical:** Like the Falcon web applications, we want our [Heroku buildpack](https://github.com/SalesforceFoundation/simple-salesforce-dx-buildpack)to stay up-to-date with the latest SF CLI to get its performance and feature enhancements. Like the Falcon web applications, this includes installing all necessary JIT plugins for the buildpack.
```
davidmreed, line 53: What criteria would you suggest for determining when it's practical to make this transition?

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Consequences
+
+-   Keeping things working as they are will cause the least disruption.
+-   By delegating backwards compatibility to the SF CLI, we can avoid making substantial changes to our command structure until it's necessary.
+-   Updating Falcon web apps and the SFDX Heroku buildpack to use the SF CLI will help us stay ahead of the change and avoid a scramble as the CLI removes deprecated commands.
+-   Installing JIT plugins in the apps and buildpack will make builds more efficient and avoid problems during runtime.
+-   However, this decision does mean that we will need to monitor Salesforce's communications closely for announcements about command deprecations, to ensure that our commands continue to function as expected.
```
davidmreed, line 61: 😬.

Our use of SFDX is really not wide. I wonder if we could just do an inventory and see if there are `sf` commands that map 1:1 with the same semantics.

There might be some stickyness around user-created tasks that use the SFDXCommandTask class, but I can see a possibility that the backwards-compatibility you mention above might be enough to address that. We could also log warnings to the user from that class.

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
jstvz: Good question. It's more accurate to say that we'd rely on the
`@salesforce/cli` package's `sfdx` alias. I'll update the text to
clarify that.

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
jstvz: Right, as mentioned above, we'd leave existing commands, which invoke
`sfdx`, as is. Since we have not yet migrated any commands to `sf` (and
I don't anticipate that we will be adding any new tasks invoking the
CLI) I think the risk is low.

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
jstvz: The main constraint I see is the team's bandwidth:

- for testing the dependency switch from `sfdx-cli` to `@salesforce/cli`
  on the Falcon web apps
- updating the buildpack to ensure the `sf` CLI is on the `PATH` and
  that the JIT plugins are available
- triggering builds to add the latest executables to the Heroku build
  slug

I'm not seeing any technical reasons not to make the switch. Users will soon start incorporating `sf`-style commands in their automation and we'll want to support that.

docs/adrs/0002-handling-migration-from-sfdx-to-sf-cli-in-cumulusci.md
jstvz: > 😬 

Yeah.

The catch is that the `sf` CLI may or may not be present, depending on the installation method: The `pkg` and other installers bundle the `sfdx` and `sf` CLIs, while NPM-based installations do not. Even if the `sf` CLI is available, the v2 SF commands that are the direct equivalents of the deprecated SFDX commands we invoke may not be available if it's v1.

In comparison, SFDXCommandTask-based tasks can remain unchanged, since v2 SF commands work with the `sfdx` alias.

vsbharath (approved):

PR #3636, opened 2023-08-03
fix: Don't pass fullName for Unlocked packages

### LINKED ISSUE #3633: Cryptic error when creating Unlocked Package Version
### Describe the bug

(Reported by @rygramer [here](https://trailhead.salesforce.com/trailblazer-community/feed/0D54V00007BFhyZ) and [here](https://trailhead.salesforce.com/trailblazer-community/feed/0D54V00007BFhyZ))

I am attempting to create a beta version of an unlocked package. While VerifyingMetadata, the package version creation fails with this error:

> "Can’t add component xyz because it’s associated with metadata from a managed 2GP package. Managed 1GP packages can’t depend on managed 2GP packages."

### Reproduction steps

1. `git clone https://github.com/rygramer/apexrolluppackagetest.git test; cd test`
2. `cci flow run release_unlocked_beta --org dev`

CumulusCI version: 3.77.0 (/Users/jestevez/.pyenv/versions/cci-d/bin/cci)
Python version: 3.8.14 (/Users/jestevez/.pyenv/versions/cci-d/bin/python)

### Operating System

Darwin Kernel Version 22.6.0

### Windows environment

_No response_

None

### Error Gist

https://gist.github.com/jstvz/dc8ce4d6219b931b66a882c1bf0f0629

### Additional information

_No response_

jstvz:
This PR fixes an issue where the creation of a beta version of an unlocked package was failing during the `create_package_version` task with a misleading error message about a 1GP dependency.

The workaround is to match the behavior of the [@salesforce/packaging](https://github.com/forcedotcom/packaging/tree/830d45bd52b80b09b7a5f96b9664961eee3ce7f1) typescript library: when the package type is 'unlocked', 'None' is passed instead of the package name. 

Fixes #3633

### DIFF
--- a/cumulusci/tasks/create_package_version.py
+++ b/cumulusci/tasks/create_package_version.py
@@ -211,7 +211,9 @@ class CreatePackageVersion(BaseSalesforceApiTask):
         package_zip_builder = None
         with convert_sfdx_source(
             self.project_config.default_package_path,
-            self.package_config.package_name,
+            None
+            if self.package_config.package_type == PackageTypeEnum.unlocked
+            else self.package_config.package_name,
             self.logger,
         ) as path:
             package_zip_builder = MetadataPackageZipBuilder(


davidmreed:
@jstvz do we have any confirmation that this is the same behavior as sfdx? There's a Deep Magic bug that manifests _roughly_ the same way but which I don't think has the same underlying cause.

jstvz:
> @jstvz do we have any confirmation that this is the same behavior as sfdx? There's a Deep Magic bug that manifests _roughly_ the same way but which I don't think has the same underlying cause.

Yes, as far as I can tell.

Source conversion takes place during `package version create` by calling [`MetadataConverter`](https://github.com/forcedotcom/source-deploy-retrieve/blob/88f711ee28af19811e10853124f7bfe990ab383e/src/convert/metadataConverter.ts#L24C20-L24C20) from source-deploy-retrieve via [`convertMetadata`](https://github.com/forcedotcom/packaging/blob/830d45bd52b80b09b7a5f96b9664961eee3ce7f1/src/package/packageVersionCreate.ts#L1107-L1129), which is only called by [`generateMDFolderForArtifact`](https://github.com/forcedotcom/packaging/blob/830d45bd52b80b09b7a5f96b9664961eee3ce7f1/src/package/packageVersionCreate.ts#L1069C31-L1069C31). Both of the `generateMDFolderForArtifact` callers that I found [construct the options](https://github.com/forcedotcom/packaging/blob/830d45bd52b80b09b7a5f96b9664961eee3ce7f1/src/package/packageVersionCreate.ts#L326-L330) [without passing](https://github.com/forcedotcom/packaging/blob/830d45bd52b80b09b7a5f96b9664961eee3ce7f1/src/package/packageVersionCreate.ts#L1058-L1062) through `packageName`. On the other hand, when we call `force:source:convert -n pkgName`, it does get passed through [here](https://github.com/salesforcecli/plugin-deploy-retrieve/blob/cf94d9b10d1256e9cad96f75df3452094b9d4865/src/commands/project/convert/source.ts#L136-L144).

davidmreed (approved):

davidmreed:
@jstvz Given your findings should this apply for _any_ 2GP package, not just Unlocked?

jstvz:
@davidmreed I believe we should limit it to just Unlocked packages because of: (1) the blast radius and, (2) no issues reporting similar behavior in 2GP packages (despite the scale at which we've been building them).

leboff:
@jstvz @davidmreed to your question, this does not just affect unlocked packages. My original bug report was for a locked 2GP, and I verified this error is still occurring on CCI 3.83.0

PR #3642, opened 2023-08-23
Add workarounds to support Vlocity local comp in CI

jstvz:
We discovered that using sf.accessToken for persistent orgs is not
compatible with the local compilation of omniscripts, which requires
this workaround. Moreover, vlocity_build does not support passing the
NPM auth token via CLI or environment variable, so we're adding support
for this to the vlocity_pack_deploy task.


jofsky:
@jstvz  what plan are you testing it against?

jofsky:
I'm finding Omniscripts to not be deployable under the api token schema

jstvz:
> I'm finding Omniscripts to not be deployable under the api token schema

@jofsky Would you say more about what error you're seeing?

jofsky (approved):

PR #3650, opened 2023-09-13
Extended 'Deploy' task to support REST API deployment

aditya-balachander:
Adding functionality to existing task "Deploy" where metadata can be deployed using REST API.


cumulusci/salesforce_api/tests/test_rest_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        mock_get.return_value = response_get
+
+        deployer = RestDeploy(
+            self.mock_task, self.mock_zip, False, False, "NoTestRun", []
+        )
+        deployer()
+
+        # Assertions to verify log messages and method calls
+        mock_post.assert_called_once()
+        self.assertEqual(
```
davidmreed, line 41: These assertions are completely fine - but they're fragile. When you do this kind of assertion, changing the order of logs or adding more logging will break your tests.

That's not _definitively_ a bad thing, but it's more likely to be noise than an opportunity to catch a real regression, which is what you want to do.

There's at least two other ways you can choose to approach this:

1. Don't assert against logs at all. If you have meaningful asserts about the API calls and outcome of your logic, you probably do not need them.
2. Use the `caplog` fixture from `pytest` and assert that the total log text contains what you expect to see.

cumulusci/salesforce_api/tests/test_rest_deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertEqual(self.mock_logger.info.call_count, 2)
+        self.assertEqual(self.mock_logger.error.call_count, 0)
+        self.assertEqual(
+            self.mock_logger.debug.call_args_list[0], call("Deployment status: Pending")
+        )
+        self.assertEqual(self.mock_logger.debug.call_count, 1)
+
+
+if __name__ == "__main__":
+    unittest.main()
```
davidmreed, line 182: Using `unittest` is fine; it's fully supported. Check out `pytest`, though. It makes it easier for you to write really lightweight tests. You don't need a class structure, just functions, and you can use _fixtures_ to inject your setup.

I'm not asking you to rewrite these perfectly fine tests! Just a pointer for the future.

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -55,6 +56,9 @@ class Deploy(BaseSalesforceMetadataApiTask):
         "transforms": {
             "description": "Apply source transforms before deploying. See the CumulusCI documentation for details on how to specify transforms."
         },
+        "rest_deploy": {
+            "description": "If True, deploy metadata with Apex Testing using REST API"
```
davidmreed, line 60: Apex tests don't _necessarily_ run on a deploy that doesn't include Apex.

```suggestion
            "description": "If True, deploy metadata using REST API"
```

davidmreed (changes requested):
@aditya-balachander Really nice work here. I'm impressed. Have you had a chance to do a hands-on test? Something I always recommend with critical code pathways like this one, since mock-based tests can only take you so far.

You have a handful of code lines I'd like to see covered in a test, particularly in `_reformat_zip()`. You can run `make coverage` to find missing coverage lines.

cumulusci/salesforce_api/rest_deploy.py
```diff
@@ -0,0 +1,152 @@
...
+import time
+import uuid
+import zipfile
+
+import requests
+
+
+class RestDeploy:
+    def __init__(
+        self, task, package_zip, purge_on_delete, check_only, test_level, run_tests
```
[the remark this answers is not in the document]
aditya-balachander: Added annotations

cumulusci/salesforce_api/tests/test_rest_deploy.py
aditya-balachander: I have handled the case where changing order of logs or adding more logs will break the tests. I have also added more test cases to test the individual functions such as `_reformat_zip` and `set_purge_on_delete` to increase coverage to 100%.

Additionally, I have added tests to `test_deploy.py` to test the functionality of setting the `api_class` to `RestDeploy`.

cumulusci/salesforce_api/rest_deploy.py
```diff
@@ -98,28 +107,19 @@ def _monitor_deploy_status(self, deploy_request_id):
             response_json = response.json()
 
             if response_json["deployResult"]["status"] != "InProgress":
```
davidmreed, line 109: I think the logic here is a little bit tricky to follow. Suggestion, if you think it works:

What if we log the state _before_ the conditional (so that the user always sees feedback during the process). Then, make the conditional check for both `InProgress` and `Pending`, so you do not need a condition around the `break`. We exit for any status other than `InProgress` or `Pending`, with the specific handling logic you have for failures.

What do you think?

davidmreed: I'm going to go ahead and approve this PR. You can make that change if you agree with it.

davidmreed (approved):

cumulusci/salesforce_api/rest_deploy.py
aditya-balachander: I understand, I will make these changes first thing tomorrow :)

davidmreed (approved):

PR #3652, opened 2023-09-13
enhancement: support spanning fields in update_data

JonnyPower:
- inbuilt isidentifier() method on string does not support strings including `.`
- Salesforce API will accept spanning fields in API query, but will return child objects / dicts
- Using pydash to navigate the dict by keys split by `.`

Draft to seek feedback early, as I don't write a lot of python and threw this together quickly.

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -37,3 +37,4 @@ For example:
 * Ed Rivas (jerivas)
 * Gustavo Tandeciarz (dcinzona)
 * Ben French (BenjaminFrench)
+* Jonny Power (JonnyPower) on behalf of Traction Rec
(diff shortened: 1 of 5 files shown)


JonnyPower:
some outstanding Qs for the team;
1. What's the preferred method of handling new dependencies? I've added pydash here to simplify the dictionary traversal; okay to add to prod dependencies? or would you rather I add custom traversal code?
2. I threw those test together quickly as I have to shoot out for PTO and wanted to gather feedback on direction for when I'm back - they are failing with `snowfakery.data_gen_exceptions.DataGenValueError: Problem rendering field FirstName:` - where do I add test data generation for snowfakery for tests?

JonnyPower:
did some more spiking on this, appears to be more complex than I though - the snowfakery render logic doesn't seem to like field names with `.` separators either. will pick up again in the future.

Issue #3466, opened 2022-12-13
Ability to specify a test suite to run instead of just test_name_match in the run_tests task


joeythomaschaske:
### Feature Type

I'm interested in functionality that solves a current problem or limitation in how CumulusCI works.

### Does your feature request solve a current problem?

We have several packages that share the same namespace and we do not have a good way to limit running tests to those in the current package. It would be useful to specify a test suite that contains the list of tests to execute.

### Describe the solution you'd like

We would like to be able to execute the `run_tests` task and specify which test suite to run. 

It would also be nice to include some "utilities" as well to maintain the list of tests in the test suite and perhaps have a dedicated subfolder in the `unpackaged/` directory to store these test suites. For our use case it would be that every test class that is in the package source gets added to the test suite instead of having to manually update it.

### Describe alternatives you've considered

We can override the `run_tests` task with a custom task where the `class_path` points to `cumulusci.tasks.sfdx.SFDXOrgTask` and execute `sfdx force:apex:test:run` while specifying which test suite to run by setting the `-s` option.

We can also create a custom flow that executes the `dx` task executing the same apex test command

jstvz:
Closing as part of a backlog triage pass against v4.10.0 / 5.0.0.dev1.

Reason: feature already implemented

If this is still reproducible against v4.10.0 (the current latest stable) or against v5.0.0.devN on the dev branch, please reopen with a fresh `cci version` block and reproduction steps. We track v5 development on the `dev` branch.

PR #3660, opened 2023-09-21
W-12214520: Adding ApexTestSuite support in run_tests

mjawadtp:
Changes made for 'run_tests' task:
`run_tests` accepts an optional list of test suites to run.
If provided, `run_tests` enqueues only those tests found in the given suites.

### DIFF
--- a/cumulusci/tasks/apex/testrunner.py
+++ b/cumulusci/tasks/apex/testrunner.py
@@ -120,7 +120,6 @@ class RunApexTests(BaseSalesforceApiTask):
                 "project__test__name_match from project config. "
                 "Comma-separated list for multiple patterns."
             ),
-            "required": True,
         },
         "test_name_exclude": {
             "description": (
@@ -171,6 +170,9 @@ class RunApexTests(BaseSalesforceApiTask):
             "description": "By default, only failures get detailed output. "
             "Set verbose to True to see all passed test methods."
         },
+        "test_suite_names": {
+            "description": "Accepts a comma-separated list of test suite names. Only runs test classes that are part of the test suites specified."
+        },
     }
 
     def _init_options(self, kwargs):
@@ -179,10 +181,15 @@ class RunApexTests(BaseSalesforceApiTask):
         self.options["test_name_match"] = self.options.get(
             "test_name_match", self.project_config.project__test__name_match
         )
-
         self.options["test_name_exclude"] = self.options.get(
             "test_name_exclude", self.project_config.project__test__name_exclude
         )
+
+        self.options["test_suite_names"] = self.options.get(
+            "test_suite_names", self.project_config.project__test__suite__names
+        )
+        if self.options["test_name_match"] is None:
+            self.options["test_name_match"] = ""
 
         if self.options["test_name_exclude"] is None:
             self.options["test_name_exclude"] = ""
@@ -236,6 +243,19 @@ class RunApexTests(BaseSalesforceApiTask):
         self.required_per_class_code_coverage_percent = int(
             self.options.get("required_per_class_code_coverage_percent", 0)
         )
+        # Raises a TaskOptionsError when the user provides both test_suite_names and test_name_match.
+        if (
+            self.options["test_suite_names"] is not None
+            and len(self.options["test_suite_names"]) > 0
+        ) and (
+            self.options["test_name_match"] is not None
+            and self.options["test_name_match"] != "%_TEST%"
+        ):
+            self.logger.info(self.options["test_suite_names"])
+            self.logger.info(self.options["test_name_match"])
+            raise TaskOptionsError(
+                "Both test_suite_names and test_name_match cannot be passed simultaneously"
+            )
 
     # pylint: disable=W0201
     def _init_class(self):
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/apex/testrunner.py
```diff
@@ -179,11 +181,16 @@ def _init_options(self, kwargs):
         self.options["test_name_match"] = self.options.get(
             "test_name_match", self.project_config.project__test__name_match
         )
-
         self.options["test_name_exclude"] = self.options.get(
             "test_name_exclude", self.project_config.project__test__name_exclude
         )
 
+        self.options["test_suite_names"] = self.options.get(
```
davidmreed, line 188: Let's use the `process_list_arg()` utility function here (you can find examples in the codebase). That will handle a situation where the user provided a comma-separated list on the command line, rather than a structured list in the YAML configuration.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -236,6 +243,19 @@ def _init_options(self, kwargs):
...
+        # Raises a TaskOptionsError when the user provides both test_suite_names and test_name_match.
+        if (
+            self.options["test_suite_names"] is not None
+            and len(self.options["test_suite_names"]) > 0
+        ) and (
+            self.options["test_name_match"] is not None
+            and self.options["test_name_match"] != "%_TEST%"
+        ):
+            self.logger.info(self.options["test_suite_names"])
+            self.logger.info(self.options["test_name_match"])
```
davidmreed, line 255:
```suggestion
```

cumulusci/tasks/apex/testrunner.py
```diff
@@ -236,6 +243,19 @@ def _init_options(self, kwargs):
         self.required_per_class_code_coverage_percent = int(
             self.options.get("required_per_class_code_coverage_percent", 0)
         )
+        # Raises a TaskOptionsError when the user provides both test_suite_names and test_name_match.
+        if (
+            self.options["test_suite_names"] is not None
+            and len(self.options["test_suite_names"]) > 0
```
davidmreed, line 249: Because of the way truthy semantics work in Python, you can reduce these conditionals to `self.options["test_suite_names"]`. An empty list is falsy, as is `None`.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -283,13 +303,81 @@ def _get_test_class_query(self):
         return query
 
     def _get_test_classes(self):
+        # If test_suite_names is provided, execute only tests that are a part of the list of test suites provided.
+        if (
+            self.options["test_suite_names"] is not None
+            and len(self.options["test_suite_names"]) > 0
```
davidmreed, line 309: Same note as above.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -283,13 +303,81 @@ def _get_test_class_query(self):
...
+        item_string = ""
+
+        for index, item in enumerate(itemlist):
+            item_string += f"'{item}'"
+            if index is not len(itemlist) - 1:
+                item_string += ","
+
+        return item_string
+
+    def _get_test_suite_ids_from_test_suite_names_query(self):
```
davidmreed, line 335: This method gets easier to test if you take `test_suite_names_arg` as a parameter, instead of pulling it from the task's state.

cumulusci/tasks/apex/testrunner.py
```diff
@@ -283,13 +303,81 @@ def _get_test_class_query(self):
...
+        query1 = self._get_test_suite_ids_from_test_suite_names_query()
+        self.logger.info("Fetching test suite metadata...")
+        result = self.tooling.query_all(query1)
+        testSuiteIds = []
+
+        for record in result["records"]:
+            testSuiteIds.append(str(record["Id"]))
+
+        query2 = self._get_test_classes_from_test_suite_ids_query(testSuiteIds)
+        self.logger.info("Fetching test classes belong to the test suite(s)...")
```
davidmreed, line 376:
```suggestion
        self.logger.info("Fetching test classes belonging to the test suite(s)...")
```

cumulusci/tasks/apex/testrunner.py
```diff
@@ -283,13 +303,81 @@ def _get_test_class_query(self):
         return query
 
     def _get_test_classes(self):
+        # If test_suite_names is provided, execute only tests that are a part of the list of test suites provided.
+        if (
+            self.options["test_suite_names"] is not None
+            and len(self.options["test_suite_names"]) > 0
+        ):
+            test_classes_from_test_suite_names = (
```
davidmreed, line 311: This early return makes the logic a bit harder to follow in my opinion. What do you think about pulling the code below in this function out into a separate function, so that you can have a simple if/else here?

davidmreed (changes requested):
Nice work here, @mjawadtp! Just a few suggestions from me.

cumulusci/tasks/apex/testrunner.py
mjawadtp: Hey David, currently, input is being fetched from the command line. Input is not read from YAML configuration as of yet. Comma-separated list provided by the user is successfully processed by the utility function _get_comma_separated_string_of_items which formats the list of strings in to a single string which is subsequently inserted into queries. 

Please let us know if we are missing anything here.

cumulusci/tasks/apex/testrunner.py
mjawadtp: Thanks for the valuable comment. Updated as requested.

cumulusci/tasks/apex/testrunner.py
mjawadtp: Thank you for pointing it out. This method as well as its respective tests has been updated as suggested.

cumulusci/tasks/apex/testrunner.py
mjawadtp: That's a fair point. We have written a separate function that fetches all the test classes from the org. The aforementioned function is run when the else block is executed. Thanks for pointing it out.

davidmreed (dismissed):
LGTM. Nice work, @mjawadtp.

Closes #3466.

davidmreed (approved):

PR #3669, opened 2023-10-05
Feature/listmetadatatypes

lakshmi2506:


cumulusci/salesforce_api/soap_envelopes.py
```diff
@@ -161,3 +161,18 @@
     </urn:create>
   </soapenv:Body>
 </soapenv:Envelope>"""
+
+
+METADATA_TYPES = """<?xml version="1.0" encoding="utf-8"?>
```
ssahay87, line 166: @lakshmi2506  one small suggestion , we should name is DESCRIBE_METADATA so that it stays in sync with its resource api, cc- @jkasturi-sf

ssahay87 (commented):
added few comments

jkasturi-sf (approved):
Looks good

PR #3670, opened 2023-10-06
Allow None override for version_base and ancestor_id in create_package_version

zenibako:
If you attempt to override the preset `version_base` and `ancestor_id` options for the `create_package_version` task (for example, to bypass GitHub resolution in the `release_2gp_beta` flow), you will be blocked by errors (`_get_base_version_number()` will attempt to parse the value as a version number, and `_resolve_ancestor_id()` doesn't recognize the value).

This PR adds a new `utils.process_none_arg` function, and updates `create_package_version` to use this new function for `version_base` and `ancestor_id`.


jstvz:

The `None`-override pattern is useful, but it touches a piece of API surface that benefits from broader design discussion before adoption. Please consider opening an issue first if you want to revisit; we would welcome a fresh PR once we have alignment on the API shape.

PR #3671, opened 2023-10-06
Don't downgrade unlocked packages

yippie:
When installing 2GP packages or any package using a package version id, CumulusCI is happily downgrading package versions.

Updated the check to pull the full version info of the package version id to be installed and then compare it to the installed packages in the org by Package ID and version number.


yippie:
@leboff https://github.com/SFDO-Tooling/CumulusCI/pull/3767 seems to be an alternative way to fix this same, and very important, issue.

vsbharath:
@yippie Closing this PR as this is similar to [3767](https://github.com/SFDO-Tooling/CumulusCI/pull/3767). And also responded [here](https://github.com/SFDO-Tooling/CumulusCI/pull/3767#issuecomment-2118209521) with few requirements.

PR #3674, opened 2023-10-12
Fix `TypeError` when service sensitive attribute is `None`

jain-naman-sf:
Fixed the exception for the command cci service info connected_app while built_in is default


jstvz (approved):
Adding a test case for this would be nice, but not a blocker.

PR #3675, opened 2023-10-12
Fixed Log line issue for cci org info

jain-naman-sf:


cumulusci/core/config/sfdx_org_config.py
```diff
@@ -25,7 +25,7 @@ def sfdx_info(self):
         username = self.config.get("username")
         assert username is not None, "SfdxOrgConfig must have a username"
 
-        self.logger.info(f"Getting org info from Salesforce CLI for {username}")
+        # self.logger.info(f"Getting org info from Salesforce CLI for {username}")
```
jkasturi-sf, line 28: @jain-naman-sf Can we suppress this line only in case of --json option is used?

jain-naman-sf: @jkasturi-sf No now it will be suppressed without —json option as well. This fix will be working in both cases

jkasturi-sf (approved):
Looks good

Issue #3089, opened 2022-02-18
task run_tests not including namespace in query


tombvo:
### Describe the bug

The tasks `run_tests` does not query for the namespace it was given.

We defined the namespace via cumulus.yml like that:
```
tasks:
  run_tests:
    options:
      required_org_code_coverage_percent: 90
      test_name_match: "%Test"
      namespace: neo_umd
```

and ran `cci task run run_tests --org dev` afterwards.
The output was the following:
`[02/18/22 15:26:28] Beginning task: RunApexTests
                    As user: test-se5xnjzaqsqn@example.com
                    In org: 00D3N0000002wKN

                    Running query: SELECT Id, Name FROM ApexClass WHERE NamespacePrefix = null AND (Name LIKE '%Test')
                    Found 0 test classes`

If i run `cci task info run_tests` "neo_umd" is set as default for namespace and "%Test" is set as default for test_name_match.
That means, the definition in the cumulus.yml is correct.

We have also tried running the command with manual parameters like that:
`cci task run run_tests --org dev --namespace 'neo_umd' --test_name_match '%Test'`

but we get the same result as above.

### Reproduction steps

1. run `cci task run run_tests` against an org with the namespace parameter attached manually or defined in cumulus.yml
2. check if output query's with correct namespace or says null instead

### Your CumulusCI and Python versions

**Test Person 1**
CumulusCI version: 3.52.0 
Python version: 3.9.0 
sfdx: sfdx-cli/7.138.1 win32-x64 node-v16.14.0

**Test Person 2**
CumulusCI version: 3.49.0
Python version: 3.9.0 
sfdx: sfdx-cli/7.120.0 win32-x64 node-v16.13.1

### Operating System

Windows 11

### Windows environment

PowerShell

### CumulusCI installation method

pipx

davisagli:
@tombvo The namespace from the `namespace` option is only used if the `managed` option is also set to True.

In the context of an org with an installed package from that namespace, the task will automatically consider `managed` to be true by default. But it looks like you are in a namespaced scratch org where you need the namespace but there is not an installed package with that namespace. You should be able to make it work by manually setting the `managed` option to `True`.

@davidmreed It looks like we could make this task work a bit more intuitively with namespaced scratch orgs.

davidmreed:
I agree that this is a bug. Tracking as W-10729382. In the meantime, the workaround @davisagli mentioned should cure the issue.

jstvz:
[...]

PR #3677, opened 2023-10-12
Allowed namespace injection without managed

jain-naman-sf:


lakshmi2506:
Hi @jkasturi-sf 

I have added the error message for the parse_url_string when the url doesn't have set origin. This is possible while we run flows.

Please review the code
Thank you

jain-naman-sf:
Hey @jkasturi-sf Please review this PR

jstvz (changes requested):
Thanks for your work here. I noticed two things we need to check:

1. Your changes look at the task options (managed and namespace) to decide about the namespace filter. But, the work item and discussion in #3089 suggests we should look at the details of the target org instead (e.g. [`self.org_config.namespace`](https://github.com/SFDO-Tooling/CumulusCI/blob/520b51128c7702a374da9a8838f38890082fd3de/cumulusci/core/config/org_config.py#L240)).
2. You don't have any tests that confirm the new behavior as specified in the acceptance criteria on the story.

So, we need the task to look at the target org's details to choose if the namespace filter is used, not just the task options.

Can you adjust this? Let me know if you need more information or if anything is unclear.

jain-naman-sf:
Hey @jstvz I have added the functionality to infer namespace from target org directly along with tests.

jkasturi-sf (approved):
Looks good

jstvz (approved):

PR #3679, opened 2023-10-13
Error message when git remote URL throws exception if no origin remote

lakshmi2506:
Fails to give the error message when run any task involving git if the URL doesn't contain the origin remote.

This is because in the project config URL is set to None due to this`url = self.git_config_remote_origin_url()` 
and when the below function is called. Parse_repo_url is called with None as argument.

`def get_github_api_for_repo(keychain, repo_url, session=None) -> GitHub:
    owner, repo_name, host = parse_repo_url(repo_url)`

Work item: [link](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001M62egYAB/view)


lakshmi2506:
Hi @jkasturi-sf 

I have added the clear error for the parse_url when the url passed doesn't have remote set as origin

please review the code
Thank you

cumulusci/utils/git.py
```diff
@@ -71,6 +72,9 @@ def parse_repo_url(url: str) -> Tuple[str, str, str]:
     Tuple: (str, str, str)
         Returns (owner, name, host)
     """
+    if not url:
+        raise error("Url is none or must have `remote` set as origin")
```
jstvz, line 76: ±: Instead of using `from os import error`, consider using Python's built-in `ValueError` for validating the URL. It's more semantically correct for this use case.

jstvz, line 76: The error message could be clearer. Consider rephrasing it to "The provided URL is empty or invalid."

cumulusci/utils/tests/test_git.py
```diff
@@ -69,3 +71,8 @@ def test_construct_release_branch_name():
 def test_parse_repo_url(repo_uri, owner, repo_name, host):
     assert parse_repo_url(repo_uri) == (owner, repo_name, host)
     assert split_repo_url(repo_uri) == (owner, repo_name)
+
+
+def test_empty_url():
+    with pytest.raises(error, match="Url is none or must have `remote` set as origin"):
```
jstvz, line 77: Use the more specific exception (`ValueError` if changed as per suggestion above) instead of the generic `error`.

jstvz, line 77: Consider putting the error message in a constant. This makes it easier to change in one place and helps keep the code clean and simple.

jstvz (changes requested):
Thanks for the changes! Have are a few suggestions below. In addition, it'd be helpful to add a meaningful PR title and description. If this change is intended to address an issue a link to that should be included as well.

Please make these updates and let's review again.

lakshmi2506:
Hi @jstvz 

Requested changes are made. Please go through the changes.
Thank you

jkasturi-sf (approved):
Looks good

PR #3681, opened 2023-10-16
Handling exception when the Tooling API returns a test result with a null method name

lakshmi2506:
Handled the situation when the Tooling API returns a test result with a null methodname by detecting the null and enqueueing all methods in the test class with the given `ApexClassId` for retry.

Retry is done for the classes with methodname None in both test case pass and fail conditions and the number of retried tests is updated accordingly.

Work item: [link](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000IMA9gYAH/view)


lakshmi2506:
Hi @jkasturi-sf 

Handled the situation when the Tooling API returns a test result with a null methodname by detecting the null and enqueueing all methods in the test class with the given `ApexClassId` for retry.

Retry is done for the classes with methodname None in both test case pass and fail conditions. 

Please review the code 

Thank you

lakshmi2506:
Hi @jkasturi-sf 

Please review the code. 
Thank you

jkasturi-sf (approved):
Looks good

PR #3683, opened 2023-10-18
Fix Github url parse error for some scenarios

mgrandhi:
Fix Github parse issue during some scenarios as mentioned below
1. https://user@github.com/owner/repo.git
2. git@github.com:/SalesforceFoundation/Subledger.git

### DIFF
--- a/cumulusci/utils/git.py
+++ b/cumulusci/utils/git.py
@@ -72,6 +72,7 @@ def parse_repo_url(url: str) -> Tuple[str, str, str]:
         Returns (owner, name, host)
     """
     url_parts = re.split("/|@|:", url.rstrip("/"))
+    url_parts = list(filter(None, url_parts))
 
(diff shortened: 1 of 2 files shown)


cumulusci/utils/git.py
```diff
@@ -81,6 +82,6 @@ def parse_repo_url(url: str) -> Tuple[str, str, str]:
 
     host = url_parts[-3]
     # Need to consider "https://api.github.com/repos/owner/repo/" pattern
-    if "http" in url_parts[0] and len(url_parts) > 6:
+    if "http" in url_parts[0] and len(url_parts) > 4 and not host.endswith(".com"):
```
jkasturi-sf, line 85: @mgrandhi This may work if hostname ends with .com, but if the user having hostname ends with different name then this may fail. Can we use regex instead of hard coding the value here?

PR #3684, opened 2023-10-18
metadeploy_publish presents clear errors when plans are not available

lakshmi2506:


cumulusci/tasks/metadeploy.py
```diff
@@ -83,12 +84,19 @@ def _init_task(self):
...
+                )
             plan_configs = {
                 plan_name: self.project_config.lookup(f"plans__{plan_name}")
             }
             self.plan_configs = plan_configs
         else:
             self.plan_configs = self.project_config.plans
+        # Handled exception for no plan
+        if self.plan_configs is None or len(self.plan_configs) == 0:
+            raise error("No plan found to publish in project configuration")
```
davidmreed, line 99: I don't think you want this `error` symbol. You can raise a `CumulusCIException` here.

cumulusci/tasks/tests/test_metadeploy.py
```diff
@@ -485,6 +494,40 @@ def test_init_task__named_plan(self):
...
+        task_config = TaskConfig({"options": {"tag": "release/1.0"}})
+        task = Publish(project_config, task_config)
+        with pytest.raises(
+            error,
+            match="No plan found to publish in project configuration",
+        ):
+            task._init_task()
+
+    @responses.activate
+    def test_init_task_no_plan_name(self):
```
davidmreed, line 515: This test case could, at your option, be parameterized (using `@pytest.mark.parametrize()`) with the previous one.

davidmreed (changes requested):
Nice work, just a tweak before we merge.

davidmreed (approved):

PR #3685, opened 2023-10-19
Docs: Python version details

allisonletts:
The Python release listed in the docs is pretty old. Can we change it to reference the latest Python 3 version instead?
Is it fair to say that in October, the previous major release of Python is going to be more compatible with CCI?


jstvz:
Thanks for the contribution and apologies for the long delay. This PR is stale enough that we are closing it to reduce queue noise. If the docs change is still useful, a fresh PR against `main` would be welcome.

PR #3688, opened 2023-10-25
Add transform to remove invalid references from profiles

aditya-balachander:
[W-11744607](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000016LSxKYAW/view)
[W-11744695](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000016LVBcYAO/view)

Implemented Functionality to clean profiles and permission sets of invalid references.

Issues (corner cases that are not supported currently)
1. Since we are using the approach of parsing the xml files in the source zip and retrieving all entities using package.xml, we are unable to retrieve:
    i. Couple of sObjects (some of which are not supported, and Document which is a metadata of it's own but comes under ObjectPermissions)
   ii. Couple of tabs (These tabs are permissioned in the profile but are not present in the TabDefinition table or the tabs page in setup of the org. Most likely these are linked tabs as part of applications or other metadata types)
  iii. Couple of fields (which are permissioned in the org but are missing in the sObject itself)
2. For User Permissions:
    i. Some UserPermissions are dependent on other UserPermissions. Say permission A depends on B (both A and B are present in the target org). There is no current check to see if B is present when A is present in the profile or permissionset file. Without B, deployment fails when you try to deploy A.

We have created a WI to resolve these issues: [W-14363935](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001d7lBSYAY/view)


cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+        self.name_xpath = name_xpath
+
+    def return_parent(self, element: ET.Element):
+        return element.find(self.name_xpath).text.split(".")[0]
+
+    def return_name(self, element: ET.Element):
+        return element.find(self.name_xpath).text
+
+
+OBJECT_PERMISSION = PermissionElementXPath(".//objectPermissions", "object")
```
davidmreed, line 43: I like the use of constants here to centralize the XPath complexity. 👍

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+APEXCLASS_PERMISSION = PermissionElementXPath(".//classAccesses", "apexClass")
+APEXPAGE_PERMISSION = PermissionElementXPath(".//pageAccesses", "apexPage")
+FLOW_PERMISSION = PermissionElementXPath(".//flowAccesses", "flow")
+CUSTOMPERM_PERMISSION = PermissionElementXPath(".//customPermissions", "name")
+CUSTOMSETTING_PERMISSION = PermissionElementXPath(".//customSettingAccesses", "name")
+CUSTOMMETADATA_PERMISSION = PermissionElementXPath(
+    ".//customMetadataTypeAccesses", "name"
+)
+
+PACKAGEXML_DICT = {
```
davidmreed, line 60: At first read, I don't understand these data structures. A comment might be helpful.

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+        }
+
+    for query_name, future in futures.items():
+        try:
+            query_result = future.result()
+            results_dict[query_name] = query_result
+        except Exception as e:
+            raise Exception(f"Error executing query '{query_name}': {type(e)}: {e}")
+        else:
+            queries.pop(query_name, None)
```
davidmreed, line 111: I'm not sure I understand what this line does, since the loop doesn't iterate over `queries`. Is the goal to mutate the data structure for the caller?

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+def entities_from_query(sf: Salesforce):
+    # Define the query
+    def run_query(path: str):
+        urlpath = sf.base_url + path
+        method = "GET"
+        return sf._call_salesforce(method, urlpath).json()
+
+    # Queries
+    queries = {}
+    queries["userPermissions"] = "sobjects/PermissionSet/describe"
```
davidmreed, line 125: I like running these queries using the sObjects rather than the Tooling API, which was what I originally had in mind. I would guess this approach is faster; it's definitely more elegant.

davidmreed: One quibble: most of these aren't _really_ queries, in the SOQL sense - they're API calls. Maybe adjust the naming to clarify? I'm also fine if you want to keep it like this, though.

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+            field_entities.extend(
+                picklistValue["value"]
+                for picklistValue in field.get("picklistValues", [])
+            )
+    return set(field_entities)
+
+
+def process_tabs(result_dict: dict) -> set:
+    tabs = [tab["Name"] for tab in result_dict["records"]]
+    return set(tabs)
```
davidmreed, line 164: Fun Python trick. You can use a _set comprehension_ here:

```suggestion
    return {tab["Name"] for tab in result_dict["records"]}
```

Curly braces are a set comprehension unless you have a `k:v` structure, which makes it a dict comprehension.

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+        if obj.get("name") == "SobjectType":
+            objects.extend(
+                picklistValue["value"]
+                for picklistValue in obj.get("picklistValues", [])
+            )
+    return set(objects)
+
+
+def return_package_xml_from_zip(zip_src: zipfile.ZipFile, api_version: str):
+    # Iterate through the zip file to generate the package.xml
```
davidmreed, line 179: Clarifying: this method scans all of the permission-related XML files in the zipfile and constructs a single, unified `package.xml` that retrieves all of the permissioned entities. Is that right? We could add a docstring here.

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+        # If object, fetch all the fields and record types present inside the file
+        if name.endswith(".object"):
+            file = zip_src.open(name)
+            root = ET.parse(file).getroot()
+            root = strip_namespace(root)
+            fields, recordTypes = get_fields_and_recordtypes(root, metadataName)
+            target_entities["fields"].update(fields)
+            target_entities["recordTypes"].update(recordTypes)
+
+        # To handle tabs which are not part of TabDefinition Table
```
davidmreed, line 263: This is a weird new facet I had not seen before. So basically you find situations where a CustomApplication contains tabs that are not found in TabDefinition?

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+    for element in root.findall(perm_entity.permission_xpath):
+        entity_names.add(
+            perm_entity.return_parent(element)
+            if parent
+            else perm_entity.return_name(element)
+        )
+    return entity_names
+
+
+def CleanXML(root: ET.Element, target_entities: Dict[str, set]):
```
davidmreed, line 318:
```suggestion
def clean_xml(root: ET.Element, target_entities: Dict[str, set]):
```

I'd use this form to match typical Python idiom. `CleanXML` looks like a class name.

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+def zip_clean_invalid_references(zf: zipfile.ZipFile, context: TaskContext):
+    # Set API version
+    api_version = context.org_config.latest_api_version
+
+    # Query and get entities
+    sf = ret_sf(context, api_version)
+    userPermissions, fields, tabs, objects = entities_from_query(sf)
+
+    # Update the target entites
+    target_entites = {}
```
davidmreed, line 357:
```suggestion
    target_entities = {}
```

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+        context.org_config,
+        api_version=api_version,
+        base_url=None,
+    )
+    return sf
+
+
+def zip_clean_invalid_references(zf: zipfile.ZipFile, context: TaskContext):
+    # Set API version
+    api_version = context.org_config.latest_api_version
```
davidmreed, line 350: Oh, there's some tricky behavior here, and I wonder if that's why you've seen unexpected outcomes around API-version-gated UserPermissions. I think what we want to do here is grab the API version _from the `package.xml` in the zip file_, because that guarantees that we clean against the same API version that will be used to run the deploy. Does that seem right?

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+
+    # Query and get entities
+    sf = ret_sf(context, api_version)
+    userPermissions, fields, tabs, objects = entities_from_query(sf)
```
davidmreed, line 354:
```suggestion
    user_permissions, fields, tabs, objects = entities_from_query(sf)
```

Takes a while to switch from Java's lowerCamelCase to Python's lower_snake_case :)

cumulusci/utils/clean_invalid_references.py
```diff
@@ -0,0 +1,365 @@
...
+
+    # Update the target entites
+    target_entites = {}
+    target_entites.update(entities_from_package(zf, context, api_version))
```
davidmreed, line 358: I admit I don't quite understand this code path. Is the intention here to clean the Profile against the contents of the zipfile?

davidmreed (commented):
@aditya-balachander  This looks great. Appreciate the depth of research you put into edge cases here. I have a few questions about aspects I didn't understand. We might actually be doing more than we need to!

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -60,10 +60,16 @@ class Deploy(BaseSalesforceMetadataApiTask):
         "clean_meta_xml": {
             "description": "Defaults to True which strips the <packageVersions/> element from all meta.xml files.  The packageVersion element gets added automatically by the target org and is set to whatever version is installed in the org.  To disable this, set this option to False"
         },
+        "clean_profiles": {
```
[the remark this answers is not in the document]
aditya-balachander: Hey @davidmreed the `clean_profiles` is not what we are using. This was previously present in @mjawadtp's code and it was changed to `clean_invalid_ref`. I'm extremely sorry for the oversight

cumulusci/tasks/salesforce/Deploy.py
```diff
@@ -220,6 +226,9 @@ def _get_package_zip(self, path) -> Union[str, dict, None]:
             "clean_meta_xml": process_bool_arg(
                 self.options.get("clean_meta_xml", True)
             ),
+            "clean_profiles": process_bool_arg(
+                self.options.get("clean_profiles", True)
```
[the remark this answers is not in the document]
aditya-balachander: The `clean_invalid_ref` defaults to False. Again sorry for not removing the old code

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Thanks :)

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Yep will put a comment there

cumulusci/utils/clean_invalid_references.py
aditya-balachander: yep I will change the variable name

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Ohh yes. That line is useless. I had earlier made this so that the same queries variable was reusable but now it doesnt matter. I will remove it

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Yep, will do that for the rest of the functions too!

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Yes that is right. I should probably add more comments everywhere. Let me do that.

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Yep will change that

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Yes there are tabs which are present in CustomApplications which are not inside TabDefinition. But I changed to using `PermissionSetTabSetting` which covers everything. This block of code is not needed

cumulusci/utils/clean_invalid_references.py
aditya-balachander: Ohh my bad

cumulusci/utils/clean_invalid_references.py
aditya-balachander: yep will change that

cumulusci/utils/clean_invalid_references.py
aditya-balachander: The logic is this:
1. I go through the .xml files and generate a package.xml containing all the entities that are needed to be deployed.
2. This package.xml I am using to retrieve all these entities in the target org using the `ApiRetrieveUnpackaged` class
3. Once I get back the zip file, this zip file will only contain those entities which are only present in the target org. This way I can compare the entities in the profile.xml and the entities from the zip file returned by package.xml and remove those which are not present in the target org.

cumulusci/tasks/salesforce/Deploy.py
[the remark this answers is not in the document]
aditya-balachander: Ohh shall I keep it to `True`?

cumulusci/utils/clean_invalid_references.py
aditya-balachander: I will make this change. But what is happening is that, when you retrieve a profile from SOAP API even if you mention an API version that these `UserPermissions` are not supported in, these `UserPermissions` get retrieved in the `profile.xml` file.

aditya-balachander:
[W-14363935](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001d7lBSYAY/view)

Most of the corner cases resolved. There is still an issue with some of the UserPermissions not being retrieved (in certain API versions). This is because in core, these UserPermissions either have a `maxApiVersion` defined (and our api version is more than that) or has `apiAccess=IsSfdcInternal` making it unexposed to public API.

The problem occurs because the SOAP call to retrieve the profile/permissionset gives all the UserPermissions regardless of which API Version is mentioned. However when we make a rest/tooling API call to retrieve these UserPermissions, only the ones present in that particular API version are retrieved.

@davidmreed kindly review the PR

jkasturi-sf (approved):
If a profile has references to an entity which doesn't exists in the org, then this change will help us to remove the references to those entities and deploy. This will allow us to deploy profiles without throwing an error

Issue #3627, opened 2023-07-19
Find And Replace Task Of CumulusCI not working


sahilconcret:
### Describe the bug

I was trying to replace the "%%%CommunityUrl%%%" with the site url via custom task in cumulus with Python Library (FindReplace).
This was the import statement
   -> from cumulusci.tasks.util import FindReplace as FindReplace

CumulusCI file task
find_replace_base_community:
        description: Replaces a valid card string with the target org's community base URL
        class_path: tasks.replace_community_url.FindReplaceCommunityURL
        options:
            find: "%%%COMMUNITY_BASE_URL%%%"
            replace: baseURL
            path: unpackaged/config/cmts/customMetadata

Expected behavior:
%%%COMMUNITY_BASE_URL%%% is being replaced by the base URL by a Python task (FindReplaceCommunityURL).
Ex: https://app-energy-9593-dev-ed.scratch.my.site.com/s/payment-router

Actual behavior:
%%%COMMUNITY_BASE_URL%%% is not being replaced by the base URL by a Python task (FindReplaceCommunityURL) instead it is replacing the %%%COMMUNITY_BASE_URL%%% to the string baseURL.
Ex:  baseURL/s/payment-router 
This error is happening after the cumulus update.
We checked the release notes and there was no update for this task

### Reproduction steps

find_replace_base_community:
        description: Replaces a valid card string with the target org's community base URL
        class_path: tasks.replace_community_url.FindReplaceCommunityURL
        options:
            find: "%%%COMMUNITY_BASE_URL%%%"
            replace: baseURL
            path: unpackaged/config/cmts/customMetadata

1. I ran the above task it was not replacing the destination file with baseUrl which was being setted in Python File instead it was 
replacing it with "baseURL" string only

### Your CumulusCI and Python versions

CumulusCI version: 3.75.1 (C:\Users\Sahil Manglani\.local\bin\cci)
Python version: 3.9.0 (C:\Users\Sahil Manglani\.local\pipx\venvs\cumulusci\Scripts\python.exe)

There is a newer version of CumulusCI available: 3.77.0
To upgrade, run `pipx upgrade cumulusci`
See the latest CumulusCI Release Notes: https://github.com/SFDO-Tooling/CumulusCI/releases/tag/v3.77.0

### Operating System

Windows

### Windows environment

Command Prompt

### CumulusCI installation method

pipx

prakash-concret:
[...]

PR #3695, opened 2023-11-03
Make task `options` read only when using Pydantic option validation

aditya-balachander:
[W-13798912](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001WfO3eYAF/view)

Made `self.options` to be Read-Only, if an `Options` class is defined in the Task. Added usage to documentation. Additionally, corrected `ListMetadataTypes` in `cumulusci/tasks/util.py` to use `parsed_options` instead of `options` (was missed in the pydantic options commit). 

Fixes: #3627


cumulusci/utils/options.py
```diff
@@ -18,6 +19,28 @@ def _describe_field(field):
...
 
+class ReadOnlyOptions(dict):
+    """To enforce self.options to be read-only"""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def __setitem__(self, key, value):
+        raise TaskOptionsError(
+            "The 'options' dictionary is read-only. Please use 'parsed_options' instead."
```
jstvz, line 30: ±: How about extracting this error message into a module-level constant? Doing so would apply the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle and be consistent with the codebase.

cumulusci/utils/tests/test_option_parsing.py
```diff
@@ -154,3 +159,33 @@ def test_multiple_errors(self):
...
+        task1 = TaskToTestTypes(self.project_config, self.task_config, self.org_config)
+        assert isinstance(task1.options, ReadOnlyOptions)
+        # Does not have an Options class
+        task2 = TaskWithoutOptions(
+            self.project_config, self.task_config, self.org_config
+        )
+        assert isinstance(task2.options, dict)
+
+    def test_init_options__options_read_only_error(self):
+        expected_error_message = "The 'options' dictionary is read-only. Please use 'parsed_options' instead."
```
jstvz, line 174: If the error message was a constant it could be imported and used here.

cumulusci/utils/tests/test_option_parsing.py
```diff
@@ -154,3 +159,33 @@ def test_multiple_errors(self):
...
+        task = TaskToTestTypes(self.project_config, self.task_config, self.org_config)
+        # Add new option
+        with pytest.raises(TaskOptionsError) as exc_info:
+            task.options["new_option"] = "something"
+        assert expected_error_message == str(exc_info.value)
```
jstvz, line 179: ↔️: Another way to write this, using [`pytest.raises`](https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises) `match` parameter:
```python
        with pytest.raises(TaskOptionsError, match=expected_error_message):
            task.options["new_option"] = "something"
```

jstvz (changes requested):
Requested changes/queries are marked with ±

---
<details>
<summary>Emoji Key</summary>


</details>

cumulusci/utils/options.py
aditya-balachander: Yes will make this change

cumulusci/utils/tests/test_option_parsing.py
aditya-balachander: Yep, makes more sense to use a constant

cumulusci/utils/tests/test_option_parsing.py
aditya-balachander: Yeah will make this change

jstvz (approved):

PR #3702, opened 2023-11-13
Error during upsert on identical DeveloperName for RecordType of same sObject

aditya-balachander:
[W-11466074](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000011wj7UYAQ/view)

Fixed a bug during upsert (load) of records where if the sObject had two record types with the same `DeveloperName` but different `IsPersonType`,  duplicate records were being generated for load and as a result, error was thrown.

Example: For the sObject `Account`, the record type `PersonAccounts` has `DeveloperName = PersonAccounts` and `IsPersonType = True`. I could also create another recordType such that `DeveloperName = PersonAccounts` but the `IsPersonType = False`. In such cases, say I was trying to upsert 1 record (any update_key) with this `PersonAccounts` (one of the two), 2 records would try to get inserted and would throw a duplicate row error.


lakshmi2506 (commented):
LGTM , just add that testcase with same developername it would be perfect

jkasturi-sf (approved):
This fix will resolve the duplicate record insertion issue when PersonAccount is enabled on the org and duplicate entries of PersonAccount are exists in RecordType entity.

jkasturi-sf (approved):

PR #3713, opened 2023-12-06
Incorrect generation of record type mapping table name in a namespaced org context

lakshmi2506:
[W-11242094](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00000yQTZ8YAO/view)

In case of  `generate_and_load_from_yaml` when inject_namespace is true and while creating the record type table, the naming format doesn't actually support a case. 

That is it is searching for the namespace_sfobject_rt_mapping but the actually table created is sfobject_rt_mapping.


mgrandhi (approved):
LGTM

jkasturi-sf (approved):
Looks good

PR #3718, opened 2023-12-11
Rollback Operation for Inserts and Upserts

aditya-balachander:
[W-14522337](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001et8KzYAI/view) and [W-14522327](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001etNspYAE/view)

The implementation is as follows:
1. For inserts, we maintain a separate table `{sobject}_insert_rollback` for records that have been created (either through insert or upsert) where we enter only the `sf_id`
2.  For upserts, before we load the records, we store the previous values of the records that will get updated (not upserted) in a table called `{sobject}_upsert_rollback`. It will have all the fields that are given in the `mapping.yml` file along with the `sf_id`.
3. When rollback occurs, it is as follows, in reverse order of tables created (and only considering the rollback tables):
    a. For insert, we delete all the `sf_id`
    b. For upsert, we upsert again the previous values


pratibhavermasalesforce (dismissed):

pratibhavermasalesforce (dismissed):

jkasturi-sf (approved):
Great work @aditya-balachander @jain-naman-sf for adding support to rollback transactions if any insert or update is failed in any part of sample data load task.

PR #3720, opened 2023-12-15
Filter on retriveable removed

mjawadtp:
Work item: https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001T8dQ8YAJ/view
This work item involves removing the filter `retrieveable` so that Metecho fetches dataset properly including the entities that are marked not `retrieveable`. Changing the Metecho codebase did not help as the filtering actually happens in CumulusCI, and the filters are added in org_schema.py. Deleting line 390 does result in non-retrieveable but `queryable` and `creteable` objects being loaded when dataset is fetched.


vsbharath:
@mjawadtp Closing this as the work item was nevered

PR #3724, opened 2023-12-19
Add content to cumulusci.yml Pydantic models and JSON Schema

jlantz:
Added title, description, and examples to all Pydantic models for the cumulusci.yml json schema to provide better help information in IDEs.

There are a few places where I added enums.


jstvz:
@jlantz Looks like the some test failures due to a schema mismatch. I'll push a copy of your branch to the remote to resolve the lint failure afterwards.

cumulusci/utils/yaml/cumulusci_yml.py
```diff
@@ -60,120 +180,731 @@ def _check(cls, values):
...
+    )
+    days: int = Field(
+        None,
+        title="Days",
+        description="The number of days the scratch org should last.",
+        min=1,
+        max=30,
+        examples=[1, 7, 30],
+    )
+    namespaced: bool = Field(
```
yippie, line 737: yay! the previous definition was wrong (claimed this was a str) and so YAML checking was failing in VSCode

Issue #3429, opened 2022-11-08
Support overriding `cumulusci.yml` to be used for configuration


attilah:
### Feature Type

I'm interested in functionality that solves a current problem or limitation in how CumulusCI works.

### Does your feature request solve a current problem?

This change could be an easier change and first step to support multiple packages in a single repository without changing the structure of the `cumulusci.yml` file by multiple project support to it.

### Describe the solution you'd like

Since `sfdx` supports multiple 2GP packages per repository in a single `sfdx-project.json` file, it would be nice if CumulusCI would provide a way to use different files for these packages.
I'd like to have support 2 types of override:
- env var `CUMULUSCI_YML` (or similar)
- `--configFile <filename>` option for any command, same level handling as `--debug` is Today.

Obviously this change could require deeper changes in `git` integration for get changes and such to filter git logs to a particular folder.
Perhaps implementation needs to find a way to support 1:1 mapping of "package aliases" and entries in `sfdx-project.json`, so instead (or beside) the filename override a `--package <alias>` could be used.

### Describe alternatives you've considered

No alternatives, beside using multiple repositories, because this is how Cumulus CI was designed.

An alternative, more involved implementation could be to support a `PROJECTS` > `<project alias>` level to the YAML file and support specifying a project name on the command line. The structure below that node would be identical what is supported Today at the root level of the file.

PR #3725, opened 2024-01-02
Add support for --load-yml to cci task and cci flow commands

jlantz:
The BaseCumulusCI runtime has long had support for passing additional_yaml which is then added to the yaml merge stack. This functionality was used extensively in MetaCI to inject build specific configuration overrides for certain build plans. However, it was never exposed to the cci cli.

This PR adds a new `--load-yml <path/to/additional.yml>` option flag for all subcommands of `cci task` and `cci flow` by adding support for the option in the main command and then implementing it as an option for the cci task and cci flow commands.

This change should have no impact on existing users and is purely additive functionality that didn't exist before. I've added a docs section to the config docs explaining the use case which is included with this PR along with full test coverage.


additional.yml
```diff
@@ -0,0 +1,14 @@
+flows:
```
jlantz, line 1: Perhaps this belongs alongside the tests instead of here?

jstvz: Yeah, I suggest adding it inline in the tests and using `tmp_path` to write the contents to disk during the test. If that's not workable, then moving it to the same subdirectory as the tests is fine.

docs/config.md
```diff
@@ -908,3 +908,25 @@ how a task or flow is _currently_ configured. The information output by
 these commands change as you make further customizations to your
 project's `cumulusci.yml` file.
 ```
+
+## Loading additional yaml configuration from a file
+
+CumulusCI supports loading in an additional yaml file from the command line with the `--load-yml <path/to/file>` option on `cci task` and `cci flow` commands. This can be useful if you have one-off automation configurations that you want to keep out of the main project's configuration and load only in special cases.
```
jstvz, line 914: Thanks for including documentation with your PR. I think that there should be a mention of this option in the configuration scopes [section above](https://github.com/SFDO-Tooling/CumulusCI/blob/a636dd1e0515ae7b3572cee839ec5f30eef2f151/docs/config.md?plain=1#L640).

jstvz (changes requested):
Thanks @jlantz. Two tweaks needed here.

docs/config.md
jlantz: Thanks for the review. I'll get these addressed shortly.

jstvz:
Fixes #3429.

jstvz:
@jlantz Following up on this PR.

jlantz:
I will get those two changes done and submitted this week, along with a whole bunch of new stuff to put in the review queue ;)

PR #3735, opened 2024-01-26
ListComponents modified to SOAP call 

lakshmi2506:
Modified listing components of non source trackable metadata types using SOAP call `APILISTMETADATA`

SOAP call takes maximum 3 queries at a time and so used the existing the soap envelope which takes 1 metadata type at a time.


jkasturi-sf (approved):
Changing method to retrieve components from SFDX to SOAP API

Issue #3738, opened 2024-01-31
is_person_type error when running snowfakery


tcdahlberg:
### Describe the bug

After upgrading to the recent version of cci (3.84.0) I get an error when running my snowfakery recipe that previously ran.  I get this error:

Error: is_person_type

Which I am guessing may be a shift to a person account model.  

### Reproduction steps

1. Add an EDA dependency to cumulus.yml
```
  dependencies:
    - github: 'https://github.com/SalesforceFoundation/EDA'
 ```
2. Add this task to cumulus.yml
```
  create_fake_school_data:
    description: Makes fake account data to mimic schools with Educaiton as industry
    class_path: cumulusci.tasks.bulkdata.snowfakery.Snowfakery
    options:
      recipe: datasets/snowfakery/fake_schools.yml
```
3. Add the dataset recipe
```
- object: Account
  count: 1000
  fields:
    Name:
      random_choice:
        - choice:
            probability: 25%
            pick: ${{fake.last_name}} University
        - choice:
            probability: 25%
            pick: ${{fake.city}} University
        - choice:
            probability: 25%
            pick: ${{fake.last_name}} College
        - choice:
            probability: 25%
            pick: ${{fake.city}} College
    BillingCity:
      fake: city
    BillingStreet:
      fake: street_address
    BillingState:
      fake: State
    BillingCountry:
      fake: CurrentCountry
    Industry: Education
    Type: College
    RecordType: Educational_Institution
- object: Account
  count: 1000
  fields:
    Name:
      random_choice:
        - choice:
            probability: 25%
            pick: ${{fake.last_name}} High School
        - choice:
            probability: 25%
            pick: ${{fake.city}} High School
        - choice:
            probability: 25%
            pick: ${{fake.last_name}} Senior High
        - choice:
            probability: 25%
            pick: ${{fake.city}} Senior High
    BillingCity:
      fake: city
    BillingStreet:
      fake: street_address
    BillingState:
      fake: State
    BillingCountry:
      fake: CurrentCountry
    Industry: Education
    Type: High School
    RecordType: Educational_Institution

```
4. Run the task after the scratch org builds
```
cci task info snowfakery
```

### Your CumulusCI and Python versions

CumulusCI version: 3.84.0 
Python version: 3.10.2

### Operating System

WIndows 11

### Windows environment

PowerShell

### CumulusCI installation method

pipx

tcdahlberg:
I built a clean cci project with the error in this repo:
[...]

PR #3739, opened 2024-01-31
Fix `is_person_type` `AttributeError` in record type handling

aditya-balachander:
Fix for [issue](https://trailhead.salesforce.com/trailblazer-community/feed/0D54V00007XIPZMSA5) related to your recent PR ([3702](https://github.com/SFDO-Tooling/CumulusCI/pull/3702)).

**Error Message:** `AttributeError: is_person_type`

**Issue Reason:**
During load, the tasks expects the column `is+person_type` to be present in the record type table `rt_mapping_table`. This column is present for sql files which do extract after the new release of CumulusCI (v3.84.0) but will not be present for existing sql files.

**Fix:**
Previously the `outerjoins_to_add` function of the `AddRecordTypesToQuery` class checks if the combination of `is_person_type` and `developer_name` is unique. Now, if the `rt_mapping_table` does not have the `is_person_type` table, the task ignores the condition to check for record types based on `is_person_type` and only checks for `developer_name` (as it was previous to this PR [3702](https://github.com/SFDO-Tooling/CumulusCI/pull/3702))


yippie:
Fixes https://github.com/SFDO-Tooling/CumulusCI/issues/3738

jstvz (approved):

PR #3740, opened 2024-02-01
Add the --no-sandbox argument for Chrome Set Headless 

xmtigra:
The last command was added to address an issue related to an error that occurred when using the script. The --no-sandbox argument disables the sandbox mode in Chrome


jstvz:
@xmtigra Thanks for your contribution. We've reviewed the addition of the `--no-sandbox` flag and decided it's not a change we can merge.  Running Chrome without a sandbox is generally not recommended due to security concerns. It's better to run Chrome as a regular user, not as the root user. 

If you need to use Chrome without the sandbox in your local setup, you can customize your Robot Framework test suite to add this argument when you open the browser.

I'm closing this PR now. If you have any questions or need to discuss this further, please let us know.

PR #3748, opened 2024-02-15
Don't add fullname tag in 2GP package.xml

### LINKED ISSUE #3415: Error creating 2GP version in Winter '23

leboff:
Fixes #3415  for the scenario of when building a locked 2GP


pasres:
We are experiencing the same issue and would really appreciate if this could be checked and hopefully merged soon.
I commented on @leboff 's other PR here: https://github.com/SFDO-Tooling/CumulusCI/pull/3416#issuecomment-2049817319

agaudoncr:
We are also experiencing this issue and would really appreciate it if this could be looked into and hopefully merged.

agaudoncr:
For additional context, we received the following response from Salesforce Support indicating that this is the root of the problem.

`CumulusCI creates the package version request it converts from SFDX to MDAPI and puts a package name (<fullName> tag) in the package.xml file. Could you please check the package.xml file of the project (manifest/package.xml) and see if that tag is being added? If so please try removing the tag and create the package.`

jstvz (approved):
@leboff Thank you for the fix, and apologies for the delay in getting it reviewed.

PR #3741, opened 2024-02-01
Polymorphic Lookup Data Extraction

aditya-balachander:
Epic: [250 - Polymorphic data extraction](https://gus.lightning.force.com/lightning/r/ADM_Epic__c/a3QEE000000szPt2AI/view)

**Polymorphic Lookups** are references that can point to multiple SObjects. Consider the `Event` SObject, where the `WhoId` field is a reference to both `Contact` and `Lead`. Previously, CumulusCI didn't handle these polymorphic lookup fields during extraction and loading. The goal of the Epic was to include this support in CumulusCI.

**Old Format: (mapping.yml)**
```yaml
Insert Contact:
    sf_object: Contact
    api: rest
    fields:
        - LastName
    lookups:
        AccountId: Account
```

**New Format: (mapping.yml)**
```yaml
Insert Event:
    sf_object: Event
    api: rest
    fields:
        - Subject
    lookups:
        WhoId:
            - Contact
            - Lead
```

The Epic spanned over multiple Work Items:

1. [W-14692643](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001gk4w3YAA/view) Identify the reference entity
2. [W-14692643](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001gk6uaYAA/view) Store the reference entity name in dataset
3. [W-14736696](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001hMGelYAG/view) Resolve the reference while loading the data

To accommodate polymorphic lookups for the data extraction and data loading, format of Id’s is changed while extraction. While loading to allow polymorphic fields multiple `sf_id` table were converted into single `sf_id` table. That is after the insertion, the `sf_ids` are queried for the lookups and all these entries are stored in a single `sf_id` table named `cumulusci_id_table`.  So, in the processing it is again a single to outer join for the polymorphic lookup field.

| File Name                             | Test Method Name                             | Description                                                                                                                      |
| ------------------------------------- | -------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
| `cumulusci/tasks/bulkdata/tests/test_load.py` | `test_query_db__joins_polymorphic_lookups` | Test if correct query is being generated for polymorphic lookup fields                                                           |
| `cumulusci/tasks/bulkdata/tests/test_load.py` | `test_generate_contact_id_map_for_person_accounts` | Test if correct `id` field is generated for load with both old format and new format                                        |
[...]

PR #3750, opened 2024-02-20
Fix for mismatch between fields

aditya-balachander:
Fixed `INVALID_CROSS_REFERENCE` error while inserting lookups and record types simultaneously. This error appeared after [PR: Polymorphic Lookup Data Extraction](https://github.com/SFDO-Tooling/CumulusCI/pull/3741). This error occurred due to mismatch in the order of fields between the step and query during generation of the json for the request.


jkasturi-sf (approved):
Looks good

PR #3761, opened 2024-03-05
Fix Id Not Specified on Update Call

aditya-balachander:
[W-15182566](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001lVihqYAC/view)

These modifications address the error reported in [Slack Thread](https://salesforce-internal.slack.com/archives/C024R3K39R6/p1709223158710319)

**Error Reason:**
The old format of sql files use the `id` field to be of type `INTEGER`:
```
CREATE TABLE "accounts" (
	id INTEGER NOT NULL, 
	"Name" VARCHAR(255), 
	"parent_id" VARCHAR(255), 
	PRIMARY KEY (id)
);
INSERT INTO "accounts" VALUES(1,'Bluth','');
INSERT INTO "accounts" VALUES(2,'Funke-Bluth',1);
```

During handling of adding the outerjoins for lookups, when we specified the condition, we were concatenating a string with an integer:
```
def join_for_lookup(lookup):
            key_field = lookup.get_lookup_key_field(self.model)
            value_column = getattr(self.model, key_field)
            if self._old_format:
                return (
                    lookup.aliased_table,
                    lookup.aliased_table.columns.id
                    == str(lookup.table) + "-" + value_column,
                    # Here value_column can be an integer in old format
                )
            else:
                return (
                    lookup.aliased_table,
                    lookup.aliased_table.columns.id == value_column,
                )
```

This resulted in the query not returning anything during the update call of the record. Hence the error, `Id not specified in an update call`

**Solution:**

We have updated the condition to cast the id column to a string so that successful concatenation happens.
```
def join_for_lookup(lookup):
            key_field = lookup.get_lookup_key_field(self.model)
            value_column = getattr(self.model, key_field)
            if self._old_format:
                return (
                    lookup.aliased_table,
                    lookup.aliased_table.columns.id
                    == str(lookup.table) + "-" + func.cast(value_column, String),
                )
            else:
                return (
                    lookup.aliased_table,
                    lookup.aliased_table.columns.id == value_column,
                )
```

### DIFF
--- a/cumulusci/tasks/bulkdata/query_transformers.py
+++ b/cumulusci/tasks/bulkdata/query_transformers.py
@@ -1,7 +1,7 @@
 import typing as T
 from functools import cached_property
 
-from sqlalchemy import and_, func, text, String
+from sqlalchemy import String, and_, func, text
 from sqlalchemy.orm import Query, aliased
 
 from cumulusci.core.exceptions import BulkDataException
--- a/cumulusci/tasks/bulkdata/query_transformers.py
+++ b/cumulusci/tasks/bulkdata/query_transformers.py
@@ -1,7 +1,7 @@
 import typing as T
 from functools import cached_property
 
-from sqlalchemy import and_, func, text
+from sqlalchemy import and_, func, text, String
 from sqlalchemy.orm import Query, aliased
 
 from cumulusci.core.exceptions import BulkDataException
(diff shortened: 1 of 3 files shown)


jstvz (approved):
LGTM

PR #3767, opened 2024-03-27
Add minimum version check for version id dependencies

leboff:
Add the same minimum version check that namespace/version dependencies get to 04t version id dependencies

### DIFF
--- a/cumulusci/core/config/org_config.py
+++ b/cumulusci/core/config/org_config.py
@@ -307,7 +307,24 @@ class OrgConfig(BaseConfig):
             )
 
         return installed_version[0].number >= version_identifier
+        
+    def get_package_from_version(self, package_version_id):
+        """Fetch and return the package details based on a provided package version ID. 
+        Find the SubscriberPackageVersion for the 04t Id provided in package_version_id
+        If no package is found return None. If a package is found return the first record
+        (assuming that the query will only return  one record for a specific Id)."""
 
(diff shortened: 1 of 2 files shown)


yippie:
This appears to be an alternative fix to the same issue we are both having that I tried solving in a different way https://github.com/SFDO-Tooling/CumulusCI/pull/3671

yippie:
@jstvz @davidmreed can we please get one of these merged so that dependency install is much more efficient and doesn't downgrade.

vsbharath:
@yippie @leboff We have reviewed the PR. Right now we don't want to change the salesforce behavior on installing unlocked packages and would recommend this as a new feature on CCI rather than considering it as a Bug. 

We can consider this feature if it satisfies the requirements below 

1. Provide an option to override the salesforce behavior of installing unlocked packages, as we don't want to restrict to not installing lower package versions on the org. 
2. Provide test coverage for get_package_from_version also branches [here](https://github.com/SFDO-Tooling/CumulusCI/pull/3767/files#diff-3e986e2e30678028ba57707bd8ac3e1c3aee3f44c252dd552849015029d0801fR524) . Also please mock the response for tests for [query](https://github.com/SFDO-Tooling/CumulusCI/pull/3767/files#diff-c32354bd415500e590831a13b1118760abdb3a749fa5f2b52eafc2c1cab3d881R325) 

Please let us know if you have any questions on this.

PR #3785, opened 2024-05-13
Allow manually triggering lint workflow

jstvz:
Needed until we update the shared workflow to run on pull requests
from forked repositories.


vsbharath (approved):
Looks good to me

PR #3795, opened 2024-05-29
Updated the PermissionSetLicenseKey to DeveloperName

lakshmi2506:
The PermissionSetLicenseKey have the naming format as `SalesforceCPQ.CPQStandardPerm` and the actual developer name is `SalesforceCPQ_CPQStandardPerm` and to avoid errors .

Replaced with the developername


jkasturi-sf (approved):
Using DeveloperName instead of PermissionSetLicenseKey

PR #3802, opened 2024-06-15
Added support for SkipHandlers

v15h9:
* Adding support for `SkipHandlers` added in `PackageInstallRequest` starting from API Version 61.0 (Summer '24)

### DIFF
--- a/cumulusci/salesforce_api/package_install.py
+++ b/cumulusci/salesforce_api/package_install.py
@@ -52,6 +52,8 @@ class UpgradeType(StrEnum):
     DEPRECATE_ONLY = "deprecate-only"
     MIXED = "mixed"
 
+class SkipHandlers(StrEnum):
+    FEATURE_ENFORCEMENT = "FeatureEnforcement"
 
 class PackageInstallOptions(CCIModel):
     """Options governing installation behavior for a managed or unlocked package."""
@@ -62,6 +64,7 @@ class PackageInstallOptions(CCIModel):
     security_type: SecurityType = SecurityType.FULL
     apex_compile_type: Optional[ApexCompileType] = None
     upgrade_type: Optional[UpgradeType] = None
+    skip_handlers: SkipHandlers = SkipHandlers.FEATURE_ENFORCEMENT
 
     @staticmethod
     def from_task_options(task_options: dict) -> "PackageInstallOptions":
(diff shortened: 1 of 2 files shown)


cumulusci/salesforce_api/package_install.py
```diff
@@ -62,6 +64,7 @@ class PackageInstallOptions(CCIModel):
     security_type: SecurityType = SecurityType.FULL
     apex_compile_type: Optional[ApexCompileType] = None
     upgrade_type: Optional[UpgradeType] = None
+    skip_handlers: SkipHandlers = SkipHandlers.FEATURE_ENFORCEMENT
```
jstvz, line 67: ±: We don't want to silently change behavior, so it shouldn't be a default. Instead, it should be an optional `Option` set to `None`.

cumulusci/salesforce_api/package_install.py
```diff
@@ -86,6 +89,8 @@ def from_task_options(task_options: dict) -> "PackageInstallOptions":
                 )
             if "upgrade_type" in task_options:
                 options.upgrade_type = UpgradeType(task_options["upgrade_type"])
+            if "skip_handlers" in task_options:
+                options.skip_handlers = SkipHandlers(task_options["skip_handlers"])
```
jstvz, line 93: 💭: Implementation note: The [`sf` flag](https://github.com/salesforcecli/plugin-packaging/blob/2421e1577f2168e4875a46361eaa2ae043b3a75c/src/commands/package/install.ts#L106-L113) allows multiple values and combines them into a [comma-separated string](https://github.com/salesforcecli/plugin-packaging/blob/2421e1577f2168e4875a46361eaa2ae043b3a75c/src/commands/package/install.ts#L140) for `PackageInstallCreateRequest`. We don't need to do this here now, but we might need to if they add more `SkipHandlers` in the future.

jstvz (changes requested):
Thanks for the contribution, @v15h9. Requested changes/queries are marked with ±

cumulusci/salesforce_api/package_install.py
v15h9: Updated to `None`

cumulusci/salesforce_api/package_install.py
v15h9: Added a comment for future consideration.

v15h9:
@jstvz Any more changes required for this PR?

PR #3807, opened 2024-06-27
Update the validation for loading to check the required fields as well

lakshmi2506:
To ensure field-level permissions and successful dataset load, there are several validations performed on the `mapping.yml` file. These validations check whether the fields and SObjects have the required permissions, and whether namespaces need to be injected, as well as handling case insensitivity for the `mapping.yml`. This functionality is already implemented in the function `validate_and_inject_mapping`.

However, there was a missing corner case where the function did not capture errors for required fields missing in the mapping.yml. This functionality has now been added, and the function is used for the preflight check.

### DIFF
--- a/cumulusci/cumulusci.yml
+++ b/cumulusci/cumulusci.yml
@@ -81,6 +81,10 @@ tasks:
         description: Waits on a batch apex or queueable apex job to finish.
         class_path: cumulusci.tasks.apex.batch.BatchApexWait
         group: Salesforce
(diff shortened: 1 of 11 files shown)


lakshmi2506:
Hi @jstvz @jkasturi-sf  

Please review the PR 

Thank you

vsbharath (approved):

jkasturi-sf (approved):
Looks good

PR #3818, opened 2024-08-19
@W-16485311: Core Logic for Selecting Records from Target Org

aditya-balachander:
[W-16485311](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001yvdnvYAA/view)

### DIFF
--- a/cumulusci/tasks/bulkdata/select_utils.py
+++ b/cumulusci/tasks/bulkdata/select_utils.py
@@ -1,0 +1,49 @@
+from cumulusci.core.enums import StrEnum
+from cumulusci.tasks.bulkdata.extract_dataset_utils.hardcoded_default_declarations import (
+    DEFAULT_DECLARATIONS,
+)
+
+
+class SelectStrategy(StrEnum):
+    """Enum defining the different selection strategies requested."""
+
+    RANDOM = "random"
+
+
+def random_generate_query(sobject: str, num_records: float):
+    """Generates the SOQL query for the random selection strategy"""
+    # Get the WHERE clause from DEFAULT_DECLARATIONS if available
+    declaration = DEFAULT_DECLARATIONS.get(sobject)
+    if declaration:
+        where_clause = declaration.where
+    else:
+        where_clause = None
+    # Construct the query with the WHERE clause (if it exists)
+    query = f"SELECT Id FROM {sobject}"
+    if where_clause:
+        query += f" WHERE {where_clause}"
+    query += f" LIMIT {num_records}"
+
+    return query, ["Id"]
+
+
+def random_post_process(records, num_records: float, sobject: str):
+    """Processes the query results for the random selection strategy"""
+    # Handle case where query returns 0 records
+    if not records:
+        error_message = f"No records found for {sobject} in the target org."
+        return [], error_message
+
+    # Add 'success: True' to each record to emulate records have been inserted
+    selected_records = [
+        {"id": record[0], "success": True, "created": False} for record in records
+    ]
+
+    # If fewer records than requested, repeat existing records to match num_records
+    if len(selected_records) < num_records:
+        original_records = selected_records.copy()
+        while len(selected_records) < num_records:
+            selected_records.extend(original_records)
+        selected_records = selected_records[:num_records]
+
+    return selected_records, None  # Return selected records and None for error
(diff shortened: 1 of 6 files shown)


cumulusci/tasks/bulkdata/select_utils.py
```diff
@@ -0,0 +1,164 @@
...
+
+    query = f"SELECT {fields_to_query} FROM {sobject}"
+    if where_clause:
+        query += f" WHERE {where_clause}"
+
+    return query, fields
+
+
+def similarity_post_process(
+    load_records: list, query_records: list, num_records: float, sobject: str
```
aditya-balachander, line 84: @mjawadtp remove the `list` typing for `load_records` as it is a generator

cumulusci/tasks/bulkdata/tests/test_step.py
```diff
@@ -879,6 +1109,452 @@ def test_get_prev_record_values(self):
...
+        assert len(results) == 3  # Expect 3 results (matching the input records count)
+        # Assert that all results have the expected ID, success, and created values
+        assert (
+            results.count(
+                DataOperationResult(
+                    id="003000000000001", success=True, error="", created=False
+                )
+            )
+            == 1
+        )
```
aditya-balachander, line 1408: Add the other two asserts for ids `003000000000002`, `003000000000003` as well as you have done for BulkApi

cumulusci/tasks/bulkdata/tests/test_step.py
```diff
@@ -879,6 +1109,452 @@ def test_get_prev_record_values(self):
...
+                ["Id: 2", "Tom", "tom@example.com"],
+            ]
+        )
+        step.start()
+        step.select_records(records)
+        step.end()
+
+        # Get the results and assert their properties
+        results = list(step.get_results())
+        assert len(results) == 3  # Expect 3 results (matching the input records count)
```
aditya-balachander, line 1492: Add the same asserts as before for each id instead of asserting the length of records as they will always be 3

cumulusci/tasks/bulkdata/select_utils.py
mjawadtp: I have removed `list` typing

cumulusci/tasks/bulkdata/tests/test_step.py
mjawadtp: Done

cumulusci/tasks/bulkdata/tests/test_step.py
mjawadtp: Done

jlantz:
Out of curiosity, would this feature enable selects against the User object? I have a current use case for querying User records and relating them to data in a dataset using the user alias field value. It would be really cool if dataset extract could do that mapping to user alias automatically :)

aditya-balachander:
Hey @jlantz,

This tool is primarily designed to load data into a Salesforce org, not extract it. While I'm not sure what you mean about user aliases, you can theoretically add the "User" object to the mapping file and set its action to "select." This means that when you load data into the target org, the tool will query for existing user records in the target org instead of creating new ones. It will then link these users to other objects based on lookup relationships. Hope this helps :)

aditya-balachander:
To further clarify, the goal is to enhance the tool by introducing a feature that allows users to select existing records in the target org. This would provide a way to reference these existing records when creating related child entities during the data load process.

cumulusci/tasks/bulkdata/select_utils.py
```diff
@@ -178,17 +223,13 @@ def similarity_post_process(
...
     complexity_constant = load_record_count * query_record_count
 
-    print(complexity_constant)
-
     closest_records = []
 
     if complexity_constant < 1000:
         closest_records = annoy_post_process(load_records, query_records)
     else:
         closest_records = levenshtein_post_process(load_records, query_records)
```
aditya-balachander: reverse this

cumulusci/tasks/bulkdata/select_utils.py
```diff
@@ -2,6 +2,12 @@
 import re
 import typing as T
 
+import numpy as np
+import pandas as pd
+from annoy import AnnoyIndex
+from sklearn.feature_extraction.text import HashingVectorizer
+from sklearn.preprocessing import StandardScaler
```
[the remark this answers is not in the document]
aditya-balachander: Discuss with James

jkasturi-sf (approved):
Thank you @aditya-balachander and @mjawadtp for putting lot of effort to bring this feature. This feature will benefit many users.

PR #3820, opened 2024-08-21
Include All Keys in standardValue XML for add_standard_value_set_entries Task from Entries Option with Backward Compatibility

lakshmi2506:
```
  task: add_standard_value_set_entries
  options:
      api_names: CaseOrigin
      entries:
          - fullName: New Account
            label: New Account
          - fullName: Questionable Contact
            label: Questionable Contact
      ui_options:
          name: Add values to Case Origin picklist
```

For the add_standard_value_set_entries task, we noticed that only the fullname and label keys were being considered when generating the XML file for deployment, while other keys were being ignored. We've implemented a fix to ensure that all provided keys are now included in the standard value entry XML file.

For example ,

If the entries were given like this , 
```
entries:
   - fullName: testerunner
      label: testrunner
      groupingString: Scheduled
```     
Now, it generates an XML like this 
```
<standardValue>
    <fullName>testerunner</fullName>
    <label>testrunner</label>
    <default>false</default>
    <groupingString>Scheduled</groupingString>
</standardValue>
```

But, there were case by case scenarios added for "OpportunityStage","CaseStatus","LeadStatus" which were left untouched to maintain backward compatibility and default will be always set as False, even though we pass in the entries as true.

Related work item : [W-16541459](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001zQu1TYAS/view)

### DIFF
--- a/cumulusci/tasks/metadata_etl/tests/test_value_sets.py
+++ b/cumulusci/tasks/metadata_etl/tests/test_value_sets.py
@@ -38,8 +38,8 @@ class TestAddValueSetEntries:
                 "api_version": "47.0",
                 "api_names": "bar,foo",
                 "entries": [
-                    {"fullName": "Test", "label": "Label"},
-                    {"fullName": "Test_2", "label": "Label 2"},
+                    {"fullName": "Test", "label": "Label", "group": "Schedule"},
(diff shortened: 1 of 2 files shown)


jkasturi-sf (approved):
Looks good

PR #3825, opened 2024-09-06
Add check recordType XML element exists before referencing text

rtoepferveruna:
Certain objects in Salesforce profile XML schema are not required to have a recordType child element of layoutAssignment node.  This fix prevents an exception being thrown when updating orgs with these objects in the profile metadata.


jlantz:
@jstvz I've reviewed this change, tested it locally, and implemented with @rtoepferveruna in a repo with active CI builds using CumulusCI. I can't see any risk to adding it as it current would be throwing an exception for anyone hitting the issue.

jstvz (approved):
Thanks for the contribution, @rtoepferveruna

jstvz:
Thank you for the contribution, @rjohnson-sfdc. The guard you added for empty `layoutAssignments` is now covered by the fix in #3925 (which handles the broader case of missing/empty XML elements in profile metadata).

Additionally, this branch introduces a regression by replacing `packaging` with the deprecated `pkg_resources` module (removed in setuptools 72+).

Closing as already-fixed. Credit to you for identifying the issue early — the root cause you spotted was correct.

PR #3827, opened 2024-09-13
fix: escape stage name when running ensure_record_type task

leo-dcfa:
Fixed a bug we recently ran into while attempting to run `ensure_record_type` task against an org where there exists an Opportunity StageName with a ampesand.


jstvz (approved):
Thanks for the contribution, @leo-dcfa!

leo-dcfa:
Glad I can contribute @jstvz

leo-dcfa:
@jstvz just wanted to check, is this change scheduled for next release?

jstvz:
> @jstvz just wanted to check, is this change scheduled for next release?

Yes.

PR #3837, opened 2024-09-30
Add `check_components` preflight to detect metadata component conflicts (#3837)

vsbharath:
### Summuray
CheckComponents task is designed to check for existing metadata components in a Salesforce org before deployment, helping to prevent deployment failures due to conflicts with existing components.

### Whats Included
1. New CheckComponents Task
- Functionality:
     Collects metadata components from specified paths or tasks within a plan or flow.
     Check if these components exist in the target Salesforce org using the Metadata API.

- Options:
```
paths: List of paths to include in the check (optional).
plan_name: Name of the current plan (optional).
flow_name: Name of the current flow (optional).
```

2. Unit Tests for CheckComponents

Sample Plan and Flow Integration:

```
plans:
  install:
    steps:
      0:
        task: check_components
        options:
          paths: 'force-app'
          plan_name: 'install'
      2:
        flow: deploy_metadata

flows:
  homehealth_metadata_setup:
    steps:
      0:
        task: check_components
        options:
          paths: 'src'
          flow_name: 'homehealth_metadata_setup'
      2:
        flow: dependencies
      3:
        flow: deploy_hh_metadata

```


cumulusci/tasks/salesforce/tests/test_check_components.py
```diff
@@ -0,0 +1,241 @@
+import os
+from unittest.mock import MagicMock, mock_open, patch
+
+from cumulusci.tasks.salesforce.check_components import CheckComponents
+
+from .util import create_task
+
+
+class TestCheckComponents:
```
jstvz, line 9: It might be worth adding a few more edge case tests, particularly for scenarios involving nested flows or when no deploy paths are found.

cumulusci/tasks/salesforce/check_components.py
```diff
@@ -0,0 +1,212 @@
...
+
+                    return [components, response_messages]
+
+                except ParseError as e:
+                    self.logger.error(f"Error parsing package.xml: {e}")
+                    return None
+            else:
+                self.logger.info(
+                    f"No package.xml found in {full_path}, scanning directories"
+                )
```
jstvz, line 177: It might be better to log a warning here if package.xml doesn't exist, instead of an info message. This would highlight that something might be wrong if expected files are missing.

jstvz (changes requested):
Thanks @vsbharath! This will  be a nice tool for pre-deployment checks. However, testing against EDA revealed some issues.

When running `cci task run check_components --plan_name install --org dev`, the task fails to detect deploy paths, even though they exist in the `customer_org` flow (e.g., `unpackaged/pre` and `unpackaged/post`). It incorrectly reports "No deploy paths found," indicating a problem with detecting paths in nested flows.

The main issue seems to be with the conditional block—it's hard to follow and doesn't correctly decompose plans and tasks. I recommend reviewing the [_freeze_steps](https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/tasks/metadeploy.py#L168) method from the metadeploy task. Lantz has already started extracting this functionality in his [PR](https://github.com/SFDO-Tooling/CumulusCI/blob/050f956a7f2da06c1375d8286363744921c48cb0/cumulusci/core/flowrunner.py#L463).

For example, both `deploy_pre` and `deploy_post` handle groups of paths and should be decomposed into individual deploy calls. If you're using the `List[dict]` returned by `_freeze_steps`, make sure to detect the `task_class` class path to handle this properly.

Another suggestion is to merge all package.xml files into one. For projects like NPSP with many deploy calls, this would be much faster. I'm not sure if we've discussed this before as a v2.

vsbharath:
> Thanks @vsbharath! This will be a nice tool for pre-deployment checks. However, testing against EDA revealed some issues.
> 
> When running `cci task run check_components --plan_name install --org dev`, the task fails to detect deploy paths, even though they exist in the `customer_org` flow (e.g., `unpackaged/pre` and `unpackaged/post`). It incorrectly reports "No deploy paths found," indicating a problem with detecting paths in nested flows.
> 
> The main issue seems to be with the conditional block—it's hard to follow and doesn't correctly decompose plans and tasks. I recommend reviewing the [_freeze_steps](https://github.com/SFDO-Tooling/CumulusCI/blob/main/cumulusci/tasks/metadeploy.py#L168) method from the metadeploy task. Lantz has already started extracting this functionality in his [PR](https://github.com/SFDO-Tooling/CumulusCI/blob/050f956a7f2da06c1375d8286363744921c48cb0/cumulusci/core/flowrunner.py#L463).
> 
> For example, both `deploy_pre` and `deploy_post` handle groups of paths and should be decomposed into individual deploy calls. If you're using the `List[dict]` returned by `_freeze_steps`, make sure to detect the `task_class` class path to handle this properly.
> 
> Another suggestion is to merge all package.xml files into one. For projects like NPSP with many deploy calls, this would be much faster. I'm not sure if we've discussed this before as a v2.

I will take a look at it. It still not resolving the steps from the source. Will address these and get back

vsbharath:
Thanks @jstvz  for taking the time to review. I have addressed all the comments. Please review and let me know if any changes are required.

vsbharath:
> Thanks @jstvz for taking the time to review. I have addressed all the comments. Please review and let me know if any changes are required.

Apologies for more changes. Added few more tests to utils and pushed the changes for review

jstvz (dismissed):
LGTM. Additional tweaks suggested below.

vsbharath:
@jstvz Addressed all the suggestions. Need your approval one more time.

jstvz (approved):

PR #3844, opened 2024-11-12
Release v4.0.0

### LINKED ISSUE #3621: Upgrade all SFDX Commands to SF Commands
### Feature Type

I'm interested in functionality that solves a current problem or limitation in how CumulusCI works.

### Does your feature request solve a current problem?

I noticed this in the SFDX update today:

`IMPORTANT ANNOUNCEMENT, PLEASE READ: We no longer update sfdx (v7); the last stable version is 7.209.6. You can keep using sfdx (v7) if you want, but it won't contain new features or bug fixes. And we no longer update this page.`

And looking at the installers, it now defaults to the SF CLI when you install via npm etc. 

I know we have had warnings for a while now but looks like the move needs to happen to ensure we are not using unsupported CLIs

### Describe the solution you'd like

Migrate all SFDX commands within CumulusCI to SF equivalents

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### LINKED ISSUE #3815: CumulusCI / Selenium 3 is incompatible with python 3.9
### Describe the bug

Currently, Running Choose File from Selenium Library in python 3.9 and above will fail with the following error when connected to selenium hub:
```
LOG FAILAttributeError: module 'base64' has no attribute 'encodestring'
```
[This is because selenium 3 does not support python 3.9](https://github.com/SeleniumHQ/selenium/issues/8762), however, this package depends on selenium 3.
### Reproduction steps

1. Install python 3.9+
2. Install cumulusci
3. Connect to Selenium Hub via Open Browser
4. Run Choose File

Cumulusci version 3.90
robotframework version 7.0
robotframework-seleniumlibrary 5.1.3
Python 3.9

### Operating System

Windows

### Windows environment

WSL

pip

### Error Gist

_No response_

### Additional information

_No response_

jstvz:
Depends on SFDO-Tooling/Snowfakery#1092
- SF CLI Migration
- Drop support for Python 3.8-3.10
- Manage dependencies via uv

Fixes #3621 
Fixes #3815


.github/workflows/release_test.yml
```diff
@@ -42,7 +42,7 @@ jobs:
               with:
                   python-version: 3.8
```
vsbharath, line 43: Do we need python version 3.8 ??

vsbharath (dismissed):

vsbharath (approved):

cumulusci/cumulusci.yml
```diff
@@ -286,18 +286,6 @@ tasks:
         options:
             src_dir: src
         group: Salesforce DX
-    dx_pull:
```
yippie, line 289: @jstvz Why was this done? These are incredibly important and useful commands all over the documentation and is the proper way to interact with scratch orgs.

jstvz: @yippie We removed the `dx_pull` and `dx_push` commands because the underlying Salesforce CLI commands they relied on—`force:source:pull` and `force:source:push`—have been deprecated by Salesforce and replaced with `project retrieve start` and `project deploy start`.

These new commands behave differently, especially in how they handle source tracking and conflict resolution. Keeping the old `dx_pull` and `dx_push` commands could mislead users into thinking everything works the same, which isn’t the case. We want to avoid confusion and ensure everyone updates their workflows to use the new commands.

We recommend updating your scripts to use `project retrieve start` and `project deploy start` directly through the `dx` task or a custom task. We thought it safest to encourage users to verify that these new commands work as expected in their environments.

PR #3845, opened 2024-11-15
Fix readthedocs configuration

jstvz:
- Rename readthedocs.yml to .readthedocs.yml, 
- update to Ubuntu 22.04 and Python 3.12, 
- add uv setup and Sphinx build commands.


mgrandhi (approved):
LGTM

PR #3855, opened 2024-12-06
@W-17366392: Fix Omnistudio issues for the sf command format 

lakshmi2506:


jkasturi-sf (dismissed):
VBT tool is not supporting SF cli. So reverting back to SFDX Cli

jstvz (changes requested):
@lakshmi2506 this PR changes from `sf` back to `sfdx`. Since CumulusCI already uses `sf`, going back might cause problems for people who depend on the newer commands. Could you please share the exact errors you saw when using `sf`?

The Vlocity team [mentioned](https://github.com/vlocityinc/vlocity_build/issues/687) that using the `useSfCli` setting and the `1.17.10` release might fix these issues. Have we tried that yet?

Could we discuss:

- The exact error messages you encountered when using sf?
- The results of testing `useSfCli` with version `1.17.10`?

We should have a plan for what happens when we run into commands that only work with `sfdx`. For example, could we make the build fail with a clear error message, rather than quietly switching back and forth between `sf` and `sfdx`?

cumulusci/tasks/vlocity/vlocity.py
```diff
@@ -105,8 +105,8 @@ def _add_token_to_sfdx(self, access_token: str, instance_url: str) -> str:
         """
         # TODO: Use the sf v2 form of this command instead (when we migrate)
         token_store_cmd = [
-            "sf",
-            "org login access-token",
+            "sfdx",
+            "force:auth:accesstoken:store",
```
vsbharath, line 109: Try this 
           ```
 "sf",
            "org", "login", "access-token",
            "--alias",
            f"{VBT_SF_ALIAS}",
            "--instance-url",
            f"{instance_url}",
```

**Review by vsbharath — CHANGES_REQUESTED** [review]

File: cumulusci/tasks/vlocity/vlocity.py [hunk 2]
```diff
@@ -105,8 +105,8 @@ def _add_token_to_sfdx(self, access_token: str, instance_url: str) -> str:
...
             "--no-prompt",
```
vsbharath, line 110:
```suggestion
            "sf",
            "org", "login", "access-token",
```

lakshmi2506: Thanks 🙇🏻‍♀️   @vsbharath, that worked.

vsbharath (approved):

vsbharath (approved):

jstvz (approved):

PR #3857, opened 2024-12-12
@W-17412267: Fix for records not being inserted when threshold 0

aditya-balachander:
Changes:
 - Made a fix for issue where when there are no records in target org and threshold is 0, the records were not being inserted.


lakshmi2506 (approved):
LGTM

PR #3865, opened 2024-12-26
@W-17440435: Filter default records during INSERT for SELECT action

aditya-balachander:
[W-17440435](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE000026pOS4YAM/view)

Changes:  
- Added a pre-step to the `select_post_process` where any default records (defined in `hardcoded_default_declarations.py`) present in the local SQL are removed if no `user_filter` is provided.  

Reason for the condition: The default records are included during selection from the target org when a `user_filter` is provided. Therefore, default records should only be excluded from the local SQL when they are also excluded during the selection process, which occurs only when no `user_filter` is specified.

### DIFF
--- a/cumulusci/tasks/bulkdata/step.py
+++ b/cumulusci/tasks/bulkdata/step.py
@@ -16,6 +16,9 @@ import salesforce_bulk
 from cumulusci.core.enums import StrEnum
 from cumulusci.core.exceptions import BulkDataException
 from cumulusci.core.utils import process_bool_arg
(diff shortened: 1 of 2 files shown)


lakshmi2506 (changes requested):
@aditya-balachander  can you please check the behavior for the records in lookup with the default records, because it will replaced with an empty string with the present behavior I guess.

aditya-balachander:
@lakshmi2506 since I cannot complete the WI, have assigned the WI to @mjawadtp and closing this WI since the new approach is entirely different

PR #3867, opened 2025-01-03
Remove default declaration for select rows query

mjawadtp:
Default records were not being selected earlier. This bug is now fixed.


lakshmi2506:
@mjawadtp can you give the description on what it solves. Or what is the new behavior.

aditya-balachander:
@mjawadtp the changes look good, however unit tests need to be modified ':-)

aditya-balachander:
Also @mjawadtp,

If possible, modify one of the existing tests for Select+Insert to include a scenario where a default record already exists locally. Ensure that the default record matches the results of the select query so that it is not inserted again during the process.

jkasturi-sf (approved):
Looks good

aditya-balachander (approved):

PR #3870, opened 2025-01-17
@W-17621812: Update _get_required_permission_types to handle SELECT operations

aditya-balachander:
[W-17621812](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002876WEYAY/view)

**Issue:**
- The `requote_uri` function did not handle the `%` character correctly when URL-encoding queries, leading to invalid query formats when passed to the Salesforce API. Specifically, it did not encode `%` as `%25`, causing issues in query string formatting. (Example and associated Error given in [W-17621812](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00002876WEYAY/view) under point 1)
- The permission type check for `SELECT` operations was incorrectly defaulting to `createable`. As a result, fields were unnecessarily required to be `createable`, even though `SELECT` operations only require them to be `queryable`.

**Changes:**  
- Updated the `_get_required_permission_types` function, used in mapping validation, to verify `queryable` permissions instead of `createable` for `SELECT` operations.
- Replaced the use of `requote_uri` with `urllib.parse.quote` to ensure proper URL encoding of special characters, including `%`, which is now correctly encoded as `%25`. This ensures the query string is properly formatted for the Salesforce API.

### DIFF
--- a/cumulusci/tasks/bulkdata/mapping_parser.py
+++ b/cumulusci/tasks/bulkdata/mapping_parser.py
@@ -338,7 +338,10 @@ class MappingStep(CCIDictModel):
         self, operation: DataOperationType
     ) -> T.Tuple[str]:
         """Return a tuple of the permission types required to execute an operation"""
-        if operation is DataOperationType.QUERY:
+        if (
+            operation is DataOperationType.QUERY
+            or self.action is DataOperationType.SELECT
+        ):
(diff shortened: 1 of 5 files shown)


jkasturi-sf (approved):
Looks good

PR #3875, opened 2025-02-07
Fix the check_sobjects_available task to take the api_version from Cumulusci.yml

lakshmi2506:
Fixed the issue where the `check_sobjects_available ` takes api_version as "48.0" and lists the sobjects available as per '48' version. So, it doesn't check for the latest sobjects as per the cumulusci.yml.

Now, removing the hardcoded value automatically picks the version from cumulusci.yml


lakshmi2506:
Hi @jkasturi-sf , since there is no API for enabling few settings, using `check_sobjects_available ` for few validations as a part of preflight checks for reference feature. So need this fix for publishing the tile in radagast for Talent recruitment Management.

PR #3892, opened 2025-04-15
Add new tasks get_assignable_permission_sets and get_assignable_licenses

vsbharath:
Added new tasks to retrieve assignable licenses and permission sets:
   - get_assignable_licenses: Retrieves a list of currently assignable license definition keys based on unused licenses.
   - get_assignable_permission_sets: Retrieves a list of currently assignable Permission Sets based on unused associated user licenses.

### DIFF
--- a/.github/workflows/release_test_sfdx.yml
+++ b/.github/workflows/release_test_sfdx.yml
@@ -45,14 +45,6 @@ jobs:
               uses: actions/setup-python@v5
               with:
                   python-version: 3.11
-                  cache: pip
-                  cache-dependency-path: "pyproject.toml"
-            - name: Check pip cache existence
-              run: |
-                  if [ ! -d ~/.cache/pip ]; then
-                      echo "No pip cache directory found, skipping cache step."
-                  fi
-
             - name: Set up uv
               uses: SFDO-Tooling/setup-uv@main
               with:
--- a/.github/workflows/release_test_sfdx.yml
+++ b/.github/workflows/release_test_sfdx.yml
@@ -47,6 +47,12 @@ jobs:
                   python-version: 3.11
                   cache: pip
                   cache-dependency-path: "pyproject.toml"
+            - name: Check pip cache existence
+              run: |
+                  if [ ! -d ~/.cache/pip ]; then
+                      echo "No pip cache directory found, skipping cache step."
+                  fi
+
             - name: Set up uv
               uses: SFDO-Tooling/setup-uv@main
               with:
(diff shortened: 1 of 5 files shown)


.github/workflows/release_test_sfdx.yml
```diff
@@ -47,6 +47,12 @@ jobs:
                   python-version: 3.11
                   cache: pip
                   cache-dependency-path: "pyproject.toml"
+            - name: Check pip cache existence
+              run: |
+                  if [ ! -d ~/.cache/pip ]; then
+                      echo "No pip cache directory found, skipping cache step."
+                  fi
+
```
jstvz, line 55: I think we can remove it entirely:
```suggestion
```
See: https://docs.astral.sh/uv/guides/integration/github/#caching

(We need to do a bit of CI cleanup.)

cumulusci/tasks/preflight/licenses.py
```diff
@@ -1,13 +1,25 @@
...
 
-class GetAvailableLicenses(BaseSalesforceApiTask):
+class BaseUserLicenseAwareTask(BaseSalesforceApiTask):
+    def get_available_user_licenses(self):
+        """Fetch active user licenses with availability."""
+        query = "SELECT Id, LicenseDefinitionKey, TotalLicenses, UsedLicenses FROM UserLicense WHERE Status = 'Active'"
+        return {
+            lic["Id"]: lic
+            for lic in self.sf.query(query)["records"]
+            if lic["TotalLicenses"] > lic["UsedLicenses"]
```
jstvz, line 11: I labeled the PR a critical change because of this change in behavior.

cumulusci/tasks/preflight/licenses.py
```diff
@@ -1,13 +1,25 @@
...
+        }
```
jstvz, line 12: Can you add some test coverage for this function?

jstvz (commented):
I'm a bit ambivalent about this PR. On one hand, It's actually a bug to return licenses that aren't available for assignment to a user. On the other hand, CumulusCI users may be relying on the current behavior because they have a use case for checking for the _existence_ of a license. 

Given that the task description is ambiguous ("Retrieves a list of the currently available license definition keys"), what about adding leaving the current behavior as-is and instead adding two new tasks (`get_assignable_licenses` and `get_assignable_permission_sets`) and clarifying the difference between them via the task description?

jstvz (approved):
LTGM, thanks. Don't forget the release note.

Issue #3849, opened 2024-11-22
Installing CumulusCI v 4 installs the latest version of urllib3 which seems to be incompatable with running ROBOT tests


liamcollerton-giveclarity:
### Describe the bug

I am running a cci flow from a Github action and I started seeing the following error for every Robot test when running the Robot task.

Parent suite setup failed:

ValueError: Timeout value connect was <object object at 0x7f9212d244f0>, but it must be an int, float or None.

After some investigation I tracked it down to the urllib3 library. The latest version is 2.2.3 which is what is being installed with cumulusCI

Collecting urllib3<3,>=1.21.1 (from requests->cumulusci)
Downloading urllib3-2.2.3-py3-none-any.whl.metadata (6.5 kB)
 

If I add the step 

      - name: Set urlib3 version
        run: |
          pip install urllib3==1.26.20
        shell: bash

after installing cumulus ci and before running the flow then it all works fine 

Note that I am not using the standard cumulus Github actions. I had created my own which install the sf cli and cumulusci and then runs a version of the ci_feature flow.

I expect that this would be an issue in any installation that doesn't already have a version of urlib between 1.21.1 and 1.26.20

### Reproduction steps

1.
2.
3.
n. ...

### Your CumulusCI and Python versions

CumulusCI version: 4.0.1 
Python version: 3.13.0 

### Operating System

Ubuntu latest on a Github Action

### CumulusCI installation method

pip

jstvz:
Thanks for the report. Looks like we pinned this in Snowfakery until recently. The issue is that we're pinning  `robotframework-seleniumlibrary<6`. We'll see whether we can remove that pin, and we'll restore the urllib3 pin if we can't.

Szandor72:
We're also currently blocked due to this issue which hit us by surprise.

liamcollerton-giveclarity:
@Szandor72 The workaround is to make sure that you are using the correct version of that library, after you have installed cumulusci.

How are you experiencing this issue and are you still blocked?

Szandor72:
We downgraded to cumulusci 3.87 for now because we don't want to be
responsible for managing CCI's internal package dependencies. We're
consumers not Pythonians.

jstvz:
[...]

PR #3896, opened 2025-04-23
pushing changes to these files

arvindbhamidipati:
Removed dependencies from Robotframework and selenium. Modified the UV version by upgrading it from 0.5.0 to 0.6.1


jstvz (changes requested):
@arvindbhamidipati You'll need to update the other workflow files referencing that action:
```console
$ rg uv .github -l
.github/workflows/feature_test.yml
.github/workflows/slow_integration_tests.yml
.github/workflows/release_test_sfdx.yml
```

Also, can you update the PR title and description to be more clear and helpful? The title should briefly explain what the change does, since it’ll end up as the commit message when we squash. The description should say why you’re making this change—what [problem](https://github.com/SFDO-Tooling/CumulusCI/issues/3849) it solves, what it's improving, or what it's exploring. Please also add any useful background or links so someone unfamiliar with the work can still follow along. This helps reviewers now and anyone reading the history later. See also:
- https://graphite.dev/guides/best-practices-pr-titles-descriptions
- https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

PR #3906, opened 2025-06-20
Fix for managed context calculation in multi package within same namespace

jorgesolebur:


jstvz:
Thank you for this fix, @jorgesolebur. The managed-context calculation improvement is solid and addresses a real gap for multi-package orgs.

I cherry-picked your changes into #3988 with test fixture updates (the existing tests needed `Name` keys in the mock SubscriberPackage data) and a backward-compat fix so that the task-level `namespace` option still works as a fallback.

Closing this in favor of #3988, which is based on current `main`. Full credit to you via Co-authored-by.

PR #3913, opened 2025-07-29
chore: pin docutils dep to 0.21.2

dipakparmar:


jstvz:
@dipakparmar Merged a separate PR, but thanks for submitting this one.

PR #3924, opened 2025-10-01
feat: Add Pydantic V2 support with V1 compatibility layer

jstvz:
## Summary

This PR adds Pydantic V2 support to CumulusCI using the V1 compatibility layer. This enables working with Pydantic V2 ecosystem while maintaining stability.

## Approach: V1 Compatibility Layer

Instead of converting all models to V2 syntax, we use Pydantic's V1 compatibility namespace (`pydantic.v1`). This approach minimizes breaking changes and allows gradual migration. FastAPI and other projects use this same approach for complex migrations.

## Changes Made

- Dependencies: Pydantic 1.10.22 → 2.11.9, point Snowfakery to `migrate_to_pydantic_v2` branch
- Core modules: Replace `from pydantic import` with `from pydantic.v1 import` across 17 files
- Test compatibility: Fix validation imports and exception handling for V1 namespace

All existing functionality is preserved. All tests pass with the compatibility layer.

### DIFF
--- a/cumulusci/core/config/project_config.py
+++ b/cumulusci/core/config/project_config.py
@@ -22,7 +22,7 @@ from cumulusci.utils.version_strings import LooseVersion
 API_VERSION_RE = re.compile(r"^\d\d+\.0$")
 
 import github3
-from pydantic import ValidationError
+from pydantic.v1 import ValidationError
 
 from cumulusci.core.config import FlowConfig, TaskConfig
 from cumulusci.core.config.base_task_flow_config import BaseTaskFlowConfig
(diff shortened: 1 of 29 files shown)


aditya-balachander (dismissed):
Hey @jstvz,

Went through all references of pydantic and you have ensured that all references have now been shifted to use `pydantic.v1` import. Everything looks good!

There is one linting error: `umulusci/utils/yaml/cumulusci_yml.py:283:5: F824 `global has_shown_yaml_error_message` is unused: name is never assigned in scope` which you can check and remove that line.

Additionally regarding the test failures in `test_select_utils.py`, they are passing on 3.12 but not on 3.11 and 3.13. And when I tested in local with 3.11 and 3.13, they seem to pass. I dont exactly know what the reason for that is.

Approving, so that you can proceed.

aditya-balachander (approved):
Hey @jstvz, 

Looks like there is a bug in Annoy package version 1.17.3 with Python 3.11 which is why the optional tests are failing. Other tests pass. So approving

PR #3925, opened 2025-10-01
Fix empty `layoutAssignments`

naicigam:
Sometimes layout assigments come empty like this `<layoutAssignments><!-- 1 children --></layoutAssignments>`. We need to check if the `recordType` element exists before accessing it.


jstvz (approved):
Required checks pass (3.11-3.13). Branch updated to current main.

PR #3934, opened 2025-10-21
ci: Add Python 3.14 to CI and pin support to <=3.13

jstvz:
This PR adds Python 3.14 to our CI test matrix as an informational/non-blocking test while pinning official package support to Python `<=3.13`. This provides early-warning detection of Python 3.14 compatibility issues without blocking development.

CumulusCI currently uses the Pydantic v1 compatibility layer (via `pydantic>=2.0,<3`), which does not support Python 3.14. A full migration to native Pydantic v2 APIs is not feasible at this time. To prevent users from encountering runtime errors when installing on Python 3.14, we need to restrict package installation while still monitoring compatibility for future planning.

### DIFF
--- a/.github/workflows/feature_test.yml
+++ b/.github/workflows/feature_test.yml
@@ -22,11 +22,11 @@ jobs:
               uses: actions/checkout@v4
               with:
                   fetch-depth: 1
-            - name: Set up Python 3.11
+            - name: Set up Python 3.13
               id: py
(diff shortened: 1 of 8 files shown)


vsbharath (dismissed):

jstvz:
Content cherry-picked into v4.10.0 release. Closing.

PR #3937, opened 2025-12-12
@W-19854606: Strict Mode & Raise validation errors for missing schema and permissions

aditya-balachander:
## Snowfakery `strict-mode`
* Added `strict-mode` support to the `snowfakery` task in CumulusCI.
* When enabled, the Snowfakery recipe is first validated using Snowfakery’s `strict-mode`.
* Standard mapping validations then run as usual.
* Data loading occurs only if both validations succeed.
* If either validation fails, errors are raised and the data load does not proceed.
---
### What changed and why
- Missing sObject: now an error (previously warning in validate-only). The run would fail at execution time, so validation should surface it as blocking.
- sObject permission denied: now an error. Lack of `createable/updateable/queryable` already fails execution; validation now mirrors that.
- Missing field or field not visible: now an error when not using `drop_missing`. Execution would fail without the field, so validation should flag it.
- Field permission denied: now an error when not using `drop_missing`. Lack of required FLS blocks execution; validation should indicate that.
- Ambiguous namespaced field (both `f` and `inject(f)` exist): **kept as warning**. The loader explicitly chooses one and can proceed; this is informational, not blocking.

### Why change warnings to errors?
- Validation-only previously gave “green” results despite conditions that would raise `BulkDataException` at runtime. Converting these to errors makes validation truthful and prevents false confidence.
- Users can still opt into `drop_missing` to auto-trim problematic fields/steps; otherwise, the validation report now matches runtime strictness.

### DIFF
--- a/cumulusci/tasks/bulkdata/generate_from_yaml.py
+++ b/cumulusci/tasks/bulkdata/generate_from_yaml.py
@@ -95,6 +95,9 @@ class GenerateDataFromYaml(BaseGenerateDataTask):
         self.working_directory = self.options.get("working_directory")
         loading_rules = process_list_arg(self.options.get("loading_rules")) or []
         self.loading_rules = [Path(path) for path in loading_rules if path]
+        # These flags may be provided internally by callers (e.g., Snowfakery task)
+        self.strict_mode = bool(self.options.get("strict_mode", False))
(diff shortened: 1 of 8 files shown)


jkasturi-sf (approved):
Looks good

PR #3940, opened 2025-12-18
Issue-3938 rest_deploy should throw errors just like SOAP deploy does

yippie:
If cumulusCI Deploy task is using REST to do the metadata deploy, errors happen silently and the task reports success. This is different from when using SOAP, which will raise an exception.

Add exception raising to the Rest_deploy handler.


jstvz (approved):
Required checks pass (3.11-3.13). Branch updated to current main.

PR #3969, opened 2026-04-23
Add --extra-yaml flag to cci flow run/info and task run/info

### LINKED ISSUE #3429: Support overriding `cumulusci.yml` to be used for configuration

jstvz:
## Summary

Adds `--extra-yaml PATH` (repeatable) to `cci flow run`, `cci flow info`,
`cci task run`, and `cci task info`. It merges extra YAML into the
project config for one command. Also reads `CUMULUSCI_EXTRA_YAML` (a
comma-separated env var) as a fallback. The flag and env var are
opt-in; nothing changes for users who do not set them.

Supersedes #3725. Closes #3429.

## Precedence (highest to lowest)

1. `-o taskname__option value`
2. `--extra-yaml` files (last wins)
3. `CUMULUSCI_EXTRA_YAML` env var
4. Local project `cumulusci.yml`
5. Project `cumulusci.yml`
6. Global `cumulusci.yml`
7. Universal (bundled) `cumulusci.yml`

## Security

Extra YAML can redefine any `class_path`, which runs arbitrary Python
when the task starts. This is the same risk as `cumulusci.yml`, but
the flag makes it easier to pass untrusted input. A stderr warning
prints on every use.

## Incidental change

The first commit pins `pyright` in `.pre-commit-config.yaml` and fixes
two pre-existing `reportPrivateImportUsage` errors from
`simple_salesforce` tightening its re-exports. Without this, the
pyright hook fails on `main`. 4 lines across 3 files. Happy to split
into a separate PR.

## Test plan

- [x] Helper: single file, multi-file deep merge, env var fallback,
      flag overrides env, missing file, empty env segments
- [x] `CliRuntime.reload_project_config`: applies, rebinds keychain,
      no-ops on None, wraps ConfigError
- [x] All four commands accept `--extra-yaml`, including `=PATH` form
- [x] End-to-end: option override preserves siblings, multi-file last
      wins, new task is resolvable, class_path override imports new
      class
- [x] `-o` wins over `--extra-yaml` for the same option
- [x] Smoke tested against this repo
- [x] All pre-commit hooks clean on touched files
- [x] Full pytest suite green

## Docs

- `docs/cli.md`: new "The `--extra-yaml` Flag" subsection
- `docs/config.md`: new per-invocation scope section, updated
  precedence list
- `docs/env-var-reference.md`: `CUMULUSCI_EXTRA_YAML` entry

## Follow-up after merge

- Update the `Cascading-Configuration-Files` wiki page to include
  `--extra-yaml` in the cascade diagram.

### DIFF
--- a/cumulusci/cli/extra_yaml.py
+++ b/cumulusci/cli/extra_yaml.py
@@ -8,8 +8,10 @@ import os
 from typing import Optional, Tuple
 
 import click
+import yaml
 
 from cumulusci.core.exceptions import CumulusCIUsageError
+from cumulusci.core.utils import dictmerge
 
 ENV_VAR = "CUMULUSCI_EXTRA_YAML"
 
@@ -23,8 +25,10 @@ def resolve_extra_yaml(paths: Tuple[str, ...]) -> Optional[str]:
             ``CUMULUSCI_EXTRA_YAML`` (colon-separated paths).
 
     Returns:
-        Concatenated YAML content with ``\\n---\\n`` separators between files,
-        or ``None`` if no paths were resolved.
+        A single YAML document representing the deep-merge of all input files
+        (later files override earlier files), or ``None`` if no paths were
+        resolved. The returned string is a valid single-document YAML stream
+        suitable for ``BaseProjectConfig(additional_yaml=...)``.
 
     Raises:
         CumulusCIUsageError: If any listed path does not exist or is unreadable.
@@ -45,13 +49,22 @@ def resolve_extra_yaml(paths: Tuple[str, ...]) -> Optional[str]:
         err=True,
     )
 
-    contents = []
+    merged: dict = {}
     for path in effective_paths:
         if not os.path.isfile(path):
             raise CumulusCIUsageError(f"--extra-yaml file not found: {path}")
         try:
             with open(path, "r", encoding="utf-8") as f:
-                contents.append(f.read())
+                raw = f.read()
         except OSError as e:
             raise CumulusCIUsageError(f"--extra-yaml could not read {path}: {e}")
-    return "\n---\n".join(contents)
+        try:
+            parsed = yaml.safe_load(raw) or {}
+        except yaml.YAMLError as e:
+            raise CumulusCIUsageError(f"--extra-yaml could not parse {path}: {e}")
+        if not isinstance(parsed, dict):
+            raise CumulusCIUsageError(
+                f"--extra-yaml expects a YAML mapping at the top level in {path}"
+            )
+        merged = dictmerge(merged, parsed)
+    return yaml.safe_dump(merged, default_flow_style=False)
--- a/cumulusci/cli/extra_yaml.py
+++ b/cumulusci/cli/extra_yaml.py
@@ -1,0 +1,57 @@
+"""Resolve ``--extra-yaml`` CLI flag and ``CUMULUSCI_EXTRA_YAML`` env var.
+
+The returned string is passed as ``BaseProjectConfig``'s ``additional_yaml``
+kwarg, which already merges into the project config via the existing YAML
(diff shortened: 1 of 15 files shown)


docs/config.md
```diff
@@ -668,6 +674,16 @@ default value for `opt2`, this new default `opt2` value takes precedence
 over the default `opt2` value specified in your global `cumulusci.yml`
 file.
 
+### Per-Invocation (`--extra-yaml`)
+
+The `--extra-yaml PATH` flag on `cci flow run`, `cci flow info`,
+`cci task run`, and `cci task info` merges an additional YAML file
+into the project config for a single command invocation. Multiple
+files are deep-merged in order (later files win); the
+`CUMULUSCI_EXTRA_YAML` environment variable (colon-separated paths)
```
abikkavilli-sf: @jstvz  should it be comma-separated instead of colon-separated

cumulusci/cli/extra_yaml.py
```diff
@@ -0,0 +1,72 @@
...
+            raise CumulusCIUsageError(f"--extra-yaml could not read {path}: {e}")
+        try:
+            parsed = yaml.safe_load(raw) or {}
+        except yaml.YAMLError as e:
+            raise CumulusCIUsageError(f"--extra-yaml could not parse {path}: {e}")
+        if not isinstance(parsed, dict):
+            raise CumulusCIUsageError(
+                f"--extra-yaml expects a YAML mapping at the top level in {path}"
+            )
+        merged = dictmerge(merged, parsed)
```
abikkavilli-sf, line 71: @jstvz dictmerge appends lists instead of replacing them. So,  if file1.yml defines a flow with 3 steps and file2.yml defines the same flow with 2 different steps, the result is 5 steps (concatenated), not 2 (replaced). Should we document this behavior explicitly in cli.md specifying the lists have this behavior

abikkavilli-sf (dismissed):
Thanks for taking care of this @jstvz 
Changes look good. Leaving some nits here for doc based changes.

docs/config.md
jstvz: Fixed in 89e325d03. The env var is comma-separated; the doc was stale.

cumulusci/cli/extra_yaml.py
jstvz: `docs/cli.md` already calls this out at the `--extra-yaml` section: "Mappings and scalars are overridden; lists are concatenated, not replaced."

jstvz:
@abikkavilli-sf Awaiting final review.

PR #3976, opened 2026-05-08
fix: SetOrgWideDefaults._poll_action fails with stale access token after deploy

yippie:
Task set_organization_wide_defaults fails 100% of the time if RTR (Refresh token rotation) is on. This is a Connected App security setting that Salesforce now requires and affects CumulusCI when embedded in MetaDeploy.

## Root cause

SetOrgWideDefaults._run_task() calls _deploy(), which instantiates a new
Deploy task and invokes api() on it. Deploy.__call__() always calls
BaseSalesforceTask._update_credentials() → org_config.refresh_oauth_token(),
rotating org_config.access_token AFTER self.sf was already frozen in _init_task().

When _post_deploy() → _poll() → _poll_action() runs, self.sf.query() uses
the pre-deploy access token. If the org has refresh token rotation or a short
session policy, this token is invalid and every poll call returns a 401.

## Sequence

| Step | Event | org_config.access_token | self.sf.session_id |
|---|---|---|---|
| SetOrgWideDefaults.__call__() | _update_credentials() → refresh | TOKEN_A | — |
| _init_task() | self.sf = _init_api() | TOKEN_A | TOKEN_A |
| _deploy() → Deploy()() | _update_credentials() → refresh again | TOKEN_B | still TOKEN_A ← stale |
| _post_deploy() → _poll_action() | self.sf.query() | TOKEN_B | TOKEN_A → 401 |

## Fix

Replace self.sf.query(...) with self.org_config.salesforce_client.query(...).

OrgConfig.salesforce_client is a @property that constructs a fresh
simple_salesforce.Salesforce instance from self.access_token on every
access, so it always uses the current (post-deploy-refresh) token.

## Tests

Updated the three _poll_action / _post_deploy tests to mock
org_config.salesforce_client via PropertyMock instead of assigning
directly to task.sf.

### DIFF
--- a/cumulusci/tasks/metadata_etl/sharing.py
+++ b/cumulusci/tasks/metadata_etl/sharing.py
@@ -73,7 +73,10 @@ class SetOrgWideDefaults(MetadataSingleEntityTransformTask, BaseSalesforceApiTas
     def _post_deploy(self, result):
         if result == "Success":
             super()._post_deploy(result)
+            # Deploy.__call__() rotates org credentials; rebuild all API clients before polling
             self.sf = self._init_api()
+            self.bulk = self._init_bulk()
+            self.tooling = self._init_api("tooling")
             self.logger.info("Waiting for sharing enablement to complete.")
             self.time_start = datetime.now()
             self._poll()
--- a/cumulusci/tasks/metadata_etl/sharing.py
+++ b/cumulusci/tasks/metadata_etl/sharing.py
@@ -73,10 +73,6 @@ class SetOrgWideDefaults(MetadataSingleEntityTransformTask, BaseSalesforceApiTas
     def _post_deploy(self, result):
         if result == "Success":
             super()._post_deploy(result)
-            # Deploy.__call__() rotates org credentials; rebuild all API clients before polling
-            self.sf = self._init_api()
-            self.bulk = self._init_bulk()
-            self.tooling = self._init_api("tooling")
             self.logger.info("Waiting for sharing enablement to complete.")
             self.time_start = datetime.now()
             self._poll()
--- a/cumulusci/tasks/metadata_etl/sharing.py
+++ b/cumulusci/tasks/metadata_etl/sharing.py
@@ -115,7 +115,7 @@ class SetOrgWideDefaults(MetadataSingleEntityTransformTask, BaseSalesforceApiTas
                 else sobject
             )
 
-            result = self.sf.query(
+            result = self.org_config.salesforce_client.query(
                 f"SELECT ExternalSharingModel, InternalSharingModel "
                 f"FROM EntityDefinition "
                 f"WHERE QualifiedApiName = '{real_api_name}'"
--- a/cumulusci/tasks/metadata_etl/sharing.py
+++ b/cumulusci/tasks/metadata_etl/sharing.py
@@ -73,6 +73,7 @@ class SetOrgWideDefaults(MetadataSingleEntityTransformTask, BaseSalesforceApiTas
     def _post_deploy(self, result):
         if result == "Success":
             super()._post_deploy(result)
+            self.sf = self._init_api()
             self.logger.info("Waiting for sharing enablement to complete.")
             self.time_start = datetime.now()
             self._poll()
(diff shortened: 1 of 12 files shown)


jstvz:
Thanks for the fix — the bug diagnosis is correct. After `_deploy()` creates a child `Deploy` task sharing the same `org_config`, the child's `_update_credentials()` rotates `org_config.access_token`, but `self.sf` on the parent `SetOrgWideDefaults` still holds the old token.

However, replacing `self.sf.query(...)` with `self.org_config.salesforce_client.query(...)` introduces a few behavioral regressions:

1. **Lost retry adapter** — `self.sf` (built by `get_simple_salesforce_connection`) installs an `HTTPAdapter` with `Retry(total=5, status_forcelist=(502, 503, 504), backoff_factor=0.3)`. `salesforce_client` creates a bare `Salesforce()` with no retry adapter. This polling loop runs up to 600 seconds post-deploy, when transient 502/503/504 errors are common.

2. **Lost `CALL_OPTS_HEADER_KEY`** — the connection utility sets a `client=...` header for server-side observability. `salesforce_client` does not.

3. **Object per iteration** — `salesforce_client` is a `@property` that creates a new `Salesforce()` on every access, so each poll iteration allocates a fresh client.

**Suggested alternative** (one-line fix, preserves all existing behavior):

Refresh `self.sf` once after deploy returns. In `_post_deploy`, before the polling loop starts:

```python
self.sf = self._init_api()
```

This re-creates the connection with the current (rotated) token while keeping the retry adapter, the `CALL_OPTS_HEADER_KEY` header, and the task-configured API version. Single allocation instead of one per poll iteration.

Would you be open to revising the approach? Happy to pair on it if useful.

yippie:
@jstvz Absolutely, I think I updated per your recommendation

cumulusci/tasks/metadata_etl/sharing.py
```diff
@@ -73,6 +73,7 @@ def _init_options(self, kwargs):
     def _post_deploy(self, result):
         if result == "Success":
             super()._post_deploy(result)
+            self.sf = self._init_api()
```
jstvz: `self.sf = self._init_api()` overwrites mocks set in `test_post_deploy_waits_for_enablement`, `test_post_deploy_waits_for_enablement__namespaced_org`, and `test_post_deploy_exception_not_found` (which pre-set `task.sf = mock.Mock()`). After this line lands inside `_post_deploy`, those tests construct a real `simple_salesforce.Salesforce` against `https://test.salesforce.com` and raise `SalesforceMalformedRequest`. CI on this SHA shows the same failure across all unit-test jobs. Either patch `_init_api` in the three tests (`task._init_api = mock.Mock(return_value=mock.Mock(query=...))`) or restructure the fix so the mock survives.

Reproduce: `pytest cumulusci/tasks/metadata_etl/tests/test_sharing.py`.

jstvz (changes requested):
Diagnosis is sound. The token-rotation race between the parent task's frozen `self.sf` and the nested `Deploy.__call__()` -> `_update_credentials` refresh is real and well-described in the PR body. The patch as posted, however, breaks the existing test suite for this task and CI is fully red.

### Blocking

- sharing.py:76 — `self.sf = self._init_api()` overwrites the mock that the three existing `test_post_deploy_*` tests rely on. CI on a256f084 shows the failure across all unit-test jobs. Either patch `_init_api` in those tests or restructure the fix. (inline)

### Non-blocking

- Re-init only refreshes `self.sf`. `self.bulk` and `self.tooling` (set in `BaseSalesforceApiTask._init_task`) still hold the pre-deploy token. Harmless here because `_poll_action` only uses `self.sf.query`, but a future edit using `self.tooling`/`self.bulk` will regress silently. Either re-call `_init_task()` or add a comment pinning the assumption.
- Repo scan confirms `SetOrgWideDefaults` is the only `_post_deploy` override in the codebase. The base stub in `metadata_etl/base.py:161` is a no-op. A per-class patch is sufficient today, but hardening would mean rebuilding `self.sf`/`self.tooling`/`self.bulk` in `BaseMetadataETLTask._run_task` between `_deploy()` and `_post_deploy(result)` to prevent regressions if future subclasses adopt this hook.

cumulusci/tasks/metadata_etl/sharing.py
yippie: I took a whack at your last comment about a longer term, more general fix, I hope the new approach looks good

Issue #3989, opened 2026-05-28
Unpackaged Metadata not included in Package Creation


yippie:
### Describe the bug

There are metadata present in my `unpackaged/post/setup` directory that Apex tests in my package depend on. When running the `create_package_version` task as part of the `release_2gp_beta` flow, it appears that the metadata in `unpackaged/post/setup` are not present because these Apex tests fail. 

I do have the unmanaged metadata setup in the `sfdx-project.json` in the primary package directory entry under `unmanagedMetadata`. 

I hadn't seen this behavior previously just using the `sfdx` commands, so I'm wondering if or how `cci` supports making unmanaged metadata accessible for Apex test executions during package version creation, as described by the [SFDX Developer Guide here](https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_dev2gp_unpackaged_md.htm).

I believe the task needs a new input parameter to include unpackaged metadata.

### Reproduction steps

1. Create a project with Apex unit tests
2. Add metadata that is referenced by the Apex tests to `unpackaged/post/setup`, e.g. a permission set
3. Run the `create_package_version` task

### Your CumulusCI and Python versions

CumulusCI version: 4.10.0
Python version: 3.12.9

### Operating System

macOS Tahoe 26.5

### CumulusCI installation method

pipx

Issue #3987, opened 2026-05-25
SfdxOrgConfig breaks with Salesforce CLI May 27 2026 security update — migrate accessToken/password reads to new dedicated commands


mayank-kumar-mishra:
## Summary

Salesforce CLI is shipping a breaking change on **May 27, 2026** that redacts sensitive
credentials from standard command output. CumulusCI's `SfdxOrgConfig` currently reads
`accessToken` and `password` directly from `sf org display --json`, which will stop
working after that release.

Reference: https://github.com/forcedotcom/cli/issues/3560

---

### `cumulusci/core/config/sfdx_org_config.py`

**1. `sfdx_info` property (line 32, 56, 60, 64–65)**
```python
p = sfdx("org display --json", self.username)
# ...
org_id = org_info["result"]["accessToken"].split("!")[0]   # line 56 — breaks
"access_token": org_info["result"]["accessToken"],         # line 60 — breaks
sfdx_info["password"] = org_info["result"]["password"]     # line 65 — breaks

```

tcdahlberg:
I updated to @salesforce/cli/2.135.7 on Friday 5/22/2026 and I believe the new commands are already in place. After upgrading my cci builds are failing on " Getting org info from Salesforce CLI for..." Are other experiencing this early?

jstvz:
Thanks for the report, @tcdahlberg and @mayank-kumar-mishra. We confirmed on our end as well.

**Workaround:** Set the environment variable `SF_TEMP_SHOW_SECRETS=true` in your shell or CI environment before running CCI. This restores the old `sf org display` behavior. Note: Salesforce will remove this workaround in summer 2026, so it is temporary.

```bash
export SF_TEMP_SHOW_SECRETS=true
cci flow run ...
```

We have a fix tracked internally (W-22684438). The permanent fix will detect the new CLI behavior and call the dedicated `sf org auth show-access-token` / `sf org auth show-user-password` commands, with graceful fallback for older CLI versions.

apretorius:
@jstvz Any update on when the fix will make it into a release?

HemangAgrawal:
@jstvz, when are we thinking of fixing this issue?

PabloRoldan:
@mgrandhi, I noticed that you merged this and released a new version. However, this new release has been launched with a prerequisite that didn't exist for version 4.9.0.
Version 4.9.0 prerequisite was pip >= 3.11
[...]

PR #3991, opened 2026-05-28
add missing unpackaged metadata option

yippie:
Add support for the Salesforce "unpackaged Metadata" option when creating package versions.

https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_dev2gp_unpackaged_md.htm

Fixes https://github.com/SFDO-Tooling/CumulusCI/issues/3989

### DIFF
--- a/cumulusci/tasks/create_package_version.py
+++ b/cumulusci/tasks/create_package_version.py
@@ -199,6 +199,14 @@ class CreatePackageVersion(BaseSalesforceApiTask):
         self.options["create_unlocked_dependency_packages"] = process_bool_arg(
             self.options.get("create_unlocked_dependency_packages") or False
         )
+        if "unpackaged_metadata_path" in self.options:
+            unpackaged_metadata_path = pathlib.Path(
+                self.options["unpackaged_metadata_path"]
+            )
+            if not unpackaged_metadata_path.is_dir():
+                raise TaskOptionsError(
+                    f"unpackaged_metadata_path does not exist or is not a directory: {unpackaged_metadata_path}"
+                )
 
     def _init_task(self):
         self.tooling = get_simple_salesforce_connection(
@@ -434,8 +442,9 @@ class CreatePackageVersion(BaseSalesforceApiTask):
                         "settings.zip", settings_zip_builder.as_bytes()
                     )
 
-            # Add unpackaged metadata if provided
-            if "unpackaged_metadata_path" in self.options:
+            # Add unpackaged metadata if provided (main package only)
+            is_dependency = package_config is not self.package_config
+            if "unpackaged_metadata_path" in self.options and not is_dependency:
                 with convert_sfdx_source(
                     pathlib.Path(self.options["unpackaged_metadata_path"]),
                     None,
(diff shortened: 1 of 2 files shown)


cumulusci/tasks/create_package_version.py
```diff
@@ -429,6 +434,25 @@ def _create_version_request(
                         "settings.zip", settings_zip_builder.as_bytes()
                     )
 
+            # Add unpackaged metadata if provided
```
jstvz: `_create_version_request` is called three times: main package (line 243), unpackaged-pre dependency (line 684), and local unpackaged dependency (line 719). The new block has no `is_dependency` gate, so the configured `unpackaged_metadata_path` will be attached to every unlocked dependency package version request, not only the main package. The existing dependencies block at line 454-463 already gates on `is_dependency`; mirroring that gate looks correct.

cumulusci/tasks/create_package_version.py
```diff
@@ -150,6 +150,11 @@ class CreatePackageVersion(BaseSalesforceApiTask):
             "description": "If True, create unlocked packages for unpackaged metadata in this project and dependencies. "
             "Defaults to False."
         },
+        "unpackaged_metadata_path": {
```
jstvz, line 153: No validation in `_init_options` that `unpackaged_metadata_path` exists or is a directory. A missing or mistyped path currently surfaces as a raw `FileNotFoundError` from `os.listdir(...)` deep inside `convert_sfdx_source`. Validating here with a clear `TaskOptionsError` would help users.

jstvz (changes requested):
Adds an `unpackaged_metadata_path` option to `create_package_version` so the build org used during package version creation can deploy extra metadata that Apex tests in the package depend on. The option writes `unpackaged-metadata.zip` into `VersionInfo` and sets `package_descriptor["unpackagedMetadata"]`, matching the SFDX `unpackagedMetadata` semantics. Tests assert presence in the positive case and absence in the negative case.

Two correctness issues to address before merge (inline):

1. The new block runs every time `_create_version_request` is called, including the two dependency calls at lines 684 and 719. The configured path will be attached to every unlocked dependency package, not only the main package. The existing dependencies block at line 432 gates on `is_dependency`; mirroring that gate looks correct.

2. A missing `unpackaged_metadata_path` directory currently surfaces as a raw `FileNotFoundError` from `convert_sfdx_source`. Validating in `_init_options` would give a clearer error.

Two minor cleanups (non-blocking): the new `version_info.writestr` call sits outside the `with convert_sfdx_source(...)` block (the analogous settings block keeps it inside; works because the zip builder buffers in BytesIO, but the asymmetry reads like a scope bug), and the new tests re-import `BasePackageZipBuilder` locally even though it is imported at module scope.

The descriptor key, the zip entry, and the test coverage all line up with the SFDX behavior the linked issue (#3989) calls out.

cumulusci/tasks/create_package_version.py
yippie: Added a user facing error

cumulusci/tasks/create_package_version.py
yippie: Added the is_dependency check

jstvz (approved):
Lint is clean locally (`ruff check` passes). The "Lint / Lint" CI failure on this SHA is a workflow infrastructure issue (`fatal: couldn't find remote ref feature/add-unpackaged-metadata-support`) unrelated to code quality.

The remaining failing checks (3.14 unit tests, Robot, Test SFDX CLI, Org-connected) reproduce on every PR right now and trace to the SF CLI 2.136 [REDACTED] token issue (#3987) plus the Python 3.14 matrix cap. Out of scope here.

LGTM.

PR #3998, opened 2026-06-26
fix(sfdx): retrieve org credentials after SF CLI 2.136+ redaction @W-22684438

### LINKED ISSUE #3987: SfdxOrgConfig breaks with Salesforce CLI May 27 2026 security update — migrate accessToken/password reads to new dedicated commands

jstvz:
## Summary

Fixes #3987. Salesforce CLI 2.136+ redacts `accessToken`, `password`, and `sfdxAuthUrl` from `sf org display --json`. `SfdxOrgConfig` reads those fields, so credential retrieval breaks the moment a user's CLI updates. CumulusCI does not pin the CLI version, so CI and MetaCI builds fail when 2.136+ reaches the stable channel.

## CLI behavior verified against shipping versions

Confirmed against installed CLI 2.140.6, `@salesforce/plugin-org` 5.11.8, and `@salesforce/core` 8.31.1. Verified against the upstream `main` HEAD of both repos.

- CLI 2.136+ does not delete or null the fields. It replaces each field with a truthy sentinel string. The source is `@salesforce/plugin-org`'s `messages/secrets-redacted.md`:
  - `accessToken` -> `[REDACTED] Use 'sf org auth show-access-token' to view`
  - `password` -> `[REDACTED] Use 'sf org auth show-user-password' to view`
  - `sfdxAuthUrl` -> `[REDACTED] Use 'sf org auth show-sfdx-auth-url' to view`
- The only gate is the env var `SF_TEMP_SHOW_SECRETS` (default false redacts). The CLI applies no version, date, or org-type gate. Scratch orgs and persistent orgs behave the same.
- `id` is never redacted. Org id is always readable.
- `password` is present (and then redacted) only when the org has a locally generated password. For passwordless orgs (web-auth sandbox or production, scratch org without `force:user:password:generate`), the field is absent, not a sentinel. The absent-vs-sentinel distinction drives the password fallback gate.
- `SF_TEMP_SHOW_SECRETS=true` re-enables the legacy output. Salesforce removes the env var in summer 2026, so it is a stopgap.

## Changes

One source file: `cumulusci/core/config/sfdx_org_config.py`.

- Detect redaction by the `[REDACTED] ` sentinel prefix. Truthiness alone does not fire because the sentinel is a non-empty string.
- Add two fallback helpers that re-fetch via the new subcommands with `--no-prompt`. The `--no-prompt` flag suppresses an interactive confirm that hangs headless processes.
  - `sf org auth show-access-token`
  - `sf org auth show-user-password`
- Access token path: fall back when the field is redacted, absent, or empty.
- Password path: fall back only when the value matches the sentinel. An absent password means the org has no password, so the code leaves it unset and runs no `show-user-password` subprocess. This avoids a pointless extra subprocess on every refresh for the most common new-CLI org class (passwordless).
- Org id: read `result["id"]` (raise a clear `SfdxOrgException` if missing). The previous `accessToken.split("!")[0]` derivation no longer works because the token is redacted. Imported-org ids now surface as the 18-character form, consistent with the base `OrgConfig`. Salesforce APIs accept both 15- and 18-character ids.
- Shell-quote every username interpolated into an `sf` command. `sfdx()` runs with `shell=True`, so an unquoted username with shell metacharacters is a command-injection footgun. Applied at all sites.

## Backward compatibility

- Pre-2.136 CLI returns the real token in `org display`. The code runs no fallback subprocess.
- On pre-2.136 CLI, the `sf org auth show-*` subcommands may not exist. The token helper raises a clear `SfdxOrgException` that names the attempted commands. The password helper degrades to "no password" and never raises.
- `force_refresh_oauth_token` is unchanged. It uses `sf org open -r`, which reads no secrets.

## Test coverage

All tests pass in `test_config_expensive.py` and `test_org.py`.

New and updated coverage:

- Sentinel token triggers the access-token fallback.
- New-CLI passwordless org does not call `show-user-password` (subprocess count == 2).
- Both fields redacted fetches both (subprocess count == 3).
- Legacy CLI passthrough (subprocess count == 1).
- Both paths fail raises an exception that names both commands.
- Username with shell metacharacters arrives shell-quoted in the command string.
- The password helper returns `None` on subprocess error, empty output, or sentinel response.

## Known limitation

The `[REDACTED] ` prefix is locale-stable on every shipping CLI today. `@salesforce/core`'s `Messages.getLocale()` is hardcoded to `en_US`, and per-locale message bundles are an unimplemented TODO upstream. The prefix lives inside the translatable message string, so if Salesforce later ships localization, a non-English `redacted.accessToken` could drop the literal `[REDACTED] ` prefix and break detection silently. Detection failure surfaces loudly: the sentinel string becomes the "access token", and the next API call returns 401 with a human-readable token value. The fix is a one-line constant change. A locale-independent signal would cover only `accessToken` (it has a shape), not `password` or `sfdxAuthUrl`, so the alternative would not remove the locale risk from the other two fields. The `sfdx_org_config.py` constant has a comment that names the failure mode and the tripwire symptom.

## Test plan

- [ ] Run `pytest cumulusci/core/config/tests/test_config_expensive.py` and confirm green.
- [ ] Run `pytest cumulusci/cli/tests/test_org.py` and confirm green.
- [ ] Exercise on a real new-CLI scratch org: `cci org info <alias>` returns access token and password.
- [ ] Exercise on a real new-CLI passwordless org: `cci org info <alias>` returns access token, password absent, no `show-user-password` subprocess in `cci --debug` log.
- [ ] Exercise on a legacy CLI (2.135 or earlier) if available: passthrough path, one subprocess.

Closes #3987.

### DIFF
--- a/cumulusci/cli/tests/test_org.py
+++ b/cumulusci/cli/tests/test_org.py
@@ -96,7 +96,7 @@ class TestOrgCommands:
         auth_code_flow.return_value = {
             "instance_url": "https://instance",
             "access_token": "BOGUS",
-            "id": "OODxxxxxxxxxxxx/user",
+            "id": "00Dxxxxxxxxxxxx/user",
         }
         runtime = mock.Mock()
         runtime.project_config = BaseProjectConfig(UniversalConfig(), config={})
@@ -109,7 +109,7 @@ class TestOrgCommands:
         )
         responses.add(
             method="GET",
-            url=f"https://instance/services/data/v{CURRENT_SF_API_VERSION}/sobjects/Organization/OODxxxxxxxxxxxx",
+            url=f"https://instance/services/data/v{CURRENT_SF_API_VERSION}/sobjects/Organization/00Dxxxxxxxxxxxx",
             json={
                 "TrialExpirationDate": None,
                 "OrganizationType": "Developer Edition",
@@ -129,7 +129,7 @@ class TestOrgCommands:
 
         name, org_config = runtime.keychain.get_default_org()
         assert name == "test"
-        assert org_config.id == "OODxxxxxxxxxxxx/user"
+        assert org_config.id == "00Dxxxxxxxxxxxx/user"
         assert org_config.connected_app == "built-in"
         assert org_config.expires == "Persistent"
         assert "Connecting org using the built-in connected app..." in result.output
@@ -141,7 +141,7 @@ class TestOrgCommands:
         auth_code_flow.return_value = {
             "instance_url": "https://instance",
             "access_token": "BOGUS",
-            "id": "OODxxxxxxxxxxxx/user",
+            "id": "00Dxxxxxxxxxxxx/user",
         }
         runtime = mock.Mock()
         runtime.project_config = BaseProjectConfig(UniversalConfig(), config={})
@@ -166,7 +166,7 @@ class TestOrgCommands:
         )
         responses.add(
             method="GET",
-            url=f"https://instance/services/data/v{CURRENT_SF_API_VERSION}/sobjects/Organization/OODxxxxxxxxxxxx",
+            url=f"https://instance/services/data/v{CURRENT_SF_API_VERSION}/sobjects/Organization/00Dxxxxxxxxxxxx",
             json={
                 "TrialExpirationDate": None,
                 "OrganizationType": "Developer Edition",
@@ -280,7 +280,7 @@ class TestOrgCommands:
         client_instance.auth_code_flow.return_value = {
             "instance_url": "https://instance",
             "access_token": "BOGUS",
-            "id": "OODxxxxxxxxxxxx/user",
+            "id": "00Dxxxxxxxxxxxx/user",
         }
         oauth2client.return_value = client_instance
         runtime = mock.Mock()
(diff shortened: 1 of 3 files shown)


yippie:
@jstvz This is very much needed.

But why try and detect if the string is redacted. Could you not simply check the SF CLI version and/or existence of the new commands and, if the commands exist, use them? This would bypass any string format detection issues.

PR #4008, opened 2026-09-03
Fix auth detection

yippie:
Updated detection to check if the sf cli had the new commands or not.


yippie:
@jstvz here is my proposed fix that bypasses the obfuscation detection issues.

PR #4009, opened 2026-09-07
Fix Handle SF CLI 2.136+ credential redaction in SFDX org config

### LINKED ISSUE #3987: SfdxOrgConfig breaks with Salesforce CLI May 27 2026 security update — migrate accessToken/password reads to new dedicated commands

sf-brahm-gaur:
# Handle SF CLI 2.136+ credential redaction in SFDX org config

  ## Summary

  SF CLI 2.136+ no longer returns raw secrets from `sf org display --json`.
  Instead of the actual value (or absence), redacted fields come back as a
  **truthy sentinel string** prefixed `"[REDACTED] "`, e.g.
  `"[REDACTED] Use 'sf org auth show-access-token' to view"`.

  CumulusCI's `SfdxOrgConfig` detected redaction with `if not value:`, which
  never fires against a truthy sentinel — so the access token and password
  were silently populated with the placeholder string, later surfacing as
  opaque API 401s. This PR corrects the detection model and adds the
  documented fallback path.

  ## Changes

  - **Detect the `[REDACTED] ` sentinel** rather than testing for
    absence/null. `sfdx_info` (access token + password) and
    `get_access_token` all route through the new detection helper.
  - **Fall back to the `show-*` subcommands** when a field is redacted:
    `sf org auth show-access-token` and `sf org auth show-user-password`.
    The legacy-CLI path (raw values, no sentinel) is unchanged.
  - **Use `result.id` directly for `org_id`** instead of splitting it out of
    the access-token prefix, which is no longer available under redaction.
  - **Shell-quote usernames** at every `--target-org={username}`
    interpolation site (these run through `sfdx()` with `shell=True`),
    closing a command-injection footgun.
  - **Gate the password fallback on the sentinel only**, not on absence, so
    passwordless orgs (web-auth sandboxes, scratch orgs without a generated
    password) don't fire a pointless extra subprocess.

  Cache shape and the 1-hour TTL are preserved. `ScratchOrgConfig.scratch_info`
  inherits the fix via the existing alias.

  ## Testing

  - Added coverage for the sentinel-detection and shell-quote contracts,
    including the passwordless-org "no fallback" case and the
    both-fields-redacted scratch-org shape.
  - Full suite green (`test_config_expensive` + `test_org`).

  Closes #3987

### DIFF
--- a/cumulusci/cli/tests/test_org.py
+++ b/cumulusci/cli/tests/test_org.py
@@ -96,7 +96,7 @@ class TestOrgCommands:
         auth_code_flow.return_value = {
             "instance_url": "https://instance",
             "access_token": "BOGUS",
-            "id": "OODxxxxxxxxxxxx/user",
+            "id": "00Dxxxxxxxxxxxx/user",
         }
         runtime = mock.Mock()
         runtime.project_config = BaseProjectConfig(UniversalConfig(), config={})
@@ -109,7 +109,7 @@ class TestOrgCommands:
         )
         responses.add(
             method="GET",
-            url=f"https://instance/services/data/v{CURRENT_SF_API_VERSION}/sobjects/Organization/OODxxxxxxxxxxxx",
+            url=f"https://instance/services/data/v{CURRENT_SF_API_VERSION}/sobjects/Organization/00Dxxxxxxxxxxxx",
             json={
                 "TrialExpirationDate": None,
                 "OrganizationType": "Developer Edition",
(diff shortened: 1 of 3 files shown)


mgrandhi (approved):
LGTM

PR #4011, opened 2026-09-08
docs(contributing): switch local setup to uv and add UTC test note

sf-brahm-gaur:
## Summary

  Updates `docs/contributing.md` to match how CumulusCI is actually
  developed today and to save new contributors from two common setup
  snags.

  ## Changes

  **Switch local setup to uv.** The setup steps still described a manual
  `python3 -m venv` + `make dev-install` + `pip` flow. This updates them
  to the current uv-based workflow:

  - Install uv (curl installer, with `brew install uv` / `pip install uv`
    noted as alternatives).
  - Use `uv sync --all-groups` to provision a compatible Python
    (3.11–3.13), create `.venv`, install CumulusCI in editable mode, and
    pull in the `dev` and `lint` dependency groups in one step.
  - Run tooling via `uv run` (pre-commit, pytest, coverage), with a note
    that activating `.venv` lets you drop the prefix.
  - Add `isort` to the listed pre-commit hooks.

  **Add a UTC test-clock note.** Some tests assume the system clock is set
  to UTC because they derive values (e.g. scratch-org age) from
  `datetime.utcnow()`. Contributors in other time zones hit off-by-one
  date failures in tests like `test_days_alive`, `test_format_days`, and
  `test_org_list`. Added a note under "Testing CumulusCI" suggesting a
  `TZ=UTC` prefix.

  ## Notes

  The legacy `make dev-install` target is intentionally left untouched;
  it's superseded by `uv sync` and no longer referenced by the guide. The
  one Makefile target the docs still call, `make coverage`, works
  unchanged under `uv run`. Cleaning up the legacy targets can be a
  separate follow-up.

  ## Impact

  Docs only. No code or behavior changes.


mgrandhi (approved):
LGTM

